PR #48, opened 2015-09-07
The start of a dark theme

mdinger:
This is definitely WIP

Here is my attempt at making a dark theme. I was mainly making a dark grey theme but decided to try a dark blue theme similar to what gitbook uses to see if I liked it better. Both need more tweaking (if you're interested in both). Blue is mostly a clone of grey (I'd have to change it some more to differentiate or consolidate it I suppose). My css is bad though so me trying to consolidate doesn't always work.

Building a doc set ignores the new css file I added (not sure why but I also didn't look). I just copy it into the folder afterwards.

This would also need a button to switch themes. I have no idea how to do that.

Also, I was thinking of trying to make a Rust theme (similar to how the playpen is themed around pale browns) since this is written in Rust and if the rust docs ever move to this, they'd probably like a Rust themed theme. But I wasn't sure so I didn't try.

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -31,7 +31,8 @@
             <div class="page">
                 <div id="menu-bar" class="menu-bar">
                     <div class="left-buttons">
-                        <i id="sidebar-toggle" class="fa fa-bars left"></i>
+                        <i id="sidebar-toggle" class="fa fa-bars"></i>
+                        <i id="theme-toggle" class="fa fa-paint-brush"></i>
                     </div>
 
                     <h1 class="menu-title">{{ title }}</h1>
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -16,6 +16,7 @@
         <link rel="stylesheet" href="http://maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
 
         <link rel="stylesheet" href="highlight.css">
+        <link rel="stylesheet" href="tomorrow-night.css">
 
         <!-- MathJax -->
         <script type="text/javascript" src="https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML"></script>
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -179,6 +179,8 @@ impl Renderer for HtmlHandlebars {
         // syntax highlighting
         let mut highlight_css = try!(File::create(book.get_dest().join("highlight.css")));
         try!(highlight_css.write_all(&theme.highlight_css));
+        let mut tomorrow_night_css = try!(File::create(book.get_dest().join("tomorrow-night.css")));
+        try!(tomorrow_night_css.write_all(&theme.tomorrow_night_css));
         let mut highlight_js = try!(File::create(book.get_dest().join("highlight.js")));
         try!(highlight_js.write_all(&theme.highlight_js));
 
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -13,46 +13,32 @@ html, body {
     .light .page-wrapper {
         background: none repeat scroll 0 0 #FFF;
     }
-    .light .chapter li {
-        color: #AAA;
-    }
-    .light .chapter li a {
-        color: #333;
-    }
-    .light .chapter li .active {
-        color: #008cff;
-    }
+    .light .chapter li .active,
     .light .chapter li a:hover {
         /* Animate color change */
         color: #008cff;
     }
-    .light .menu-bar {
+    .light .menu-bar,
+    .light .nav-chapters {
         color: #CCC;
     }
-    .light .menu-bar :hover {
+    .light .chapter li a,
+    .light .menu-bar i:hover,
+    .light .nav-chapters:hover {
         /* Animate color change */
         color: #333;
-    }
-    .light .menu-title {
-        color: #7E7E7E;
-    }
-    .light .menu-bar:hover .menu-title {
-        color: #7E7E7E;
     }
     .light .content a {
         color: #4183c4;
     }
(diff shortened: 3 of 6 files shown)


src/theme/index.hbs
```diff
@@ -13,10 +13,11 @@
         <link href='http://fonts.googleapis.com/css?family=Open+Sans:300italic,400italic,600italic,700italic,800italic,400,300,600,700,800' rel='stylesheet' type='text/css'>
         <link rel="stylesheet" href="http://maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
         <link rel="stylesheet" href="highlight.css">
+        <link rel="stylesheet" href="tomorrow-night.css">
 
 
     </head>
-    <body>
+    <body class="theme-dark-grey">
```
azerupi: I think the light theme should be the default one

azerupi:
Nice! Thank you very much :)

> This would also need a button to switch themes. I have no idea how to do that.

I will do it, but basically you just have to add a button to the menu-bar with a click event-listener, that opens some kind of popup to choose a theme.

Do you want me to merge or do you want some extra time to tweak?

mdinger:
I actually want a button that I can switch themes with so that I know how to do it in the future. If you can put up a branch with a button switcher that I can see, I can try to add it into this and just add the two others as options.

Should I try to make a Rust theme? Do you want a bunch of themes or just a light and a dark? Do you have a preference between the two dark themes?

azerupi:
Ok I will do that tomorrow

> Should I try to make a Rust theme? Do you want a bunch of themes or just a light and a dark?

If you make one I can definitely add it. But I need to think a little bit about how to handle multiple themes, because the current design does not work very well.

mdinger:
Okay cool. Also, if you can split off the color parts of the white theme into a separate "theme" file, then I can rebase onto that and put the different color themes into separate files (I'm not sure if you can do that or if you can, how you split css files apart like that but it seems like a good idea).

Man, you'd think css tutorials would mention the difference between commas, spaces, and nesting but they don't (see the following example). The only way I could get that theme to actually was commas. Web (html, css, jquery, ...) documentation is so confusing:
```css
// What is the difference between all of these?
outer_name {
    name1 name2 { color: #000000; }
    name3, name4 { color: #000000; }
    name5 {
        name6 { color: #000000; }
    }
}
```

azerupi:
It's going to take a little longer than expected because my internet is
down at home since yesterday. Hopefully it will be back up soon.

azerupi:
I have added a [branch with a theme button](https://github.com/azerupi/mdBook/tree/js-theme-button), all the javascript to change the theme works, it already toggles the body class to `light` and `dark`. You just have to plug your css in and tweak it a little bit.

> Also, if you can split off the color parts of the white theme into a separate "theme" files

I will probably change from css to stylus (css pre-processor), when I do the change it will be a lot easier to split off the themes.

Also like I said, I think the theme part needs a complete overhaul. It's not dynamic enough. I will try to do this for *v0.3*

> difference between commas, spaces, and nesting

It's quite simple actually

- **spaces** indicate a nested relationship: `div a` matches all links that are in a div element, the link does not have to be the direct child of the div, but can be nested further. Like `<div><ul><li><a>...`

- **Commas** indicate that you want to apply the styling to multiple elements, there is no relationship between the elements. `div, a` translates to *"apply the styling to all `div` elements and all `a` elements."*

- **Nesting**, in pure css you can't use nesting to designate an element. But it is widely used in css-preprocessors (Sass, Less, Stylus). The only form of "nesting" in pure css is [media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries) that gives you the ability to style differently depending on properties of the media used to browse the website, like screen size or whether is a screen or for printing purposes , etc.

mdinger:
Okay, that's good to know. I tried nesting for like an hour but it just didn't work.

Cool. I'll try to mess with it tomorrow.

mdinger:
Here I finished it. It does a few things I don't know how to fix but this should be a really good start I think.

Added localStorage so the theme would persist across pages but it doesn't always work. Not sure why. If you toggle the theme and then click the browser back button, the theme reverts until you refresh. If you go from a subchapter to a different chapter, the theme may revert to the default requiring you to pick it again. I think it may be seeing it as a different website or something but I'm not sure.

If you refresh the page and watch the "next chapter" buttons, they flash purple seemingly before the final css take over. I thought I had it fixed but it came back. I'm not sure why they're doing that but adding a `a { color: #fff; }` at the upper level of the css fixed but since I didn't want them to all be that color I backed it out. Not sure what the solution to it is.

I considered making the chapters **bold** but I wasn't sure so I didn't.

Finally, I kinda think the navy needs a blue based highlighting theme but I didn't find one I liked so I used the tomorrow-night I was using for the coal. Maybe a new one can be manually created later but I'm not sure. I think it still looks decent.

azerupi:
Very nice themes, I love the rust theme!

I have fixed the flashing on page load by setting the correct class on `body` before any content is loaded. Other than that it seems to work without any problems.

Many thanks for this PR, this really adds a lot of value.

mdinger:
Okay cool. Then I'm not sure if [this](https://github.com/azerupi/mdBook/blob/master/src/theme/book.js#L6-L10) is needed anymore or not. Does it need to be checked at page load or will your updated script handle it well.

Also, if you set the theme I think you want to call `set_theme()` (though I don't know how to scope into `book.js` to get that function) because it correctly disables/enables the correct highlight.js stylesheets for the theme (I couldn't figure out another way to select between multiples styles). It seems to work okay so maybe it's getting corrected in the linked function.

@steveklabnik You should check out the Rust theme [here](http://azerupi.github.io/mdBook/). I was thinking specifically of the rust book at the time.

mdinger:
Okay, I looked into it further and I see now and it seems okay as is. [`import`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) doesn't seem possible yet. Maybe in the future.

PR #53, opened 2015-09-17
Fix many minor warnings found by clippy

mdinger:
Mostly I took whatever suggestions I found because the result seemed good.

I thought dereferencing a match would result in a move so I wasn't sure about:
```rust
// This
match *a {
    A::B => {},
    A::C => {},
}
// Instead of
match a {
    &A::B => {},
    &A::C => {},
}
```
But I tried it and didn't get any errors. I'll have to experiment with it in the future sometime later.

### DIFF
--- a/src/book/bookconfig.rs
+++ b/src/book/bookconfig.rs
@@ -47,10 +47,7 @@ impl BookConfig {
 
         // Just return if an error occured.
         // I would like to propagate the error, but I have to return `&self`
-        match config_file.read_to_string(&mut data) {
-            Err(_) => return self,
-            _ => {},
-        }
+        if let Err(_) = config_file.read_to_string(&mut data) { return self }
 
         // Convert to JSON
         if let Ok(config) = Json::from_str(&data) {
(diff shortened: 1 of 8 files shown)


azerupi:
Thanks! Looks good. 
Do you think it's worth integrating clippy in the Travis testing?

mdinger:
You could but I probably wouldn't. I wouldn't follow all the suggestions so you may always get warnings if you do. Also, since they're only warnings they won't kill a travis build so you probably wouldn't notice it.

For example, I ignored a warning like:

```rust
// Replace this
match a.foo().bar() {
    Ok(x) => x,
    _ => println!(""),
}
// With
if let Some(x) = a.foo().bar() { x }
else { println!(""); }
```

Both are clear but I probably prefer the first in cases like this.

azerupi:
Ok :) In that case, the best thing to do is run it every so often.

> Both are clear but I probably prefer the first in cases like this.

I do prefer n° 1 too

PR #65, opened 2015-11-20
adds version for deps to make cargo installable

kbknapp:
Removes the `*` so this can be installed via `cargo install mdbook`

Also, I locked the `handlebars` dep to `0.11.2` because the current `0.11.3` fails to build.


azerupi:
Thanks!  :)

PR #71, opened 2015-12-15
Implement 'mdbook test'

### LINKED ISSUE #69: `mdbook test`
This would be a subcommand that runs `rustdoc --test` on each of the source files and reports what happened.

steveklabnik:
Fixes #69 

Let me know if this style is off at all, since I'm not very familiar with the codebase!

### DIFF
--- a/src/bin/mdbook.rs
+++ b/src/bin/mdbook.rs
@@ -33,6 +33,8 @@ fn main() {
                         .arg_from_usage("[dir] 'A directory for your book{n}(Defaults to Current Directory when ommitted)'"))
                     .subcommand(SubCommand::with_name("watch")
                         .about("Watch the files for changes"))
+                    .subcommand(SubCommand::with_name("test")
+                        .about("Test that code samples compile"))
                     .get_matches();
 
     // Check which subcomamnd the user ran...
@@ -40,6 +42,7 @@ fn main() {
         ("init", Some(sub_matches))  => init(sub_matches),
         ("build", Some(sub_matches)) => build(sub_matches),
         ("watch", _)                 => unimplemented!(),
+        ("test", Some(sub_matches)) => test(sub_matches),
         (_, _)                       => unreachable!()
     };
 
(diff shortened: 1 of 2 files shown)


azerupi:
Seems good for now :)

Thinking about a more generic way to have tests I was thinking about something like:

- We could keep a map that stores the tests (a function / closure associated with a name). Similar to how `helpers` work in handlebars. 
- In the API we could expose a method to add custom tests. (for people who use the crate as a lib)
- The user should be able to choose the tests he wants to run in the `book.json`

Any thoughts on this?

steveklabnik:
I like these ideas, but one hard part is that tests don't really have names, exactly, or they're auto-generated. You usually control what you want to be tested in the source, via annotations.

Using the crate as a lib might need work though.

azerupi:
Sorry I think I expressed myself poorly.

By "tests" I meant different tests mdbook could perform, checking rust code being one of them. Other tests could include checking for broken links for example (I am making this up as an example, it's not something I am planning to do at the moment).

So I was thinking that all different tests would be a function stored in a map associated to a keyword
```
"rust": test_rust_examples()
"broken_links": check_broken_links()
...
```

So that the user could have something like this in his `book.json`
```json
"test": ["rust", "broken_links"] 
```

And we would just run the tests that match from the map. I don't know if that makes any sense..? 

> Using the crate as a lib might need work though.

I'm interested to hear more about this :)

All the functionality is in the "lib" part of the crate, the binary is essentially a wrapper around the "lib" providing a cli. It should be possible to use mdbook as a lib to incorporate it in another project.   The API may be terrible though, I am not very experienced in making good APIs ;)

steveklabnik:
Ah yes!

Here's a fun story about "broken links" tests. When working on Rust by Example, I accidentally screwed up a deploy, and the whole site was 404ing. Frantically, I fixed the issue, and tried to redeploy... but since the deploy process relied on the tests passing, and the pages had links to other pages, every intra-site link 404d, failing the deploy! I had to turn them off, and now we have them as just warnings.

Anyway, I do think that style makes a lot more sense than what I thought you were talking about. :)

I mis-spoke myself, I meant to say "might work", not "might need work" :sweat_smile:

azerupi:
Haha sometime's it feels like everything is against you :wink: 

>I mis-spoke myself, I meant to say "might work", not "might need work" 

That's ok, it happens :)

PR #75, opened 2015-12-22
Generalize inline code to all themes

### LINKED ISSUE #74: Dark theme messes up inline code blocks
Source:

```
    This `let` statement has this form:
```

light (and Rust):

2015-12-21-130315_264x38_scrot

navy (and coal):

2015-12-21-130411_205x77_scrot

mdinger:
As far as I can tell, the top two blocks in `highlight.css` are added/tweaked from the original (included in hightlight.js) while the remaining part of the original css file basically remains identical. The problem is the `:not(pre) > .hljs` isn't included for dark themes which causes issues like https://github.com/azerupi/mdBook/issues/74.

If possible, I'd move away from modified themes so they can just be updated to the latest version of whatever highlight.js is using. I tried backing out the theme tweak but reverted it because the code backgrounds were a weird yellow.

Fixes #74

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -270,8 +270,16 @@ h3 {
   }
 }
 .light {
+/* Inline code */
   color: #333;
   background-color: #fff;
+}
+.light :not(pre) > .hljs {
+  display: inline-block;
+  vertical-align: middle;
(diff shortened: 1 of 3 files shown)


azerupi:
Thanks! :)

Issue #76, opened 2015-12-23
Newlines in markdown files aren't turned into spaces


mdinger:
[Go here](http://azerupi.github.io/mdBook/) and CTRL-F `liketo`. This is a mangling of [this file](https://github.com/azerupi/mdBook/blob/a2b25232d34ffeaa0506bfee26df5c695e94dbaa/book-example/src/README.md) where the newline between `like` and `to` is removed without any space being added. It shows up in issues like https://github.com/rust-lang/book/issues/39. The newline should be parsed into a space. Maybe this needs to go upstream but this is a good jumpoff point.

azerupi:
Ow, that's bad :disappointed: 
That should definitely be handled in the markdown parser though. I really hope someone picks-up pulldown-cmark or that an alternative pops up. 

In the mean time, I am not sure what I could do about it

steveklabnik:
> In the mean time, I am not sure what I could do about it 

Well, do you know why pulldown-cmark hasn't been developed? Did they just drop off the map?

azerupi:
Yes the author had [stated his interest](https://users.rust-lang.org/t/new-commonmark-parser/1690) in replacing the use of hoedown by pulldown-cmark in rustdocs and there was a lot of enthusiasm about it. 

>I'd like for this to become the engine in rustdoc. Right now it implements CommonMark (an effort to standardize Markdown) with no extensions, but if extensions are necessary I'm open to adding them.

There were a couple of features missing though before it could replace hoedown. And some pull-request were made. The pull request adding support for tables was the last one that got merged in, I think. After that the author made one more commit, changing the license to MIT.

He hasn't made any comments, he just stopped responding (on github). I am not sure if he is coming back or not.

-----------------------------------

Since I started with Rust I have been interested in creating a markdown parser. Mainly because there wasn't any and because I wanted to learn about parsers. But I don't have enough time nor the experience to develop / maintain one on my own. 

If however a couple of people would be interested to team up, I would be glad to be one of them ;)

steveklabnik:
Well, let's see if we can get him to at least acknowledge what's up, maybe if not we can fork.

azerupi:
If you can contact him, that would be sweet! Then at least  we know what we should expect :)

asolove:
@azerupi @steveklabnik I'm pretty sure this isn't a problem with pulldown-cmark, but with a little bit of logic here in mdBook.

[...]

PR #78, opened 2015-12-27
Add newlines back in to generated index.html files.

asolove:
This is an inelegant fix for #76. I read about better ways to do this using `intersperse` and `join` but couldn't make the types work out. Happy for someone else to show me a better way.


azerupi:
Thanks a lot for spotting this bug, I would have never thought of that :)

Issue #72, opened 2015-12-16
Automatically create anchors for section headers


steveklabnik:
If I write something like this:

```
## Foo
```

and I want to link to it later, I have to do this:

```
<a href="#foo">
## Foo
</a>
```

or the equivalent. GitHub and rustbook generate these kinds of links automatically, which keeps source files cleaner. Example: https://github.com/rust-lang/rust/#quick-start

azerupi:
If I implement this in mdbook it's going to be an ugly hack, I would have to either parse the markdown a second time myself or parse the html outputted by pulldown-cmark.

I think this should be implemented in the markdown parser. But pulldown-cmark hasn't seen any activity in a while, so I doubt that is going to happen anytime soon. :/

steveklabnik:
Ah interesting, I should check up on `pulldown-cmark`.

asolove:
Since `pulldown-cmark` is striving to match the CommonMark spec, I think it's unlikely they would accept a PR for this non-standard behavior. It would be pretty trivial to do this in JS after the fact, though, as long as you had a clear rule for how to turn a run of text into an anchor name that can't have spaces.

azerupi:
>It would be pretty trivial to do this in JS after the fact

Yeah, but I had in mind that the browser wasn't going to scroll down if anchors were added after page load. Obviously I was wrong :wink:  Thanks for the PR!

>Since pulldown-cmark is striving to match the CommonMark spec, I think it's unlikely they would accept a PR for this non-standard behavior.

It's pretty common to add anchors to headings in web-dev, I don't understand why this isn't part of the spec yet..

And one neat thing about Rust, is that you can easily gate some parts of your code behind `Features`, non-standard additions could be feature-gated and off by default. :)

PR #79, opened 2015-12-27
Add anchors around all headers in the content.

asolove:
This is one way to address #72, creating anchors for sections in the text.

- Just uses the header's text as its anchor name. Spaces in the names work fine in URLs. In the browsers I tested, scrolling to the anchor in the URL works even when the anchor is added after the DOM loads.
- Adjust theme css to only style links, not <a> tags used as anchors.

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -315,7 +315,7 @@ h3 {
 .light .mobile-nav-chapters {
   background-color: #fafafa;
 }
-.light .content a {
+.light .content a:link {
   color: #4183c4;
 }
 .light .theme-popup {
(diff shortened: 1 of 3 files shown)


azerupi:
You changed the CSS file directly, but I use [stylus](http://stylus-lang.com/) (css pre-processor) to have more modular stylesheets. The stylesheet is added to the git repo so that people don't have to install node.js to regenerate it.

Would you mind editing the [stylus files](https://github.com/azerupi/mdBook/tree/master/src/theme/stylus) instead?  

>scrolling to the anchor in the URL works even when the anchor is added after the DOM loads.

I never thought that would work, but if it does, everything else is good. I will merge once you change the stylus files :wink:

asolove:
Oh, hah, didn't even see them. Yeah, I'll update this a little later today.

azerupi:
Sure I will merge it tomorrow morning then (time-zone difference UTC+01:00 here :wink: )

steveklabnik:
:confetti_ball:

azerupi:
Perfect, merging now!

PR #84, opened 2015-12-30
Switch to SASS, get rid of node dependency

killercup:
Hi, I wanted to see how well sass-rs (bindings to libsass) is working and thought mdBook could profit from this. I semi-manually converted all you `.styl` files to `.scss` (mostly just adding semicolons and changing `$name = val` to `$name: val`) and changed `build.rs` to use `sass-rs`.

This way, there is no need to globally install any packages from npm (it would've been nicer to just add a `package.json` and install them locally, btw). One downside is that currently, there is no autoprefixing happening (e.g., to automatically generate definitions of older `display: flex` syntaxes), but this can also be done with some simple mixins.

Another downside is that currently Cargo cannot deal with optional build-dependencies, so sass-rs will _always_ be compiled.

Like I said, I mostly did this to play with the libsass bindings. Feel free to ignore this PR if this is not something you'd like to do. :smile:


azerupi:
Thanks for that :)

However I think I will stay with Stylus for the moment. I have used Sass in the past, but I prefer stylus because it gives you more freedom. This is personal preference though :wink: 

Unless there are real advantages to using sass-rs over stylus?

I would stay with Stylus becauseI like it better and I don't feel like it's a burden for the user, I recompile the css when I change it and upload the generated css with the repo. Therefore, no user has to recompile the css. Only developers do have to regenerate it when they change the stylesheets.

>it would've been nicer to just add a package.json and install them locally, btw

You are right, I just happened to have it installed already, and it was one less file "polluting" the repo. :smile:  I can change that if contributors prefer to have local installs.

killercup:
> I would stay with Stylus becauseI like it better and I don't feel like it's a burden for the user

Sure, that sounds good. You are probably going to be touching these files the most. :)

> Unless there are real advantages to using sass-rs over stylus?

Not having to use node/npm is the only one I can think of (and that one doesn't even bother me, my day job is writing JS). This was the first time I interacted with Stylus, so I can't really compare it to Sass (the code here is pretty trivial as it's 'just' nested CSS with variables, which maps 1:1 to Sass). Oh, one thing I noticed when reading the docs for sass-rs: You can define Sass functions in Rust.

> I can change that if contributors prefer to have local installs.

I can't speak for other developers but I _hate_ to globally install stuff I only use in one project. (By defining a `package.json` you could also move the CSS compile step to an npm script and just call `npm run build` in your `build.rs`. That way, you could later also add stuff to compile JS files and fonts.)

azerupi:
>pretty trivial as it's 'just' nested CSS with variables, which maps 1:1 to Sass

I am also using "nib" but I guess this could be replaced with "Compass"? And in the [color themes](https://github.com/azerupi/mdBook/tree/master/src/theme/stylus/themes) I use a slightly more advanced feature, I don't know if Sass has an equivalent. 

Basically I have a `base.styl` where I have all the repetitive css enclosed in a dynamic class

```styl
.{unquote($theme-name)} {
    // Repetitive css using variables $bg, $fg, ...
}
```

and in every theme I define the `$theme-name` variable and customize all the variables used in `base.styl` and then import it.

```styl
$theme-name = 'light'

$bg = #ffffff
$fg = #333333
...
@import 'base'
```

And this will create 

```css
.light {
    // Repetitive css using variables $bg, $fg, ...
}
```

Does that make sense? :smile: 

>You can define Sass functions in Rust

That is pretty cool :)

>By defining a package.json you could also move the CSS compile step to an npm script and just call npm run build in your build.rs. That way, you could later also add stuff to compile JS files and fonts.

Yeah, that's a good point too. As my javascript begins to grow I should probably split it into smaller chunks for maintainability. Also It would probably be better to compress it for production.. 

However I would prefer doing that in Rust, because else I would have to include non-compressed and compressed files in the repo

killercup:
> I am also using "nib" but I guess this could be replaced with "Compass"?

What does nib do? Compass is a _huge_ set of mixins (up to and including functions to generate image sprites and gradients als SVG files…).

> And in the color themes I use a slightly more advanced feature, I don't know if Sass has an equivalent. […] Does that make sense?

Yep, makes perfect sense! And that works fine with this code, too, as you can see in the `book.css`. The Sass syntax for `.{unquote($theme-name)} { … }` is `.#{$theme-name} { … }`.

> However I would prefer doing that in Rust, because else I would have to include non-compressed and compressed files in the repo

Isn't that exactly what you are doing with the CSS? :wink: You could, of course, move the static files (CSS, JS, fonts, …) to a separate repository (e.g., `mdbook-style`) and only include the compiled versions here.

azerupi:
>What does nib do?

It overrides a lot of CSS3 properties so that when you use one it will automatically add all the vendor-prefixed versions too. This without the user having to use an alternative syntax.

[First example they give on their site](http://tj.github.io/nib/)
```styl
body {
  background: linear-gradient(top, white, black);
}
```
converts to
```css
body {
  background: -webkit-gradient(linear,
    left top,
    left bottom,
    color-stop(0, #fff),
    color-stop(1, #000));
  background: -webkit-linear-gradient(top, #fff 0%, #000 100%);
  background: -moz-linear-gradient(top, #fff 0%, #000 100%);
  background: linear-gradient(top, #fff 0%, #000 100%);
}
```

>Isn't that exactly what you are doing with the CSS?

It is :blush: But in the future I would like to include some parts of the JS as plugins configured in `book.json` therefore (I think) I would need to concatenate and compress at run-time.

Actually now that I think about it, I could do the same for color-themes, so that people can link a css / sass file to be included. Therefore I would need to use a rust solution > sass-rs :yum:

Issue #81, opened 2015-12-29
Bind right and left arrow keys to next and previous pages


mdinger:

azerupi:
This would be a nice addition :)

azerupi:
#87 Closes this

PR #85, opened 2015-12-30
Add navigation by keyboard using alt + left/right arrows

funkill:
Add navigation to next and previous pages by 'alt + &rarr;' and 'alt + &larr;' combinations. Issue #81 with modification (add alt key)

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -23,6 +23,23 @@ $( document ).ready(function() {
         hljs.highlightBlock(block);
     });
 
+    var KEY_CODES = {
+        PREVIOUS_KEY: 37,
+        NEXT_KEY: 39
+    };
+
+    $(document).on('keydown', function (e) {
+        if (e.altKey === true) {
+            switch (e.keyCode) {
+                case KEY_CODES.NEXT_KEY:
+                    window.location.href = $('.nav-chapters.next').attr('href');
+                    break;
+                case KEY_CODES.PREV_KEY:
+                    window.location.href = $('.nav-chapters.previous').attr('href');
+                    break;
+            }
+        }
+    });
 
     // Interesting DOM Elements
     var html = $("html");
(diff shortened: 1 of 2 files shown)


azerupi:
I know that sometimes people have trouble with non-english keyboards when using hardcoded keycodes. I don't expect `alt` and `arrows` to cause trouble though. Have you tested it with different keyboards?

I have an Azerty keyboard, so I will let you know if it works fine here.

Thanks a lot for the Pull Request! :smile:

azerupi:
Sorry, I merged too quickly. This causes some trouble with the `Go back` and `Go forward` page navigation in Google Chrome.

funkill:
I tested in russian keyboard, all works. 
About back/forward navigation: now i tests in firefox, he has same behavior. I often use `backspace` for previous page and dont known about `alt-arrow` combinations :smile: 
I think if use only one key, we have some trouble with errors key press. What are you think about possible combinations?

azerupi:
Yeah arrows could be pressed by mistake, but I don't think it would be that problematic.

On Chrome, both `ctrl` and `shift` combined with the arrow keys don't seem to do anything. Shift seems to me like the safest modifier key, but it's not very intuitive. `ctrl` would have to be tested on mac, to see if it stays `ctrl` as they use `command`. 

I would just bind to the arrow keys without modifier, if we get negative feedback it can easily be changed later. What do you think?

funkill:
Combination of `ctrl` and `shift` can be used for change keyboard layout (e.g. on Windows use `shift+alt` for change english and other layouts, some people change it combination to `ctrl+shift`).
Only arrow: why not?

funkill:
Are need some help message in title for shortcuts (e.g. For see next page, you can use `->` key) or only character?

azerupi:
Yeah I will add a help icon on the menu-bar in the future to spawn a help popup. If you want you can add a tooltip on the navigation arrows for now.

PR #87, opened 2015-12-30
add navigation by arrows

funkill:


azerupi:
Thanks! :)

src/theme/book.js
```diff
@@ -23,6 +23,23 @@ $( document ).ready(function() {
...
+    var KEY_CODES = {
+        PREVIOUS_KEY: 37,
+        NEXT_KEY: 39
+    };
+
+    $(document).on('keydown', function (e) {
+        switch (e.keyCode) {
+            case KEY_CODES.NEXT_KEY:
+                e.preventDefault();
+                window.location.href = $('.nav-chapters.next').attr('href');
```
azerupi: I missed it, but you have to check that it actually exists. Because when you are on the first page and then press left arrow you get `undefined` and that page does not exist. I will fix it now ;)

funkill: Oh, sorry. Thank you!

Issue #88, opened 2015-12-30
Support ebooks and pdf export


mdinger:
Gitbook supports export to [ebooks and pdfs](https://github.com/GitbookIO/gitbook#output-formats) via [calibre](http://calibre-ebook.com/). This might be easy to hook into.

See also https://github.com/rust-lang/rust-by-example/issues/684 for problems this implementation creates for rustbyexample.

azerupi:
I would like to support pdf and ebook format. I think this could already be developed out of tree if you use the Renderer trait from mdBook. 

I am not sure I want to depend on a full blown Gui tool though. There must surely be a better alternative for that.

mdinger:
Not familiar with many tools like this but [pandoc](http://pandoc.org/) also seems like a plausible option.

azerupi:
Yeah pandoc seems a lot better!

asolove:
Did some exploration on this and seems doable. [Here's the default epub version of the Rust book](https://github.com/azerupi/mdBook/files/85834/sample.epub.zip). Note the chapters out of order and links not working.

To get good output, I think we would need to:
- parse the ToC to get the list of md files, in the right order
- concat and transform the markdown files, replacing file links with internal links
- match the themes with epub versions of the styles

I'm interested in working on this but will be a bit slow.

Useful info here: [Pandoc commands and styling options](http://pandoc.org/epub.html)

killercup:
> - parse the ToC to get the list of md files, in the right order
> - concat and transform the markdown files, replacing file links with internal links

@asolove, I have implemented this (among other transformations) in <https://github.com/killercup/trpl-ebook>, feel free to use my code.

asolove:
@killercup great, thanks!

azerupi:
Great! Thanks for doing this :)

>parse the ToC to get the list of md files, in the right order

This is already done in the Rust code, the `MDBook` struct can be iterated on. If you make a new `Renderer` you have access to that.

>concat and transform the markdown files, replacing file links with internal links

Concatenating the markdown files is also not that hard, I do it for the print page.

Replacing the links could be a little trickier, what should internal links look like for pandoc? I know that pulldown-cmark gives you the ability to transform the parsed markdown events before rendering, but it's not well documented. Maybe link replacing is in it's capabilities. 

Static files, like images, will probably also need some special treatment to be included correctly?

-----

>I'm interested in working on this but will be a bit slow.

[...]

PR #94, opened 2016-01-12
WIP on ePub

asolove:
:warning: Not at all ready, just putting this here for review as I work through #88.

I think a minimal useful version of this would:

- [ ] get the ordered list of files from the `MDBook` and concat them
- [ ] parse the combination of files in pulldown-cmark
- [ ] transform the AST adding anchors to headings
- [ ] transform the AST mapping external links to internal anchors
- [ ] match the themes with epub versions of the styles

### DIFF
--- a/src/book/mdbook.rs
+++ b/src/book/mdbook.rs
@@ -8,7 +8,7 @@ use std::process::Command;
 
 use {BookConfig, BookItem, theme, parse, utils};
 use book::BookItems;
-use renderer::{Renderer, HtmlHandlebars};
+use renderer::{Renderer, HtmlHandlebars, Pandoc};
 
 
 pub struct MDBook {
@@ -38,7 +38,7 @@ impl MDBook {
                         .set_src(&root.join("src"))
                         .set_dest(&root.join("book"))
                         .to_owned(),
-            renderer: Box::new(HtmlHandlebars::new()),
+            renderer: Box::new(Pandoc::new()),
         }
     }
 
--- a/src/renderer/pandoc/mod.rs
+++ b/src/renderer/pandoc/mod.rs
@@ -3,34 +3,40 @@ use book::MDBook;
 use book::bookitem::BookItem;
 
 use std::error::Error;
+use std::process::Command;
 
-enum PandocOutput {
-    Epub,
-    Pdf
-}
-
-pub struct Pandoc {
-    output: PandocOutput
-}
+pub struct Pandoc;
 
 impl Pandoc {
   pub fn new() -> Pandoc {
-    Pandoc { output: PandocOutput::Epub }
+    Pandoc
   }
 }
 
 impl Renderer for Pandoc {
     fn render(&self, book: &MDBook) -> Result<(), Box<Error>> {
+        let mut paths = vec!();
 
         for item in book.iter() {
             match *item {
-                BookItem::Chapter(ref title, _) => {
-                    println!("{}", title)
+                BookItem::Chapter(_, ref ch) => {
+                    paths.push(book.get_src().join(&ch.path).into_os_string());
                 },
-                _ => println!("Something I don't understand")
+                _ => println!("FIXME: don't understand this kind of BookItem")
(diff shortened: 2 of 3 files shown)


src/book/mdbook.rs
```diff
@@ -38,7 +38,7 @@ impl MDBook {
                         .set_src(&root.join("src"))
                         .set_dest(&root.join("book"))
                         .to_owned(),
-            renderer: Box::new(HtmlHandlebars::new()),
+            renderer: Box::new(Pandoc::new()),
```
asolove: This is just temporary to make local testing easier, but curious how we see this playing out long term. Should the mdbook command line support a new argument? Default to using all the available renderers?

src/renderer/pandoc/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+            .arg("-osample.epub")
+            .args(&paths)
+            .output();
+
+        match output {
+            Ok(_) => Ok(()),
+            Err(e) => Err(Box::new(e))
+        }
+        // FIXME: why doesn't this work
+        // output.map(|_| ()).map_err(|e| Box::new(e))
```
asolove: I don't understand why the output of `match` and `map`/`map_err` here are different.

In the `map`/`map_err` version, the compiler complains:

```
 expected `core::result::Result<(), Box<std::error::Error + 'static>>`,
    found `core::result::Result<(), Box<std::io::error::Error>>`
```

But it seems to me like both the types and the lifetimes should be the same in both?

src/book/mdbook.rs
azerupi: In the long term I guess it would be nice if the user could choose a couple of formats in his `book.json` and on build all formats would be exported. This needs a little bit of thinking to come up with a good design:

- How to handle the automatic switching of renderers
- Where to export, currently the html version is exported in `book` dir but where would pdf, epub, ... go?

src/renderer/pandoc/mod.rs
azerupi: You don't need to be afraid to use the `match` statement, it is not considered "a last resort bazooka" that should be avoided when possible. Generally there are multiple ways to handle a certain pattern, in this case the match works. You could also have written:

```rust
if let Err(e) = output {
    return Err(Box::new(e))
}

Ok(())
```

Maybe there are other solutions, but it is generally a matter of personal preference :)

mkpankov:
@asolove @azerupi Hi!

We over at @ruRust/mdbook plan to implement EPUB, MOBI and PDF output so we wanted to know if this PR is being worked on.

Anyway, we should probably discuss the specifics of configuration and choice of renderers in advance. @azerupi do you hang out on Mozilla's IRC (#rust)?

azerupi:
>do you hang out on Mozilla's IRC (#rust)?

Not often, but if we agree on some date / time I can come online ;)

>We over at @ruRust (@ruRust/mdbook) plan to implement EPUB, MOBI and PDF output so we wanted to know if this PR is being worked on.

That would be awesome! 

I have been wanting to rewrite how mdBook works internally so that it would be easier to implement / enable / disable different renderers, themes and plug-ins for extra functionality. Because currently you can only have one renderer and this rendererer is not completely isolated into its own module to just swap it out. I have experimented with some designs but didn't get far.

The problem is that I feel like it would be a shame to put a lot of effort in writing the EPUB / PDF / MOBI renderer now and have to rewrite it when that new API is implemented. On the other hand I don't want to block new features as I have nothing concrete yet and it will probably take more time than I expect.

Maybe we could work together on that rewrite, we could discuss ideas about the new API. You could concentrate more on implementing the renderer and I could get useful feedback about how bad my ideas are. :wink: It will probably take more time, but I think it could be beneficial in the end and I would feel less guilty of making you do unnecessary work! 

Now that you know that, I leave the choice up to you. Don't feel obligated in any way! :)

azerupi:
I am going to close this, as it has been inactive for some time

Issue #102, opened 2016-02-08
Table generation is broken


JIghtuse:
I tried to generate a book with a simple table from TRPL. Here is a complete file:

    # Chapter 1

    Something something

    |  Target                       | std |rustc|cargo| notes                      |
    |-------------------------------|-----|-----|-----|----------------------------|
    | `x86_64-unknown-linux-musl`   |  ✓  |     |     | 64-bit Linux with MUSL     |
    | `arm-linux-androideabi`       |  ✓  |     |     | ARM Android                |
    | `arm-unknown-linux-gnueabi`   |  ✓  |  ✓  |     | ARM Linux (2.6.18+)        |
    | `arm-unknown-linux-gnueabihf` |  ✓  |  ✓  |     | ARM Linux (2.6.18+)        |
    | `aarch64-unknown-linux-gnu`   |  ✓  |     |     | ARM64 Linux (2.6.18+)      |
    | `mips-unknown-linux-gnu`      |  ✓  |     |     | MIPS Linux (2.6.18+)       |
    | `mipsel-unknown-linux-gnu`    |  ✓  |     |     | MIPS (LE) Linux (2.6.18+)  |

And here is output:

    thread '<main>' panicked at 'index 232 and/or 781 in `<complete file contents>` do not lie on character boundary', ../src/libcore/str/mod.rs:1598
    note: Run with `RUST_BACKTRACE=1` for a backtrace.

So I decided that tables can't have unicode (non-ascii?) characters and replaced `✓` with `+`. Now mdbook completes generation, but tables are incorrect. rustbook generates correct output from the same source (with minor change in title `#` => `%`).

rustbook output:
http://i.imgur.com/JA0Pd1F.png

mdbook output:
http://i.imgur.com/KGyC9lx.png

JIghtuse:
Seems like a pulldown-cmark problem.

    $ pulldown-cmark --enable-tables

    # Chapter 1

    Something something

    |  Target                       | std |rustc|cargo| notes                      |
    |-------------------------------|-----|-----|-----|----------------------------|
    | `x86_64-unknown-linux-musl`   |  ✓  |     |     | 64-bit Linux with MUSL     |
    | `arm-linux-androideabi`       |  ✓  |     |     | ARM Android                |
    | `arm-unknown-linux-gnueabi`   |  ✓  |  ✓  |     | ARM Linux (2.6.18+)        |
[...]

PR #106, opened 2016-02-15
pulldown-mark version bump

funkill:
Pulldown-mark with fixed table support: #102 and [google/pulldown-cmark#16](https://github.com/google/pulldown-cmark#16)


azerupi:
Thanks! I will release a new version of mdBook to crates.io with the update

PR #108, opened 2016-02-16
add rounding for first and last items in theme selector

funkill:


azerupi:
What does this change for you? I tried to apply this in browser (firefox) and it doesn't seem to change much for me.

funkill:
it's not critical and changes sees only in very big zoom: hovered area has rectangle figure without rounded angles, but theme selector's angles is rounded
in picture: up - in master, down - in this commit
theme selector

azerupi:
Ah yes, I wasn't zoomed in enough to see it clearly. It's good, I will
merge when I come home. Thanks! :)

PR #109, opened 2016-02-22
Add theme support for favicon

jessestricker:
~~This PR adds support for an optional config option:~~
~~`"favicon": "path/image.png"`~~
~~Users can specify their own favicon and mdBook will copy it and set up the html for it.~~

This PR adds support for indicating a custom favicon.png via the theme folder.


azerupi:
This is a good idea :+1:

jessestricker:
So, what about a default image for the default theme?
The one in `book-example` was made by me, you could use that one.

azerupi:
That's a gorgeous logo!

Just one little nitpick, the "D" does not seem as wide as the M, would it be possible to make them the same width?
image3398

It's because when you look at it in 16x16px the "D" seems a little more fade than the "M". If it's too much work, don't bother! It's only a detail and I am not even sure it will make it better.

jessestricker:
Oops, I didn't even notice. This should be better.

mdbook-logo

azerupi:
That's perfect! :)

jessestricker:
So, this could be merged, if you are d'accord.

azerupi:
Did you not want to add it in the default theme?

jessestricker:
I was asking you :smile:. I will add it.
Should the favicon be overidable by a theme?

azerupi:
>I was asking you

I implicitly agreed, I should have been more explicit :wink:

>Should the favicon be overidable by a theme?

In my opinion, there shouldn't be two ways to indicate a custom favicon. So either through the config file or via the theme directory. I would be personally more inclined to override the default favicon by just putting a file named `favicon.*` in the theme directory. What do you think?

jessestricker:
> What do you think?

Yes, I agree, that is better. :)

azerupi:
Ok, let's do that! Sorry for making you change everything :)

jessestricker:
No problem, that's what PRs are for.

jessestricker:
This PR is now finished.

azerupi:
Thanks a lot!

jessestricker:
You're welcome, don't forget to add this to the [documentation](http://azerupi.github.io/mdBook/format/theme/theme.html)!

azerupi:
Right! thanks for the reminder

pickfire:
The favicon seems unreadable in dark theme. 
2018-02-20-120123_20x20_scrot

Michael-F-Bryan:
@sorin-davidoi, do you know what may have caused this?

sorin-davidoi:
This is not a bug in `mdBook`, but rather a consequence of using a dark theme. Browsers display favicons as they are, even if it is black (or grey) on a black background.

This is how the GitHub favicon looks when using the dark there in Firefox Nightly.
image

pickfire:
@sorin-davidoi Is it possible to have a color that displays well in both dark and light theme at this time of writting (I bet switching colors won't be that easy)? As well, we can think for future whether there will be grey theme (possibly not).

sorin-davidoi:
The Bugzilla logo works nice with light and dark themes, so I guess a more vivid color would work better. That being said, maybe we should open bugs report on projects that use `mdBook` (e.g. https://rustbyexample.com/) which don't provide their own favicon, since the default one doesn't make sense for them anyway.

PR #111, opened 2016-02-23
Generate simple .gitignore on init

vrinek:
Handles: https://github.com/azerupi/mdBook/issues/110

Generates a simple `.gitignore` file inside the book's destination folder with the following contents:

```
# Ignore everything
*

# Except this file
!.gitignore
```

It is intended to instruct git to ignore the auto-generated destination folder's contents.

### DIFF
--- a/src/bin/mdbook.rs
+++ b/src/bin/mdbook.rs
@@ -94,7 +94,7 @@ fn init(args: &ArgMatches) -> Result<(), Box<Error>> {
     // If flag `--theme` is present, copy theme to src
     if args.is_present("theme") {
 
-        // Skip this id `--force` is present
+        // Skip this if `--force` is present
         if !args.is_present("force") {
             // Print warning
             print!("\nCopying the default theme to {:?}", book.get_src());
@@ -113,6 +113,22 @@ fn init(args: &ArgMatches) -> Result<(), Box<Error>> {
         try!(book.copy_theme());
         println!("\nTheme copied.");
 
+    }
+
+    // Because of `src/book/mdbook.rs#L37-L39`, the following will always evaluate to `true`
+    let is_dest_inside_root = book.get_dest().starts_with(book.get_root());
+
+    if !args.is_present("force") && is_dest_inside_root {
+        let gitignore = book.get_dest().join(".gitignore");
+        println!("\nCreating default .gitignore at {:?}", gitignore);
+        print!("\nAre you sure you want to continue? (y/n) ");
+
+        if confirm() {
+            book.create_gitignore();
+            println!("\n.gitignore created.");
+        } else {
+            println!("\nSkipping...\n");
+        }
     }
 
     println!("\nAll done, no errors...");
--- a/src/book/mdbook.rs
+++ b/src/book/mdbook.rs
@@ -101,32 +101,6 @@ impl MDBook {
         }
 
         {
-            let root = self.config.get_root();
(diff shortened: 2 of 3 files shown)


vrinek:
Travis can only build on nightly because of the unstable feature.

azerupi:
First of all, thanks for doing this!

There are a couple of issues with this though:

1. You can't assume every book is going to be under version control, this should definitely be an opt-in behavior (activated through a command line flag)

2. What if the output destination is somewhere outside the git repo? For example:

    ```
    src:   /home/azerupi/Programming/Rust/mdBook/book-example/src/
    dest:  /home/azerupi/Documents/book/
    ```
   And that is assuming that the root of the book is the root of the git repository. A book could be a small sub-part of a bigger project.

3. I can't accept patches that don't build on stable. I test on all channels, but I target mostly stable. Unfortunately that means I can't use unstable features and until this feature gets stabilized a workaround should be used.

All in all, I think this requires a bit of work for very little gain. For now the only thing that has to be put in a `.gitignore` is the output path and it has to be done only once. Doing it manually is not that big of a burden :wink: 

In the future (after the rewrite) you will be able to set individual output paths for different renderers. Maybe it will be more useful then.

What do you think? If you still want to work on this I would be happy to discuss it further to solve the remaining issues and help you with the implementation.

vrinek:
> I can't accept patches that don't build on stable.

I agree. (I'll add it to the README :wink:)

> A book could be a small sub-part of a bigger project.

Git can handle multiple `.gitignore` files within its hierarchy (http://git-scm.com/docs/gitignore).

I don't mind putting the extra effort to get this to work because my 1st priority this period is to learn Rust. I'm also using mdbook for my own project's documentation under version control and I assume more would like to do that too as it is described as _"Like Gitbook but implemented in Rust"_.

---

One course of action would be to 1. check for git, 2. make sure the `dest` dir is under the `root` dir. If any of the above is false, do not create a `.gitignore`.

Alternatively, we could check for git and display a helpful message like this:

> If you intend on using version control for this book, I suggest you ignore the `book` directory because that's where the auto-generated assets are stored.

What do you think?

azerupi:
>Git can handle multiple .gitignore files within its hierarchy

Good to know!

>One course of action would be to 1. check for git, 2. make sure the dest dir is under the root dir. If any of the above is false, do not create a `.gitignore`.

That is one possibility. I think the option I like the most is:

When running `mdbook init` prompt the user if he wants to create a gitignore (y/n), this prompt has to be skipped if the `--force` flag is passed. The default behavior is to not create a gitignore.
If the destination folder is outside the root folder skip the prompt.

With this strategy you don't have to check for a git folder, the user can choose for himself if he needs it or not.

Does that sound like a good plan to you?

vrinek:
:+1:

vrinek:
Updated PR description to reflect code changes.

azerupi:
Thanks for this! :)
I have changed the printed output slightly

vrinek:
Thanks! :smile:

vrinek:
Whoops! Looks like putting the `.gitignore` inside the dest folder was not such a good idea after all. `mdbook build` removes it on every build.

I'll work on a fix.

azerupi:
Ow yes, I forgot about that. The content of the destination folder is removed before every render to avoid having ghost files.

I would put the .gitignore in the root folder and add a line with the destination path in it. That should do the job.

PR #112, opened 2016-02-25
Add description config option

jessestricker:
There seems to be a bug in the index.hbs template.

```
<meta name="description" content="{% block description %}{% endblock %}">
```

I fixed this, adding a `"description": "..."` option to `book.json`.


azerupi:
Ah yes, I think I think I forgot to implement this. In the rewrite (v0.1.0) there are / will be more fields in the config file.

Thanks for implementing this! :)

PR #115, opened 2016-03-07
Move `.gitignore` directly under the root folder

vrinek:
Continuation of https://github.com/azerupi/mdBook/pull/111 after [realising a bug in the logic of the implementation](https://github.com/azerupi/mdBook/pull/111#issuecomment-191642716).

---

Generates a `.gitignore` file directly under the `root` folder instead of under the `dest` folder.

Makes use of the newly stabilised ([as of 1.7.0](https://github.com/rust-lang/rust/blob/stable/RELEASES.md#version-170-2016-03-03) :smile:) [Path.strip_prefix](http://doc.rust-lang.org/nightly/std/path/struct.Path.html#method.strip_prefix) method.

### DIFF
--- a/src/bin/mdbook.rs
+++ b/src/bin/mdbook.rs
@@ -115,7 +115,7 @@ fn init(args: &ArgMatches) -> Result<(), Box<Error>> {
 
     }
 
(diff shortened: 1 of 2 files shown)


azerupi:
Looks good, thank you :)

PR #118, opened 2016-03-07
Travis CI: expand to test and deploy for Linux and Mac

japaric:
This one is more complex. I hope I got it right on the first try. *crosses fingers*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,10 @@ env:
 matrix:
   include:
     # Stable channel
-    - os: osx
-      env: TARGET=i686-apple-darwin
-      rust: stable
+    # TODO re-enable the i686-apple-darwin target when passcod/rsnotify#53 is fixed
+    #- os: osx
+      #env: TARGET=i686-apple-darwin
+      #rust: stable
     - os: linux
       env: TARGET=i686-unknown-linux-gnu
       rust: stable
@@ -34,9 +35,9 @@ matrix:
       env: TARGET=x86_64-unknown-linux-musl
       rust: stable
     # Beta channel
-    - os: osx
-      env: TARGET=i686-apple-darwin
-      rust: beta
+    #- os: osx
+      #env: TARGET=i686-apple-darwin
+      #rust: beta
     - os: linux
       env: TARGET=i686-unknown-linux-gnu
       rust: beta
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,20 +1,106 @@
...
 install:
-    - npm install stylus nib
+  - sh ci/install.sh
+
+script:
+  - sh ci/script.sh
+
 after_success:
-  - test $TRAVIS_PULL_REQUEST == "false" && test $TRAVIS_BRANCH == "master" && bash deploy.sh
+  - test $TRAVIS_PULL_REQUEST == "false" &&
```
japaric: I hope this right, this is meant to:

- run on a single build job, and
- the `npm install stylus nib` is only needed for this deploy script

?

deploy.sh
```diff
@@ -15,13 +15,13 @@ cargo doc
 
 echo -e "${CYAN}Running mdbook build${NC}"
 # Run mdbook to generate the book
-target/debug/mdbook build book-example/
+target/$TARGET/debug/mdbook build book-example/
```
japaric: note: we need target/$TARGET instead of just target in all these instances because mdbook was compiled with `cargo build --target $TARGET` instead of just `cargo build`

japaric:
Argh, about the Travis errors. doctests are known to not work correctly when `--target` is used with `cargo test`. Let me think about how to workaround this problem.

japaric:
Fixed the doc test problem, but we now have a legit failure. The notify crate can't be correctly compiled for i686-apple-darwin because its fsevent{,-sys} dependencies aren't compiled for that target (see [1], note that i686-apple-darwin is missing). I propose not targeting i686-apple-darwin until that is sorted out.

[1]: https://github.com/passcod/rsnotify/blob/b2c32e1b4efd38e411f9f663c4543c51318a694a/Cargo.toml#L56-L58

japaric:
All Travis CI tests are green. @azerupi r?

azerupi:
>I propose not targeting i686-apple-darwin until that is sorted out.

I suppose that is acceptable, 32-bits Macs are quite old.. There shouldn't be too much of them (I think) :yum: 

I will merge when I get home

azerupi:
Merged, I will push a new release to test it :)

PR #121, opened 2016-03-19
Update dependencies

Bobo1239:


azerupi:
I was about to do the same! Thanks for that :)

PR #122, opened 2016-03-19
Fix watch event cooldown

Bobo1239:
This also fixes

    Error while building: Error { repr: Os { code: 2, message: "No such file or directory" } }

as that was caused by several file events happening directly after each other (e.g. swap file created by editor) and trying to clean the target directory simultaneously.


azerupi:
Thanks a lot for this!

PR #124, opened 2016-03-23
Fix for insecure content on HTTPS enabled sites

LucioFranco:
Chrome and Firefox will not allow `http` content to be loaded when the site is accessed from `https`. This PR just moves the font, bootstrap, and jQuery to `https`. All the links support `https` and I have tested them.

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -10,12 +10,12 @@
         <base href="{{ path_to_root }}">
 
         <link rel="stylesheet" href="book.css">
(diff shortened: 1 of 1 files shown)


ebkalderon:
After testing on a local version of [www.amethyst.rs](https://www.amethyst.rs/), I can confirm that the fonts still load fine on HTTP and should fix the issues with HTTPS if merged.

screenshot from 2016-03-23 16-28-57

azerupi:
I wasn't even aware this was an issue :wink: 

Thanks for the fix!

PR #130, opened 2016-04-13
travis: use rustup instead of Travis built-in Rust support

japaric:
this ensures we install the correct set of standard crates when working
in the beta channel


azerupi:
Thanks!

Issue #136, opened 2016-05-08
`cargo build --no-default-features` errors


japaric:
with:

```
$ cargo build --no-default-features
   Compiling mdbook v0.0.12 (file:///home/japaric/tmp/mdbook)
error: structure `book::MDBook` has no field named `livereload`
  --> src/book/mod.rs:60:13
60 |>             livereload: None,
   |>             ^^^^^^^^^^

error: attempted access of field `livereload` on type `&mut book::MDBook`, but no field with that name was found
   --> src/book/mod.rs:431:9
431 |>         self.livereload = Some(livereload);
    |>         ^^^^^^^^^^^^^^^

error: attempted access of field `livereload` on type `&mut book::MDBook`, but no field with that name was found
   --> src/book/mod.rs:436:9
436 |>         self.livereload = None;
    |>         ^^^^^^^^^^^^^^^

error: attempted access of field `livereload` on type `&book::MDBook`, but no field with that name was found
   --> src/book/mod.rs:441:15
441 |>         match self.livereload {
    |>               ^^^^^^^^^^^^^^^

error: aborting due to 4 previous errors
error: Could not compile `mdbook`.

To learn more, run the command again with --verbose.
```

### Meta

```
$ git rev-parse HEAD
cfcf6d952fd7c5b08ba030ba93a8a77d5039ac10

$ rustc -V
rustc 1.10.0-nightly (62e2b2fb7 2016-05-06)

$ cargo -V
cargo 0.11.0-nightly (3a7a620 2016-05-04)
```

PR #137, opened 2016-05-08
Fix no-default-features build

Bobo1239:
I believe it's currently not possible to do leave out the livereload field as it's used in other parts of the code where we can't utilize conditional compilation without ugly hacks.

Fixes: #136 

related: https://github.com/rust-lang/rust/issues/15701


azerupi:
Thanks!
I should probably test this with CI but there are so much targets / combinations to test it takes forever.. And my Travis-fu is limited :confused:

PR #139, opened 2016-05-08
ci: test without default features

### LINKED ISSUE #138: Add CI test for compilation with no-defaults-features
As I never build with `no-defaults-features` it's easy to break. This should probably be tested with Travis. Ideally only for one platform to avoid doubling the time it takes to run the tests for all platforms...

japaric:
closes #138 

I think this should do it


azerupi:
Thank you so much! :)

PR #142, opened 2016-06-10
Add print media query

Bobo1239:
Before (only one page; content is cut off):
chrome_2016-06-10_19-32-41
After:
chrome_2016-06-10_19-33-41


azerupi:
I thought this worked before. But I might have only tested with the print button in the top right.

Thanks!

Issue #144, opened 2016-06-15
Links looks ugly when javascript is disabled


onur:
Javascript is disabled on my browser and links in a page generated by mdBook don't have any default style and they look so ugly. Example book without javascript:

[![book](https://i.imgur.com/0KPnbxal.png)](https://i.imgur.com/0KPnbxa.png)

This is not an issue on gitbook, and I believe it would be much better if mdBook uses `light` theme by default, and without any javascript.

And thanks for this awesome project!

azerupi:
Indeed, I noticed that part of the style is not applied when JS is disabled. I haven't had the time yet to really look into it, but I have an idea what might be causing this.

It's probably just the class of the theme that isn't set when JS is disabled, and since all the theme styling is applied only on that class, part of it is not applied. The fix is probably as easy as setting a default class in the handlebars template.

I will look into it when my exams are over if no one else fixes it until then :wink: 

>And thanks for this awesome project!

:blush: Thanks for the kind words

PR #145, opened 2016-06-15
Use light theme when javascript is disabled

onur:
mdBook is setting theme (by adding a class attribute to body tag) with javascript.
Page is not using any theme by default and page is not using any styling unless
javascript is enabled.

This patch is adding class attribute to body tag and making mdBook to use `light`
theme when javascript is disabled.

Fixes: #144

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -31,7 +31,7 @@
             }
         </script>
     </head>
-    <body>
+    <body class="light">
         <!-- Set the theme before any content is loaded, prevents flash -->
         <script type="text/javascript">
             var theme = localStorage.getItem('theme');


azerupi:
Hah thanks! :smile:

onur:
Awesome! That was a big issue for me, I thought it was a design choice and never reported the issue :no_mouth:

azerupi:
I encourage you to report things you think could be better. :blush: 

Specifically, fully supporting no-JS browsers is not an absolute goal. But if they can be supported without resorting to ugly hacks or added complexity, it's definitely something I want to do. And if it's too much trouble it should at least degrade gracefully when possible.

Issue #146, opened 2016-06-28
[Discussion] Book representation


azerupi:
Currently, the internal representation is very uni-language focused. We have one struct `MDBook` containing all the information and the chapters for the book.

To support multiple languages we have to store multiple books with different metadata. There should be one default language, the rest are assumed to be translations.

I propose that `MDBook` holds a hashmap with language codes like `"en"` or `"fr"` as keys and `Book`s as values.

``` rust
pub struct MDBook<'a> {
    books: HashMap<&'a str, Book>,
    // Other fields omitted 
}         
```

The `Book` structure would be defined as follows:

``` rust
pub struct Book {
    metadata: BookMetadata,

    preface: Vec<Chapter>,
    chapters: Vec<Chapter>,
    appendix: Vec<Chapter>,
}
```

`Chapter` would be 

``` rust
pub struct Chapter {
    title: String,
    file: path::PathBuf,

    sub_chapters: Vec<Chapter>,
}
```

And `BookMetadata` would be

``` rust
pub struct BookMetadata {
    pub title: String,
    pub description: String,

    pub language: Language,

    authors: Vec<Author>,
    translators: Vec<Author>,
}
```

And finally `Author` and `Language` would be

``` rust
pub struct Author {
    name: String,
    email: Option<String>,
}

pub struct Language {
    name: String,
    code: String,
}
```

This is what PR #147 implements. (not merged)

It allows to represent multiple languages, each with different book structures, authors / translators etc.
### Open questions
- Would there be a better way to represent the books?
- How do we represent the default language?
  Store the key (language code) for the default language in `MDBook`?

gambhiro:
This is a good start!

Based on books I produced in the past, I'd recommend to extend this a bit to recognize other common features of a book's structure.

See the structure of the Table of Contents in the EPUB and PDF of these books for example, to get an idea of some more challenging structures. I'd like to be able to reproduce these with `mdbook` (someday?).

- [Wolves, Dreams, Transformations](https://github.com/profound-labs/wolves-dreams/tree/master/publish/final)
- [Forest Path](https://github.com/profound-labs/forest-path/tree/master/publish/final)
- [Darma da Floresta (Portuguese)](https://github.com/profound-labs/darma-da-floresta/tree/master/publish/final)

In `Book`, I'd recommend different key names:

```
pub struct Book {
    metadata: BookMetadata,

    frontmatter: Vec<Chapter>,
    mainmatter: Vec<Chapter>,
[...]

PR #147, opened 2016-06-28
[WIP] New book struct

azerupi:
This PR refactors the internal representation of the book to allow for more and better metadata and leads the way to multi-language books.
#146

### DIFF
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -1,0 +1,77 @@
+use book::metadata::BookMetadata;
+use book::chapter::Chapter;
+
+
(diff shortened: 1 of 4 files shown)


steveklabnik:
@azerupi with https://github.com/azerupi/mdBook/pull/200 closed, should we close this too? Do you think this old PR is worth updating at all, or should we start again afresh?

azerupi:
This unfinished PR can serve as a good basis for future work. But it could be a good idea to start from a blank slate again as the changes in this PR are minimal and it may not be worth the trouble to update it to current master.

I think it's ok to close it and keep the branch as a reference for later.

Issue #148, opened 2016-06-28
Rewrite the parser for SUMMARY.md with nom


azerupi:
The parser for the `SUMMARY.md` file should be very simple but I feel like it is overly complicated. Maybe rewriting it with [nom](https://github.com/Geal/nom) can reduce some of the complexity?

I have experimented with nom but made very little progress, the macros give very poor error explanations when you mess up :confused: 

If someone can help me with this, feel free to chime in!

azerupi:
I'm going to abandon this idea. Nom is currently too difficult for me...

The parser needs to be rewritten to be more independent of the representation of the `MDBook` struct and this is blocking further progress. I am going to rewrite the custom parser and split it into multiple functions in order for it to be more manageable.

PR #150, opened 2016-06-28
[WIP] nom parser for SUMMARY.md

azerupi:
This implements the parsers for the `SUMMARY.md` file with nom

#148


azerupi:
The last commit has one test that fails, because I would like to allow nested brackets in link, e.g.

```
[This is an array[]](array.md)
```

And the tile would be `"This is an array[]"`. If any one want to tackle that last test, feel free to do so!

azerupi:
I'm going to abandon this idea.. Nom is currently too difficult for me and this is blocking all the rest. I will rewrite the custom parser and split it into multiple functions and structs to make it more maintainable.

PR #157, opened 2016-07-31
Fixes #156 - anchors are now URI encoded

### LINKED ISSUE #156: anchors with spaces
Hello. Clicking on anchors generate urls with spaces which seem unusable when trying to link to, for example here:
https://rust-lang.github.io/book/ch04-02-ownership.html#Variable binding scope

Is there a way to make the spaces be %20 or whatever is idiomatic in these cases?
Thanks.

icanrealizeum:
also fixes https://github.com/rust-lang/book/issues/166 anchors duplication

Thanks @azerupi for mentoring in #156 !
Cheers!

Note: `split("#")[0]` does return the full url if there are no `#` (anchor) to strip out - just making sure that case was handled :)

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -57,7 +57,9 @@ $( document ).ready(function() {
         var wrapper = $("<a class=\"header\">");
         wrapper.attr("name", $(this).text());
         // Add so that when you click the link actually shows up in the url bar...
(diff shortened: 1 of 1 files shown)


azerupi:
Thanks! This looks good :+1: 

I will try to push a new release to crates.io soon

PR #160, opened 2016-08-05
Added option to configure serve interface and public websocket address.

code-ape:
Made the following changes:

* Added option to specify interface for http and websocket server to listen on. `-i` flag.
* Added option to specify host for browser to connect to websocket server from. `-a` flag
* Added page to the `book-example` for serve command.

Let me know if anything is missing or needs to be done!

Cheers,
code-ape


azerupi:
Looks good, thanks! :)

Issue #151, opened 2016-07-08
Use the log crate instead of the home-made macros


azerupi:
The [log](https://github.com/rust-lang-nursery/log) crate provides macros like `debug!`, `info!`, `trace!`, etc. There is no need to recreate this in mdBook. 

Another advantage of using the `log` crate is that the final behavior of the logging macros is defined by the binary. If mdBook were to be used as a lib, the user could choose how the logging reports the information.

PR #164, opened 2016-08-14
use macros from the log crate, issue #151

gambhiro:
This wasn't too hard :)

Please comment if there is more to it.


src/book/mod.rs
```diff
@@ -42,7 +42,7 @@ impl MDBook {
     pub fn new(root: &Path) -> MDBook {
 
         if !root.exists() || !root.is_dir() {
-            output!("{:?} No directory with that name", root);
+            info!("{:?} No directory with that name", root);
```
azerupi: This one could be changed to `warn`

src/renderer/html_handlebars/helpers/playpen.rs
```diff
@@ -20,7 +20,7 @@ pub fn render_playpen(s: &str, path: &Path) -> String {
 
         // Check if the file exists
         if !playpen.rust_file.exists() || !playpen.rust_file.is_file() {
-            output!("[-] No file exists for {{{{#playpen }}}}\n    {}", playpen.rust_file.to_str().unwrap());
+            info!("[-] No file exists for {{{{#playpen }}}}\n    {}", playpen.rust_file.to_str().unwrap());
```
azerupi: `warn` here would also be better

azerupi:
Thanks for doing this!

One thing that is missing is the actual logger in the binary. We could use [env_logger/](http://rust-lang-nursery.github.io/log/env_logger/) (I'm not familiar with any others). You would have to add env_logger to the dependencies and only import it in the binary (`src/bin/mdbook.rs`), initialize it in the main and I think it would be best to set the default log level to info.

Once you have done that it looks good enough to merge :)

azerupi:
Looks good, thanks! :+1:

gambhiro:
It was pretty small, but thanks for the instructions. Good experience for the beginner like me.

PR #170, opened 2016-09-03
Make line-height for chapter greater than section

### LINKED ISSUE #166: Spacing in sidebar
selection_007

The chapters have a line height of 1.9em and the sections (and subsections) have a line height of 2.5em. This seems backward - I would expect chapters to be spaced at least as far apart as sections.

JIghtuse:
Fixes #166

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -72,7 +72,7 @@ table thead td {
 .chapter {
   list-style: none outside none;
   padding-left: 0;
-  line-height: 1.9em;
+  line-height: 2.2em;
 }
 .chapter li a {
   padding: 5px 0;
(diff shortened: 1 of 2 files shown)


JIghtuse:
@azerupi got it.
"regenerate-css" was failing on my machine due to missing `stylus`, so I've installed it and also some `nib` thing with npm. Now it builds fine, but throws away some stuff not related to line-heights. Should I amend this to include only relevant changes?

azerupi:
Sorry for the delay! 
The stuff that gets thrown away is probably a PR I accepted that modified the css instead of the stylus files. I'm going to copy that part and commit it properly.

Thanks!

PR #171, opened 2016-09-23
Copy files with no extension

rzlourenco:
I needed this feature recently (e.g. to have a README file copied automatically), so here's a patch, if it is useful to someone.


azerupi:
Good catch! Thanks a lot :smiley:

PR #173, opened 2016-10-21
slugify section headers

### LINKED ISSUE #168: Slugify section anchors
For a section like `## Hello World` mdBook currently creates an anchor like `page.html#Hello World`. The anchor should be [slugified](https://github.com/SamyPesse/github-slugid) to be more URL-friendly: `page.html#hello-world`.

HParker:
The current section headers are url encoded.  Because of that they
have some funny characters like %20.  We can clean that up by removing
all of the non-word characters before placing them in the anchor.

I originally planned to use `/\s/g` as the regex to preserve other url encoded characters like ‰ or ¥ that might be useful, but I double checked what Github does on gists and it looks like they strip everything.

hi ‰ there

becomes #hi--there in the url.

mdbook urls will now match that style.

fixes #168

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -55,11 +55,12 @@ $( document ).ready(function() {
     // Add anchors for all content headers
     content.find("h1, h2, h3, h4, h5").wrap(function(){
         var wrapper = $("<a class=\"header\">");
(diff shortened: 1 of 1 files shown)


azerupi:
Looks good! Thanks you :)

Issue #103, opened 2016-02-09
Add YAML syntax highlighting


ebkalderon:
It seems that with isagalaev/highlight.js#119, the Highlight.js library now supports syntax highlighting for YAML. However, when I annotate some YAML with ````yaml`, no syntax highlighting appears.

ebkalderon:
@azerupi Could you perhaps explain why this is happening? Does mdBook use an older version of Highlight.js for syntax highlighting?

azerupi:
Yes, I haven't updates highlight.js since August, I think.

I am rewriting mdBook to have a sort of plugin api and will extract the
code that manages the syntax highlighting into a more isolated plugin. When
that is done I will probably switch to a dependency manager like Bower to
update dependencies on a regular basis.

Also I have disabled auto-detect, because in my experience it doesn't
always work like it is supposed too. When it's a plugin I will probably add
an option in the config file to indicate a default language.

But.. at the moment I am on vacation ;)

ebkalderon:
@azerupi Sounds good. Until then, please relax and enjoy your vacation!

Luthaf:
YAML support is not activated by default in highlight.js code. As a temporary workaround, you could [download highlight.js](https://highlightjs.org/download/) (Custom package section) from there website, activating all the languages you need, and then putting the file in `src/theme/highlight.js`.

azerupi:
Yes, @ebkalderon you can do like @Luthaf suggests until I update the in tree version. You can read more about the process of overriding default theme files in [the documentation](http://azerupi.github.io/mdBook/format/theme/theme.html).

If you do, let me know if it fixes the issue

PR #174, opened 2016-10-28
Highlight js update

rnkaufman:
Updates highlight js to include new languages such as YAML, as requested in issue #103 

Selected languages:
- apache
- bash
- cs
- cpp
- css
- coffeescript
- diff
- xml
- http
- ini
- json
- java
- javascript
- makefile
- markdown
- nginx
- objectivec
- php
- perl
- python
- ruby
- sql
- ada
- arduino
- basic
- clojure
- clojure-repl
- django
- dockerfile
- dust
- erb
- elixir
- elm
- erlang
- erlang-repl    
- excel
- fsharp
- fortran
- go
- haml
- handlebars
- haskell
- x86asm
- julia
- less
- lisp
- livescript
- lua
- mipsasm
- mathematica
- matlab
- ocaml
- powershell
- puppet
- r
- roboconf
- rust
- scss
- scala
- smalltalk
- swift
- vbnet
- vbscript
- html
- verilog
- vim
- yaml


azerupi:
Thanks a lot for this!

Just to make sur, does the stylesheet needs to be updated as well or should the old one still be valid?

rnkaufman:
I think the old one should still be valid. This was tested out and it looked like the colors are still as expected based on the themes css file.

azerupi:
Great! Thanks :)

PR #175, opened 2016-11-01
Remove unused metadata import

DenisKolodin:
PR removes warning:

```rust
warning: unused import, #[warn(unused_imports)] on by default
 --> src/utils/fs.rs:4:21
  |
4 | use std::fs::{self, metadata, File};
  |                     ^^^^^^^^
```

It's really unused.


azerupi:
Thanks :)

PR #181, opened 2016-11-09
Use fixed positioning and remove overflow-x's for smoother scrolling …

### LINKED ISSUE #180: Bug: Incorrect scrolling behaviour on iOS 10.0.2/10.1.1 Safari
When reading mdBook on iOS 10.0.2/10.1.1 Safari, the address bar and the bottom bar (containing the back and forward arrows, as well as the share button, etc.) don't properly retract when scrolling down (swipe up).

Furthermore, if using a connected bluetooth keyboard, pressing the arrow keys either results in no scrolling at all, or it'll scroll about 1-2 lines at the maximum and stop. Scrolling up and back down doesn't provide more of a view, so you're stuck seeing the same part of the document. (This may be related to the top/bottom bars not retracting, as the amount scrolled seems about equal to the height of the top bar.) This probably generalises to any hardware keyboard.

If anyone has a higher iOS version (the most updated version is iOS 10.1.2, though this should not be a factor), could you be kind enough to test this scrolling behaviour on your phone? Thanks!

thomastanck:
…experience in iOS

Fixes #180 \o/

I've only tested this on Firefox 49.0.2 and iOS 10.0.2 Safari (iPhone 5). Please check that the intended behaviour is preserved on all supported platforms! At least for me, it seems to work perfectly (scrolling behaviour and element positioning) on both Firefox and iOS Safari on many different screen sizes.

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -37,7 +37,7 @@ table thead td {
   font-weight: 700;
 }
 .sidebar {
-  position: absolute;
+  position: fixed;
   left: 0;
   top: 0;
   bottom: 0;
@@ -99,7 +99,6 @@ table thead td {
 }
 .page-wrapper {
   position: absolute;
-  overflow-y: auto;
   left: 315px;
   right: 0;
   top: 0;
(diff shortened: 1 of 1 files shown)


azerupi:
Thanks! 
I will test this in the coming days. :)

thomastanck:
How are the tests going?

azerupi:
Sorry for the delay..
I have not seen any regressions, on the contrary, it solves a scrolling bug with the arrow keys on desktop too! :+1: 

Thanks a lot! :)

PR #182, opened 2016-11-14
Add hljs class to all code blocks, regardless of highlighting

### LINKED ISSUE #179: Possible to have code block with gray background, no syntax highlighting, and not run with rustc?
What I'm trying to do is get a `code` block that has the gray background that comes with:

```
.hljs {
   background: #f1f1f1;
}
```

(and the rest of the `.hljs` css too, probably) but *not* have any syntax highlighting applied, and also *not* have mdbook try and run the code block with `rustdoc --test`.

My use case is code blocks [like this](https://github.com/rust-lang/book/blame/master/src/ch09-01-unrecoverable-errors-with-panic.md#L34-L42) that are showing command line output:

```
$ cargo run
   Compiling panic v0.1.0 (file:///projects/panic)
    Finished debug [unoptimized + debuginfo] target(s) in 0.27 secs
     Running `target/debug/panic`
thread 'main' panicked at 'index out of bounds: the len is 3 but the index is
100', ../src/libcollections/vec.rs:1265
note: Run with `RUST_BACKTRACE=1` for a backtrace.
error: Process didn't exit successfully: `target/debug/panic` (exit code: 101)
```

I've tried:

- Setting the language as `bash`, which uses the gray background, but since it's meant for bash *scripts* not bash *command line interactions*, it [confusingly](https://github.com/rust-lang/book/pull/134#discussion_r81432171) will [highlight arbitrary words like "in"](http://rust-lang.github.io/book/ch09-01-unrecoverable-errors-with-panic.html).
- Setting the language as `text`, [which turns off highlighting](https://github.com/isagalaev/highlight.js/blob/master/src/highlight.js#L39), but that also means that block doesn't get the `hljs` class since it technically didn't go through highlight.js.
- Setting the language as something nonexistent like `blah`, which has the same effect as `text`.
- Not setting a language at all, which *does* get me the grey background and lack of syntax highlighting, but `mdbook test` tries to run the contents as rust :(
- Setting the language to just `ignore`, which makes `mdbook test` ignore it as desired but has the same display without the grey background as `text`, `blah`, etc

Do you know of any other workarounds that would result in the behavior I'm looking for?

Alternatively, would you accept a PR that adds this (or similar) to highlight.css, that matches the [words in highlight.js's regex to not do syntax highlighting](https://github.com/isagalaev/highlight.js/blob/master/src/highlight.js#L39)? 

```
.language-no-highlight, .language-nohighlight, .language-plain, .language-text {
  display: block;
  overflow-x: auto;
  background: #f1f1f1;
  color: #6e6b5e;
  padding: 0.5em;
-webkit-text-size-adjust: none;
}
```

carols10cents:
Fixes #179.

Highlight.js does not apply syntax highlighting to code blocks marked
no-highlight, nohighlight, plain, or text. When it finds blocks of those
languages, it does not add the `hljs` class to those code blocks either.

highlight.css and tomorrow-night.css use the `hljs` class to give code
blocks their backrgound color and text color, and we want that to apply
even if the code doesn't get syntax highlighting markup.

This is a somewhat hacky solution to get just that behavior! After this
commit, code blocks with no-highlight, nohighlight, plain, or text
language set on them will indeed get the hljs colors.

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -22,6 +22,10 @@ $( document ).ready(function() {
     $('code').each(function(i, block) {
         hljs.highlightBlock(block);
     });
+    
+    // Adding the hljs class gives code blocks the color css
+    // even if highlighting doesn't apply
+    $('code').addClass('hljs');
 
     var KEY_CODES = {
         PREVIOUS_KEY: 37,


carols10cents:
Tested this out with the rust book and it works marvelously! I also tried putting the `addClass` inside the loop, but that didn't add the class to the node, i think because `block` gets replaced with whatever highlight.js does to it? This does the job, though. I'm happy to make any modifications if there are better ways of doing this! Thank you!!! ❤️❤️❤️❤️

azerupi:
Thanks! :)

PR #188, opened 2016-12-20
Book struct

gambhiro:
This is taking the book-struct branch for #147, rebasing with updates from master, and adding the structs discussed in #146.

The only API-level change is calling the Book chapters frontmatter, mainmatter and backmatter.

The new structs and fields are just placeholders, there is no real effect at the moment.

It compiles, and builds the example book.

I see that it includes the 58 commits since July, I don't know if that's normal.


src/book/chapter.rs
```diff
@@ -1,24 +1,44 @@
...
+    file: PathBuf,
+
+    /// TODO The author of the chapter, or the book.
+    author: Author,
+    /// TODO The description of the chapter.
+    description: String,
+    /// TODO Index number of the chapter in its level. This is the Vec index + 1.
+    index: i32,
+    /// TODO CSS class that will be added to the page-level wrap div to allow customized chapter styles.
+    class: String,
```
azerupi: I am uncomfortable with the idea of storing a class attribute in the chapter struct. Because the notion of class is **very** renderer specific. Currently it would be OK, but what happens when we have more renderers? I think chapter should only contain information that is generic over the renderers.

We have already discussed the notion of chapter specific configurations, using YAML headers or another method. I think per-chapter styling would fit better in those configurations.

src/book/chapter.rs
```diff
@@ -1,24 +1,44 @@
...
-    file: path::PathBuf,
+    /// Path to chapter's markdown file.
+    file: PathBuf,
+
+    /// TODO The author of the chapter, or the book.
+    author: Author,
+    /// TODO The description of the chapter.
+    description: String,
+    /// TODO Index number of the chapter in its level. This is the Vec index + 1.
+    index: i32,
```
azerupi: What do you plan to do with this field? 
If this is just `Vec::len + 1`, what is the purpose of keeping it in the struct? By duplicating indices you add a synchronisation complexity. For example, if you add a chapter in the middle you need to update all the `index` fields.

src/book/chapter.rs
```diff
@@ -1,24 +1,44 @@
...
 #[derive(Debug, Clone)]
 pub struct Chapter {
+    /// The title of the chapter.
     title: String,
-    file: path::PathBuf,
+    /// Path to chapter's markdown file.
+    file: PathBuf,
+
+    /// TODO The author of the chapter, or the book.
+    author: Author,
```
azerupi: Maybe this should be an `Option`? Or a `Vec` because multiple authors are possible?
Can you maybe explain shortly how you see this author property interact with the author we store in the book?

src/book/chapter.rs
```diff
@@ -1,24 +1,44 @@
...
+    /// TODO The description of the chapter.
+    description: String,
```
azerupi: I think this should be an `Option` too.

src/book/metadata.rs
```diff
@@ -1,12 +1,32 @@
...
     pub language: Language,
 
     authors: Vec<Author>,
     translators: Vec<Author>,
+
+    /// TODO Chapter numbering scheme
+    number_format: NumberFormat,
+    /// TODO Section names for nested Vec<Chapter> structures, such as `[
+    /// "Part", "Chapter", "Section" ]`
+    section_names: Vec<String>,
```
azerupi: This is probably a little too much anticipated. I am not against a high level of customisation, but I want to nail the basics first. 

I think we should concentrate on the fundamental necessities at first. Make the renderer abstraction work, make the plugin abstraction work. And after that we can add fine grained customisations. Trying to do everything at once will make it so much more difficult.

gambhiro:
I reckon I better cancel this. I have been doing more work on it since, and it changes of course as one understands it better. I'm putting it all in a branch which I want to take as far as rendering multiple books (i.e. translations) in html and epub.

I will send a note when that's ready to see. It will probably make more sense to read and reason about than half-finished stages.

PR #189, opened 2016-12-22
refactoring docs

gambhiro:
This is what I'm planning to do. Basically just refactoring how the data moves and stored, so that they express the intention of the structs better and allow cleaner separation.

Eventually this can go to the example book to document the app internals.

It is probably easier to read by [opening the branch](https://github.com/gambhiro/mdBook/blob/docs/doc/doc.md) in the browser.


gambhiro:
Please keep in mind this is before actually doing the work. I started on it and for example BookMetadata can be just part of BookConfig, it's better for parsing and they seem to always go together.

PR #195, opened 2017-01-01
Various refactoring and cleanup

mbrubeck:
These changes factor out some common error-handling code, to make the rendering code shorter and (I hope) more readable.


src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -81,47 +81,27 @@ impl Renderer for HtmlHandlebars {
...
-                                data.insert("path".to_owned(), p.to_json());
-                            },
-                            None => {
-                                return Err(Box::new(io::Error::new(io::ErrorKind::Other,
-                                                                   "Could not convert path to str")))
-                            },
-                        }
-
-                        // Remove content from previous file and render content for this one
-                        data.remove("content");
```
azerupi: I assume the calls to `remove` here were superfluous, because `insert` updates existing keys?
I can't recall why I added them in the first place, but if it works the same without them :+1:

mbrubeck: > I assume the calls to `remove` here were superfluous, because `insert` updates existing keys?

Yes, that’s correct.

azerupi:
Thanks! :)

PR #198, opened 2017-01-02
Update watch command to use `notify` 3.0

mbrubeck:


azerupi:
Looks very nice!
Thank you :)

PR #199, opened 2017-01-12
Add --dest-dir option to build, watch and serve subcommands

petehayes102:
If you're building a book for multiple contexts, it is helpful to be able to override the output dir dynamically, rather than statically in config.

I didn't add this option to `mdbook init` because I assumed that if you wanted to set an output dir when you initialise a project, it's probably going to be a permanent thing. Thus you should be encouraged to do this via config.

Also, I couldn't help but correct a bit of spelling!


azerupi:
This is very useful, thanks! :)

> Also, I couldn't help but correct a bit of spelling!

I won't hold it against you :wink: 
I highly appreciate spelling corrections, I think they are as important as other bug and style fixes because in my opinion, correct spelling sends a message of care and attention to detail, which I find important :)

petehayes102:
No worries mate :) I totally agree about spelling too. If I get time I'll have a browse for any other typos.

PR #202, opened 2017-02-11
widen hit regions of menu buttons

paiv:
Wider hit regions are easier to pick with mouse.

see here


azerupi:
Indeed, thank you! :)

PR #207, opened 2017-02-16
Generate links at compile-time rather than use JS

### LINKED ISSUE #204: Generate anchors when building instead of using JavaScript
The Rust team uses a linkchecker tool, but since anchors are generated by JavaScript code, it can't check if those anchors are valid.

It would be a better solution to generate anchors when building the book. This would also be a step in the right direction for no-js people too.

rust-lang/rust#39856

steveklabnik:
Fixes #204 

/cc @frewsxcv

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ serde = "0.9"
 serde_json = "0.9"
 pulldown-cmark = "0.0.8"
 log = "0.3"
-env_logger = "0.3"
+env_logger = "0.4.0"
 toml = { version = "0.2", features = ["serde"] }
 open = "1.1"
 regex = "0.2.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ log = "0.3"
 env_logger = "0.3"
 toml = { version = "0.2", features = ["serde"] }
 open = "1.1"
-regex = "0.1.80"
+regex = "0.2.1"
 
 # Watch feature
 notify = { version = "3.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ log = "0.3"
 env_logger = "0.3"
 toml = { version = "0.2", features = ["serde"] }
 open = "1.1"
+regex = "0.1.80"
 
 # Watch feature
 notify = { version = "3.0", optional = true }
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -242,7 +242,7 @@ fn build_header_links(mut html: String) -> String {
             }).collect::<String>();
 
             format!("<a class=\"header\" href=\"#{id}\" name=\"{id}\"><{h}>{text}</{h}></a>", h=header, id=id, text=text)
-        });
+        }).into_owned();
     }
 
     html
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -3,7 +3,9 @@ use renderer::Renderer;
 use book::MDBook;
 use book::bookitem::BookItem;
 use {utils, theme};
+use regex::{Regex, Captures};
 
+use std::ascii::AsciiExt;
 use std::path::{Path, PathBuf};
 use std::fs::{self, File};
 use std::error::Error;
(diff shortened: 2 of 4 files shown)


steveklabnik:
oh wait i bet i can do this in one pass....

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -209,3 +214,37 @@ fn make_data(book: &MDBook) -> Result<serde_json::Map<String, serde_json::Value>
...
     Ok(data)
 }
+
+fn build_header_links(mut html: String) -> String {
+    for header in &["h1", "h2", "h3", "h4", "h5"] {
+        let regex = Regex::new(&format!("<{h}>(.*?)</{h}>", h=header)).unwrap();
+
+        html = regex.replace_all(&html, |caps: &Captures| {
+            let text = &caps[1];
+            let mut id = text.to_string();
```
steveklabnik: i am almost tempted to pull this shenanigans out into a microcrate so that rustdoc and mdbook can share it

steveklabnik:
regex doesn't support back-references, which is what's need to do this in one pass. so we can stick with this, or we can add a dependency on https://github.com/google/fancy-regex. I'm tempted to just stick with it.

steveklabnik:
actually https://docs.rs/regex/0.2.0/regex/struct.RegexSet.html might be able to do it in one pass...

steveklabnik:
Nevermind, I don't think so. This should be good to go.

Cargo.toml
```diff
@@ -24,6 +24,7 @@ log = "0.3"
 env_logger = "0.3"
 toml = { version = "0.2", features = ["serde"] }
 open = "1.1"
+regex = "0.1.80"
```
azerupi: The last version of regex seems to be `0.2.1` is there a reason for using this specific version?

steveklabnik: It was the version that was already in cargo.lock; I'd be happy to upgrade it too, but I was trying to not mess with dependencies 😄

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -209,3 +214,37 @@ fn make_data(book: &MDBook) -> Result<serde_json::Map<String, serde_json::Value>
...
+            let repl_sub = vec!["<em>", "</em>", "<code>", "</code>",
+                                "<strong>", "</strong>",
+                                "&lt;", "&gt;", "&amp;", "&#39;", "&quot;"];
```
azerupi: I don't think this list is exhaustive and future proof. Maybe it would be better to match against anything that could look like a tag? I am not sure how GitHub does it.

steveklabnik: i copied this code directly from rustdoc so that it'd have the same behavior; i agree that it's probably not exhaustive.

> I am not sure how GitHub does it.


I don't know either :/

Cargo.toml
azerupi: I think it's ok to go for the latest, it shouldn't cause any problems right?

src/renderer/html_handlebars/hbs_renderer.rs
azerupi: I guess it's good enough for now :)
We can improve it later if necessary.

Cargo.toml
steveklabnik: the only thing it might cause is two version of regex, given that iirc this is a dependency of something.

I could also just update all deps; https://github.com/azerupi/mdBook/pull/206 is doing some of that

azerupi: Updating deps is good, but maybe it's better if I first accept #206 then to avoid conflicts

steveklabnik: Sounds great!

azerupi: #206 is merged, you can go ahead and update the rest :)

steveklabnik:
Okay! I have updated all the deps required to have one version of regex; I did not update _all_ of the deps, but might send in a new PR later :smile:

azerupi:
Great! Once the little lights turn green it will be good to merge :wink:

steveklabnik:
@azerupi travis was taking a while, and in the meantime, I figured out how to make it one pass. Given that we know it always generates valid tags, we can just do this. LMK if you'd prefer the more explicit version with the extra passes.

azerupi:
Looks good to me, thanks! :)

steveklabnik:
@azerupi any chance we can get a crates.io release for this? We need it to ship https://github.com/rust-lang/rust/pull/39855

azerupi:
@steveklabnik Done :)

steveklabnik:
❤️

PR #212, opened 2017-02-24
Fix anchors #211

### LINKED ISSUE #211: <base href="../"> messes with anchor links
**Symptom:**

Go to http://azerupi.github.io/mdBook/cli/build.html, click on the title "The build command". You are redirected to the the introduction page. I'd expect this link to just be an anchor on the same page.

This is most likely due to the `<base href="../">` at the top of the page.

azerupi:
This should fix #211 

\cc @steveklabnik 

You can merge this, unless you have a cleaner way to solve this :)

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -95,14 +95,15 @@ impl Renderer for HtmlHandlebars {
                         debug!("[*]: Render template");
                         let rendered = try!(handlebars.render("index", &data));
 
+                        let filename = Path::new(&ch.path).with_extension("html");
+
                         // create links for headers
-                        let rendered = build_header_links(rendered);
+                        let rendered = build_header_links(rendered, filename.to_str().unwrap_or(""));
 
                         // fix code blocks
                         let rendered = fix_code_blocks(rendered);
 
                         // Write to file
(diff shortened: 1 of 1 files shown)


vberger:
#211 also manifest itself on footnotes which links on `<a href='#1'>...</a>` for example. I may be misunderstanding, but it looks like this PR only handles titles?

azerupi:
Ah, thanks for mentioning ! I will take a look at footnotes later today :)

steveklabnik:
so, https://github.com/azerupi/mdBook/pull/207/commits/d6099882640931b83ad3336097f1652b502c8bf3 was the JS code that was removed, and this was only for headers. My PR didn't touch footnotes, or at least, not that I'm aware.

The bug happened because the JS did `$(location).attr('href').split("#")[0] + "#" + header_name` and I missed the first part, I guess ☹️

steveklabnik:
Anyway, looks good to me, I think this is the cleanest way ❤️

azerupi:
This should now fix all anchors. It's not the most elegant solution, but it's functional.

steveklabnik:
It looks like travis is just broken, i might investigate tomorrow. All prs are failing the same way.

azerupi:
Yes, there is definitely something wrong with the Travis builds. :/

PR #213, opened 2017-02-27
Improve hidden code detection

steveklabnik:
This doesn't _fix_ https://github.com/azerupi/mdBook/issues/155, but it
does paper over it :wink:

What do you think about this, @azerupi ?

My understanding is that the new docs based on mdbook will be deployed tonight, and I'm expecting we'll get bugs filed about this, so it'd be nice to fix, even if it is just a hack for now. Open to other perspectives as well.


steveklabnik:
https://github.com/azerupi/mdBook/pull/214 is much better

PR #214, opened 2017-02-28
Fix code blocks with comma separated classes

### LINKED ISSUE #155: Multiple classes for code blocks
The rust book uses annotations for rustdoc to test code snippets, like the following:

``````
```rust,should_panic
fn main() {
    // Code here
}
```
``````

But when rendering, the language class ends-up being `language-rust,should_panic` instead of just `language-rust`. Stripping the part after the comma or adding spaces would help.

See https://github.com/rust-lang/book/issues/96

azerupi:
Fixes #155 

This PR iterates over all code blocks and replace all commas in the class attribute by spaces. This allows to specify multiple classes in code blocks. This is mostly for the Rust book that uses:

    ```rust,should_panic
    fn main() {
        // Code here
    }
    ```

@steveklabnik Is this better than #213 ?

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -94,9 +94,12 @@ impl Renderer for HtmlHandlebars {
                         // Render the handlebars template with the data
                         debug!("[*]: Render template");
                         let rendered = try!(handlebars.render("index", &data));
-                        
+
                         // create links for headers
                         let rendered = build_header_links(rendered);
+
+                        // fix code blocks
+                        let rendered = fix_code_blocks(rendered);
 
                         // Write to file
                         let filename = Path::new(&ch.path).with_extension("html");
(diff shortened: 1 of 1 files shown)


steveklabnik:
yes!

steveklabnik:
https://doc.rust-lang.org/nightly/book/

🎊 🎊 🎊 🎊 🎊 🎊 🎊 🎊 🎊 

Do you think we could get a release cut with this? https://doc.rust-lang.org/nightly/book/associated-types.html#trait-objects-with-associated-types is a thing now 😄

azerupi:
Yes, sure :)
I will take this opportunity to test out [cargo-release](https://github.com/sunng87/cargo-release) which should make it easier.

Also, I think it would be a good idea to add you to the crate maintainers on crates.io so that you can push releases too. I'm not sure how I should do that but I'm sure it is documented somewhere :)

**Edit:** Waoh! That was smooth :open_mouth: I am impressed by cargo-release

PR #215, opened 2017-02-28
Try to fix Travis

steveklabnik:
I don't if this will work or not, since CI has a lot going on; let's see.


azerupi:
That seems to do it! Thanks! :)

PR #222, opened 2017-03-06
Implement playpen support for ```rust

### LINKED ISSUE #29: Support embedding of rust playpen examples which are runnable
For example, just like http://rustbyexample.com/

steveklabnik:
Fixes #29 

So, this all works, and fixes an important regression in mdbook for the rust book.

However.

The big issue remaining is the "wrap in `main`" rules from rustdoc. In other words, something like this:

    ```rust
    let x = 5;
    ```

will fail, as it needs to be inside of `main`.

I'm not sure what the right way is to go here, exactly. I'm imagining duplicating the logic of rustdoc here makes sense.

Thoughts @azerupi @carols10cents @frewsxcv ?

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -307,7 +307,9 @@ fn add_playpen_pre(html: String) -> String {
                 format!("<pre class=\"playpen\">{}</pre>", text)
             } else {
                 // we need to inject our own main
-                format!("<pre class=\"playpen\"><code class=\"{}\">#fn main() {{
+                format!("<pre class=\"playpen\"><code class=\"{}\"># #![allow(unused_variables)]
+# 
+#fn main() {{
 {}
 #}}</code></pre>", classes, code)
             }
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -294,14 +294,23 @@ fn fix_code_blocks(html: String) -> String {
 }
 
 fn add_playpen_pre(html: String) -> String {
-    let regex = Regex::new(r##"((?s)<code[^>]?class="([^"]+)".*?>.*?</code>)"##).unwrap();
+    let regex = Regex::new(r##"((?s)<code[^>]?class="([^"]+)".*?>(.*?)</code>)"##).unwrap();
     regex.replace_all(&html, |caps: &Captures| {
         let text = &caps[1];
         let classes = &caps[2];
+        let code = &caps[3];
 
         if classes.contains("language-rust") && !classes.contains("ignore") {
             // wrap the contents in an external pre block
-            format!("<pre class=\"playpen\">{}</pre>", text)
+
+            if text.contains("fn main") {
+                format!("<pre class=\"playpen\">{}</pre>", text)
+            } else {
+                // we need to inject our own main
+                format!("<pre class=\"playpen\"><code class=\"{}\">#fn main() {{
+{}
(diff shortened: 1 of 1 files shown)


src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -292,3 +292,19 @@ fn fix_code_blocks(html: String) -> String {
...
     }).into_owned()
 }
+
+fn add_playpen_pre(html: String) -> String {
+    let regex = Regex::new(r##"((?s)<code[^>]?class="([^"]+)".*?>.*?</code>)"##).unwrap();
+    regex.replace_all(&html, |caps: &Captures| {
+        let text = &caps[1];
+        let classes = &caps[2];
+
+        if classes.contains("language-rust") && !classes.contains("ignore") {
```
carols10cents: what about should_panic? i can't decide whether not having a run button or having a run button that then causes a panic would be more confusing, just want to make sure we consider it :)

steveklabnik: this is a good question.

I'm not sure of the answer.

steveklabnik:
Here's the logic from rustdoc:

```rust
pub fn maketest(s: &str, cratename: Option<&str>, dont_insert_main: bool,
                opts: &TestOptions) -> String {
    let (crate_attrs, everything_else) = partition_source(s);

    let mut prog = String::new();

    // First push any outer attributes from the example, assuming they
    // are intended to be crate attributes.
    prog.push_str(&crate_attrs);

    // Next, any attributes for other aspects such as lints.
    for attr in &opts.attrs {
        prog.push_str(&format!("#![{}]\n", attr));
    }

    // Don't inject `extern crate std` because it's already injected by the
    // compiler.
    if !s.contains("extern crate") && !opts.no_crate_inject && cratename != Some("std
") {
        if let Some(cratename) = cratename {
            if s.contains(cratename) {
                prog.push_str(&format!("extern crate {};\n", cratename));
            }
        }
    }
    if dont_insert_main || s.contains("fn main") {
        prog.push_str(&everything_else);
    } else {
        prog.push_str("fn main() {\n");
        prog.push_str(&everything_else);
        prog = prog.trim().into();
        prog.push_str("\n}");
    }

    info!("final test program: {}", prog);

    prog
```

so, doing this too shouldn't be too hard.

steveklabnik:
I pushed up a commit to inject mains. The attribute stuff is a little harder, but not necessary.

One issue...

untitled

I have not verified how rustdoc does this.

steveklabnik:
Discussing ⬆️ with @carols10cents , we think auto-injecting `allow(unused_variables)` is probably the best route. I'll add another commit soonish.

steveklabnik:
Done. @azerupi how are you feeling about this PR?

azerupi:
It's good :+1:

Obviously, we can't continue to add special cases like this forever. But until we tackle the bigger architectural changes, it's fine to handle it this way :)

steveklabnik:
> Obviously, we can't continue to add special cases like this forever. But until we tackle the bigger architectural changes, it's fine to handle it this way :)

Awesome, I feel the same way. My intention is to try to look at the bigger stuff after we get all of this little stuff fixed as we go 😄

steveklabnik:
Oh, and do you think we could get a release? We have two or three things now; it'd be good to get this out in front of people

azerupi:
>Oh, and do you think we could get a release?

Well, it's a perfect occasion to test if you can publish to crates.io too ;)
I added you to the maintainers, so it should work. Let me know how it goes.

steveklabnik:
0.0.18 published! ❤️

PR #225, opened 2017-03-10
Update to highlight.js 9.10.0

carols10cents:
eddyb noticed some weird things being highlighted in the book, which led to [him sending this PR to highlight.js](https://github.com/isagalaev/highlight.js/pull/1454), which has made it into highlight.js 9.10.0, which has led to me sending in this pr :) 

I generated the js using [this page](https://highlightjs.org/download/) and choosing the same languages that were listed in [this pr](https://github.com/azerupi/mdBook/pull/174).

I also updated highlight.css using the Atelier Dune Light Theme that comes with highlight.js, since that's what the comment at the top seemed to indicate.... there's a lot in the diff though, and [atelier dune light doesn't seem to change very often](https://github.com/isagalaev/highlight.js/blame/master/src/styles/atelier-dune-light.css) but it has changed a little, soooo i'm not sure how much change is due to the "modifications" [since it was introduced as modified](https://github.com/azerupi/mdBook/commit/58d18d467cc015a06cbc4e6b59a5abcdafc3a523#diff-fd5c4b47d5b72534cb049aeb66740033). I tried to keep what looked important, and I definitely kept the gray background of code blocks because it tried to change it to yellow 😂😂😂 Please let me know if there's anything in particular I should put back the way it was!

It looks pretty good trying it out locally on the book, wdyt?

screen shot 2017-03-09 at 11 01 19 pm


azerupi:
Looks good to me! 
I think highlight.js had not been updated in a while. The theme looks a bit different but it's fine :) I think I made a custom theme back then, but it's easier to have as little deviation from the original theme as possible for maintenance reasons. So :+1:

steveklabnik:
@azerupi is happy, Travis is happy, and so I am happy. Merging!

PR #228, opened 2017-03-19
Relative links for non-JS browsers

jmillikan:
mdBook is a mostly pleasant experience in emacs' 'eww' browser (and probably other non-JS browsers). Next and prev links make it a bit more seemless.

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -21,11 +21,9 @@
         <link rel="stylesheet" href="tomorrow-night.css">
 
         {{#previous}}
-	<link rel="prev" href="{{link}}#content">
         {{/previous}}
 	
         {{#next}}
-	<link rel="next" href="{{link}}#content">
         {{/next}}
 
         <!-- MathJax -->
@@ -80,13 +78,13 @@
 
                 <!-- Mobile navigation buttons -->
                 {{#previous}}
-                    <a href="{{link}}" class="mobile-nav-chapters previous">
+                    <a rel="prev" href="{{link}}#content" class="mobile-nav-chapters previous">
                         <i class="fa fa-angle-left"></i>
                     </a>
                 {{/previous}}
 
                 {{#next}}
-                    <a href="{{link}}" class="mobile-nav-chapters next">
+                    <a rel="next" href="{{link}}#content" class="mobile-nav-chapters next">
                         <i class="fa fa-angle-right"></i>
                     </a>
                 {{/next}}
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -20,6 +20,14 @@
         <link rel="stylesheet" href="highlight.css">
         <link rel="stylesheet" href="tomorrow-night.css">
 
+        {{#previous}}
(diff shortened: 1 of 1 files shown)


src/theme/index.hbs
```diff
@@ -20,6 +20,14 @@
         <link rel="stylesheet" href="highlight.css">
         <link rel="stylesheet" href="tomorrow-night.css">
 
+        {{#previous}}
+	<link rel="prev" href="{{link}}#content">
```
steveklabnik: I have a feeling that this should be connected to https://github.com/azerupi/mdBook/pull/228/files#diff-9f0d373829e789d9a6923dd290c5b174R83

src/theme/index.hbs
```diff
@@ -20,6 +20,14 @@
...
+        {{/previous}}
+	
+        {{#next}}
+	<link rel="next" href="{{link}}#content">
```
steveklabnik: ... and this to https://github.com/azerupi/mdBook/pull/228/files#diff-9f0d373829e789d9a6923dd290c5b174R89

jmillikan:
Fair point. Here's a simplification. I added #content for convenience, but this doesn't seem to disturb chrome or firefox.

steveklabnik:
@azerupi what do you think?

src/theme/index.hbs
```diff
@@ -20,6 +20,12 @@
         <link rel="stylesheet" href="highlight.css">
         <link rel="stylesheet" href="tomorrow-night.css">
 
+        {{#previous}}
+        {{/previous}}
+	
+        {{#next}}
+        {{/next}}
```
azerupi: This doesn't seem to do anything, is this a relic you forgot to remove?

src/theme/index.hbs
```diff
@@ -72,13 +78,13 @@
 
                 <!-- Mobile navigation buttons -->
                 {{#previous}}
-                    <a href="{{link}}" class="mobile-nav-chapters previous">
+                    <a rel="prev" href="{{link}}#content" class="mobile-nav-chapters previous">
```
azerupi: What is `#content` supposed to do? When I test it locally it doesn't seem to change anything.

src/theme/index.hbs
jmillikan: Whoops, yes.

src/theme/index.hbs
jmillikan: In 'eww', #content causes prev/next to jump past the TOC to the content. A "skip to content" link would probably be better for supporting screen readers etc, this is just what I'm using to generate books locally.

azerupi: Ah I see, focusing on the content rather than the TOC has been requested in #235. But `#content` does not solve that issue, could you maybe try adding `tabindex=1` to the content to see if it solves both issues at the same time? :)

azerupi:
I merged this with a small modification, I left out the `#content` anchor links as I have recently merged another PR fixing that issue (with js) and I'm not sure if it would collide or not. 

Thanks a lot for the PR and I'm terribly sorry for the delay.  :)

PR #230, opened 2017-03-23
Fix keyboard navigation to trigger only if no modifier key is pressed

### LINKED ISSUE #224: Don't trigger next/prev chapter nav when modifier key was pressed
People on HN (rightfully) [complain](https://news.ycombinator.com/item?id=13815331) that mdbook's keyboard navigation is a bit too eager. I think it should only be triggered by pressing the left/right cursor buttons by themselves, but not when also pressing any modifier key (like ctrl or cmd).

You can accomplish this by adding

```js
if (e.altKey || e.ctrlKey || e.metaKey || e.shiftKey) { return; }
```

above [this line](https://github.com/azerupi/mdBook/blob/677fa42458df658463593150ad62395da8105b9c/src/theme/book.js#L36) (see [here](https://developer.mozilla.org/de/docs/Web/API/KeyboardEvent) for docs).

crazymerlyn:
Fixes #224.

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -33,6 +33,7 @@ $( document ).ready(function() {
     };
 
     $(document).on('keydown', function (e) {
+        if (e.altKey || e.ctrlKey || e.metaKey || e.shiftKey) { return; }
         switch (e.keyCode) {
             case KEY_CODES.NEXT_KEY:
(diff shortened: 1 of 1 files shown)


azerupi:
Thanks! :)

Issue #220, opened 2017-03-05
Fragment identifiers aren't unique


ollie27:
Fragment identifiers should be [unique](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id) but this currently isn't the case when a document contains multiple identical headers. They can also clash with footnotes.

azerupi:
This should be fixed now, closing.

PR #231, opened 2017-03-23
Fix header links

### LINKED ISSUE #219: Header fragment links use depreciated `name` attribute
Headers are currently surrounded by an `a` tag which uses a `name` attribute but they are [obsolete](https://www.w3.org/TR/html5/obsolete.html). `id` should be used instead.

crazymerlyn:
Header fragment links now use "id" attribute instead of the depreciated "name" attribute.

Similar headers are given numbered ids to avoid id collisions.
For instance, if there are three headers named "Example", their ids would be "example", "example-1", and "example-2" respectively.

Fixes #219 and #220.

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -11,6 +11,7 @@ use std::fs::{self, File};
 use std::error::Error;
 use std::io::{self, Read};
 use std::collections::BTreeMap;
+use std::collections::HashMap;
 
 use handlebars::Handlebars;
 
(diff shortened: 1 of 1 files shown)


steveklabnik:
Ah, this is a nice solution!

steveklabnik:
@azerupi @frewsxcv any thoughts here?

azerupi:
Yes, my thoughts are that it is good to merge :wink: 
Thanks for the PR!

steveklabnik:
I'm not gonna cut a release with this just yet, as I'd like to do some more stuff first. But soon.

Issue #11, opened 2015-07-29
Add tests


azerupi:
Add tests in the code. 

I have never really used tests, if someone could help me make meaningful tests that would be wonderful!

azerupi:
On hold, is a part of #90

rbuckland:
As there are tests - this issue should be closed now. If there are any specific tests that are thought needed (that I am sure there are), suggest a new issue is raised.

mattico:
Agreed!

PR #237, opened 2017-03-28
Introduce smoke test

steveklabnik:
This checks that we render the example book correctly. In order to do
that, we put a known-good copy of the book under tests/book, and then in
tests/smoke.rs, we generate a copy of it, and then diff the two
directories.

This means that PRs that change the generated output will need to update
this fixture, but it also means we get to see an easy way of what they
actually update.

Part of #11


azerupi:
Oh yes yes yes! I am all in for better tests :)
I have accepted a couple of PRs so make sure that the test passed before merging.

steveklabnik:
>  I have accepted a couple of PRs so make sure that the test passed before merging.

👍 ❤️

steveklabnik:
only a small amount of change, as you can see in the second commit 😄

steveklabnik:
Ahh, line endings are the issue here.... I made this PR on Windows, so windows stuff succeeds, but linux stuff fails. Hrm.

steveklabnik:
..... I am not sure how to fix this, given that if I change it the other way, it will fail on Windows. Hrm.

carols10cents:
Sounds like something that should be fixed in dir-diff, I [filed an issue for you](https://github.com/steveklabnik/dir-diff/issues/9) ❤️

steveklabnik:
:heart:

tests/book/index.html
```diff
@@ -0,0 +1,109 @@
...
+
+                    <h1 class="menu-title">mdBook Documentation</h1>
+
+                    <div class="right-buttons">
+                        <i id="print-button" class="fa fa-print" title="Print this book"></i>
+                    </div>
+                </div>
+
+                <div id="content" class="content">
+                    <a class="header" href="misc\introduction.html#introduction" id="introduction"><h1>Introduction</h1></a>
```
ollie27: This is the actual problem. Backslashes shouldn't be used in URLs. I've opened a more relevant issue: #263

azerupi:
Thinking about this a bit, this kind of test might be a bit too restrictive. Since the output often changes with small PRs I think it would be better to drop this idea and instead implement a lot of small tests checking for one particular property in the generated output. It would probably not catch everything, but it would be more granular than this solution. 

Now the major drawback would be to actually write all of the small tests. But we have had a lot of contributions recently and I think that if we make easy issues, we can get most of them contributed as mentored PRs.

What do you think?

Michael-F-Bryan:
Hey @azerupi and @steveklabnik I just noticed this PR.

> I think it would be better to drop this idea and instead implement a lot of small tests checking for one particular property in the generated output

I've started something similar in #374. In particular the `/tests/rendered_output.rs` test file seems to do exactly what you are talking about. If I can abstract the checking out into a nice macro, would it be worth merging that and then supplementing it with a "fuzzier" version of this smoke test (perhaps something which compares simplified HTML ASTs)?

I'd also be happy to mentor a PR for asserting particular patterns appear in the output as expected.

steveklabnik:
Let's do it! 😄

PR #243, opened 2017-03-31
Accept nightly examples.

### LINKED ISSUE #241: Allow unstable features for the unstable book
The home page of the Unstable book shows an example with the box syntax. But if I try to run it in the page, the compilation fails because unstable features are not supported.

steveklabnik:
This also brings us to parity with rustdoc regarding attributes in
general; while this PR was focused on enabling nightly, that was a
happy accident.

Fixes #241 

Check it out:

mdbook

This is after clicking "run" on each. Top goes to stable, bottom to nightly.

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -318,15 +318,37 @@ fn add_playpen_pre(html: String) -> String {
                 format!("<pre class=\"playpen\">{}</pre>", text)
             } else {
                 // we need to inject our own main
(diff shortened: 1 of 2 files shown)


azerupi:
This means that `#![...` can only occur at the top, right?
Otherwise it looks good, you may merge it :)

steveklabnik:
> This means that #![... can only occur at the top, right?

Yeah, it will put them up at the top. This is exactly how it's worked in rustdoc for a long time, so even though it's not perfect, seems fine 😄

steveklabnik:
AppVeyor was spurious, so I merged anyway. 👍

PR #248, opened 2017-04-07
Fix alignement of chapters with three digit numbering

### LINKED ISSUE #242: Three digit numbering in ordered list has strange alignment
https://doc.rust-lang.org/nightly/unstable-book/

screen shot 2017-03-30 at 4 38 39 pm

mthh:
Should fix #242 .
mdbook
Instead of relying on the length of the chapter number, it now rely on the number of dot(s) it contains. 
(i'm not really used to the internals of mdBook, so I may have missed something that would make this patch completely invalid, sorry if is the case!)


frewsxcv:
Relying on the number of dots in the string seems a little hacky, but it doesn't seem much worse than what we were doing previously. I'm also not familiar with the internals of mdbook at all to give a better suggestion, so this seems okay to me. @steveklabnik @azerupi Have any thoughts?

azerupi:
Indeed, it is not ideal but it is better than what we have right now. Maybe we can find a better solution in the future, but let's go with this in the mean time :)

Thanks for the PR!

PR #250, opened 2017-04-13
Added monospace font with support for box-drawing chars

regexident:
Even Github seams to not be capable of properly showing curses-style box-drawing characters:

```
  ╭────────────────────╮
╭→│ boxes are awesome! ├─╮
│ ╰────────────────────╯ │
╰────────────────────────╯
```

With this PR mdBook would be (`$ cargo doc` does support them, too!).

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -3,6 +3,10 @@ body {
   font-family: "Open Sans", sans-serif;
   color: #333;
 }
+code {
+  font-family: "Source Code Pro", "Menlo", "DejaVu Sans Mono", monospace;
+  font-size: 0.875em;
+}
 .left {
   float: left;
 }
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -10,7 +10,8 @@
         <base href="{{ path_to_root }}">
 
         <link rel="stylesheet" href="book.css">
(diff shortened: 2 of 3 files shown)


azerupi:
Looks cool!
Two nitpicks though :)

1. The font looks much bigger than before. Is it possible to tweak the font-size so that it stays roughly the same size? Ideally we should check for every back-up font that it looks ok at that size.
2. Could you modify the [stylus files](https://github.com/azerupi/mdBook/tree/master/src/theme/stylus) and regenerate the css, instead of editing the css file directly? Because changes in the css file will be overwritten on the next regeneration.

regexident:
> 1. The font looks much bigger than before. Is it possible to tweak the font-size so that it stays roughly the same size? Ideally we should check for every back-up font that it looks ok at that size.

Apple's "Menlo" and Ubuntu's "DejaVu Sans Mono" are both based on "Vera Sans Mono". As such they are near-identical in general appearance and size, while "Source Code Pro" has a slightly taller line-height.

There is [`font-size-adjust`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-size-adjust), but unfortunately it's only supported by Firefox right now. However adding `font-size: 0.875em;` should do the trick, I think, bringing the font-size down roughly to where it was before (on my machine's Chrome, at least).

> 2. Could you modify the stylus files and regenerate the css, instead of editing the css file directly? Because changes in the css file will be overwritten on the next regeneration.

Oh, I missed that step. Done.

azerupi:
Thank you very much :)

PR #252, opened 2017-04-16
Remove the extra run button on runnable rust files

### LINKED ISSUE #251: Run button is displayed twice on runnable rust files
image

crazymerlyn:
The playpen helper now uses a simple pre block instead of a pre block
with class playpen as it led to nested playpens.

Fixes #251.


steveklabnik:
Yup, my mistake! Thank you so much.

PR #260, opened 2017-04-30
Allow static files to be served using mdbook serve

derekdreery:
This patch allows the user to mount asset directories when using the `serve` command. It's still necessary to move the folder manually into the site for serving it externally.

Let me know if you think this would be a good addition or not :)


azerupi:
My apologies for the delay! 🙁 
This looks good, I would not be strongly opposed to merging this. But I'm wondering if we should. I'm a little bit afraid that by adding such functionality we risk adding a lot of new expectations and maintenance burden for features that are not the main functionality.

I wonder if this particular feature is not best served by one of the many simple http servers out there (SimpleHTTPServer from python comes to mind)? I may be unnecessarily worried though. And like I said, I am not specifically against this feature.

@steveklabnik @frewsxcv @budziq do you have any thoughts on this? :)

derekdreery:
You could include this, then if you find that the problems you mention materialize, you can remove it again before 1.0. :)

budziq:
Hi @derekdreery,

Thanks for your effort!  

Please note that both `build` and transitively `serve` commands already copy the directory structure recursively under `src_dir` to the `output_dir` which gives us way to have arbitrary number of served `asset` directories. 

Before I form an opinion I would like to hear your part of the story. Namely what is your exact usecase that cannot be easily realized with current features?

If your aim is to:
1. Have an `assets` dir outside of `src_dir. -  Then I would rather have additional `asset_dirs` (name is arbitrary) entry in `book.toml` that would give a list of dirs to copy into the `output_dir`. Then these directories would be both build and served. 
2. Serve arbitrary dir - I would be very hesitant to add such feature as:
    - `serve` command is strictly a feature to allow book writing nicer (auto reload) and not to make it a incomplete content server. As @azerupi already noted a better tool would be easily found for this job and we would like to keep the mdbook [lean and maintanable](http://www.catb.org/~esr/writings/taoup/html/ch01s06.html#id2878022).
    - we loose symmetry between `build` and `serve` commands. User would expect for the option to somehow affect `build` also 

But I might be missing something so please describe what is the itch that you are trying to scratch and we'll try to help :smile:!

derekdreery:
Don't worry - it wasn't too much effort :). Yes it was to have some assets for the book, so (1) would solve the problem.

Is there a way to do this with the current codebase? If not I'll rework the PR.

azerupi:
>Yes it was to have some assets for the book

Out of curiosity, what is the reason that you can't move your assets into the `src` directory? :)
Because as @budziq mentioned, assets are already copied over from src to the output directory automatically today. 

>Is there a way to do this with the current codebase?

Not currently, but it should be relatively simple to add. You would have to add a new key to the configuration, much like the `additional-css` and `additional-js` keys. And then just copy over the assets at the specified paths when building.

derekdreery:
I'll look to see if I can do it without any code changes.

azerupi:
Ok :)
If you can't, let us know! That might be a very valid reason to incorporate this feature in mdBook.

steveklabnik:
So, I personally never use `mdbook serve`. I probably wouldn't have put it in mdbook in the first place 😄 But I don't mind it being there, as it doesn't really hurt anything.

I guess this is a long way of saying "If you're fine with this PR I am"

derekdreery:
Hmm, I'm not sure what to argue. I understand the value of simplicity, but then assets shared between mdbook and other parts of the app is a valid use case.

I think I'll rework this PR to use the asset_dirs suggestion and see how other people find this. I think the API should be 

```toml
["...".asset_paths]
"output/path/to/asset" = "actual/path/to/asset"
```
, where `asset` could be a file or directory (if directory, then everything is recursively copied in). This would fail if any of the following are true:

 1. `output/path/to/asset` already exists in `src`.
 2. Two assets have the same `output/path/to/asset`.
 3. `actual/path/to/asset` doesn't exists, or there are permission problems etc.

with an appropriate error message. Symlink behaviour is the same as for current files in src (ignore).

Can I ask what people think of this before I spend time implementing it, as it's slightly more complicated than the original PR?

budziq:
@derekdreery 
> ` ["...".asset_paths]
"output/path/to/asset" = "actual/path/to/asset" 
`
I would rather suggest api similar to `additional-css`
```
[output.html]
destination = "book"
theme = "theme"
additional-css = ["theme/custom.css"]
additional-js = ["theme/custom.js"]
additional-assets = ["path/to/assetdir_a/", "other/path/to/assetdir_b/"]
```

A list of paths to dirs and the dirs themselves would be copied instead of contents. Then the user would not have to worry about collisions in the asset dirs content names. And just make sure that the dirs themselves do not collide and no dir in src has identical name. This would be explicitly stated in docs and acompanied by a verbose `mdbook build` error if such situation occurs.

Do you think this variant is too restrictive?

derekdreery:
> A list of paths to dirs and the dirs themselves would be copied instead of contents.

This trades flexibility (renaming assets/dirs) for simplicity (array rather than map, more unified book.toml structure, simpler implementation). For me, this would be fine.

frewsxcv:
Sorry for the very long delay here.

> So, I personally never use `mdbook serve`. I probably wouldn't have put it in mdbook in the first place 😄 But I don't mind it being there, as it doesn't really hurt anything.

Ditto what Steve said. I probably won't use it, but I'm alright with it merging in.

budziq:
> This trades flexibility (renaming assets/dirs) for simplicity (array rather than map, more unified book.toml structure, simpler implementation). For me, this would be fine.

Let me know if you encounter any problems :)

azerupi:
I am going to close this PR for now. If you want to resume work on this, feel free to open up a new PR! 😊

PR #262, opened 2017-05-02
Fix broken link in book-example/src/lib/lib.md

Rufflewind:
The template sets `<base href="../">` so it's not necessary to have `../` in the link.


azerupi:
Thanks!

PR #267, opened 2017-05-05
Change MathJax to use CDNJS

Rufflewind:
Because the [MathJax CDN will soon be retired](https://www.mathjax.org/cdn-shutting-down/).


azerupi:
Thank you!

PR #283, opened 2017-05-17
Do not add playpen boilerplate when using quick_main!

budziq:
Code snippets using quick_main! macro from error-chain
https://docs.rs/error-chain/0.10.0/error_chain/macro.quick_main.html
no longer have `fn main` implicitly added

resolves https://github.com/azerupi/mdBook/issues/282


azerupi:
That was quick, thank you! :)

PR #284, opened 2017-05-17
Fix indentation of hidden code blocks

budziq:
Hidden code blocks are no longer indented with
one additional space (required for doctests to compile in some cases)
Now the behavior is similar to the rustdoc's

Please note that I have zero experience with js so I am really open to criticism here ;)

resolves https://github.com/azerupi/mdBook/issues/281


azerupi:
Looks good to me :)
Thanks for the patch!

Issue #266, opened 2017-05-05
Add CSS without overriding theme


rap2hpoutre:
Maybe I misunderstood something but it seems the CSS theme must be fully replaced if I want to add a rule. So if I did understand well, if I just want to add a simple rule (ex: `h1 { color: #f00; }`) I must copy the original file then add my rule at the end of the file. 

If so, maybe we should add the possibility to create an optional `custom.css` where we could add our simple rules over the existing ones? I could submit a PR if you want.

steveklabnik:
You are right, but this is almost an exact duplicate of https://github.com/azerupi/mdBook/issues/178, down to the file name! 😄 So thanks, but closing as a duplicate.

rap2hpoutre:
Oops.

PR #287, opened 2017-05-18
Revamp config code

### LINKED ISSUE #178: Add an empty, but included last, `custom.css` file
I'd like to be able to add just a few little custom style rules, but I don't want to maintain my own version of book.css or highlight.css that might diverge from the mdbook provided styles.

What do you think about mdbook always including, or trying to include if it exists, a file named `custom.css` or similar, that mdbook would intentionally never add anything to?

This file would need to be included last, so that the rules in `custom.css` would be able to override rules in `book.css` or `highlight.css`.

I'd be happy to work on a PR for this, but wanted to see if this was something you'd be interested in first.

### LINKED ISSUE #240: some values are overwritten by read_config
I get the following output:

```
"foo"
"foo/bar"
"foo/baz"

"foo"
"foo/src"
"foo/book"
```

That's when I run this code:

```rust
extern crate mdbook;

use std::path::Path;
use mdbook::MDBook;

fn main() {
    let book = MDBook::new(Path::new("foo"))
        .set_src(Path::new("bar"))
        .set_dest(Path::new("baz"));
    println!("{:#?}", book.get_root());
    println!("{:#?}", book.get_src());
    println!("{:#?}", book.get_dest());
    println!();

    let book = book.read_config();
    println!("{:#?}", book.get_root());
    println!("{:#?}", book.get_src());
    println!("{:#?}", book.get_dest());
}
```

### LINKED ISSUE #285: Push the TOML configuration as default
As part of #275 we would like to push the TOML configuration as the default and deprecate the JSON config. We are not going to remove the JSON config just yet, but we are going to show a deprecation warning.

Here are some the things that need to happen:

- [x] Parity between the JSON config and TOML config
- [x] Move and rename `dest`, `src` and `theme_path` to
  ```toml
  source = "path/to/sources"
  
  [output.html]
  destination = "path/to/output"
  theme = "path/to/directory"
  ```
- [x] Add a key for an additional custom stylesheet #266 #178 
  ```toml
  [output.html]
  additional-css = ["path/to/file.css"]
  ```
- [x] Add tests for each config key
- [x] Reading the config should not overwrite unspecified keys #240 
- [x] Document the TOML config

azerupi:
This is a rewrite of the configuration code for the TOML configuration push #285 
The JSON configuration should remain unchanged, but is deprecated. Unless we want to support both?

The supported TOML format has been changed to the following:

```toml
title = "A book about nothing"
description = "A large book that tells no story"
author = "John Doe"
# authors = ["John Doe", "Jane Doe"]   <-- multiple authors
source = "path/to/src"

[output.html]
destination = "path/to/dest"
theme = "path/to/them"
google-analytics = "123456"
```

I have also added tests for each key, both in TOML and JSON.

The other files that use the configs have only gotten changes necessary to compile and work. But by going through the code for the changes I realised how ugly my code is... 😄  I think changing the rest is out of scope for this PR, but there is a lot of opportunity for clean-ups.

Closes #285, closes #178, closes #240 and improves the test coverage.


azerupi:
Ok, this PR is ready for review. I have implemented all the things from #285 

- [x] Parity between the JSON config and TOML config
- [x] Move and rename `dest`, `src` and `theme_path` to
  ```toml
  source = "path/to/sources"

  [output.html]
  destination = "path/to/output"
  theme = "path/to/directory"
  ```
- [x] Add a key for an additional custom stylesheet #266 #178 
  ```toml
  [output.html]
  additional-css = ["path/to/file.css"]
  ```
- [x] Add tests for each config key
- [x] Reading the config should not overwrite unspecified keys #240 
- [x] Document the TOML config

src/book/mod.rs
```diff
@@ -395,52 +396,49 @@ impl MDBook {
...
         self
     }
 
     pub fn get_title(&self) -> &str {
-        &self.title
+        self.config.get_title()
     }
 
+/*
     pub fn set_author(mut self, author: &str) -> Self {
```
frewsxcv: do we still need these methods?

azerupi: get and set author?
No, we can remove them. I would like to remove the other methods as well and provide a better way to access the configuration, but they are too pervasively used throughout the code and a little to ambitious for this PR.

azerupi:
Thanks! Let's wait to see if Steve wants to have a look before we merge :)

budziq:
Hi,

I was thinking about removing duplication of fields between Book and BookConfig (most likely keep Option<BookConfig> within Book) add getters instead of direct field access (most likely also add builder here and there) for https://github.com/azerupi/mdBook/issues/301 but such refactor would clash with this work.

I'll wait untill this gets merged and dust settles before making any proposals.

azerupi:
Oh yes, that would be a good refactor, when making this PR it was pretty painfull to update all the getters everywhere. Actually now I think `MDBook` should not even provide getters and setters but just have a method to get the config.

It's pretty safe to say that this PR is going to land soon without much changes. I'm just waiting to see if @steveklabnik wants to have a look before we do.

budziq:
> I think MDBook should not even provide getters and setters but just have a method to get the config.

My thoughts exactly :)

steveklabnik:
> It's pretty safe to say that this PR is going to land soon without much changes. I'm just waiting to see if @steveklabnik wants to have a look before we do.

Yup! 👍

PR #288, opened 2017-05-19
Made changes with rustfmt including `use_try_shorthand`

budziq:
Updated the project rustfmt.toml to include `use_try_shorthand = true`.
Run rustfmt on all rust sources.

Resolves: https://github.com/azerupi/mdBook/issues/286

### DIFF
--- a/src/bin/mdbook.rs
+++ b/src/bin/mdbook.rs
@@ -121,7 +121,7 @@ fn init(args: &ArgMatches) -> Result<(), Box<Error>> {
     let mut book = MDBook::new(&book_dir);
 
     // Call the function that does the initialization
-    try!(book.init());
+    book.init()?;
 
     // If flag `--theme` is present, copy theme to src
     if args.is_present("theme") {
@@ -142,7 +142,7 @@ fn init(args: &ArgMatches) -> Result<(), Box<Error>> {
         }
 
         // Call the function that copies the theme
-        try!(book.copy_theme());
+        book.copy_theme()?;
         println!("\nTheme copied.");
 
     }
@@ -172,14 +172,14 @@ fn build(args: &ArgMatches) -> Result<(), Box<Error>> {
 
     let mut book = match args.value_of("dest-dir") {
         Some(dest_dir) => book.set_dest(Path::new(dest_dir)),
-        None => book
+        None => book,
     };
 
     if args.is_present("no-create") {
         book.create_missing = false;
     }
 
-    try!(book.build());
+    book.build()?;
 
     if args.is_present("open") {
         open(book.get_dest().join("index.html"));
@@ -197,11 +197,11 @@ fn watch(args: &ArgMatches) -> Result<(), Box<Error>> {
 
     let mut book = match args.value_of("dest-dir") {
         Some(dest_dir) => book.set_dest(Path::new(dest_dir)),
-        None => book
+        None => book,
     };
 
     if args.is_present("open") {
-        try!(book.build());
+        book.build()?;
         open(book.get_dest().join("index.html"));
     }
 
@@ -227,7 +227,7 @@ fn serve(args: &ArgMatches) -> Result<(), Box<Error>> {
 
     let mut book = match args.value_of("dest-dir") {
         Some(dest_dir) => book.set_dest(Path::new(dest_dir)),
-        None => book
+        None => book,
     };
 
     let port = args.value_of("port").unwrap_or("3000");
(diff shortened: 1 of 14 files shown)


src/bin/mdbook.rs
```diff
@@ -253,19 +253,19 @@ fn serve(args: &ArgMatches) -> Result<(), Box<Error>> {
                 socket.close();
             }}
         </script>
-    "#, public_address, ws_port, RELOAD_COMMAND).to_owned());
+    "#,
+                                public_address,
+                                ws_port,
+                                RELOAD_COMMAND)
+        .to_owned());
```
azerupi: These 5 lines are a little strange, I'm going to check rustfmt to see if there is an option to fix this or if I should file this as a bug.

src/renderer/html_handlebars/helpers/toc.rs
```diff
@@ -40,48 +40,49 @@ impl HelperDef for RenderToc {
...
+                    rc.writer
+                        .write_all(Path::new(item.get("path")
                                                        .expect("Error: path should be Some(_)"))
                                              .with_extension("html")
                                              .to_str()
                                              .unwrap()
                                              // Hack for windows who tends to use `\` as separator instead of `/`
                                              .replace("\\", "/")
-                                             .as_bytes()));
+                                             .as_bytes())?;
```
azerupi: This bit is also a little weird, I think this can be easily fixed by moving the `Path::new()...` to a temporary variable just above.

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -106,14 +107,15 @@ impl Renderer for HtmlHandlebars {
...
+                        book.write_file(filename, &rendered.into_bytes())?;
 
                         // Create an index.html from the first element in SUMMARY.md
                         if index {
                             debug!("[*]: index.html");
 
                             let mut content = String::new();
-                            let _source = try!(File::open(book.get_dest().join(&ch.path.with_extension("html"))))
+                            let _source = File::open(book.get_dest().join(&ch.path.with_extension("html")))
+                                ?
```
azerupi: The questionmark should definitely not go on its own line here. This is probably a bug in rustfmt, but it can certainly be solved on our side by moving `book.get_dest()...` into a temporary variable.

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -301,39 +323,45 @@ fn fix_anchor_links(html: String, filename: &str) -> String {
...
-        let after = &caps[3];
+            let before = &caps[1];
+            let classes = &caps[2].replace(",", " ");
+            let after = &caps[3];
 
-        format!("<code{before}class=\"{classes}\"{after}>", before=before, classes=classes, after=after)
-    }).into_owned()
+            format!("<code{before}class=\"{classes}\"{after}>", before = before, classes = classes, after = after)
+        })
+        .into_owned()
```
azerupi: This block seems to be wrongly indented, bug in rustfmt?

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -301,39 +323,45 @@ fn fix_anchor_links(html: String, filename: &str) -> String {
...
+                // not language-rust, so no-op
+                format!("{}", text)
             }
-        } else {
-            // not language-rust, so no-op
-            format!("{}", text)
-        }
-    }).into_owned()
+        })
+        .into_owned()
```
azerupi: Same here, indentation is one level too high.

azerupi:
Thank you for this!
It really improves readability in most places.

I found a couple of places where formatting is a little weird. It's very much possible that those are bugs in rustfmt. If they are, it's worth filing an issue on their repo. I will investigate that a little bit further.

Otherwise I'm really happy! A couple of small changes should improve the formatting even more in some places (see comments) and then it is good to merge!

budziq:
Ok I'll work on this later today :)

budziq:
I have redone the formatting with newest rustfmt 0.8.4 and at least 2 of the mentioned bugs are fixed but unfortunately rustfmt seams to have acquired some new funky rules by default which you might not like :(  (including breaking up long comments in strangest places which I have manually restructured to make more sense).
Also I have applied actionable fixes.
Regarding your last two comments for hbs_renderer.rs I guess that these are not bugs but wierd rules as `into_owned` is indented to the level of previous call on the variable.

@azerupi  I'm not sure if you like the newest version and I should not try to just push the try! related ones :/

azerupi:
I think it's okay :)
There is much more 'fixed' than what it breaks. I can live with the couple of spots that have weird formatting. One of these days I will go through the configuration options to see if I need to change any. But for now this is good enough.

Thanks for the PR!

PR #289, opened 2017-05-19
Inline code with hyperlink is now highlighted

budziq:
Inline code with hyperlink has now the same color and on hover underline as standard hyperlink.

This is how it looks in rust-cookbook
hilight

Please note that I have zero experience with css / webdev so I am really open to criticism.
solves: https://github.com/azerupi/mdBook/issues/261

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -258,7 +258,6 @@ table thead td {
   position: relative;
   left: 10px;
   z-index: 1000;
-  -webkit-border-radius: 4px;
   border-radius: 4px;
   font-size: 0.7em;
 }
@@ -295,7 +294,6 @@ table thead td {
     position: relative;
     display: inline-block;
     margin-bottom: 50px;
-    -webkit-border-radius: 5px;
     border-radius: 5px;
   }
   .next {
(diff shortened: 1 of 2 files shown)


budziq:
Hmm I looks like both travis and appveyor had some infrastructure problems on some builds

steveklabnik:
lets encrypt has had some downtime so all the things are failing.

On Fri, May 19, 2017 at 12:35 PM, Michał Budzyński <notifications@github.com

azerupi:
@budziq thanks for the PR!

For the styling we use a CSS pre-processor called Stylus; so you shouldn't modify the CSS manually because it will get overwritten. Instead you should modify the stylus files and regenerate the CSS.

Check out our brand new [contribution guide](https://github.com/azerupi/mdBook/blob/master/CONTRIBUTING.md) and let us know if you encountered any difficulties we should document. 😄

budziq:
Argh! Newbie mistake :( I'll look into stylus later today. Thanks for your tim!

budziq:
I've updated the PR with changes into base.styl and the generated book.css (as seams to be the convention basing on git history).
I have no idea why stylus decided to drop `-webkit-border-radius` so I might still be doing something wrong.

azerupi:
Thank you! :)
I wouldn't worry too much about the stylus output. It might just be that the webkit property is so old that they decided it wasn't worth inserting it anymore.

budziq:
Thanks!

Issue #271, opened 2017-05-15
Disable CDNs


azerupi:
Some resources are loaded through a CDN if possible with a local fallback. An option to deactivate the CDNs when rendering is wanted. See #46 

An option could be given in the config file, but I think it would fit better as a CLI argument.
On the template side, we could add a `cdn` variable that evaluates to true or false and use `if ... ` tags in handlebars.

tjis:
Ideally, we have more flexibility than just a true or false for cdn. When resources cannot be sourced through a cdn, they'll still need to be found in another way.
In debian, we ship javascript libraries in separate packages, which are then symlinked or copied. mdbook cannot have any knowledge of this process though, so I imagine we'll need to configure our intentions somehow.
A simple solution would be to make the resource url configurable somewhere, but default to the cdn.

azerupi:
So I am a little lost with what would be actually required. Currently, the strategy is to fetch from a CDN to have cache benefits and parallel loading. If the CDN fails we send the local version. This strategy does not really benefit books that are not hosted online though. But I am not sure it hurts either. 

If someone could clearly formulate the problem that needs to be solved and what they expect mdBook to do or allow, that would help us immensely! :)

tjis:
Alright. The problem is as follows.
Debian packages aren't supposed to make network connections without the user's consent. Also, packages aren't supposed to contain embedded resources that could be provided by other packages instead. Therefore, both the CDN and the fallback are problematic for Debian.
We are running into this issue while packaging rust, which uses mdBook to build the various rust books. In order to make the generated books comply with debian policy, we strip the embedded resources from mdBook and alter the templates so that the generated documentation complies.
But this is far from ideal. Not only might future mdBook changes break our present patch, it also means we have to ship a patched copy of mdBook with the rust source package, rather than using a pristine packaged version of mdBook and build-depend on that.

The solution to our problem would be if mdbook could be supplied with alternative means to retrieve third-party dependencies, rather than having the current CDN+fallback method hardcoded. Then debian could ship rust with the books configured to use the debian locations.

If this is something your project is willing to consider, I could provide a patch and we could work from there.

azerupi:
Ooh I see now, thank you for the explanation!
I would totally be ok with adding an option to allow this. I am not sure if this is common enough to put a key in the config file, but controlling this with environment variables seems like a good plan to begin with. :)

Michael-F-Bryan:
[...]

PR #291, opened 2017-05-19
rendering configuration from environment variables

tjis:
As discussed in #271, here's a patch that makes the mdBook output configurable through environment variables.

For details on the implemented scheme, see the comments under resources/.


azerupi:
I apologise for the wait!
I'm going to start reviewing this now. :)

src/resources/env_config.rs
```diff
@@ -0,0 +1,233 @@
...
+    fn must_embed(&self) -> bool;
+    /// returns the url for this resource. This panics if must_render_url returns false.
+    fn url(&self) -> String;
+    /// returns the source location for this resource, or None if none was configured.
+    /// This panics if must_embed returns false.
+    fn source(&self) -> Option<String>;
+}
+
+/// A third-party resource with nothing special to it.
+pub enum BasicResource {
```
Michael-F-Bryan: Any chance you can make this guy derive `Debug` and `Clone`? Not having a debug impl makes things annoying when you are debugging.

src/resources/env_config.rs
```diff
@@ -0,0 +1,233 @@
...
+            &BasicResource::UrlWithFallback{url: _, ref source} => source.clone(),
+            _ => panic!("no source available")
+        }
+    }
+}
+
+/// Special struct for font-awesome, as it consists of multiple embedded files
+/// but is configured as a directory. Methods implemented for this struct
+/// calculate the paths of all the files in this directory.
+pub struct Awesome {
```
Michael-F-Bryan: This should probably also have a `#[derive(Debug, Clone)]` attribute.

src/resources/env_config.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    pub fn otf_source(&self) -> Option<String> {
+        self.resource.source().map(|p| p + "/fonts/FontAwesome.otf")
+    }
+}
+
+/// Configuration information for all the third-party resources. This specifies
+/// for every resource whether it should be included at all, and if so, just as
+/// an URL or also as an embedded resource.
+pub struct Configuration {
```
Michael-F-Bryan: Similar to `Awesome` and `BasicResource`, `Configuration` should also implement `Debug`. `Clone` is also a useful thing to have.

src/resources/env_config.rs
```diff
@@ -0,0 +1,233 @@
...
+}
+
+/// Returns a variable as var would. If the variable could not be found, return
+/// the given default.
+fn var_default(resource: &str, key: &str, default: &str) -> String {
+    var(resource, key).unwrap_or(String::from(default))
+}
+
+/// The rendering strategy for a resource
+enum Strategy {
```
Michael-F-Bryan: This could probably derive `Debug`, `Clone`, `Copy`, and `PartialEq` (the last one makes testing nicer).

src/resources/env_config.rs
```diff
@@ -0,0 +1,233 @@
...
+    }
+}
+
+/// parse font_awesome configuration from environment variables.
+fn awesome() -> Awesome {
+    Awesome { resource: resource("AWESOME", AWESOME_URL) }
+}
+
+/// Parse the configuration from the environment variables.
+pub fn configuration_from_env() -> Configuration {
```
Michael-F-Bryan: Can we add a couple integration tests (most probably in the `tests/` directory) which set some environment variables and make sure `Configuration_from_env()` picks up the appropriate resources? The implementation itself is fairly straightforward, but it's always nice to have a couple sanity tests to make sure it all works and prevent any regressions from being introduced down the track.

src/resources/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+                Vec::from(default)
+            }
+        }
+        else {
+            Vec::from(default)
+        }
+    }
+
+    pub fn highlight_js(&self) -> Vec<u8> {
+        assert!(self.conf.highlight.must_embed());
```
Michael-F-Bryan: Just out of curiosity, what do we stand to gain from having these `assert!()` statements instead of returning something like a `Result`?

(continuing on my previous comment on the `resource_content()` method)

PR #295, opened 2017-05-22
Friendlier errors from build.rs and support for Windows stylus build

budziq:
`cargo build --features=regenerate-css` should now have nicer errors and work on windows

It is written mostly to solve: https://github.com/azerupi/mdBook/issues/293
but also fixes: https://github.com/azerupi/mdBook/issues/294

Please note that I have very little Rust experience so I would be very glad for critical review. 
Build.rs has become quite wordy and error-chain was added as build-dependency which is quite hevyweight for what the script does.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,9 @@ ws = { version = "0.7", optional = true}
 [dev-dependencies]
 tempdir = "0.3.4"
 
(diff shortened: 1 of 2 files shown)


azerupi:
Thank you! 
Looks really good :)

PR #298, opened 2017-05-26
Fix websocket port option for serve

superstring:
Websocket port option does not work since mdbook.rs use --websocket-port in arguments and use ws-port in implementation.


azerupi:
Thank you for the PR!
So looking at it now, I would actually prefer to keep the more verbose name `--websocket-port`.
If you change that I can merge the PR :)

superstring:
Done. :)

azerupi:
Thank you!

PR #300, opened 2017-05-27
Reformatted "Serving on " message for easier consumption

budziq:
Now we have: `Serving on: http://localhost:3000`

I am aware of the `-o` option but it has been raised as desirable to be able to just copy and paste the address from terminal.

fixes: https://github.com/azerupi/mdBook/issues/299


azerupi:
Thanks !

PR #303, opened 2017-05-30
RFC - Initial implementation of clipboard handling

budziq:
Connected to https://github.com/azerupi/mdBook/issues/302

Here is a hacky backport of my rust-cookbook implementing clipboard handling.

The main idea is to add a third playpen button http://fontawesome.io/3.2.1/icon/copy/ which on click transfers playpen contents to system clipboard (and some tooltip notifying about success).

I do not feel confident with js and stylus at all (stylus part is shaky at best)
I wanted someone to have a look on the code prior to moving forward (or handing it over to someone competent).


azerupi:
It's pretty good and an awesome feature, thanks!
It's so useful that I would actually enable it on all code blocks and not only playpen code.

A couple of small requests:

1. Can you add a local fallback for clipboard.js for when there is no internet access? You can see how I have done it for the rest.
2. The tooltip looks a bit strange with the serif font, could you make it use the same font(s) as the normal text? Also I think it looks better a little smaller, like `font-size: 0.8em;` and for centering I would use
    ```css
    .tooltiptext {
        position: absolute;
        left: -8px;   /* Half of the width of the icon */
        transform: translateX(-50%);  /* Center by moving the tooltip 50% of its width to the left */
        top: -35px;
        ...
    }
    ```
3. Is there a specific reason you use `opacity: 0.8;`? If the only purpose is to not have pure black, then you can directly set the background color to the equivalent grey: `#333333`:)

budziq:
Thanks :) The CSS was a product of my addled brain while late night googling.
Actually I wanted the popup to be a little translucent but honestly don't remember why ;)

I'll try to fix that later today.

budziq:
@azerupi fixed suggested problems, added cdn fallback for clipboard.js and rebased

azerupi:
Thanks a lot ! :)

PR #304, opened 2017-05-31
Updated highlight.js to v9.12.0

budziq:
Fixing problem with raw strings syntax highlighting https://github.com/azerupi/mdBook/issues/269
Also backported updates to atelier-dune-light.css

The highlight.js was build with support for:
- All the commons
- Rust
- Scala
- Swift
- Go
- Haskel
- Assembly (x86 and ARM)
- Yaml
- D
- Handlebars


azerupi:
Thanks!

PR #305, opened 2017-06-01
Convert to curly quotes when rendering markdown

jimmydo:
This updates `render_markdown` in `utils/mod.rs` to convert all straight quotes to curly quotes,
except for those that occur in code blocks and code spans.

This was motivated by my search for the best way to fix [inconsistent curly quotes in the Rust book](https://github.com/rust-lang/book/issues/363), and it seemed like enhancing mdbook might be the cleanest way to do it. Let me know what you think!

### DIFF
--- a/src/config/jsonconfig.rs
+++ b/src/config/jsonconfig.rs
@@ -13,6 +13,7 @@ pub struct JsonConfig {
     pub description: Option<String>,
 
     pub theme_path: Option<PathBuf>,
+    pub curly_quotes: Option<bool>,
     pub google_analytics: Option<String>,
 }
 
--- a/book-example/src/format/config.md
+++ b/book-example/src/format/config.md
@@ -66,6 +66,7 @@ The following configuration options are available:
 - **destination:** By default, the HTML book will be rendered in the `root/book` directory, but this option lets you specify another
   destination fodler.
 - **theme:** mdBook comes with a default theme and all the resource files needed for it. But if this option is set, mdBook will selectively overwrite the theme files with the ones found in the specified folder. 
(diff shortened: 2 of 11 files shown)


azerupi:
Thank you!
I'm ok with adding this, but I would prefer for it to be behind a configuration option and probably off by default to begin with. I don't think we should alter what the user has written without him deliberately opting into it, but offering this functionality as an extension seems like a nice idea :)

jimmydo:
Yup, that makes sense – thanks for the feedback. I'll work on making it an opt-in feature.

jimmydo:
@azerupi It's now off by default, and the user can opt-in using either `curly_quotes = true` in `book.{toml,json}` or `--curly-quotes` from the command line.

jimmydo:
Pushed some updates that fixes merge conflicts.

azerupi:
Thanks! I will have a look at this in the weekend :)

src/config/jsonconfig.rs
```diff
@@ -13,6 +13,7 @@ pub struct JsonConfig {
     pub description: Option<String>,
 
     pub theme_path: Option<PathBuf>,
+    pub curly_quotes: Option<bool>,
```
azerupi: The only reason the JSON config is still there is to give people the time to migrate to the TOML one. So, I'm not sure if I would add the new configuration options to it.

azerupi:
Everything looks good!
I left a small note about the JSON config, but I took care of it. It was only a minor detail. :)

Thanks a lot for this PR!

jimmydo:
Yup, your comment makes sense. Thanks for fixing and merging. :)

PR #307, opened 2017-06-01
Now changes to `theme` directory trigger rebuild for `mdbook serve`

budziq:
As `theme` dir is no longer under `src`. Updates to "theme" did not
trigger book rebuild.
Also fixed misleading docs about `theme` dir being located in `src`

fixes https://github.com/azerupi/mdBook/issues/306


azerupi:
Thanks!
Just to be sure, how does it behave when the theme dir does not exist? :)

budziq:
Argh rookie mistake. Fixed now :)

Issue #185, opened 2016-12-05
iOS jumping to page top is broken


aaaxx:
A minor nit but I thought I'd report it anyway.

Something, in javascript most likely, breaks the standard functionality in iOS of jumping to the top of the page when the top of the screen is tapped.

azerupi:
Thanks for reporting! I will probably need some help for this one though, as I have no iOS device to test this on :)

aaaxx:
Well, I can test it if you can figure out what's causing it.

Dylan-DPC:
This was fixed in #310

PR #310, opened 2017-06-05
On iOS, allow scrolling to the top of the page by tapping the top of the screen

jimmydo:
Hello! This fixes the iOS scroll-to-top bug mentioned in issue #185.

Notes:

* Scroll-to-top is a built-in function of iOS Safari that didn't work because the
  page content was inside absolutely-positioned, scrollable divs.

* The fix is to stop using absolute positioning on `.page-wrapper` and
  `.page`, so that the content uses static positioning and flows
  naturally down the page.

* Consequently, `.sidebar` and `.nav-chapter` now have to use `position:
  fixed` in order to be positioned relative to the viewport.

* This fix also enables Safari's built-in behavior of automatically
  hiding the top and bottom toolbars when scrolling down the page.

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -3,6 +3,9 @@ body {
   font-family: "Open Sans", sans-serif;
   color: #333;
 }
+body {
+  margin: 0;
+}
 code {
   font-family: "Source Code Pro", Consolas, "Ubuntu Mono", Menlo, "DejaVu Sans Mono", monospace, monospace;
   font-size: 0.875em;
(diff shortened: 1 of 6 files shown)


jimmydo:
@azerupi Any thoughts on this pull request? It's a very annoying papercut on iOS, so it would be great to get this merged. I've tested this on my iPhone and it seems to work well. Let me know if there's anything I can do to help. Thanks!

azerupi:
I apologise for the wait! 
I am just about to finish my exam period, so I will try to go through all the open PRs in the next couple of days. 

Stay tuned! 😊

jimmydo:
Oh no worries – thanks for the prompt reply. Good luck on exams!

azerupi:
I didn't find any regressions, so it's good for me!

Many thanks for this PR!

PR #312, opened 2017-06-06
Minor cleanup

budziq:
- removing need to explicitly use `Path::new` all over the place
- removed warnings from doctests (normally invisible unless `cargo test -- --nocapture`)
- no doctests are norun/ignore now
- updated docs both in book-example and in docs not to refer to nonexisting API's

This is mostly due to my fear of introducing any regressions / warnings to the code when I try to contribute something.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,16 +21,16 @@
 //! extern crate mdbook;
 //!
 //! use mdbook::MDBook;
-//! use std::path::Path;
 //!
+//! # #[allow(unused_variables)]
 //! fn main() {
-//!     let mut book =  MDBook::new(Path::new("my-book"))           // Path to root
-//!                         .with_source(Path::new("src"))          // Path from root to source directory
-//!                         .with_destination(Path::new("book"))    // Path from root to output directory
-//!                         .read_config()                          // Parse book.json file for configuration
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -21,16 +21,16 @@
...
-//!     let mut book =  MDBook::new(Path::new("my-book"))           // Path to root
-//!                         .with_source(Path::new("src"))          // Path from root to source directory
-//!                         .with_destination(Path::new("book"))    // Path from root to output directory
-//!                         .read_config()                          // Parse book.json file for configuration
-//!                         .expect("I don't handle the error for the configuration file, but you should!");
+//!     let mut book =  MDBook::new("my-book")        // Path to root
+//!                         .with_source("src")       // Path from root to source directory
+//!                         .with_destination("book") // Path from root to output directory
+//!                         .read_config()            // Parse book.toml configuration file
+//!                         .expect("I don't handle configuration file error, but you should!");
```
azerupi: "error" should be plural here :)

azerupi:
Apart from that little typo, it is ready to merge. So I will fix it during the merge, there is no need to bother you with that.

Thanks for this PR! The changes are really great, makes to code tidier overall.

PR #314, opened 2017-06-06
Fixes missing the default "theme" dir location

budziq:
if one is not specified in book.toml

fixes https://github.com/azerupi/mdBook/issues/313

### DIFF
--- a/src/theme/mod.rs
+++ b/src/theme/mod.rs
@@ -1,4 +1,4 @@
-use std::path::PathBuf;
+use std::path::Path;
 use std::fs::File;
 use std::io::Read;
 
@@ -46,7 +46,7 @@ pub struct Theme {
 }
 
 impl Theme {
-    pub fn new(src: Option<&PathBuf>) -> Self {
+    pub fn new(src: Option<&Path>) -> Self {
 
         // Default theme
         let mut theme = Theme {
--- a/src/book/mod.rs
+++ b/src/book/mod.rs
@@ -13,7 +13,7 @@ use std::process::Command;
 use {theme, parse, utils};
 use renderer::{Renderer, HtmlHandlebars};
 
-use config::{BookConfig, HtmlConfig};
+use config::BookConfig;
 use config::tomlconfig::TomlConfig;
 use config::jsonconfig::JsonConfig;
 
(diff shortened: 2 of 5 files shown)


azerupi:
I think this is better fixed directly by giving the [`HtmlConfig` a default value for the theme](https://github.com/azerupi/mdBook/blob/master/src/config/htmlconfig.rs#L30). What do you think?

budziq:
Yep that's much better idea :)

I have pushed a version that changes HtmlConfig.theme from `Option<&PathBuf>` to `PathBuf` as now it is guaranteed to always be set (unfortunately a lot of the code had to be adapted to this change). Otherwise I can submit a super minimal version:
```diff
-------------------------- src/config/htmlconfig.rs ---------------------------
index 1c42a34..c90c76b 100644
@@ -25,9 +25,10 @@ impl HtmlConfig {
     /// assert_eq!(config.get_destination(), &output);
     /// ```
     pub fn new<T: Into<PathBuf>>(root: T) -> Self {
+        let root = root.into();
         HtmlConfig {
-            destination: root.into().join("book"),
-            theme: None,
+            destination: root.clone().join("book"),
+            theme: Some(root.join("theme")),
             curly_quotes: false,
             google_analytics: None,
             additional_css: Vec::new(),
```

src/theme/mod.rs
```diff
@@ -46,7 +46,7 @@ pub struct Theme {
 }
 
 impl Theme {
-    pub fn new(src: Option<&PathBuf>) -> Self {
+    pub fn new(src: Option<&Path>) -> Self {
```
azerupi: I suppose we can remove the option here too, no?

budziq: I was thinking about it but but there is this problem that while in the context of `HtmlConfig` the `theme` is guaranteed, it is not in the context of `MdBook` as the `HtmlConfig` is optional itself

```rust
    pub fn get_theme_path(&self) -> Option<&Path> {
        if let Some(htmlconfig) = self.config.get_html_config() {
            return Some(htmlconfig.get_theme());
        }

        None
    }
```

So removing this option will make us jump through hoops here:
```rust
  --> src/renderer/html_handlebars/hbs_renderer.rs:36:39
   |
36 |         let theme = theme::Theme::new(book.get_theme_path());
```

azerupi: Right, let's keep it like it is for now then. Maybe it will get refactored with all the planned changes anyway.

PR #315, opened 2017-06-06
Add `Ayu` theme

### LINKED ISSUE #223: Would you be interested in a PR for this theme?
Hello,

I recently put together a theme for [various Rustlang-related websites](https://github.com/Cldfire/ayu-rs) based on [this ST3 theme](https://github.com/dempfi/ayu). I was looking into porting it to the upcoming Rust book as well, which is, as you are probably aware, built with this program. I thought it would also be worth asking if you would be interested in a PR to have this theme added to mdBook's out-of-the-box themes?

If not, that's perfectly fine (I understand you already have a blueish dark theme).

Cldfire:
Closes #223.

I finally got this done! Here is a comparison between the new theme (left) and the Navy theme (right):

comparison

The only modification I had to make to mdBook itself (beyond the addition of the theme throughout the codebase, of course) was to introduce a variable so that I could change inline code's text color. If you don't want this it can certainly be removed (although I feel that it makes the theme more readable).

And finally, to be clear: if you don't feel like mdBook should ship with this theme, feel free to say so. I'm not trying to force it on anyone :)

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -177,6 +177,7 @@ impl Renderer for HtmlHandlebars {
         book.write_file("jquery.js", &theme.jquery)?;
         book.write_file("highlight.css", &theme.highlight_css)?;
         book.write_file("tomorrow-night.css", &theme.tomorrow_night_css)?;
+        book.write_file("ayu-highlight.css", &theme.ayu_highlight_css)?;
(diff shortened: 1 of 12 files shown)


azerupi:
Looks very nice! Thank you 😊 

I will take a look at this PR when I have more time, but from a quick glance through the changes everything looks fine to merge.

azerupi:
Merged. Thanks a lot!

PR #320, opened 2017-06-09
Fix ES5 compatibility.

pravic:
Currently book.js contains pure ES5 code, so it is not a big deal to backport just two lines to make old browsers happy.

Thanks!


azerupi:
Thanks for catching this! It slipped through review ;)

PR #321, opened 2017-06-09
One more ES5 fix.

pravic:
Yet another fix for missing [String.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes). Actually, there is a polyfill for it, but since it is just one place - it is easier to use `indexOf` here.

Sorry for two PRs, I didn't expect previous to be merged so fast :)

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -242,7 +242,7 @@ function run_rust_code(code_block) {
         code: text,
     };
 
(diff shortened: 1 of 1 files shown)


azerupi:
Don't worry, I happened to be checking GitHub at that moment and it was a small enough change to merge directly :)

Thank you for the PRs!

PR #324, opened 2017-06-12
Move from localStorage to store.js (v2.0.3) - also hide sidebar on mobile

budziq:
Fixes a lot of browser incompatibilities in localStorage/cookie handling
Including but not limited to:

- loss of styling and functionality on chromium private mode
- loss of styling and functionality on safari and safari private mode
- awaiting verification if problems in mobile safari are solved.

fixes: https://github.com/azerupi/mdBook/issues/317
fixes: https://github.com/azerupi/mdBook/issues/322

due to my laziness I've included changes to also fix "no way out of sidebar on mobile"
fixes: https://github.com/azerupi/mdBook/issues/318
fixes: https://github.com/azerupi/mdBook/issues/125
as this heavily uses localStorage. but this can be removed from the PR


azerupi:
Thanks!

PR #326, opened 2017-06-12
copying to clipboard no longer copies the compilation results from "play"

budziq:
fixes https://github.com/azerupi/mdBook/issues/323

might not be the prettiest solution but it gets the job done.


azerupi:
Thanks! :)

PR #329, opened 2017-06-13
Fix for playpen buttons missing on mobile safari and chrome IOS

budziq:
some googling and a hunch later...
it looks like a long standing bug in ios webkit with absolutely possitioned elements without z-index
I've verified it in a coworkers iphone (safari and chrome) and remotely on ios simulator

fixes https://github.com/azerupi/mdBook/issues/325


azerupi:
Nice! Thanks a bunch :)

PR #334, opened 2017-06-14
Hide theme selection popup after interaction

budziq:
fixes https://github.com/azerupi/mdBook/issues/330


budziq:
Both commits are orthogonal so I can drop the first one.

azerupi:
>Both commits are orthogonal so I can drop the first one.

I think that would be best. If in the future we do find we want it anyway, we can still add it later.

budziq:
done :)

azerupi:
Thanks!

PR #328, opened 2017-06-13
Update handlebars and some helpers

sunng87:
In 0.27, I made some change to handlebars `RenderContext` API based on how mdbook uses it. Now it's time to update it. Also I had some cleanup in helpers.

PR #335, opened 2017-06-15
Refactor hbs renderer

Michael-F-Bryan:
In my "plugins" branch I'm wanting to experiment with different ways of doing plugins (for pre-processing, post-processing, different backends, etc) and in the process I needed to refactor the `HTMLHandlebars` renderer a little bit. If you reckon they'll be useful, feel free to merge the changes into master :)

All I did was pull bits and pieces of the `HTMLHandlebars::render()` method out into their own helper functions so it's easier to wrap your head around.

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -25,6 +25,96 @@ impl HtmlHandlebars {
     pub fn new() -> Self {
         HtmlHandlebars
     }
+
+    fn render_item(&self, item: &BookItem, book: &MDBook, data: &mut serde_json::Map<String, serde_json::Value>,
+                   print_content: &mut String, handlebars: &mut Handlebars, index: &mut bool)
+                   -> Result<(), Box<Error>> {
+        match *item {
+            BookItem::Chapter(_, ref ch) |
+            BookItem::Affix(ref ch) => {
+                if ch.path != PathBuf::new() {
+
+                    let path = book.get_source().join(&ch.path);
+
+                    debug!("[*]: Opening file: {:?}", path);
+                    let mut f = File::open(&path)?;
+                    let mut content: String = String::new();
+
+                    debug!("[*]: Reading file");
+                    f.read_to_string(&mut content)?;
+
+                    // Parse for playpen links
+                    if let Some(p) = path.parent() {
+                        content = helpers::playpen::render_playpen(&content, p);
+                    }
+
+                    // Render markdown using the pulldown-cmark crate
+                    content = utils::render_markdown(&content);
+                    print_content.push_str(&content);
+
+                    // Update the context with data for this file
+                    let path =
+                        ch.path
+                            .to_str()
+                            .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "Could not convert path to str"))?;
+                    data.insert("path".to_owned(), json!(path));
+                    data.insert("content".to_owned(), json!(content));
+                    data.insert("chapter_title".to_owned(), json!(ch.name));
+                    data.insert("path_to_root".to_owned(), json!(utils::fs::path_to_root(&ch.path)));
+
+                    // Render the handlebars template with the data
+                    debug!("[*]: Render template");
+                    let rendered = handlebars.render("index", &data)?;
+
+                    let filename = Path::new(&ch.path).with_extension("html");
+
+                    // Do several kinds of post-processing
+                    let rendered = build_header_links(rendered, filename.to_str().unwrap_or(""));
+                    let rendered = fix_anchor_links(rendered, filename.to_str().unwrap_or(""));
+                    let rendered = fix_code_blocks(rendered);
+                    let rendered = add_playpen_pre(rendered);
+
+                    // Write to file
+                    info!("[*] Creating {:?} ✓", filename.display());
+                    book.write_file(filename, &rendered.into_bytes())?;
+
+                    // Create an index.html from the first element in SUMMARY.md
+                    if *index {
+                        debug!("[*]: index.html");
+
+                        let mut content = String::new();
+
+                        let _source = File::open(
+                                book.get_destination()
+                                    .expect("If the HTML renderer is called, one would assume the HtmlConfig is set... (3)")
+                                    .join(&ch.path.with_extension("html"))
+                            )?.read_to_string(&mut content);
+
+                        // This could cause a problem when someone displays
+                        // code containing <base href=...>
+                        // on the front page, however this case should be very very rare...
+                        content = content
+                            .lines()
+                            .filter(|line| !line.contains("<base href="))
+                            .collect::<Vec<&str>>()
+                            .join("\n");
+
+                        book.write_file("index.html", content.as_bytes())?;
+
+                        info!("[*] Creating index.html from {:?} ✓",
+                                  book.get_destination()
+                                      .expect("If the HTML renderer is called, one would assume the HtmlConfig is set... (4)")
+                                      .join(&ch.path.with_extension("html"))
+                            );
+                        *index = false;
+                    }
+                }
+            },
+            _ => {},
+        }
+
+        Ok(())
+    }
 }
 
 impl Renderer for HtmlHandlebars {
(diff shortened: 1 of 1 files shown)


steveklabnik:
> All I did was pull bits and pieces of the HTMLHandlebars::render() method out into their own helper functions so it's easier to wrap your head around.

Sounds great to me! Tests are failing though.

Michael-F-Bryan:
Hmm, that's awfully strange. I looked at the Travis build and apparently that "X" is for commit cdeec22, yet the failing commit for this PR is b7aa78c... Additionally, when you follow the link to Travis, the PR is my "[WIP] Plugin Experiment" (#336). Maybe Travis hit a bug because there are multiple PRs from the same person, both sharing a couple commits? 

I should be able to fix this up tomorrow so it's ready to merge.

Michael-F-Bryan:
@steveklabnik, travis should be happy now. I think it was just getting confused because my PRs had a similar history and were made around the same time. The first 3 or 4 builds have all passed without any issues, so when they're all finished this should be good to merge.

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -182,6 +182,16 @@ impl HtmlHandlebars {
         handlebars.register_helper("previous", Box::new(helpers::navigation::previous));
         handlebars.register_helper("next", Box::new(helpers::navigation::next));
     }
+
+    fn copy_additional_css(&self, book: &MDBook) -> Result<(), Box<Error>> {
```
budziq: Hi just minor suggestion. This function copies also additional js files so maybe rename it.

budziq:
Wow that's a lot of awesome work!

src/renderer/html_handlebars/hbs_renderer.rs
Michael-F-Bryan: Cheers for picking up on that! I had another look at the names I'm using to make sure they reflect what the function is actually doing. In the end I went with the slightly longer `copy_additional_css_and_js`.

Michael-F-Bryan:
I'm not quite happy with the `render_item()` method (just look at that function signature!), but I think other than that I've left the `hbs_renderer.rs` module tidier than I found it. @budziq or @steveklabnik, any chance you can give this one final review?

budziq:
Hi. I'm on mobile only for upcoming few days so only a birds eye view on my part (and nothing speccific I'm afraid as viewing diffs on mobile github is quite a chore). 

- idealy we would loose the out params
- have a minimal `RenderContext` param that might contain all required info and enclose mutable state if there is no other choice.
- return something consumable in the Result instead of `()`. Maybe a some `RenderItem` type if we can avoid mutable state.
- most likely the index handling part could be extracted to another function (possibly dropping the `index` out param)
- Possibly it might be nice to split the rendering and write/build phases of the `Renderer` trait. One would accumulate the state that might me unavoidable for other output types like pdf. And second could write/build.

These are just some random thoughts without much insight so fell free to disregard any of those.

azerupi:
I think this PR is going to have some conflicts with #328 
Maybe we could try to combine both?

steveklabnik:
I wanted to say that I largely agree with @budziq

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -189,33 +137,109 @@ impl Renderer for HtmlHandlebars {
...
+        debug!("[*]: Register handlebars template");
+        handlebars.register_template_string("index", String::from_utf8(theme.index.clone())?)?;
+
+        debug!("[*]: Register handlebars helpers");
+        self.register_hbs_helpers(&mut handlebars);
+
+        let mut data = make_data(book)?;
+
+        // Print version
+        let mut print_content = String::new();
```
Michael-F-Bryan: I believe this is for when you'll try to print the html page from your browser, so it's actually a subsection of the HTML renderer. That was my interpretation anyway, either way at the minimum I should pull all print-specific stuff out into its own method.

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -25,152 +26,99 @@ impl HtmlHandlebars {
...
+                    info!("[*] Creating {:?} ✓", filename.display());
+                    book.write_file(filename, &rendered.into_bytes())?;
+
+                    // Create an index.html from the first element in SUMMARY.md
+                    if *index {
+                        debug!("[*]: index.html");
+
+                        let mut content = String::new();
+
+                        let _source = File::open(destination.join(&ch.path.with_extension("html")))
```
Michael-F-Bryan: Oops! That's probably a bug in `rustfmt-nightly` and I didn't take the time to look over the code for formatting issues before committing. 

I think the `_source` was left over from the original `HtmlHandlebars` implementation and I didn't notice it wasn't being used because the leading underscore silences unused variable warnings.

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -324,16 +357,16 @@ fn build_header_links(html: String, filename: &str) -> String {
...
-                                    Some(c.to_ascii_lowercase())
-                                } else {
-                                    Some(c)
-                                }
-                            } else if c.is_whitespace() && c.is_ascii() {
-                                Some('-')
-                            } else {
-                                None
-                            })
+                    if c.is_ascii() {
```
Michael-F-Bryan: Yeah, definitely. I was mainly trying to break things up into smaller chunks so it's easier to understand what `HtmlHandlebars::render()` does, but I may as well clean this up too.

Michael-F-Bryan:
Sorry it's taken me a while to get back to you guys! In #336 I was discussing with @azerupi that I'd like to do a lot of refactoring/restructuring of `mdbook` over the holidays (my last exam is tomorrow) to make it easier to work on and testable. 

As such, the entire `Renderer` system will probably be getting an overhaul because `mdbook` is really tightly coupled to the existing `HtmlHandlebars` renderer. A while back I tried to create a pdf renderer (which eventually spawned [latex-rs]), but couldn't make much headway because of how the the library is structured and a lot of the implicit assumptions in `MDBook` and configuration.

> have a minimal RenderContext param that might contain all required info and enclose mutable state if there is no other choice.

Good idea, at the very least it'll help make that function signature cleaner, but I don't think it'll deal with the underlying issue that rendering is tightly coupled to the `MDBook::write_file()` method and a lot of state. Which in turn makes it very hard to test and understand the code. I'm hoping to fix a lot of those state and testability issues in this upcoming refactoring/restructuring.

> return something consumable in the Result instead of (). Maybe a some RenderItem type if we can avoid mutable state.

> Possibly it might be nice to split the rendering and write/build phases of the Renderer trait. One would accumulate the state that might me unavoidable for other output types like pdf. And second could write/build.

Eventually that's the plan. Rendering should be broken up into a two-step process where you'll convert the in-memory `Book` into an in-memory representation of the rendered book, doing all the work to turn markdown into html. Then the second step would be to walk your in-memory representation, writing all of it to disk. The way the playpen and other bits of rendering are currently designed it assumes the book has been incrementally written to the file system, so if I were to return an intermediate `RenderedItem` playpen links would probably be broken because the item hasn't actually been written to disk yet.

There are a couple other things like that in the `HtmlHandlebars::render()` method and the only real way to fix them would be to rewrite most of the html renderer module to make it more flexible and testable (without needing to create an entire book in a temporary directory). I'll probably end up doing that over the next month or so anyway, but this PR was mainly to make the existing implementation more readable and help future work.

[latex-rs]: https://github.com/Michael-F-Bryan/latex-rs

src/renderer/html_handlebars/hbs_renderer.rs
Michael-F-Bryan: I just had a shot at fixing this up. The issue I'm having is that there aren't any tests to go off of and it's difficult to understand what the underlying algorithm is.

I *think* it's trying to:

- generate an id for each header tag (h1, h2, ...) from its content by
  - removing any inner tags
  - ascii lowercasing
  - replacing spaces with dashes
  - appending an auto-incremented counter to make sure the id is unique
- then wrap the entire thing in an anchor tag so regex can replace the old header with the new one

Can you or @azerupi let me know if I'm on the right track?

budziq:
@Michael-F-Bryan I haven't had time to really look into your changes yet. Please note that the nits so far were to the legacy implementation that you are trying to refactor so do not feel discouraged. So far the changes seem sound  :+1:  but taking into consideration the shear volume it will need some time to scan through (hopefully later this week).

sunng87:
@budziq @azerupi #328 didn't  touch any file in this patch. It should be ok to merge both, I think.

Michael-F-Bryan:
@sunng87, when I get back from work I'll make a local copy of this pull request and try to merge it with what you've done and see if that still compiles. 

The files I imported might use something you changed (or vice versa), skimming through what you've done it doesn't look like we'll affect each other but it's probably a good idea to check regardless.

sunng87:
@Michael-F-Bryan Got it. It's probably caused by some internal changes in handlebars 0.27. Let me take a look later today.

sunng87:
I can see no conflict and no error in `cargo build` and `cargo test` by merging #328 and this. It should be OK to merge both. @Michael-F-Bryan @budziq @azerupi

azerupi:
Thanks @sunng87 for testing!
I will take a look at the two PRs and try to merge them :)

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -25,197 +26,274 @@ impl HtmlHandlebars {
...
+        )?;
+        book.write_file(
+            "_FontAwesome/fonts/FontAwesome.ttf",
+            theme::FONT_AWESOME_TTF,
+        )?;
+
+        Ok(())
+    }
+
+    fn write_custom_file(&self, custom_file: &Path, book: &MDBook) -> Result<(), Box<Error>> {
```
azerupi: The name of this function is quite vague and doesn't really tell you what it does. Could you add a short comment above it explaining what it does for other people? Or if you have an idea for a better name..? :)

Michael-F-Bryan: I'm pretty sure that's just a helper function which will take some path relative to the book's `src` directory and copy it to the output directory. I've added a doc-comment explaining what it does.

Michael-F-Bryan:
@azerupi, I've solved the merge conflict on this (#305 altered a line I also refactored) and added an extra doc comment to explain what that `write_custom_file()` method does. 

Do you think anything else needs to be done on this PR?

azerupi:
No, it's all good for me! :)

Issue #163, opened 2016-08-12
[Discussion] Plug-ins


azerupi:
Non-core functionality should be isolated into plug-ins. Plug-ins can then be enabled / disabled depending on the users need. *e.g. syntax highlighting could be a plug-in*

Plug-ins should:
- be able to alter the markdown before rendering
- be able to alter the outputted format after rendering
- have access to the important configuration options
- have a configuration of their own
- be able to specify which renderers they target
- ???

How could we design this?

dtolnay:
Other abilities I would expect from plugins:

- serve local javascript (a search plugin)
- insert remote javascript (MathJax or Google Analytics)
- add rules to book.css

azerupi:
That is definitely something I would like too! And I have thought about it already. 
But I am not sure how I can make the plugin interface generic over multiple renderers (html, pdf, ...) and still be able to allow injecting arbitrary dependencies.

If you have any idea on how we could make this work, I am open for suggestions.

Also, I think it would be nice if plugins could be developed by third-parties. But since Rusts ABI is not stable that would probably require going through a C api..

Many open questions about this design

Evrey:
> Plug-ins should:

LLVM compilers should...

- > be able to alter the markdown before rendering
  
  be able to optimise the IR before generating machine code.
- > be able to alter the outputted format after rendering
  
  be able to optimise machine code. (Why?)
- > have access to the important configuration options
  
  Just give each plugin a read-only map of configuration options or a JSON value or whatever. Read-only as the order in which plugins would run is unclear.
- > have a configuration of their own
  
  Also pass a private read-write configuration in addition to the previous one. From there on it would be easy to build a single configuration object where plugins could apply local changes to the global configuration.
- > be able to specify which renderers they target
  
  check `--target=triple`. Back-end plugins might just have an additional name identifying what they generate, e.g. `my_plugin.target() == "HTML"` or `my_plugin.target() == MdBookTargets::HTML`, depending on how extensible MdBook will be. An MdBook plugin might have multiple backends available.
- > ???
  
  !!!

-----

So, yeah, to me, this list sounds a lot like basic compiler design. And with basic compiler design comes this workflow: Tokenise inputs, generate some sort of IR, modify IR, spit code out. Now, translated to MdBook, this means:

- **Generate Markdown IR**, e.g.:
  ```
  MdToken::H1[ MdToken::Text("Why "), MdToken::InlineCode("mdbook"), MdToken::Text(" ?") ],
[...]

Issue #90, opened 2016-01-02
Refactoring


azerupi:
# Refactoring mdBook

The current implementation has some design flaws that causes some limitations. Implementations of new features are sometimes more complex and scattered than I want them to be and mdBook is not flexible enough.

In this post I will try to order my thoughts, reflect on what I want to change and how I might do that. If you have any ideas / feedback to share, feel free to hop in the discussion!
## Goals
1. **Book representation**
   Internal representation (data structures) should change to allow for multi-language books. 
2. **Renderers**
   There should be a generic `Renderer` trait that can be implemented to provide renderers for all kind of formats: _html_, _pdf_, _epub_, ... **The core should be renderer agnostic!**
   Multiple renderers should be able to run (in parallel?) to export to multiple formats
3. **TOML config file**
   The configuration file should switch to TOML
4. **Improve Error handling**
   Some errors are handled well, others aren't at all. I would like to keep a list of the errors that happened so that at the end I can say: `3 errors occurred` instead of terminating on the first error. Of course some errors, like not being able to read the configuration, will cause termination. But a small parsing error in one file does not warrant to stop the whole rendering. Not clear how I will implement this.
5. **Plug-ins**
   Non-core functionality should be available in _"plug-in"_ form. By plug-in is meant an isolated piece of code that can be enabled / disabled by the user without affecting other functionality. Plug-ins could / should be able to specify the renderer they target.
   _e.g. syntax highlighting could be a plug-in_
6. **Tests**
   A lot more tests should be included, at the moment it is difficult because functions are long and do a lot of things at the same time. Those functions should be split up in functions that do one thing only so that they can be tested more easily. Ouput should be tested in some way too, without being to constraining.
7. **Make mdBook's code base overall simpler**
   Shorten long functions, group similar functionality in more general purpose functions. Make the code more idiomatic, remove dead code and add more comments.
#### Book representation
- [x] [Discussion:](146) come up with a good internal representation
- [x] Write a decent parser for the `SUMMARY.md` file
#### Renderers
- [ ] [Discussion:](149) come up with a design to allow multiple renderers
#### TOML config file
- [x] [Discussion:](96) settle on a good and future proof format
- [x] Implement TOML configuration
- [x] Switch from the JSON configuration to TOML
#### Improve Error handling
- [ ] [Discussion:](14) come up with a design to store and report better errors
#### Plug-ins
- [ ] [Discussion:](163) come up with a design for Plug-ins
#### Tests
[...]

PR #336, opened 2017-06-15
[WIP] Plugin Experiment

Michael-F-Bryan:
I thought I'd play around with the concept of plugins (#163) in a private fork to see how something like this could look. 

I doubt anything from this PR will end up in master, but it'll give us a more informed idea for what works, what doesn't work, how much effort could be involved in implementing it properly, etc. 

Note: I've started this branch on top of #335 because I needed some of the refactorings done there. I'm not sure if this is a no-no in terms of git branching and all that, so let me know if I need to fix any history.


Michael-F-Bryan:
To make it easier to integrate plugins into configuration and the rendering process, it may be a good idea to break the `MDBook` up into a couple smaller types. Off the top of my head, I'd probably break it up like this:

- `Book`: a plain old container which keeps track of its chapters and configuration
  - Has no real behaviour or responsibilities except from a bunch of getters and setters
- `DirectoryManager`: the object in charge of the `src/` directory, it will
  - create a new book and source tree from a template
  - Parse a `SUMMARY.md` and source files into the `Book` struct
- `Orchestrator`: the object in charge of tying everything together, it
  - Defers to the `DirectoryManager` to load the `Book` from disk
  - sets up the rendering pipeline so the `Renderer` can transform source text into the final product, while making sure each `Plugin` is called so they can do their pre/post-processing

I feel like `MDBook` has grown over time and gained functionality, and now it's trying to do too much, making it hard to extend or add different things like an alternate renderer backend or plugins. It's also fairly closely tied to the HTML renderer.

For example a `MDBook` probably doesn't need to know about CSS or JS explicitly. Instead, when the `Orchestrator` instantiates a new `Renderer` (it might have looked in the config and seen a `renderer = "html"` entry), it'll pass in a reference to the config. Then the `Renderer` can figure out how to set itself up, as well as look for a table with its name on it for `Renderer`-specific options. So the `HTMLRenderer` would look for a table in the config called "html", the `PDFRenderer` would look for a "pdf" table, etc.

@azerupi, what are your thoughts? After next Wednesday I've got a month or two off uni, so I'll have plenty of time to work on this.

Michael-F-Bryan:
I believe solving this (#163) in an extensible way will probably end up also having to deal with #90. So we'd be potentially killing 2 birds with 1 stone.

azerupi:
This is very exciting!
I haven't had the time to look at the code yet, so I can't give a lot of feedback.

>I feel like MDBook has grown over time and gained functionality, and now it's trying to do too much, making it hard to extend or add different things like an alternate renderer backend or plugins. It's also fairly closely tied to the HTML renderer.

That is a correct assessment 😉  
When I started this project I didn't really think it would come this far, so I didn't put a lot of thought into it. And then people got interested and asked for new features, so I tried to deliver. Unfortunately I often went the quick route instead of the durable one. Anyways, a big refactor is absolutely necessary at this point.

>I believe solving this (#163) in an extensible way will probably end up also having to deal with #90. 

I think so too, plugins would probably need most of the other changes to work.

>For example a MDBook probably doesn't need to know about CSS or JS explicitly.
It's also fairly closely tied to the HTML renderer.

Absolutely, I didn't start with the idea of having multiple renderers. We need to progressively decouple the html renderer so that it is completely isolated. But I think it not going to be a quick fix. 😋  And renderers should know what resources they need, mdBook should only tie everything together, like you say.

I am very very interested in this! So if you want to take on the work, I can only encourage you! I'm open for discussions and questions. Feel free to make new issues when you want to discuss a certain topic.

Michael-F-Bryan:
It's an achievement in itself that `mdbook` has been so popular and evolved to the point where you need to rethink its design! The vast majority of my personal projects lose momentum after a couple weeks and end up being forgotten and half finished. As it is, I've used `mdbook` several times now and found it to be excellent so it's nice to give something back.

Because this will largely be a structural change with me mainly rearranging existing code, I thought I might approach the task by creating an empty sub-crate (arbitrarily named `mdbook-core`) and then copy/pasting things across one at a time. It'll probably be a bit more time-consuming, but that way I've got the flexibility to structure things any way I want and it'll be easier to discover which components are unnecessarily coupled in the existing system. Do you think this is the right approach to take?

So far I've copied across the summary parsing and set up a lot of the base types (`Runner`, `Book`, `Loader`, etc). I've also added a couple tests here and there to help me figure out what various things are doing, I figure they'll be useful in the future in preventing regression.

azerupi:
>I thought I might approach the task by creating an empty sub-crate (arbitrarily named mdbook-core)

Seems like a sensible approach :)
I propose we try to merge and discuss changes on a regular basis because I'm terrible at reviewing large diffs.. 😉 

>I've also added a couple tests here and there to help me figure out what various things are doing, I figure they'll be useful in the future in preventing regression.

That's awesome, a lot of things are untested right now and it sometimes comes back to bite us. So adding more tests is already a big improvement in itself! 😊

Michael-F-Bryan:
> I propose we try to merge and discuss changes on a regular basis because I'm terrible at reviewing large diffs.. 

I was thinking about this as well. I don't really want to do the entire refactoring in a single pull request because a lot of stuff will be moved around and it'll probably end up with a 6000-odd line diff. Plus if I'm incrementally moving things across then other people can help out, I can use the issue tracker, and it'll help to break the task down into manageable chunks.

What are your thoughts on merging an initial implementation that's pretty much just a skeleton with a bunch of stubbed out types and `unimplimented!()` functions? 

Also, after we find some nice way of laying things out, what are your thoughts on asking for help in the [Call for Participation][cfp] thread in the forums? We could easily make a bunch of easy issues along the lines of "copy X to mdbook-core and make sure everything compiles", which would be nice as an easy first PR.

[cfp]: https://users.rust-lang.org/t/twir-call-for-participation/4821

Michael-F-Bryan:
@azerupi, when you have a spare couple minutes, would you be able to have a look at what I've done so far? 

The easiest way will probably be to clone my fork (`git clone https://github.com/Michael-F-Bryan/mdBook`), checkout the `plugins` branch, then go into the `mdbook-core` directory and build the docs. That's what I've been doing when I want to get a high-level view of what I've done so far (plus it forces me to document what things do).

azerupi:
I will check it out in the next couple of days. I still have a big project that is due Wednesday for uni, but I should be able to take some time to check out your changes. Stay tuned!

>What are your thoughts on merging an initial implementation that's pretty much just a skeleton with a bunch of stubbed out types and unimplemented!() functions?

Sure I'm ok with that. So if I understand correctly, we would keep the current mdbook crate and continue parallel development in the mdbook-core crate?

If we do this, I think mdbook should relatively rapidly remove some of its components and use the core as soon as possible in order to avoid diverging too much. If people continue to fix issues and contribute features we don't want to have to back-port everything constantly. What do you think?

>what are your thoughts on asking for help in the Call for Participation thread in the forums?

Definitely, I have made changes to the repo recently to make it more contributor friendly. I am always happy to mentor and help others contribute. :)

Michael-F-Bryan:
> If we do this, I think mdbook should relatively rapidly remove some of its components and use the core as soon as possible in order to avoid diverging too much.

I was worried about this too. After thinking a bit more about this I'm not sure my `mdbook-core` approach is the right one. It'll mean we have two versions of `mdbook` being worked on concurrently and is pretty much the equivalent of a rewrite, just with a lot more copy/paste (obligatory [Joel on Software article][jos] on why this may be a bad idea). Instead I think I'll try to do lots of incremental restructurings of the existing codebase to make it less coupled and more testable. With the end goal to have it structured similar to what I've done in this PR. 

I'll still keep my `plugins` branch around because that contains an "ideal" version of how I think `mdbook` could be structured in the future and it'll be a nice reference for where I want to end up. My last exam is on wednesday, so expect to see a lot of smaller pull requests coming in over the next month or so.

> I doubt anything from this PR will end up in master, but it'll give us a more informed idea for what works, what doesn't work, how much effort could be involved in implementing it properly, etc.

Haha, turns out my comment at the very start of this PR was pretty accurate! It *did* give me a much better understanding of the task on hand and how `mdbook` works under the covers though, so this definitely hasn't been a waste of time.

[jos]: https://www.joelonsoftware.com/2000/04/06/things-you-should-never-do-part-i/

azerupi:
Perfect, let's do that :)
You can leave the PR open so that other people can see it and join in.
I'm going to make a new milestone for plug-ins, don't hesitate to create a bunch of issues / todos on the  issue tracker as you see fit :)

PR #337, opened 2017-06-15
Select right pane content on page load

budziq:
fixing problems with keyboard navigation

fixes https://github.com/azerupi/mdBook/issues/235

I also have a variant that selects the right pane whenever sidebar gains focus which is nice as the keyboard navigation always works even if someone misclicks on sidebar not hitting a link. But then keyboard navigation on sidebar itself stops being possible. So I've decided to just post this minimal version as it causes no regressions.

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -137,6 +137,7 @@ table thead td {
   right: 0;
   left: 0;
   bottom: 0;
+  outline: 0;
   padding-right: 15px;
   overflow-y: auto;
 }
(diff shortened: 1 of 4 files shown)


azerupi:
Was setting the `tabindex` on the content div to `1` not fixing it?

budziq:
Nope. Any value of tabindex (0,1) on `content` or `page` did not produce results by itself (at least not in chromium nor firefox). While the tabindex="-1" + javascript oneliner produces consistent good results.

azerupi:
In that case 👍 
Thanks!

PR #340, opened 2017-06-19
Add library path argument for `mdbook test`

### LINKED ISSUE #339: `mdbook test` with extern crates
[When invoking `rustdoc`](https://github.com/azerupi/mdBook/blob/master/src/book/mod.rs#L382) it should be possible to specific some options namely `-L` and `--extern` for `rustdoc` to allow testing examples with `extern crate xxx;`

messense:
Add `-L`/`--library-path` argument and pass it into `rustdoc` should be sufficient.

Closes #339

### DIFF
--- a/src/book/mod.rs
+++ b/src/book/mod.rs
@@ -351,9 +351,13 @@ impl MDBook {
         self
     }
 
-    pub fn test(&mut self) -> Result<()> {
+    pub fn test(&mut self, library_paths: Vec<&str>) -> Result<()> {
         // read in the chapters
         self.parse_summary().chain_err(|| "Couldn't parse summary")?;
+        let library_args: Vec<&str> = (0..library_paths.len()).map(|_| "-L")
+                                                              .zip(library_paths.into_iter())
+                                                              .flat_map(|x| vec![x.0, x.1])
+                                                              .collect();
         for item in self.iter() {
 
             if let BookItem::Chapter(_, ref ch) = *item {
@@ -363,7 +367,7 @@ impl MDBook {
 
                     println!("[*]: Testing file: {:?}", path);
 
-                    let output = Command::new("rustdoc").arg(&path).arg("--test").output()?;
+                    let output = Command::new("rustdoc").arg(&path).arg("--test").args(&library_args).output()?;
 
                     if !output.status.success() {
                         bail!(ErrorKind::Subprocess("Rustdoc returned an error".to_string(), output));
--- a/src/bin/test.rs
+++ b/src/bin/test.rs
@@ -5,15 +5,18 @@ use get_book_dir;
 
 // Create clap subcommand arguments
 pub fn make_subcommand<'a, 'b>() -> App<'a, 'b> {
-    SubCommand::with_name("test").about("Test that code samples compile")
+    SubCommand::with_name("test")
+        .about("Test that code samples compile")
+        .arg_from_usage("-L, --library-path [DIR]... 'directory to add to crate search path'")
 }
 
 // test command implementation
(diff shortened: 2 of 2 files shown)


src/book/mod.rs
```diff
@@ -379,7 +383,7 @@ impl MDBook {
 
                     println!("[*]: Testing file: {:?}", path);
 
-                    let output_result = Command::new("rustdoc").arg(&path).arg("--test").output();
+                    let output_result = Command::new("rustdoc").arg(&path).arg("--test").args(&library_args).output();
```
azerupi: Is this going to work okay when no arguments are passed?

azerupi:
Thanks!
I just have a simple question before merging, see inline comment. :)

src/book/mod.rs
messense: Yes. With no/empty arguments, it simply return `self`.

https://github.com/rust-lang/rust/blob/master/src/libstd/process.rs#L403

src/book/mod.rs
```diff
@@ -367,9 +367,13 @@ impl MDBook {
         self
     }
 
-    pub fn test(&mut self) -> Result<(), Box<Error>> {
+    pub fn test(&mut self, library_paths: Vec<&str>) -> Result<(), Box<Error>> {
         // read in the chapters
         self.parse_summary()?;
+        let library_args: Vec<&str> = (0..library_paths.len()).map(|_| "-L")
+                                                              .zip(library_paths.into_iter())
+                                                              .flat_map(|x| vec![x.0, x.1])
```
messense: It's a little annoying that `tuple` does not implement `IntoIterator` trait, causing this `vec![x.0, x.1]` nonsense.

azerupi:
Thanks! And sorry for the delay :)

Michael-F-Bryan:
It may be a good idea to thread this through the configuration system so you aren't continually writing the "-L" arguments. 

When #371 lands I'll move onto tidying up configuration and see if I can add an extra "tests" table to the `book.toml` file for adding parameters to tests like external dependencies. We've just got to think of a way to configuration so it's scalable to multiple backends and plugin systems (html, pdf, rustdoc testing, mathjax, etc) and not special casing everything like we are now.

azerupi:
@Michael-F-Bryan I don't see this as a permanent solution. The test command is currently just a wrapper around rustdoc, maybe with some additional checks, because it was needed for the Rust book. But ideally we would want this to be configurable and extendable by the user. Because there is no way we can anticipate everything the user would want to "tests".

Michael-F-Bryan:
> But ideally we would want this to be configurable and extendable by the user. 

My thoughts were that you could add something like this to your `book.toml`:

```toml
[test]
runner = "rustdoc"
dependencies = ["error-chain", "rayon"]
```

To me the test runner is pretty much just an alternate renderer, except instead of rendering content it'll give you a pass or a fail (because a `Renderer` returns a `Result`).

Otherwise if you have your own custom test script it might be invoked like this:

```toml
[test]
runner = "shell"
command = "test_docs.py"
```

azerupi:
That would be nice indeed!

budziq:
On the other hand rust-cookbook uses https://github.com/brson/rust-skeptic/. We never invoke `mdbook test` just using `cargo test` (extern crates are handled by cargo dependencies then) which imho is generally better approach if the mdbook is kind of insource/inrepo documentation.

budziq:
Hmm on the other hand I do nit see a better way (I mean this general approach suggested by @Michael-F-Bryan) to support spell checking and link checking which are very much desired. So big 👍

Issue #345, opened 2017-06-21
Reserved characters are not replaced by character entities when using files for playpens


projektir:
I've tested this on [rust-lang/book](https://github.com/rust-lang/book) with latest `mdBook` code on `master` with this page: [Creating Reference Cycles and Leaking Memory is Safe](https://doc.rust-lang.org/stable/book/second-edition/ch15-06-reference-cycles.html).

I moved the first block of code to a file called `file.rs`, and all code was preserved in that file, i.e., `RefCell<Rc<List>>` and such are present in the file itself. Then, I replaced what used to be that block of code with `{{#playpen file.rs}}` on the main page.

The resulting playpen got all of its generics either removed or garbled.

You can update the file and put various text between `<` and `>` and it will disappear. If you use `<` and `<`, things just get plain weird.

Looks like `<` and various other characters should be replaced by character entities, but this is not happening for playpens rendered from files.

projektir:
I've talked to @cbreeden about this and they had an interesting suggestion: instead of adding `<pre><code>` tags around the file contents, surround them with ```` ``` ````. Then they will get fed through the same process as normal code blocks and the Markdown parser will take care of it.

Right now, since code blocks and files are handled separately, there's a bit of duplication of effort.

azerupi:
That's a clever solution!

PR #346, opened 2017-06-21
Creating markdown code from playpen files instead of HTML #345

projektir:
Potential code for #345 for reference. Adding `editable` passthrough while we're at it.

Doing something like may have unintended consequences, of course.


azerupi:
Looks good! 
I don't expect to see any negative consequences, but if any do occur we can easily revert this fix :)

Thanks!

Issue #359, opened 2017-06-24
Refactor the book structure into its own type


Michael-F-Bryan:
The in-memory representation of a book is currently really tightly tied into `MdBook` and has a lot of functionality that it probably doesn't need. This complicates the current library quite a bit, makes testing difficult because you often need to create a temporary book on disk, and makes it hard to implement much-desired features like plugins and alternate renderers.

My thoughts are that a "book" should be a super simple tree data structure which represents the source files making up the book, then contain the chapters' text as a simple string. It shouldn't really have any behaviour, instead it'll use a `Loader` object (sorry,  names are hard :disappointed:  ) which will parse the `SUMMARY.md` and use it to construct a `Book` in memory.

This issue should act as an overall tracker for the changes necessary to make working with the book structure easier, dealing with the "book representation" part of #90. I'm planning to make a separate PR for each of the following:

- [x] Pulling the book representation out into a separate `Book` struct
- [x] Converting the `SUMMARY.md` parser to generate a "recipe" which can be used to load source files from disk and populate the `Book` tree structure
- [x] Actually populate the `Book` tree structure from disk using the `Summary` found earlier
- [ ] Updating `MdBook` so it can use the new `Book`

Some questions which may need addressing:

- How does multi-language support work? My thoughts are that it shouldn't know/care about multi-language, the low level internal representation should consider chapters to be blobs of text and leave differentiating languages to higher layers.
- Should individual chapters contain extra metadata? @gambhiro in #146 ([comment]) mentions it may be advantageous to attach metadata like the author, description, or a class to give the end renderer more information. This was also discussed in #277.
- Should the `Loader` be exposed publicly? This question was raised in #265. I'm of the opinion that it should be, being able to read the `SUMMARY.md` file will be useful for 3rd party tools.

Are there any other questions which should be addressed, or tasks which may fall under refactoring the internal representation of a book?

cc: @azerupi @budziq 

[comment]: https://github.com/azerupi/mdBook/issues/146#issuecomment-257905723

Michael-F-Bryan:
This was fixed in #491.

PR #360, opened 2017-06-24
Initial `Book` loader and summary parsing

Michael-F-Bryan:
This is the first step towards #359, creating a `Loader` type which is in charge of reading the source directory, parsing the `SUMMARY.md`, and then constructing an internal representation of the book.

This PR will refactor the existing summary parsing to return some `Summary` struct which indicates how the book is structured so that later steps can locate files and populate the book from disk.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ ws = { version = "0.7", optional = true}
 # Tests
 [dev-dependencies]
 tempdir = "0.3.4"
+pretty_assertions = "0.2.1"
 
 [build-dependencies]
 error-chain = "0.10"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -76,13 +76,17 @@ extern crate serde_derive;
 extern crate serde;
 #[macro_use] 
 extern crate serde_json;
+#[macro_use]
+extern crate log;
 
 extern crate handlebars;
 extern crate pulldown_cmark;
 extern crate regex;
 
+#[cfg(test)]
 #[macro_use]
-extern crate log;
+extern crate pretty_assertions;
+
 pub mod book;
 pub mod config;
 mod parse;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -89,6 +89,7 @@ mod parse;
 pub mod renderer;
 pub mod theme;
 pub mod utils;
+pub mod loader;
 
 pub use book::MDBook;
 pub use book::BookItem;
--- a/src/loader/mod.rs
+++ b/src/loader/mod.rs
@@ -27,14 +27,15 @@ impl Loader {
 
     /// Parse the summary file and use it to load a book from disk.
     pub fn load(&self) -> Result<()> {
-        let summary = self.parse_summary()
-            .chain_err(|| "Couldn't parse `SUMMARY.md`")?;
+        let summary = self.parse_summary().chain_err(
+            || "Couldn't parse `SUMMARY.md`",
+        )?;
 
         unimplemented!()
     }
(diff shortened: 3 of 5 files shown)


azerupi:
Is the loader's only purpose to look for and parse the summary? In that case I would make the name more like: `SummaryLoader` or put it in the `summary` module.

Also, I feel like we might want the parsing method to not return immediately on the first syntax error. It might be more interesting to continue parsing to report multiple errors. I'm not sure how that is best handled, but since this exploratory work we can think a little more about this :)

Michael-F-Bryan:
I was thinking the `Loader` can parse the summary and then using that parsing result it'll load the book from disk.

So in very rough pseudo-code:

```rust
impl Loader {
  ...

  pub fn load(&self) -> Result<Book> {
    let summary = self.parse_summary().chain_err(|| "Invalid SUMMARY.md")?;
    
    let mut book = Book::new();

    for item in summary {
      match item {
        SummaryItem::Chapter(ch) => {
          let chapter = self.load_chapter()?;
          book.add_section(chapter);
        }
        SummaryItem::Separator => book.add_section(BookItem::Separator),
     }
    }
  }
  
  Ok(book)
}
```

Another idea would be to accumulate errors in some `Vec` while loading, then add a variant to the `error-chain` called "LoadError" which contains a `Vec<Error>` to indicate all the errors encountered while parsing.

Michael-F-Bryan:
Oops... I think I accidentally did `git merge` instead of `git rebase` to pull in the latest changes from master. 346ddef upped the changes from ~300 lines and 2 files touched to ~30 files. I'll have to do a little cherry picking to fix up the history.

Anyways, I think the easiest thing for me to do to parse the summary is take advantage of `pulldown_cmark` and create a simple state machine which will continually munch new events from `pulldown` until there are none left, changing behaviour and incrementally building the `Summary` depending on what state you are in. After the `Summary` is created (it's just a tree representing of `SUMMARY.md`), you can walk it recursively and generate the corresponding `BookItem` node using the visitor pattern(e.g. a method may be `fn visit_link(&self, node: Link) -> BookItem`).

@budziq, @azerupi, and @steveklabnik, what are your thoughts on taking that approach compared to [the current implementation][1] which recursively walks the `SUMMARY.md` file, building up the `MdBook` all in one hit?

[1]: https://github.com/azerupi/mdBook/blob/master/src/parse/summary.rs

Michael-F-Bryan:
@azerupi, I've finally gotten things to the point where you can correctly parse a `SUMMARY.md` file and [the full integration test][test] passes! :smile: 

This PR is actually a lot smaller than it looks. There are only 362 executable lines of code in the `loader` module and something like 300-400 lines of tests. `cargo kcov` says the new summary parser has the largest number of lines covered by tests.

screenshot_2017-06-29_210528

Now I need to clean things up, deal with as many `TODO`s and `FIXME`s as possible, and get other people to review so I know what bits can be improved.

[test]: https://github.com/Michael-F-Bryan/mdBook/blob/01bab56fafb9e8b66386e3a76a4087fb0e8ccb72/tests/loading.rs

azerupi:
Nice!
I will try to take a look at this later today :)

Michael-F-Bryan:
> I'm not sure if its worth the effort to keep two separate PR's Why not incorporate all of the changes to #360 change it's title and close this one?

After [@budziq's comment][comment] I think I'll closing this PR and will put merge #360 and #371 into a single pull request.

[comment]: https://github.com/azerupi/mdBook/pull/371#issuecomment-312628571

Issue #14, opened 2015-07-29
[Discussion] Add solid Error messages


azerupi:
When using the cli tool there should be solid error messages explaining exactly what went wrong, if it is a parsing error add line numbers, ...

Non-fatal errors should not cause mdBook to stop. They should be stored and reported at the end. 

How can this be done? What would be a good design?
#### Inspiration
- [Cargo](https://github.com/rust-lang/cargo)

azerupi:
On hold, is a part of #90

Michael-F-Bryan:
It looks like [error-chain] is quickly becoming the de-facto tool for doing good error handling. In particular, the [rust cookbook] is pushing it as the way to do error handling in all examples.

Would it be a good idea to switch to using `error-chain` instead of `std::io::Error` or `Box<Error>` everywhere? I feel like the `chain_err()` method in particular would help with returning a nice human-readable message to the user, while still retaining all the information necessary to help with debugging. 

Coupling `error-chain` with more robust algorithms (e.g. trying fallbacks when an operation returns an `Err` instead of erroring out) would probably greatly improve the error handling story for `mdbook`.

[error-chain]: https://github.com/brson/error-chain

azerupi:
Yes, error-chain was already pulled in as a dependency for the build script to 'test out'. I haven't had the time yet to catch up with what error-chain can do, but I've heard a lot of good. So I'm ok to start using it pervasively throughout the code base :)

Dylan-DPC:
Closing this issue due to inactivity. If needed we can discuss this in an new issue. Thanks

Michael-F-Bryan:
Thanks @Dylan-DPC. I think we should have closed the issue after merging #361.

PR #361, opened 2017-06-24
Add error-chain throughout the codebase

Michael-F-Bryan:
To help with #14 (better error messages), I've included `error-chain` as a dependency and updated almost all the places we return a `Result` to use the new error type instead of `Box<Error>` or `std::io::Error`.

I skipped the handlebars helpers though. Handlebars expects you to return a specific error type, so you can't use the `error-chain` error type there.

Now we just need to let people know to use the new error type when returning a `Result`, and that they should use `chain_err()` to return an error while also attaching a human-readable explanation of what went wrong.


Michael-F-Bryan:
Travis just errored on the `x86_64-unknown-linux-musl` build with a rather interesting error message. Apparently while trying to compile `backtrace-sys` (a dependency of `error-chain`) the build script failed with:

```
configure: error: in `/home/travis/build/azerupi/mdBook/target/x86_64-unknown-linux-musl/debug/build/backtrace-sys-204dc57c91e9a514/out':
configure: error: C compiler cannot create executables
```

This may be an issue in the environment travis uses because it happened on stable, beta, and nightly for just that platform. For now I'll see if excluding the `backtrace` feature for that target platform works.

Michael-F-Bryan:
Yay, travis is passing again :) 

@azerupi and @budziq, I had to add a couple things to the `x86_64-unknown-linux-musl` build rules in `travis.yml` to get this to work. `error-chain` depends on `backtrace-sys` and on the default travis image you can't compile C libraries using musl. I had to add the following to `travis.yml` for all `x86_64-unknown-linux-musl` targets so `backtrace-sys` will build properly.

```yml
dist: trusty
addons:
  apt:
    packages: &musl_packages
      - musl
      - musl-dev
      - musl-tools
```

azerupi:
Thanks! Looks good :)

src/lib.rs
```diff
@@ -90,3 +93,21 @@ pub mod utils;
...
+        foreign_links {
+            Io(::std::io::Error);
+            HandlebarsRender(::handlebars::RenderError);
+            HandlebarsTemplate(::handlebars::TemplateError);
+            Utf8(::std::string::FromUtf8Error);
+        }
+
+        errors {
+            Subprocess(message: String, output: ::std::process::Output) {
+                description("A subprocess failed")
```
messense: This caused `mdbook test` only print on failure

```
[*]: Testing file: "/Users/messense/Projects/PyO3/guide/src/./conversions.md"
An error occured:
A subprocess failed
```

budziq:
*@messense this is serious regression. would you be willing to open an
issue?*

PR #362, opened 2017-06-24
Make HtmlConfig is no longer optional

budziq:
`HtmlConfig` was both guaranteed to exist within `BookConfig`
and `expect`ed in few places.
This simplifies the API a little by representing the fact that
`HtmlConfig` is curently mandatory for proper mdbook binary operation.

Please feel free to drop this PR if it does not match the project vision (it did not take a lot of effort anyway ;) ).

### DIFF
--- a/src/config/bookconfig.rs
+++ b/src/config/bookconfig.rs
@@ -5,7 +5,7 @@ use super::tomlconfig::TomlConfig;
 use super::jsonconfig::JsonConfig;
 
 /// Configuration struct containing all the configuration options available in mdBook.
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct BookConfig {
     root: PathBuf,
     source: PathBuf,
@@ -17,7 +17,7 @@ pub struct BookConfig {
     multilingual: bool,
     indent_spaces: i32,
 
-    html_config: Option<HtmlConfig>,
+    html_config: HtmlConfig,
 }
 
 impl BookConfig {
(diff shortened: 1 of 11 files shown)


src/config/bookconfig.rs
```diff
@@ -5,7 +5,7 @@ use super::tomlconfig::TomlConfig;
 use super::jsonconfig::JsonConfig;
 
 /// Configuration struct containing all the configuration options available in mdBook.
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
```
Havvy: Why are these being removed?

budziq: `BookConfig` is never serialized or deserialized directly (only through filling the fields ``)`BookConfig` is never directly serialized nor deserialized. This is done by copying fields to/from `TomlConfig`. So I've decided to clean it up to avoid misconception that it actually is.

I should probably find other instances of such spurious derives.

Havvy: That should probably be in a different commit than this one then, just to keep commit history clean. Course, I'm just a nobody, so take my recommendation with a grain of salt.

budziq: > That should probably be in a different commit than this one then, just to keep commit history clean. 

Good point. ATM this was just an exploratory PR for further review (and honestly I've removed the derives to check if this will sill compile). I'll cleanup if the PR will ever have a chance of merging.

> Course, I'm just a nobody, so take my recommendation with a grain of salt.

Nope. All help and another pair of eyes is always welcome. Thanks :+1: !

azerupi:
The reason `HtmlConfig` is optional is in anticipation of having multiple renderers. When having multiple renderers, we can't assume that the HTML renderer will be enabled. But the actual design will certainly change before multiple renderers can work, so maybe I'm skipping too much ahead.

budziq:
> But the actual design will certainly change before multiple renderers can work, so maybe I'm skipping too much ahead.

I venture that the renderer would be some kind of trait object and not having one set would be a hard unrecoverable error anyway that would bailout to the user immediately. Well unless we want to handle more than one renderer at any given time but then still a list of trait objects might be prefferable. 

If the design is likely to change we might consider making it simpler for rewrite? On the other hand  I'm not really attached to this PR so I won't complain if it is dropped :)

azerupi:
>unless we want to handle more than one renderer at any given time

That is the idea, but the details are still a little vague.

We can make it easier for now indeed. I don't really mind as it will be almost certainly changed again in the future. :)

budziq:
rebased

azerupi:
Damn, I think your clippy PR made this one have conflicts again... 😋  Sorry about that.

budziq:
Done. No problem fixing these conflicts takes only a second :smile:

azerupi:
Ugh.. I'm sorry. It should be the last time 😕

budziq:
done :)

PR #363, opened 2017-06-24
Make MathJax support optional

budziq:
to enable add following to book.toml
```toml
[output.html]
mathjax-support = true
```

fixes https://github.com/azerupi/mdBook/issues/347


azerupi:
Thank you :)

PR #364, opened 2017-06-25
Correct clippy complaints

budziq:
I've run clippy over the code and corrected all of its warnings.

Transition from [`ok_or`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or) to [`ok_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else) might be non obvious but the first one calls the "or" branch unconditionally (suggested only if Err value is already available) while the second one evaluates it lazily.

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -77,7 +77,7 @@ impl HtmlHandlebars {
                     ctx.book.write_file(filename, &rendered.into_bytes())?;
 
                     if ctx.is_index {
-                        self.render_index(&ctx.book, ch, &ctx.destination)?;
+                        self.render_index(ctx.book, ch, &ctx.destination)?;
                     }
                 }
             },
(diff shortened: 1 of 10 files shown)


src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -121,12 +121,10 @@ impl HtmlHandlebars {
...
-        let rendered = build_header_links(rendered, "print.html");
-        let rendered = fix_anchor_links(rendered, "print.html");
-        let rendered = fix_code_blocks(rendered);
-        let rendered = add_playpen_pre(rendered);
-
-        rendered
+        let rendered = build_header_links(&rendered, "print.html");
+        let rendered = fix_anchor_links(&rendered, "print.html");
+        let rendered = fix_code_blocks(&rendered);
+        add_playpen_pre(&rendered)
```
azerupi: I understand why Clippy suggests this, but I am of the opinion that the previous code is actually somewhat clearer and more coherent in this case. No?

budziq: Well some of the changes might be arguable but these things tend to be subjective. I'm not really attached to this work As I mostly wanted to take clippy and rustfix for a spin. Just tell me which changes you don't like and I'll drop these :)

budziq:
> could you run rustfmt on the changes perhaps?

Gladly but I would rather wait for some of the PR's to be merged first I would not like to make spam all other contributors with merge conflicts ;)

src/renderer/html_handlebars/hbs_renderer.rs
azerupi: Just this one, the rest looks good 👍

azerupi:
Yes, indeed. In that case let's merge this and make a new rustfmt PR that we can merge when the PR count is a little lower :)

budziq:
updated :)

azerupi:
Thanks!

PR #369, opened 2017-06-28
Increased default logging level to info unless RUST_LOG is set

budziq:
Sets the `INFO` as the default logging level shown to stdout unless `RUST_LOG` env var is set.
fixes https://github.com/azerupi/mdBook/issues/358

TLNR:
---
- we can merge this PR (no colors and verbose output)
- if we can live with colored output working only on unix'y systems then adding it will take ~10lines with ansi_term
- if we want colors but can live without support for RUST_LOG we can use [simplelog](https://github.com/drakulix/simplelog.rs)
- we might leave the logging as is and add "cargo like" split logging based on [termcolor](https://github.com/BurntSushi/ripgrep/tree/master/termcolor) or [term](https://github.com/Stebalien/term) for the mdBook binary commands and error output) 

Discussion:
---

I'm not really happy with the effects of this PR. The output starts to be overly verbose.
The original suggestion to use [pretty_env_logger](https://github.com/seanmonstar/pretty-env-logger) was misplaced due to two problems:
- it does not expose the [`LogBuilder::filter`](https://docs.rs/env_logger/*/env_logger/struct.LogBuilder.html#method.filter) so we cannot set the default level
- it uses [ansi_term](https://github.com/ogham/rust-ansi-term) for coloring which will not work on no-unix platforms

While the first is trivially circumvented by using env_logger directly, the second is not solvable if we want to stick to env_logger or any of its derivatives as these are incompatible with portable colored printing (all of them take preformatted string while on windows you need to directly manipulate terminal and stdio to have colored output)

On the other hand other colored loggers don't support log triggering/filtering based on RUST_LOG which might be important for consistency.

Cargo uses [term](https://github.com/Stebalien/term) while ripgrep uses [termcolor](https://github.com/BurntSushi/ripgrep/tree/master/termcolor) crate to achieve colors in a portable manner.

We could use one of these in our own logger (or a backend for one of the more advanced loggin frameworks like fern or log4rs) But it feels too involved and we would need to reimplement env_logger's env var parsing.

All of these approaches seam forced at best and now I can see why cargo is using split logging (1 standard log for library part and 2 direct printing with [termcolor](https://github.com/BurntSushi/ripgrep/tree/master/termcolor) for direct printing to stdout of cargo binary messages)

### DIFF
--- a/src/bin/mdbook.rs
+++ b/src/bin/mdbook.rs
@@ -10,6 +10,8 @@ use std::ffi::OsStr;
 use std::io::{self, Write};
 use std::path::{Path, PathBuf};
 use clap::{App, ArgMatches, AppSettings};
+use log::{LogRecord, LogLevelFilter};
+use env_logger::LogBuilder;
 
 pub mod build;
 pub mod init;
@@ -22,7 +24,7 @@ pub mod watch;
 const NAME: &'static str = "mdbook";
 
 fn main() {
-    env_logger::init().unwrap();
+    init_logger();
 
     // Create a list of valid arguments and sub-commands
     let app = App::new(NAME)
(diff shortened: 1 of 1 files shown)


azerupi:
>I'm not really happy with the effects of this PR. The output starts to be overly verbose.

We can change that by using using `info`, `warn` and `error` sparingly. I think there are a lot of messages that can be moved from `info` to `debug` to hide them. I will check this a little later.

I agree! Coloured output should work on all the platforms we support, I don't want to play favourites. 😊  I'm fine with having uncoloured output for now until we find a better solution.

PR #372, opened 2017-07-04
Rewrite of {{#}} links handling in preprocess module

budziq:
First commit is just moving playpen.rs `renderer` helper to `preprocess/links.rs` module

Second commit rewrites the {{#...}} links handling:
- Replaced link parser with a Regex
- Implemented {{#include ...}} links
- now {{#playpen...}} supports arbitrary number of arbitrary properties (like normal fenced code blocks)
- Will display relatively nice error when cannot open {{#...}} linked file
- Escaped links no longer render with escape char

Additionally:
- utils::fs::file_to_path now takes AsRef<Path>
- sorted export/mod in lib.rs

fixes https://github.com/azerupi/mdBook/issues/140
fixes https://github.com/azerupi/mdBook/issues/245


budziq:
@azerupi could you look at it? I would not merge new functionality without someone else's approval.

src/preprocess/links.rs
```diff
@@ -0,0 +1,240 @@
...
+        replaced.push_str(&s[previous_end_index..playpen.start_index]);
+        replaced.push_str(&playpen.render_with_path(&path)?);
+        previous_end_index = playpen.end_index;
+    }
+
+    replaced.push_str(&s[previous_end_index..]);
+    Ok(replaced)
+}
+
+#[derive(PartialOrd, PartialEq, Debug)]
```
azerupi: `Clone` should probably be derived too

src/preprocess/links.rs
```diff
@@ -0,0 +1,240 @@
...
+enum LinkType<'a> {
+    Escaped,
+    Include(PathBuf),
+    Playpen(PathBuf, Vec<&'a str>),
+}
+
+#[derive(PartialOrd, PartialEq, Debug)]
```
azerupi: Idem, derive `Clone`

src/preprocess/links.rs
```diff
@@ -0,0 +1,240 @@
...
+                return Some(inc);
+            }
+        }
+        None
+    }
+}
+
+fn find_links(contents: &str) -> LinkIter {
+    // lazily compute following regex
+    // r"\\\{\{#.*\}\}|\{\{#([a-zA-Z0-9]+)\s*([a-zA-Z0-9_.\-:/\\\s]+)\}\}")?;
```
azerupi: That is some particularly complex regex, are you sure you want to include it like this?
Most of the complexity comes from the backslashes, but still I am not sure it is very maintainable.

As a matter of fact, when trying to understand the regex I noticed that the space between the `link type` and the `target path` is optional. Making this: `{{#playc}}` match as `play`and `c` which is probably not what you want?

src/preprocess/links.rs
```diff
@@ -0,0 +1,240 @@
...
+        }
+    }
+}
+
+struct LinkIter<'a>(CaptureMatches<'a, 'a>);
+
+impl<'a> Iterator for LinkIter<'a> {
+    type Item = Link<'a>;
+    fn next(&mut self) -> Option<Link<'a>> {
+        for cap in &mut self.0 {
```
azerupi: Is this not going to start again from the beginning when next is called? Or is the 'state' kept by the `CaptureMatches`?

src/preprocess/links.rs
```diff
@@ -0,0 +1,240 @@
...
+fn find_links(contents: &str) -> LinkIter {
+    // lazily compute following regex
+    // r"\\\{\{#.*\}\}|\{\{#([a-zA-Z0-9]+)\s*([a-zA-Z0-9_.\-:/\\\s]+)\}\}")?;
+    lazy_static! {
+        static ref RE: Regex = Regex::new(r"(?x) # insignificant whitespace mode
+                    \\\{\{\#.*\}\}               # match escaped link
+                    |                            # or
+                    \{\{\s*                      # link opening parens and whitespace
+                      \#([a-zA-Z0-9]+)           # link type
+                      \s*                        # separating whitespace
```
budziq: yep it should be a `\s+`

src/preprocess/links.rs
budziq: right `\s+` should be used here

src/preprocess/links.rs
budziq: right `CaptureMatches` is keeps the iteration state here

src/preprocess/links.rs
budziq: > Most of the complexity comes from the backslashes, but still I am not sure it is very maintainable.

Right it might not be the easiest to understand or modify but we are expressing quite complex grammar here (and I don;t think the regex approach could be made much simpler).

We can either:

- keep the regex approach, add better comments and a set of comprehensive tests 
- add a separate parser code using some high level library or parsing and add a set of comprehensive tests ;)

src/preprocess/links.rs
azerupi: Ok, thanks :)

src/preprocess/links.rs
budziq: will do

src/preprocess/links.rs
budziq: will do

src/preprocess/links.rs
azerupi: I am fine with keeping the regex, other potential contributors might not be as comfortable with it though, so it's something to keep in mind. :) 

Ultimately, I think we should change the syntax we use for special mdBook syntax (plugins). But until then, we can keep this regex and a good set of tests.

budziq:
ok pushed the actionable changes. I can add more/better comments to the regex and parser if needed (but most likely today)

azerupi:
Thanks, for me it is good to merge :)

budziq:
Thanks!

PR #377, opened 2017-07-10
Make themes module more DRY

Michael-F-Bryan:
I had a quick look at the themes module and found some easy things to clean up. I'm just replacing 10 copy/pasted "populate this variable with a file's contents if it exists" blocks with a loop and helper function.

I also created a `Default` impl for `Theme` which will use the static variables embedded in the binary.


Michael-F-Bryan:
Ooh, I think I found a bug in the old `Theme::new()` method! 

I added a test where it'll go to a temporary directory and create empty versions all the non-Rust files vendored in `themes`... It looks like `jquery` doesn't get overridden when the user provides their own version.

Michael-F-Bryan:
@azerupi or @budziq, any chance you can have a look at this? 

I've gone over the `Theme::new()` constructor, removing a lot of the repetition by removing the copy/pasted `File::open(); thing.clear(); file.read_to_end(thing)` blocks and instead used a table of `(filename, &mut destination)`. 

I also found a bug in the previous version where the user couldn't override the version of `jquery` to use. Now it'll search for something which looks like `jquery-2.1.4.min.js` (where the version number and `.min` are optional) and use that file if it exists (the `search_dir_for_js` and `looks_like_js` functions).

Hopefully this preserves all the old behaviour (excluding the `jquery` bug), but it definitely needs another set of eyes to look at it in case I did something silly.

azerupi:
Looks very good to me :)
I'm not sure if @budziq  wants to take a look, but for me it is good to merge.

budziq:
I'd love to take a look but I'll have time later today. :)

src/theme/mod.rs
```diff
@@ -25,6 +25,7 @@ pub static FONT_AWESOME_WOFF: &'static [u8] = include_bytes!("_FontAwesome/fonts
 pub static FONT_AWESOME_WOFF2: &'static [u8] = include_bytes!("_FontAwesome/fonts/fontawesome-webfont.woff2");
 pub static FONT_AWESOME_OTF: &'static [u8] = include_bytes!("_FontAwesome/fonts/FontAwesome.otf");
 
+
```
budziq: well this commit was unneeded ;)
It was enough to amend previous commit and push with force. Travis would pick it up again

src/theme/mod.rs
```diff
@@ -126,4 +126,38 @@ mod tests {
...
+        // "touch" all of the special files so we have empty copies
+        for special_file in special_files {
+            let filename = temp.path().join(special_file.file_name().unwrap());
+            let _ = File::create(&filename);
+        }
+
+        let got = Theme::new(temp.path());
+
+        // if you can make this less repetitive, please do!
+        assert!(got.index.is_empty());
```
budziq: This might get hard to keep in sync with the actual `Theme` struct contents.
We might make an additional empty constructor for Theme or just construct empty instance. It is minimally less repetitive (subjectively) and forces us to keep the test in sync.

how about?
```rust
        let empty = Theme {
            index: vec![],
            css: vec![],
            favicon: vec![],
            js: vec![],
            highlight_css: vec![],
            tomorrow_night_css: vec![],
            ayu_highlight_css: vec![],
            highlight_js: vec![],
            clipboard_js: vec![],
            store_js: vec![],
            jquery: vec![],
        };

        let got = Theme::new(temp.path());

        assert_eq!(got, empty);
```

src/theme/mod.rs
```diff
@@ -69,7 +72,18 @@ impl Theme {
...
+                if let Err(e) = load_file_contents(&filename, dest) {
+                    warn!("Couldn't load custom file, {}: {}", filename.display(), e);
+                }
+            }
+        }
+
+        // jquery is special because it can often have multiple forms
+        // ("jquery-2.1.4.min.js", "jquery.js", "jquery.min.js", etc), so we
+        // need to manually check the theme_dir for a "valid" filename.
+        if let Some(jquery) = search_dir_for_js(&theme_dir, "jquery") {
```
budziq: I have a problem with all this jquerry related machinery. I would just remove all of it (sadly large part of this commit) and rename `jquery-2.1.4.min.js` to simple `jquery.js` we don't keep the original version numbers in the filenames for other js anyway. Also tracking the version via filename is quite brittle (most likely we could use one of the js dependency managers like npm - we use it for stylus anyway).

src/theme/mod.rs
```diff
@@ -61,75 +110,130 @@ impl Theme {
...
+        let temp = TempDir::new("mdbook").unwrap();
+
+        // "touch" all of the special files so we have empty copies
+        for special_file in special_files {
+            let filename = temp.path().join(special_file.file_name().unwrap());
+            let _ = File::create(&filename);
         }
 
-        theme
+        let got = Theme::new(temp.path());
```
budziq: How about?

```rust
        let empty = Theme {
            index: vec![],
            css: vec![],
            favicon: vec![],
            js: vec![],
            highlight_css: vec![],
            tomorrow_night_css: vec![],
            ayu_highlight_css: vec![],
            highlight_js: vec![],
            clipboard_js: vec![],
            store_js: vec![],
            jquery: vec![],
        };

        let got = Theme::new(temp.path());

        assert_eq!(got, empty);
```

This might be subjectively a little less repetitive and forces us to actually keep the test in sync with `Theme` struct impl which otherwise be hard

budziq:
Hmm also I'm not sure if `Theme` should not live closer to `HbsRenderrer` as it makes sense only in its context. I venture that even other hypothetical html based renderers (like Epub/Mobi) might prefer their own configuration.
Whatever the decision, this might be left for another PR.

Michael-F-Bryan:
> Hmm also I'm not sure if Theme should not live closer to HbsRenderrer as it makes sense only in its context.

I agree. It'd probably be better for each renderer to look after its own themes although that's an issue we can deal with later on.

>  I would just remove all of it (sadly large part of this commit) and rename jquery-2.1.4.min.js to simple jquery.js

I wasn't sure whether keeping the version number for jquery was important, so I erred on the side of caution and added all that regex stuff. I'm fine with removing it in favor of using the simpler `jquery.js` filename, after all the easiest code to maintain is code that doesn't exist.

src/lib.rs
```diff
@@ -3,8 +3,7 @@
 //! **mdBook** is similar to Gitbook but implemented in Rust.
 //! It offers a command line interface, but can also be used as a regular crate.
 //!
-//! This is the API doc, but you can find a [less "low-level" documentation
```
Michael-F-Bryan: haha yeah, I've got rustfmt set up to run every time I save. `wrap_comments = true` is set in the `rustfmt.toml` so it'll try to wrap all comments at 80 characters, which adds a bunch of noise to PRs because `lib.rs` has a few really long lines.

budziq:
This looks very nice and ready to merge :) Great work :+1:

PR #380, opened 2017-07-24
Fixed wrong filename when post processing html

ffissore:
Since b7aa78c3c0c739b8166b6e21927ae522ade80f7b, the file name `print.html` was hard coded into `HtmlHandlebars.post_process`, hence anchors in chapters were all in the form `http://web site/print.html#paragraph title`

This PR fixes it


azerupi:
Sorry for the delay!
Good catch, thanks a lot for the fix :)

PR #381, opened 2017-07-25
Better ignores

dns2utf8:
I made a list inside the function because I think it is more maintainable than a long &'static str

### DIFF
--- a/src/book/mod.rs
+++ b/src/book/mod.rs
@@ -221,7 +221,14 @@ impl MDBook {
 
             debug!("[*]: Writing to .gitignore");
 
+            // Write the destination folder to file
             writeln!(f, "{}", relative).expect("Could not write to file.");
+            
+            // Add common files
+            let ignores = vec!("*~", ".DS_Store");
+            for pattern in ignores {
+                writeln!(f, "{}", pattern).expect("Could not write to file.");
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


dns2utf8:
The build currently fails on mac with the following error:
```
+rustup default beta
info: syncing channel updates for 'beta-x86_64-unknown-linux-gnu'
error: missing key: 'url'
```
It looks like a rustup error. Right?

budziq:
yep this is a rustup problem. I'll look into your PR's when I'm back from holidays next week. Cheers!

budziq:
yep this is a rustup problem. I'll look into your PR's when I'm back from holidays next week. Cheers!

src/book/mod.rs
```diff
@@ -221,7 +221,14 @@ impl MDBook {
 
             debug!("[*]: Writing to .gitignore");
 
+            // Write the destination folder to file
             writeln!(f, "{}", relative).expect("Could not write to file.");
+            
+            // Add common files
+            let ignores = vec!("*~", ".DS_Store");
```
budziq: Hi @dns2utf8 thanks for the submission.

I'm not convinced that we should try to add "common" (or any for that matter) ignore patterns here. We would not like to start catering to requests for adding everyone's favorite temp file formats as this is trivially handled by [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore). This `.gitignore` serves a specific purpose and it cannot be easily handled via global ignore patterns (actually you can build a whole hierarchy of gitignores).

Could you give a usecase that cannot be handled by global patterns?.

azerupi: I agree, temporary files are very specific to the environment of the user, we can not possibly support all environments and it should not be our responsibility to do so.

dns2utf8:
I can see that.

PR #382, opened 2017-07-26
fix travis ci for beta channel

budziq:
travis started to fail for beta channel. This is a first idea for fix (do not merge yet as I have no way to test it on mobile)


budziq:
it is reemergence of https://github.com/rust-lang-nursery/rustup.rs/issues/1062

PR #393, opened 2017-08-06
handle play.rust-lang.org communication errors in playpens

budziq:
also add 15s communication timeout

fixes https://github.com/azerupi/mdBook/issues/392

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -329,8 +329,12 @@ function run_rust_code(code_block) {
         dataType: "json",
         contentType: "application/json",
         data: JSON.stringify(params),
+        timeout: 15000,
         success: function(response){
             result_block.text(response.result);
-        }
+        },
+        error: function(qXHR, textStatus, errorThrown){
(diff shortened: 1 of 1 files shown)


azerupi:
What does this print for the different cases? :)

budziq:
"Playpen communication error"
"Playpen communication timeout"
"Playpen communication abort"

actually I did not find much interesting info in http://api.jquery.com/jquery.ajax/ that would allow to further differentiate.

azerupi:
Seems good to me! :+1: 
We can always refine the messages over time. :)

Thanks!

shepmaster:
> Playpen

FWIW, it's normally called the *playground*. "Playpen" was a Very Old name for it

budziq:
> FWIW, it's normally called the playground. "Playpen" was a Very Old name for it

fixed in 
 https://github.com/azerupi/mdBook/commit/c36eca15c2f8d35005dd27571c22043b5bdac4ff

Thanks @shepmaster !

PR #403, opened 2017-08-11
fixed missing playpen css class when codeblock properties had whitespace

budziq:
fixes https://github.com/azerupi/mdBook/issues/397


azerupi:
Nice! Looks good to me :)

If we add more markdown preprocessors we will need to think about a better place to put them than the utils module.

budziq:
yep I was surprised to find it there ;)

PR #404, opened 2017-08-11
Dynamic play button based on `no_run` property and used crates availability 

budziq:
fixes https://github.com/azerupi/mdBook/issues/396

- list of available crates is dynamically loaded from play.rust-lang.org
- play button is enabled only if crates used in snippet are available on playground
- ACE editor's play button is dynamically updated on each text change
- `no_run` is honored by always disabling the play button
- minor cleanups/refactors

This might not be ready for prime time as currently the play button is just hidden if both on `no_run` and when using crates unavailable on play.rust-lang.org. I would like to eventually just disable the play button with some visual cue (dim/color change) and tooltip describing the state (I guess that It'll wait for https://github.com/azerupi/mdBook/issues/388 ;)).

But I wanted to push the WIP for review as soon as possible as I'll not have much free time on my hands for the next two weeks.

But OTOH it might be worth merging as it is anyway with later fix for nicer play button disabling.

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -1,5 +1,3 @@
-playground_crates = [];
-
 $( document ).ready(function() {
 
     // url
@@ -19,7 +17,7 @@ $( document ).ready(function() {
         tabReplace: '    ', // 4 spaces
         languages: [],      // Languages used for auto-detection
     });
-    
+
     if (window.ace) {
         // language-rust class needs to be removed for editable
         // blocks or highlightjs will capture events
@@ -33,7 +31,7 @@ $( document ).ready(function() {
             hljs.highlightBlock(block);
         });
     }
-    
+
     // Adding the hljs class gives code blocks the color css
     // even if highlighting doesn't apply
     $('code').addClass('hljs');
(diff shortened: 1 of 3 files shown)


src/theme/book.js
```diff
@@ -262,8 +255,79 @@ $( document ).ready(function() {
...
             showTooltip(e.trigger, "Clipboard error!");
     });
+
+    $.ajax({
+        url: "https://play.rust-lang.org/meta/crates",
+        method: "POST",
+        crossDomain: true,
+        dataType: "json",
+        contentType: "application/json",
+        success: function(response){
```
azerupi: What do we do on error?

budziq: @azerupi  Currently nothing. The play buttons will not appear which imho is a safer default than showing play buttons on content that might be unsupported by play.rlo. 
I did not have a reasonable fallback scenario, do you have anything in mind?

azerupi: No, that is perfect :+1: 
Not showing the play button when the playground can not be reached makes a lot of sense :)

src/theme/book.js
```diff
@@ -262,8 +255,79 @@ $( document ).ready(function() {
...
+    if (classes.indexOf("no_run") > -1) {
+        play_button.addClass("hidden");
+        return;
+    }
+
+    // get list of `extern crate`'s from snippet
+    var txt = playpen_text(pre_block);
+    var re = /extern\s+crate\s+([a-zA-Z_0-9]+)\s*;/g;
+    var snippet_crates = [];
+    while (item = re.exec(txt))
```
azerupi: Could you add braces for the while loop?
I tend to avoid brace-less control flow as it is error-prone when refactoring.

src/theme/book.js
```diff
@@ -262,8 +255,79 @@ $( document ).ready(function() {
...
+        snippet_crates.push(item[1]);
+
+    // check if all used crates are available on play.rust-lang.org
+    var all_available = snippet_crates.every(elem => playground_crates.indexOf(elem) > -1);
```
azerupi: I am not sure we should use the arrow syntax (introduced in ES6) for compatibility reasons.

azerupi:
Sorry for the delay, I was quite busy lately.
I went through the PR and except for the two comments, I think it's good! :+1:

budziq:
Both fixes pushed.

> Sorry for the delay, I was quite busy lately.

No problem. I'm the one out of touch these days :)

budziq:
hmm travis does not seam to be likely to finish anytime soon (and the change is js only anyway). I'll merge this one base on appveyor passing.

PR #411, opened 2017-08-29
Fix toml problem and release 0.0.23

steveklabnik:
The toml crate removed its serde dependency in 0.3, but in 0.4, it
went away. Cargo didn't warn on this, but now it does. As such, we
need a release so that rust's build doesn't warn constantly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mdbook"
-version = "0.0.22-pre"
+version = "0.0.23"
 authors = ["Mathieu David <mathieudavid@mathieudavid.org>"]
 description = "create books from markdown files (like Gitbook)"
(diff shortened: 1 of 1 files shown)


steveklabnik:
published https://github.com/azerupi/mdBook/releases/tag/0.0.23

azerupi:
Does this include all the breaking changes?
I guess it doesn't matter much because at this stage every patch version is considered breaking but.. 

We should get the docs in order and push the `0.1.0` release :)

steveklabnik:
It does, yeah. given the point releases, that seemed okay to me.

> We should get the docs in order and push the 0.1.0 release :)

Agreed; this is a release week so it's very busy for me. Maybe I could pitch in some next week.

steveklabnik:
So, re-reading this:

> Does this include all the breaking changes?

Do we have a summary of what these are?

PR #413, opened 2017-08-30
[book] Prevent over-matching in gitignore rule

behnam:
To only ignore the output destination (default: `book`) and no other
file/directory with the same name under the mdbook root, we should
prefix the gitignore rule with a leading slash (default: `/book`).


azerupi:
Thank you very much! :)

PR #415, opened 2017-09-01
Fix the print title that was using the title from the last rendered chapter

### LINKED ISSUE #414: Incorrect title for print.html
Looks like `print.html` inherits the title of the last item in the SUMMARY, instead of only only showing the book title.

Example: <http://code.behnam.es/rust-cargo/src/doc/book/book/print.html> has
```html
<title>FAQ - The Cargo Manual</title>
```
where it's expected to be just
```html
<title>The Cargo Manual</title>
```

azerupi:
This should fix #414

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -51,9 +51,17 @@ impl HtmlHandlebars {
                     io::Error::new(io::ErrorKind::Other, "Could not convert path to str")
                 })?;
 
+                // Non-lexical lifetimes needed :'( 
+                let mut title: String;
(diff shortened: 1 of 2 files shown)


azerupi:
>Most likely I would go with 3 variables "chapter_title", "book_title" and "title" 

We can do that, it's a good idea :+1:

azerupi:
I made a new commit, but it doesn't seem to show up here... I'm going to try closing and reopening to see if it changes. Otherwise I will make a new PR.

azerupi:
This new commit implements your suggestions and works as expected when I tested. So I am going to merge this so that it makes it into the 0.0.24.

PR #418, opened 2017-09-01
[Cargo.toml] Fix package.exclude warnings

behnam:
IIUC, the existing exclude rule has meant to do what it will do in the
future (gitignore-like matching, not glob-only matching). This fix makes
the rule to do what it was expected to do, and is forward-compatible,
therefore fixing the warning messages.


azerupi:
Thanks!

PR #419, opened 2017-09-02
Fix heading links in nested pages

behnam:
Plus fixing the whitespace chars not being replaced by hyphen.

Also expand tests for link creations, and add test for nested pages.

Fixes <https://github.com/azerupi/mdBook/issues/416>
Fixes <https://github.com/azerupi/mdBook/issues/417>

And, also fix some test warnings by moving non-test module files into their own directories.


azerupi:
Thanks for the PR!
I did a very quick review pass and it looks very good for most parts! Travis fails because of rustup, but appveyor seems to run the tests and fail at `check_correct_cross_links_in_nested_dir`. I will do a more in depth review soon. :)

behnam:
Uh, on windows the `filepath` needs to be changes to use forward-slash, instead of backslash. I'll submit an update.

behnam:
Travis is completely broken right now. I filed this: https://github.com/azerupi/mdBook/issues/421

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -457,7 +458,7 @@ fn id_from_content(content: &str) -> String {
     let mut content = content.to_string();
 
     // Skip any tags or html-encoded stuff
-    let repl_sub = vec![
+    static REPL_SUB: &[&str] = &[
```
Michael-F-Bryan: I'm assuming you meant `const REPL_SUB ...` here. Using `static` means it actually gets embedded in the binary and is typically only done for global mutable objects or if your hardware requires it (like the exception vector when writing a kernel).

I'd also pull it up to the top of the file and name it to something like `HTML_ENCODE_BLACKLIST` (feel free to give it a better name) so it's easy to update later on.

tests/rendered_output.rs
```diff
@@ -28,62 +32,76 @@ fn by_default_mdbook_generates_rendered_content_in_the_book_directory() {
...
     md.build().unwrap();
 
     let dest = temp.path().join("book");
     let links = vec![
-        "intro.html",
-        "first/index.html",
-        "first/nested.html",
-        "second.html",
-        "conclusion.html",
+        r#"href="intro.html""#,
```
Michael-F-Bryan: Nice fix! Without the `href="` this test wouldn't have actually been checking for links, just that the filenames are present, which means it's not actually testing what we want it to.

src/lib.rs
```diff
@@ -100,6 +100,8 @@ pub use book::BookItem;
 pub use renderer::Renderer;
 
 /// The error types used through out this crate.
+// TODO: Drop after error_chain is fixed
+#[allow(unused_doc_comment)]
```
Michael-F-Bryan: This should probably be pulled out into its own PR. `error-chain` has actually fixed this, so bumping the dependency version to `0.11.0-rc.2` would be sufficient.

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -63,14 +63,15 @@ impl HtmlHandlebars {
                 debug!("[*]: Render template");
                 let rendered = ctx.handlebars.render("index", &ctx.data)?;
 
-                let filename = Path::new(&ch.path).with_extension("html");
+                let filepath = Path::new(&ch.path).with_extension("html");
                 let rendered = self.post_process(rendered,
-                    filename.file_name().unwrap().to_str().unwrap_or(""),
+                    &normalize_path(filepath.to_str()
+                        .expect(&format!("Bad file name: {}", filepath.display()))),
```
Michael-F-Bryan: Is it a good idea to panic here? Another option may be to assume a safe default or return an error.

If we were to return an error, I imagine it'd look something like this:

```rust
filepath.to_str()
  .ok_or_else(Error::from("Bad file name: {}", filepath.display()))?`
```

Michael-F-Bryan:
I just had another look through this PR... Are there any tests to make sure `normalize_id()` correctly translates titles containing spaces get converted to using hyphens? 

It might also be useful to add an integration test in `tests/rendered_output.rs` to make sure the chapter title `Chapter 1` gets a hyphenated `id` (i.e. `<h1 id="chapter-1">Chapter 1</h1>`).

src/lib.rs
steveklabnik: yeah; I did this on one of my recent projects

src/renderer/html_handlebars/hbs_renderer.rs
behnam: Good point! I would expect non-Unicode file names would be caught somewhere earlier in the lib, but in case they are not, makes sense to return a nice error here.

src/renderer/html_handlebars/hbs_renderer.rs
behnam: Ooop! Right.

behnam:
> I just had another look through this PR... Are there any tests to make sure normalize_id() correctly translates titles containing spaces get converted to using hyphens?

Yes, the newly added "## Some Section" to the nested `first` pages actually tests SPACE being converted to HYPHEN, and in fact makes sure both `href` and `id` are set correctly.

I'm submitting a fix for the inline comments, and I think it would be good to land.

azerupi:
Thank you for fixing the issues raised by @Michael-F-Bryan !
I will try to take a last look at this PR tonight to merge it ASAP.

steveklabnik:
Awesome, and then we should cut a release.

azerupi:
Thanks a lot for this PR @behnam!
I will publish a new release with the fixes in a couple of minutes.

PR #422, opened 2017-09-03
Regression tests

Michael-F-Bryan:
I've started writing a bunch of tests which build the example book and check to see "significant" parts are still present. This *should* help us with regression testing because if something is accidentally broken with in the rendered output, Travis will be able to pick it up. Similar to @steveklabnik's #237.

At the moment I'm just checking the TOC structure, but I'll keep adding other things. I want to also check samples from random chapters are present in the output, and make sure things like the playground buttons and next/previous links are present.

This'll hopefully help with #409 because I know I've broken at least some of the output. It does add some maintenance cost though, seeing as when we update the example book or the rendered output you may need to revisit these tests and update them.


Michael-F-Bryan:
@azerupi or @budziq, any chance you can give this a quick review? It looks like I've let it bitrot a bit, but it'll be nice to know if I'm going in the right direction :)

tests/regression.rs
```diff
@@ -0,0 +1,151 @@
...
+
+mod dummy;
+
+use std::path::Path;
+use walkdir::{DirEntry, WalkDir, WalkDirIterator};
+use select::document::Document;
+use select::predicate::{Class, Name, Predicate};
+
+
+const BOOK_ROOT: &'static str = concat!(env!("CARGO_MANIFEST_DIR"), "/book-example");
```
budziq: I would suggest to using something more stable than the book-example. It is our de facto documentation which will naturally grow and having it rigorously tested against static tests might make a little painful to edit.
How about using something a little more synthetic and static with which we could test several ugly edge-cases and for instance check if:
- playpens with given info-strings results in required css classes
- if js libs are imported if given option in book.toml is given.
- ???

Actually, we might consider using some non rust solutions for web regression testing (I'm no webdev but I've seen some really impressive magic done with selenium)

Michael-F-Bryan: Yeah, my first thought was to use Python and BeautifulSoup because it makes it loads easier to navigate HTML documents and find the things you want, I was a little hesitant to add more complication and a second language to the testing process though. I'll have a look at what other projects do to verify rendered HTML/CSS and check for regressions.

In terms of making things more static, would it be a good idea to rewrite this based on the "dummy book" instead? It'd mean the integration tests need to be updated whenever you change it so you can test for regressions, but that should give a nicer experience than tests breaking whenever we update documentation.

steveklabnik: There are libraries similar to BeautifulSoup in Rust

budziq: > In terms of making things more static, would it be a good idea to rewrite this based on the "dummy book" instead?

Yeah. Im not even sure we need to create it dynamically. Why not have the dummy book as a directory structure fixture similar to book-example which we could easily expand 

> There are libraries similar to BeautifulSoup in Rust

These tests already use the [`select`](https://crates.io/crates/select) (nice wrapper over `html5ever`) crate which is as close to bs4 as we can get now. I was thinking more in terms of actual usability tests with headless [selenium](http://www.seleniumhq.org) but might be too ambitious for now.

Michael-F-Bryan: > Why not have the dummy book as a directory structure fixture similar to book-example which we could easily expand

I like this idea. Originally I was dynamically generating the dummy book in a temporary directory so each test is completely independent from the others, but there's no reason why we can't just do a `cp -r` into a temp directory.

Michael-F-Bryan:
@steveklabnik, do you have some sort of link checker or other method for checking rendered documentation in The Rust Book? 

I'm trying to think of various ways we can do sanity checks to make sure a change in mdbook doesn't accidentally break rendered documents without us noticing.

steveklabnik:
It's here: https://github.com/rust-lang/rust/tree/master/src/tools/linkchecker

There's talk of putting it on crates.io, but I don't know if that's happened.

Michael-F-Bryan:
**Note to self:** the regression tests need to be rewritten so they depend upon a dummy book instead of the example documentation. 

Also rewrite the `dummy` and `helper` modules so creating a new copy of the dummy book is a simple `cp -r` to a temporary directory instead of the current `include_str!()` madness. I'd still like to make sure each regression test uses its own copy of the dummy book in a temp dir though. That way tests can't accidentally overwrite the original or interfere with each other.

Michael-F-Bryan:
I revised this so it now exercises a contrived "dummy book" instead of the example book we use as the de-facto documentation for mdbook. So far it's just checking the TOC is correct and that a bunch of known links and CSS classes are present in files. 

It's by no means complete, but now the foundations are established we can add to the regression tests over time when people report bugs. @budziq or @steveklabnik, any chance you can skim through this and let me know what you think? 

The regression tests themselves are in `tests/rendered_output.rs`. There's a bit of code churn because I renamed `tests/dummy` to `tests/dummy_book` to make it clearer this is a contrived book for testing.

budziq:
> budziq or steveklabnik, any chance you can skim through this and let me know what you think?

@Michael-F-Bryan Wow you're going fast :) 
I most likely will not be able to keep up with the pace but will try to find some time next week (sadly I could steal only few minute timeslices this weekend)

Michael-F-Bryan:
> @Michael-F-Bryan Wow you're going fast :)

I think it was just good timing in that I had the entire weekend free when things started happening :smiley: I'm not sure how much I'll get done during the week due to work and all that though.

I'm also pretty keen to get the `Book` refactoring PR back in motion, but that depends on this PR so I can detect if I accidentally broke stuff... Which I have a sneaking suspicion that I have.

tests/rendered_output.rs
```diff
@@ -111,3 +124,100 @@ fn chapter_content_appears_in_rendered_document() {
...
+
+#[test]
+fn check_second_toc_level() {
+    let doc = root_index_html();
+    let mut should_be = Vec::from(TOC_SECOND_LEVEL);
+    should_be.sort();
+
+    let pred = descendants!(Class("chapter"), Name("li"), Name("li"), Name("a"));
+
+    let mut children_of_children: Vec<String> =
```
budziq: `Vec<_>` type ascription should be enough

tests/rendered_output.rs
```diff
@@ -111,3 +124,100 @@ fn chapter_content_appears_in_rendered_document() {
...
+fn check_first_toc_level() {
+    let doc = root_index_html();
+    let mut should_be = Vec::from(TOC_TOP_LEVEL);
+
+    should_be.extend(TOC_SECOND_LEVEL);
+    should_be.sort();
+
+    let pred = descendants!(Class("chapter"), Name("li"), Name("a"));
+
+    let mut children: Vec<String> = doc.find(pred).map(|elem| elem.text().trim().to_string())
```
budziq: `Vec<_>` type ascription should be enough

tests/dummy_book/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+fn poor_mans_sed(filename: &Path, from: &str, to: &str) {
+    let contents = read_file(filename).unwrap();
+    File::create(filename).unwrap()
+                          .write_all(contents.replace(from, to).as_bytes())
+                          .unwrap();
+}
+
+impl Default for DummyBook {
```
budziq: do we need a separate `Default` impl? Why not just implement `new` directly?

tests/dummy_book/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+/// functionality, `$TEST_STATUS` can be substitute for either `true` or
+/// `false`. This is done using the `passing_test` parameter.
+#[derive(Clone, Debug, PartialEq)]
+pub struct DummyBook {
+    passing_test: bool,
+}
+
+impl DummyBook {
+    /// Create a new `DummyBook` with all the defaults.
+    pub fn new() -> DummyBook {
```
budziq: why not implement it directly? It's a single bool.

tests/dummy_book/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /// Write a book to a temporary directory using the provided settings.
+    ///
+    /// # Note
+    ///
+    /// If this fails for any reason it will `panic!()`. If we can't write to a
+    /// temporary directory then chances are you've got bigger problems...
+    pub fn build(&self) -> TempDir {
+        let temp = TempDir::new("dummy_book").unwrap();
```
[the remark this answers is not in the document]
Michael-F-Bryan: Good point, I think I was just being lazy and not wanting to deal with errors properly. It does just end up pushing the `unwrap()`'s up one level to the tests though...

Michael-F-Bryan:
@budziq I went back through and added better error handling to the `DummyBook` mock, now it does proper error propagation and adds more high level context using `chain_err()`.

What do you usually do in test code when the setup stages fail? For example, there are quite a few tests which look like the example below. Do you reckon it's fine to just leave the `unwrap()` as-is (the idea being that our error chain should give sufficient context), or would you usually use an `expect()` instead?

```rust
#[test]
fn chapter_files_were_rendered_to_html() {
    let temp = DummyBook::new().build().unwrap();

    // do stuff and make assertions
}
```

budziq:
@Michael-F-Bryan 
> What do you usually do in test code when the setup stages fail? 

IMHO panic is the best option but I usually prefer to do it at top level if possible or at least with expects instead of unwraps.

Michael-F-Bryan:
@budziq, any idea why the beta mac build is failing? I thought it might just be a temporary thing due to a flaky build, but restarting it gave the exact same error message.

https://travis-ci.org/rust-lang-nursery/mdBook/jobs/301898364

screenshot_2017-11-16_150757

PR #429, opened 2017-09-08
Revive Travis

azerupi:
This PR greatly simplifies the Travis builds and uses the preinstalled Rust packages provided by Travis.

There are less architectures that are tested, no 32-bits and no musl. But it speeds up the builds considerably. The doc deployment and release artefact deployment *should* remain intact but as I can't test it I can't guarantee it works.


.travis.yml
```diff
@@ -1,110 +1,42 @@
...
+  - cargo build --verbose
+  - cargo test --verbose
 
 after_success:
+  # Deploy the docs if the commit is on master
   - test "$TRAVIS_PULL_REQUEST" == "false" &&
     test "$TRAVIS_BRANCH" == "master" &&
-    test "$TARGET" == "x86_64-unknown-linux-gnu" &&
-    test "$CHANNEL" = "stable" &&
+    "$TRAVIS_RUST_VERSION" == "stable" &&
```
behnam: Shouldn't this line start with `test `? I think it's resulting in some false-positives right now.

azerupi: Ah yes, good catch!

Issue #398, opened 2017-08-10
run rustfmt on the repository


budziq:
The codebase could use some minor cleanups:
We should run `rustfmt` on the codebase.

It is a nice introductory task to get someones feet wet in the project with some light mentoring.

please use [rustfmt-nightly](https://crates.io/crates/rustfmt-nightly) crate for the job

azerupi:
The new version of rustfmt that only runs on nightly at the moment. I heard the style changed quite a bit for the better.

budziq:
Good idea. I've updated the description

budziq:
merged

PR #437, opened 2017-09-13
run rustfmt on the repository #398

prertik:
Repository was formatted using rustfmt-nightly. The max_width needed to be increased to remove errors of width being more than the default 120. Some error still persist like:
`error: left behind trailing whitespace`
`---> /mdBook/src/utils/mod.rs:178`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,7 +3,8 @@
 //! **mdBook** is similar to Gitbook but implemented in Rust.
 //! It offers a command line interface, but can also be used as a regular crate.
 //!
-//! This is the API doc, but you can find a [less "low-level" documentation here](../index.html) that
+//! This is the API doc, but you can find a [less "low-level" documentation
+//! here](../index.html) that
 //! contains information about the command line tool, format, structure etc.
 //! It is also rendered with mdBook to showcase the features and default theme.
 //!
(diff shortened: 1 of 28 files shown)


azerupi:
Thank you for this PR!
I am away from home today, but I will go through the changes as soon as possible.

src/lib.rs
```diff
@@ -25,18 +26,23 @@
 //! # #[allow(unused_variables)]
 //! fn main() {
 //!     let mut book =  MDBook::new("my-book")        // Path to root
-//!                         .with_source("src")       // Path from root to source directory
-//!                         .with_destination("book") // Path from root to output directory
-//!                         .read_config()            // Parse book.toml configuration file
-//!                         .expect("I don't handle configuration file errors, but you should!");
+//! .with_source("src")       // Path from root to
```
azerupi: rustfmt seems to have destroyed the code snippets in the doc comments and this results in test failures.
This is certainly a bug in rustfmt.

prertik: Yeah, that seems to be the case.

prertik: So, this means we can't use rustfmt currently?

budziq: > So, this means we can't use rustfmt currently?

Not really. We can pull only some of the changes that improve the code readability. Also it might be a good moment to submit bug report to rustfmt

prertik: Hey @budziq, I closed this pull request and then opened a new pull request fixing the problem. I did write an issue on rustfmt and found the solution to this.

PR #438, opened 2017-09-13
run rustfmt on the repository #398(Updated)

prertik:
Repository was formatted using rustfmt-nightly. The max_width needed to be increased to remove errors of width being more than the default 120. Some error still persist like:
`error: left behind trailing whitespace`
`---> /mdBook/src/utils/mod.rs:178`

and with the fix of code snippets being destroyed from doc comments.


prertik:
@budziq @azerupi how about this one?

azerupi:
The tests passed, I will take a look at the output to see if everything is good :+1:

azerupi:
@prertik I think we should remove the rustfmt.toml file altogether. It was meant for an old version of rustfmt, but it has changed significantly since then and most of the options are probably irrelevant now.

Could I ask you to remove the config file and run rustfmt again? :)

prertik:
Yeah about rustfmt.toml, I had to configure it to pass the test as you can see certain changes. If we remove rustfmt.toml, the rustfmt would operate in default condition which is having max_width:120 which would fail to format [mdBook/src/preprocess/links.rs](https://github.com/azerupi/mdBook/blob/master/src/preprocess/links.rs) as it has codes with max_width more than 120 and secondly, the problem in previous pull request would again arise i.e. rustfmt by default would again format code snippets in the doc comments. I opened an issue on rustfmt about this, a collaborator told me make changes in the rustfmt.toml file.

Michael-F-Bryan:
Wouldn't it make more sense to just shorten the overly long lines instead of increasing the max width above 120 characters? If you have a really long string literal, you can break it over a newline by adding a `\` to the very end.

Like this:

```rust
let some_really_long_string: &'static str = "this is a very long and uninteresting string \ 
  which I've wrapped over multiple lines.";
```

Also, in doc comments you typically want to wrap them at 80 characters anyway. I know there are one or two examples with really long overhanging lines, but it'd probably be better to either move the comments above or delete them outright if they don't add anything extra to the example... I think uncle bob is quoted as saying, "comments are apologies for badly written code [or poor naming]".

azerupi:
Yes, I think it would be better to drop the config file. Resolve the issues with lines being too long and see if the output is acceptable. If the default style works for us, we should embrace it.

prertik:
Thanks @Michael-F-Bryan  for the feedback. Alright @azerupi, I will do that. I will send a new pull request and will close this one.

budziq:
@prertik For future reference you can use `git rebase` and `git push --force` to completely rewrite the PR without need to close and open another one. Actually, I would suggest to reopen this PR and try this way yourself. If you'll have any problems with git I'll help you :+1: .

prertik:
Thanks @budziq. Actually, i used to do `git rebase` before. But, in one repo, I ran into some problems due to some merge conflicts which i couldn't solve. So, i ignored it. But, i will definitely try it again. Surely, will ask you for any git help. 👍

prertik:
hey @azerupi @budziq could you look at the failing `preprocess::links::test_find_playpens_with_properties` test. Why is it failing?

prertik:
If i just decrease the `start_index: 90` and `end_index: 137` to `start_index: 89` and `end_index: 136`. The test passes. Will this solution be feasible?

azerupi:
Yes, that is totally acceptable :+1:

prertik:
Alright I'll commit the changes as soon as I reach home.

prertik:
So, @budziq  what should I do?

prertik:
Actually, the changes done here are done by default config of rustfmt. If we edit the changes currently to our liking which contrasts to default of rustfmt then I guess while running rustfmt later after adding some new codes will revert the lines back to default implementation of rustfmt.

prertik:
Isn't the point of using rustfmt is to follow community standard?

budziq:
@prertik I've just listed my concern, please od not treat my opinion on style as a blocker (if @azerupi likes the new style I'm game).

But in the mean time you might look into [rustfmt configuration docs](https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md) to check if there is a way to alleviate this behaviour (again if we decide that we really don't like it :smile: )

On the other hand in terms of the large numer of commits, I would squash them irregardless of what we decide :smile:

Michael-F-Bryan:
I've found `rustfmt-nightly` tends to be pretty eager with breaking up longer lines as well. I'm still trying to figure out how to stop it from doing this:

```diff
-            debug!("[*]: {:?} does not exist, trying to create .gitignore", gitignore);
+            debug!(
+                "[*]: {:?} does not exist, trying to create .gitignore",
+                gitignore
+            );
```

I was going to make a comment about the large number of commits too, but github gives you an option to "Squash and Merge" when you merge the PR, doesn't it?

prertik:
Yeah @Michael-F-Bryan squash and merge is available. Sorry for the large number of commits. I had a commit for every file I edited using rustfmt or manually.

Michael-F-Bryan:
No stress, it's a big job so whatever workflow works for you is fine :smile: We can always fix up the history when we're done anyway.

Thanks for taking this on by the way! Its something I've been thinking of doing because with the old formatting my editor would reformat files every time I hit save, adding loads of annoying code churn. I just never got around to making a PR and it's not exactly a job which can be done during a lunch break.

budziq:
> github gives you an option to "Squash and Merge" when you merge the PR, doesn't it?

Yeah but this is one more thing for the maintainer to spot and remember, and if we forget one has to fix the repository with `--force` which no one likes (as it leaves people stranded with situation that is hard to recover)

prertik:
Sorry @budziq for the large amount of commits. Next time I will try to keep it at minimum.

budziq:
> Sorry @budziq for the large amount of commits. Next time I will try to keep it at minimum.

No reason to be sorry :smile:. I've just given my rationale (its not even that big of an issue)

prertik:
The only way to stop rustfmt doing this: 
screenshot_20170919_173649 
is to set `fn_call_width` to large value in rustfmt.toml. The only problem with this is this will then keep long lines everywhere. Breaking down some long lines are desired and look readable but this will then ignore it and again to do as we desire, it is a lot of work because of editing every file.

budziq:
> The only way to stop rustfmt doing this

And how about setting [fn_call_style](https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md#fn_call_style) to `visual`?
@prertik

prertik:
Yeah, I guess it takes care of the extra line from block(default). But, I think the block has been made default for a reason i.e. being community standard.

prertik:
Shouldn't we follow community standard? In Ruby, community standard is held at highest order.

budziq:
> But, I think the block has been made default for a reason i.e. being community standard.

Be that as it may, but rustftmt.toml facility was also made for a reason ;). Please note that the rust community standard is in constant flux (4months ago its output on our repo was different and arguably/subjectively nicer). Imho additional 2lines per invocation is just adding noise.
Just my 0.02$ :)

prertik:
So, what should I do @budziq? Should I add rustfmt.toml and redo everything? :(

budziq:
>  Should I add rustfmt.toml and redo everything?

If there would be some changes it would not be redoing everything a single run of 
```bash
find ./src  -iname "*.rs"  | xargs rustfmt
```
 should do the trick (if you're on a nix'y environment)

But anyhow I would wait on a decision from owner :+1:

azerupi:
>and redo everything

Redoing everything should be relatively easy, no? rustfmt comes with a cargo sub-command to run it on the whole crate: `cargo fmt`.

I agree with everyone that some lines that are expanded would be better left as one line.If we can play with the configuration file and find some options that work better, great! Otherwise, I would be inclined to merge with the options that result in the best formatting and open an issue on rustfmt to give feedback.

prertik:
So, just to be clear, I have to change the `fn_call_style` to visual by adding rustfmt.toml. Is that all?

Michael-F-Bryan:
See what everything looks like when you tweak `fn_call_style` and run `cargo fmt` to reformat the entire project again. Then look at what changed with `git diff HEAD` and see if the changes seem alright to you.

Personally I don't mind what style we use, the biggest thing for me, and most probably the others, is that hitting save in my editor (I have format-on-save turned on) or running `cargo fmt` to format the entire project won't pollute PRs with unnecessary code churn.

prertik:
How about I change every indentation to `visual` rather than `block`. Looks like everyone prefers `visual`?

prertik:
This looks fine to me. How about you @budziq, @azerupi, @Michael-F-Bryan?

Michael-F-Bryan:
looks good to me :+1:

prertik:
This line `format_strings = true` is for future. When new lines of code are pushed containing long strings. The strings would be formatted automatically using `format_strings` otherwise manual formatting needs to be done for the long strings because of `max_width` error.

budziq:
:+1:

src/theme/mod.rs
```diff
@@ -59,19 +64,17 @@ impl Theme {
...
-                (theme_dir.join("favicon.png"), &mut theme.favicon),
-                (theme_dir.join("highlight.js"), &mut theme.highlight_js),
-                (theme_dir.join("clipboard.min.js"), &mut theme.clipboard_js),
-                (theme_dir.join("store.js"), &mut theme.store_js),
-                (theme_dir.join("highlight.css"), &mut theme.highlight_css),
-                (theme_dir.join("tomorrow-night.css"), &mut theme.tomorrow_night_css),
-                (theme_dir.join("ayu-highlight.css"), &mut theme.ayu_highlight_css),
-                (theme_dir.join("jquery.js"), &mut theme.jquery),
-            ];
+            let files = vec![(theme_dir.join("index.hbs"), &mut theme.index),
```
azerupi: This looks a little strange.

azerupi:
I think it's pretty good. There are still a couple of weird edge cases, but we can refine the config later if we find better defaults.

src/theme/mod.rs
prertik: Oh yeah, about this I made a `array_layout = "Visual"` in the rustfmt.toml because everyone here seemed to prefer the visual layout rather than the block. I think it looks OK.

Michael-F-Bryan:
> I think it's pretty good. There are still a couple of weird edge cases, but we can refine the config later if we find better defaults.

@azerupi, does that mean this is good to merge?

azerupi:
Yes, I think we can merge this PR. We can still tweak the settings when we discover better defaults. 

I wonder if we can find a way to force or encourage the use of rustfmt for PRs. Maybe a pass in Travis? The only problem is that I don't want to block PRs on style, but I do want to make it obvious for us and the author of the PR that rustfmt was not run.

Michael-F-Bryan:
Hmm... that's hard. Usually I have my editor set up so every time I hit save it'll format the document. Then every now and again you run `cargo fmt` and `cargo clippy` to clean things up. 

I guess we could periodically make a `cargo fmt` PR, but that's not exactly ideal...

prertik:
Yeah, I totally agree with @Michael-F-Bryan. Periodically running cargo fmt and using the configuration file present as rustfmt.toml in this repo would be ideal.

Michael-F-Bryan:
> Yes, I think we can merge this PR. We can still tweak the settings when we discover better defaults.

@azerupi, awesome! I'll let you or @budziq hit the big green "merge" button then because I don't have collaborator/owner status.

budziq:
done. Thanks @prertik !

PR #439, opened 2017-09-13
Fix the issue with pages named print not at the root

### LINKED ISSUE #435: Chapter named "print.html" brings up print dialog
A chapter like this:

    - [Formatted print](hello/print.md) 

when navigated to, brings up the print dialog.

I think we should check that it's the root's `print.html`, instead of any chapter named `print`.

azerupi:
Should fix #435

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -220,6 +220,7 @@ impl HtmlHandlebars {
         // Make sure that the Print chapter does not display the title from
         // the last rendered chapter by removing it from its context
         data.remove("title");
+        data.insert("is_print".to_owned(), json!(true));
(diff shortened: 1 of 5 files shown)


steveklabnik:
Just tested this locally with RBE, works for me! 👍

steveklabnik:
> The only problem that I see is that we have some styling regression in print media css (for some time since 0.0.23)

Is this a blocker then, or not?

steveklabnik:
> The only problem that I see is that we have some styling regression in print media css (for some time since 0.0.23)

Is there a separate bug tracking this? If not, it should be opened!

budziq:
> Is there a separate bug tracking this? If not, it should be opened!

Openend https://github.com/azerupi/mdBook/issues/444 :)

PR #440, opened 2017-09-13
added `mdbook-runnable` infostring support

budziq:
makes `ignore`'d playpens runnable

fixes https://github.com/azerupi/mdBook/issues/434

this might require some refactoring as our `ignore` handling is not cleanest at the moment


steveklabnik:
Works for me!

PR #446, opened 2017-09-18
Fix styling regression on print media - chromium

### LINKED ISSUE #444: styling regression on print media
We seam to have a styling regression in post 0.0.22 versions.
print_style

The printed page should not have a left margin that big.

Listwon:
Fixes #444

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -931,7 +931,7 @@ table thead td {
     left: 0;
     overflow-y: initial;
   }
-  .page-wrapper {
+  #page-wrapper.page-wrapper {
     padding-left: 0px;
   }
   #content {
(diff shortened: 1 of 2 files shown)


Listwon:
Hi @budziq! I've just tested this on Chrome. Have you tried to rebuild example book with rebuilt mdBook? Or maybe I just don't understand where this bug persists :) 
mdbook_print

budziq:
Ok I think I've got the root cause.

The print output looks exactly as expected when the sidebar is hidden.
good_print

While when its shown prior to triggering the print dialog.
bad_print

Listwon:
Oh, I've just reproduced it. Thanks for the hint!

Listwon:
@budziq please test it now :)

budziq:
Merged! Thanks a lot @Listwon! :+1:

PR #447, opened 2017-09-18
Fix copy to clipboard is available only for rust snippets

### LINKED ISSUE #432: copy to clipboard is available only for rust snippets

Listwon:
Fixes #432

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -7,7 +7,7 @@ $( document ).ready(function() {
     window.onunload = function(){};
 
     // Set theme
-    var theme = store.get('theme');
+    var theme = store.get('mdbook-theme');
     if (theme === null || theme === undefined) { theme = 'light'; }
 
     set_theme(theme);
@@ -145,7 +145,7 @@ $( document ).ready(function() {
             });
         }
 
-        store.set('theme', theme);
+        store.set('mdbook-theme', theme);
 
         $('body').removeClass().addClass(theme);
     }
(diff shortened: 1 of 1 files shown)


budziq:
Thanks for taking this up @Listwon !

There is just one caveat. The added clip button collides with already present buttons for rust code snippets.
untitled

src/theme/book.js
```diff
@@ -234,7 +246,7 @@ $( document ).ready(function() {
     var clipboardSnippets = new Clipboard('.clip-button', {
         text: function(trigger) {
             hideTooltip(trigger);
-            let playpen = $(trigger).parents(".playpen");
+            let playpen = $(trigger).parents("pre");
```
budziq: most likely we would no longer name this variable `playpen`

src/theme/book.js
```diff
@@ -199,6 +199,18 @@ $( document ).ready(function() {
             }
         });
     });
+    
+    $("pre code").each(function(i, block){
+        var pre_block = $(this).parent();
+        if( !pre_block.hasClass('playpen') ) {
```
budziq: I guess that we would prefer to have the button dock for each code/pre instead of specifically in the playgrounds. And only add additional buttons to (other than clipboard) to already existing button bar for rust specific snippets

Listwon:
@budziq yeah, my bad, it needs fixing and refactoring. I need to study the code more deeply :) Thanks for quick testing and feedback!

src/theme/book.js
Listwon: I'll wait for your comments then.

Michael-F-Bryan:
Hi @Listwon, did you ever get a chance to revisit this? It seems like a nice little feature to have. Especially for things like *The Book*, being able to copy an entire snippet at the click of a button would be quite convenient.

Listwon:
Hi @Michael-F-Bryan, sorry I didn't revisit it earlier. I've just updated book.js to the newest version and fixed the bug found by @budziq with duplicating copy buttons. @budziq please test this again, I don't see other bugs for now. If I'm not mistaken, adding opt-in requires changes in the Rust code in which I'm not fluent yet. 

I'll try to find the time for some refactoring in this part, but for now it's working as I wanted ;)

Michael-F-Bryan:
> If I'm not mistaken, adding opt-in requires changes in the Rust code in which I'm not fluent yet.

If you need anything from the Rust side of things, let me know. We just landed a PR (#457) which makes configuration a lot easier to work with, so that should greatly simplify the job of adding a new configuration item and making sure the renderer has access to it.

donald-pinckney:
I think this would also be a great feature to have, and in fact something I am writing right now would benefit from this. Are there any plans to get the ball rolling on this again, or is there anything I can do to help?

Michael-F-Bryan:
@donald-pinckney, it looks like most of the work for this feature is done and as far as I can tell everything *looks* okay. 

I've got a couple spare hours, so I'm going to check out the changes locally and do a proper review now. If there aren't any issues then this PR will probably be good to merge.

Michael-F-Bryan:
From what I can tell this looks good to me. Copying works, it pops up a little "Copied!" tooltip, and the clipboard icon no longer collides with the other buttons.

donald-pinckney:
Awesome, thanks!

PR #449, opened 2017-09-19
Change key for theme to not clobber old books

steveklabnik:
Fixes https://github.com/azerupi/mdBook/issues/448

Is this the path we want to take? It sucks to choose a different key, but it's an easy fix, and not *that* bad.

What about the `sidebar` key?

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -7,7 +7,7 @@ $( document ).ready(function() {
     window.onunload = function(){};
 
     // Set theme
-    var theme = store.get('theme');
+    var theme = store.get('mdbook-theme');
     if (theme === null || theme === undefined) { theme = 'light'; }
 
     set_theme(theme);
@@ -145,7 +145,7 @@ $( document ).ready(function() {
             });
         }
 
-        store.set('theme', theme);
+        store.set('mdbook-theme', theme);
 
         $('body').removeClass().addClass(theme);
     }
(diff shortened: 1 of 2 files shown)


azerupi:
Is this really needed? Can't we find a way to go back to the value without quotes? 
I assume that store.js is doing this. Would there be a way to disable that behavior? Before merging this PR adding prefixes, I would like to see if there isn't another way to solve this issue. The ideal fix would be to revert to using the unquoted value.

budziq:
@azerupi I'm guessing that the quoting comes from the fact that store.js actually serializes/deserializes complex objects into local storage as strings.

azerupi:
Yes, I just found that out too. Store.js calls [JSON.stringify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) internally, inserting quotes in the strings. There doesn't seem to be much we can do about it.

Actually, now that I think about it a bit more, prefixing might actually be the right move. Because localstorage is per domain and our keys might clash with other scripts using localstorage on the same domain.

steveklabnik:
> Is this really needed? Can't we find a way to go back to the value without quotes?
I assume that store.js is doing this. Would there be a way to disable that behavior? Before merging this PR adding prefixes, I would like to see if there isn't another way to solve this issue. The ideal fix would be to revert to using the unquoted value.

All these are good questions, which is why I wanted to start a discussion 😄 

It's basically "is it simpler to customize store.js, or is it simpler to namespace a key?" The latter seems cleaner to me overall, but I could be convinced, I guess.

budziq:
I'm no webdev but judging from the fact that localstorage is a shared for the whole domain. I would go for some namespacing anyway.

steveklabnik:
@budziq updated!

Is there any chance we can cut a release for this sometime soon?

PR #450, opened 2017-09-21
Call playground with /execute

Zengor:
This PR switches to using https://play.rust-lang.org/execute for the execution of rust code. It seems like the older evaluate.json format will not (understandably) support newer features. This now allows adding support to run as a lib crate or run with test configuration.


azerupi:
Thanks a lot for the PR!

Issue #453, opened 2017-09-26
Add header.hbs


Havvy:
Both the second edition of The Rust Programming Language and The Rust Reference put up a warning at the top of each page. To do this, we currently have our own `theme/index.hbs`. While this works, it does sort of pin us to a specific release of mdbook. We can occasionally rebase against index.hbs, but that has to be remembered to be done and takes away time from working on other things.

It'd be nice if there was a `theme/header.hbs` that we could override instead. Then at least the Rust Reference wouldn't need to override `index.hbs`, and TRPL would have a lot smaller diff to rebase with.

budziq:
Actually I was thinking about the exact same feature for some time to split the index.hbs into few sections with default implementation. Candidates for such extraction would probably be:
- sidebar
- button bar (this one actoulalby be required by "Rust Cookbook")
- header (needed by "The Rust Programming Language" and "The Rust Reference")
- footer

steveklabnik:
The Cookbook also does this, but for othe things.

Havvy:
This has been fixed. 😄

PR #454, opened 2017-09-27
Adding a header partial integration #453

projektir:
Addresses #453.

Giving this an initial stab.

I've tested it and it works: you need to point the theme at the location of your theme folder with the modified `header.hbs`, otherwise the default one is blank.

This still doesn't address the problem of additional classes/CSS, though...

Let me know what you think!

### DIFF
--- a/src/book/mod.rs
+++ b/src/book/mod.rs
@@ -263,6 +263,10 @@ impl MDBook {
         let mut index = File::create(themedir.join("index.hbs"))?;
         index.write_all(theme::INDEX)?;
 
+        // header.hbs
+        let mut header = File::create(themedir.join("header.hbs"))?;
+        header.write_all(theme::HEADER)?;
+
         // book.css
         let mut css = File::create(themedir.join("book.css"))?;
         css.write_all(theme::CSS)?;
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -253,8 +253,17 @@ impl Renderer for HtmlHandlebars {
 
         let theme = theme::Theme::new(theme_dir);
 
-        debug!("[*]: Register handlebars template");
-        handlebars.register_template_string("index", String::from_utf8(theme.index.clone())?)?;
+        debug!("[*]: Register the index handlebars template");
+        handlebars.register_template_string(
+            "index",
+            String::from_utf8(theme.index.clone())?,
+        )?;
+
+        debug!("[*]: Register the header handlebars template");
+        handlebars.register_partial(
+            "header",
+            String::from_utf8(theme.header.clone())?,
+        )?;
 
         debug!("[*]: Register handlebars helpers");
         self.register_hbs_helpers(&mut handlebars);
--- a/src/theme/header.hbs
+++ b/src/theme/header.hbs
@@ -1,0 +1,1 @@
+{{!-- Put your header HTML text here --}}
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -74,6 +74,7 @@
         <div id="page-wrapper" class="page-wrapper">
 
             <div class="page" tabindex="-1">
+                {{> header}}
                 <div id="menu-bar" class="menu-bar">
                     <div class="left-buttons">
                         <i id="sidebar-toggle" class="fa fa-bars" title="Toggle sidebar"></i>
--- a/src/theme/mod.rs
+++ b/src/theme/mod.rs
@@ -7,6 +7,7 @@ use std::io::Read;
 use errors::*;
 
 pub static INDEX: &'static [u8] = include_bytes!("index.hbs");
+pub static HEADER: &'static [u8] = include_bytes!("header.hbs");
 pub static CSS: &'static [u8] = include_bytes!("book.css");
 pub static FAVICON: &'static [u8] = include_bytes!("favicon.png");
 pub static JS: &'static [u8] = include_bytes!("book.js");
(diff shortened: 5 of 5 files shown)


src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -256,12 +256,18 @@ impl Renderer for HtmlHandlebars {
...
 
-        debug!("[*]: Register handlebars template");
+        debug!("[*]: Register the index handlebars template");
         handlebars.register_template_string(
             "index",
             String::from_utf8(theme.index.clone())?,
         )?;
 
+        debug!("[*]: Register the header handlebars template");
+        handlebars.register_template_string(
```
budziq: As far as I understand the [`register_partial`](https://docs.rs/handlebars/0.29.1/handlebars/struct.Handlebars.html#method.register_partial) would be contextually more appropriate (yes currently these two functions have identical implementation but this can change in the future)

src/book/mod.rs
```diff
@@ -262,6 +262,10 @@ impl MDBook {
         let mut index = File::create(&themedir.join("index.hbs"))?;
         index.write_all(theme::INDEX)?;
 
+        // header.hbs
+        let mut header = File::create(&themedir.join("header.hbs"))?;
```
budziq: and when the "header.hbs" is missing?
I understand that this should be handled similarly to "additional.css/js". Hmm We might be interested in extracting some type hbs partials registry.

budziq:
Thanks @projektir ! I think that we might be interested in generalizing the implementation a little and possibly extraction of some hbs blocks from index.hbs to separate files. But we would need to wait on word from @azerupi if he is interested in such feature at all (I know I am :)).

src/book/mod.rs
projektir: So this works similar to other /present/ files in the theme, in that there's already a blank `header.hbs` file and when the user doesn't supply one, the blank one is used, so it does nothing. I don't think I'm clear on what the difference is.

If we want to be more free form we could just let the user specify partials and then they can override the `index.hbs` itself to use their own partials in their own places (maybe this is one way to address CSS problems).

projektir:
I believe @azerupi marked this as `Wishlist`, so I think they want it! We just need to agree on how general it should be.

projektir:
@azerupi 
So the initial culprit for this was the Rust books which add a warning header. But what they also do is add a CSS class to one of nav elements (i.e., not part of the header), and then add a separate CSS style for that class.

Here's the relevant commit: https://github.com/rust-lang/book/commit/bcb1cc59bb63cf3c223e70c591b7acaddc85a968#diff-ebf1c3b21d73d0e04f32752d0b51ad38

azerupi:
Ok, but what problem does that cause exactly? Is it the fact that they overwrite the index.hbs file to add a class? If that is the problem they can just add custom css rules to the `#page-wrapper` element. Using the new keys to add additional css and js files it should be fairly trivial.

If that is not the problem I need more clarifications to understand.

projektir:
It means they still need to override the `header.hbs` directly, which means it still needs to be rebased once in a while to keep it in sync with any changes introduced to mdBook. That's what spurred @Havvy to make this issue originally I believe.

I'm not sure why you're reacting like that. If you think there's no problem that's a good thing, but I don't don't necessarily know your view nor do I have the full picture of what mdBook does, so I don't think it's unreasonable for me to want make sure that the original issue is indeed addressed.

Additional CSS and JS can be used to add a header even without a partial, also, but this is not what happened, so potentially there's a barrier to using those. But, yes, it can be done.

Havvy:
For TRPL, since there's other changes to `index.hbs`, it'd be okay to leave the CSS thing there. Just minimizing the diff we have to upkeep would be beneficial in and of itself.

For The Reference, we can remove the class, and just add the CSS to the page no matter what - leaving a comment in both the header.hbs and whichever CSS file saying what they are for.

---

The CSS we use is:

```css
.page-wrapper.has-warning > .nav-chapters {
   /* add height for warning content & margin */
  top: 120px;
}
```

So if the .nav-chapters could be below whatever content is in `header.hbs` without having to specify, we wouldn't need the CSS class at all.

projektir:
> For TRPL, since there's other changes to index.hbs, it'd be okay to leave the CSS thing there.

Oh, I didn't know that. Yeah that makes sense then.

I'm not sure if I'm quite following the thing with the reference but I guess that's mostly going to be figured out on that side.

If someone wants me to add the other parts to be partials let me know, otherwise we can perhaps close this?

azerupi:
>I'm not sure why you're reacting like that. If you think there's no problem that's a good thing

I'm sorry, I think we misunderstood each other. I truly did not understand what the problem was and I want to understand the issue before suggesting any solution. Now that I better understand the problem, I can try to find the best possible solution.

I thought that the problem you were talking about was simply adding some new classes to elements and give them a style. So I was trying to propose an alternate solution for the commit you linked to. Now I see that you want the underlying problem to be solved, namely to not have to override the included .hbs files for simple changes.

I think it is a problem worth solving, so I am going to think about this. 

*An idea for exploration:* It should be possible to extend templates and include the original inside the extended template. At least in other templating languages it is possible if I remember correctly. This would allow users to extend the templates without overwriting them.

projektir:
@azerupi so that sounds like a separate effort. Can we merge this PR for now and close the issue about the header and open a different one, perhaps?

Michael-F-Bryan:
@projektir, other than doing a `rebase` and resolving any merge conflicts is there anything else you think we need to do before this is ready to merge?

projektir:
Whoops, sorry, didn't notice this update. I'll rebase this up in a day or two.

projektir:
I rebased it, but I'm having trouble replacing the header now... not sure if I am doing something wrong or something else changed with theme replacement...

projektir:
I went back and fiddled with the config and seems like it _is_ working, so this should be good to merge. There's somewhere I wouldn't mind using this, as well.

Michael-F-Bryan:
LGTM :+1:

It seems like all mac jobs for travis are blocked indefinitely, but that's most probably just Travis struggling with availability. I'm going to merge this.

PR #456, opened 2017-09-30
Conditional Google Analytics

Michael-F-Bryan:
This makes sure google analytics isn't triggered when you are inspecting the document on your local computer. This will only ever really happen during development, so it doesn't make sense to pollute google analytics whenever you are working on the book.

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -146,7 +146,8 @@
             // make sure we don't activate google analytics if the developer is 
             // inspecting the book locally...
             if (localAddrs.indexOf(document.location.hostname) === -1) {
-                (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){ (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+                (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+                (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
                 m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
                 })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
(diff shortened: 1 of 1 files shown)


Michael-F-Bryan:
@budziq or @azerupi, any chance you can have a look at this and give it a yea or nay?

src/theme/index.hbs
```diff
@@ -139,14 +139,20 @@
...
-        m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-        })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+            var localAddrs = ["localhost", "127.0.0.1", ""];
 
-        ga('create', '{{google_analytics}}', 'auto');
-        ga('send', 'pageview');
+            // make sure we don't activate google analytics if the developer is 
+            // inspecting the book locally...
+            if (localAddrs.indexOf(document.location.hostname) === -1) {
+                (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){ (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
```
budziq: it looks like the GA has become slightly more garbled than before ;) Maybe we should pass it through some JS reindenter/beautifier to pretend that its human readable ;)

Otherwise to my limited knowledge, this looks sane

Michael-F-Bryan: Haha, good point. Although the ["official" snippet][o] on their site is just as ugly. I've fixed up that first extra-long line, but I don't think it's worth prettifying the snippet because it'll take up lots of vertical space and isn't exactly meant to be read/edited by humans anyway.

[o]: https://developers.google.com/analytics/devguides/collection/analyticsjs/#the_javascript_tracking_snippet

Michael-F-Bryan:
Damn, I thought adding `(skip ci)` would skip the travis step (it's just a whitespace change). Turns out you're meant to use `[skip ci]`.

src/theme/index.hbs
Michael-F-Bryan: I also checked this on my local environment with a book I've got set up for Google Analytics. Everything *seems* okay.

Michael-F-Bryan:
@azerupi, can this be merged?

azerupi:
Sorry for the delay!
Thanks for the PR :)

Issue #51, opened 2015-09-14
Have a search bar


mdinger:
Having a search bar would be very very useful. [rustbyexample](http://rustbyexample.com/) has one for example. I'm not sure how it would be implemented. I'm sure this has long been wanted by the [rust book](doc.rust-lang.org/book/).

Being able to search for programming keywords would actually make it very unusual and better than most too (outside the scope of a regular search bar though)

azerupi:
A search bar would indeed be very useful, but I don't think it's trivial to implement.

Ideally there would be some priority, like this:

`Sidebar title > h1 > h2 > h3 > h4 >  ... > bold > other text`

that would require to parse the markdown a second time, maybe we can hook into the parser from pulldown-cmark without the renderer?

> Being able to search for programming keywords would actually make it very unusual and better than most too (outside the scope of a regular search bar though)

A more general approach could be to specify high priority keywords in `book.json`. I am not sure how difficult that would be and how well that would work though.

steveklabnik:
:+1: this is something people have asked from rustbook forever see https://github.com/rust-lang/rust/issues/22786

asolove:
Is anyone else interested in working on this? If not, I might take a crack at generating a naive search index and seeing how it works on the rust book.

azerupi:
I have exams until end of January, so.. be my guest!  :wink: 

I think using [lunr.js](http://lunrjs.com/) has been considered in the past, so maybe take a look. But if you have other ideas, go ahead. I haven't done a lot of research about the subject.

If you have any question about the project, or you want to share advancement on this matter, feel free to post here! Good luck :+1:

s3bk:
So I just noticed I can't search the Rust reference anymore, which isn't really helpful.

Looking already existing search frameworks has resulted in these:

 - [Tantivy](https://github.com/tantivy-search/tantivy) (nightly Rust and optional C++ for SIMD)
 - [lunr.js](http://lunrjs.com/)
 - [http://elasticlunr.com/](Elasticlunr.js)

Using Tantivy would require compiling to asmjs (I guess around 10MB), or rewriting the search functionality in JS. And the search index needs to be deployed with the book.

With the two JS variants the situation is a bit different: The index could be generated during compilation with node.js, or when the book is opened.

I personally prefer the last option. It saves space for the Index, does not introduce additional build dependencies and the search framework can be easily updated.

azerupi:
>So I just noticed I can't search the Rust reference anymore

[...]

PR #457, opened 2017-09-30
Making configuration more flexible

Michael-F-Bryan:
This attempts to start cleaning up the configuration API. The idea is that all items should be public seeing as a `Config` struct is just a bag of options with no behaviour (besides convenience methods for loading from disk and such).

Removing the getters and setters and builder pattern may be a bit controversial, but my reasoning is you usually use getters/setters to allow for forward compatibility without breaking downstream. However, if we remove a config item that someone downstream is using, I'd *want* their build to break. This indicates they are using an option which is no longer available and they should update their stuff accordingly.

I'm also trying to make sure the html renderer isn't special. You still specify html renderer-specific options using the `output.html` table, but the `output` table itself is just a `BTreeMap<String, toml::Value>`. To make life more convenient for downstream I've added a `try_get_output()` method which allows you to get a table according to key and deserialize it automatically using `serde`.

This way, later on when we give the rendering and plugin pipeline a makeover, you can pass in a reference to the global `Config` struct and individual renderers/plugins can get the options specific to it.


Michael-F-Bryan:
So far I've just created the type definitions and a couple tests to make sure it deserializes as I expect it to. I'm going to start migrating the existing mdbook across to use the new configs now.

Michael-F-Bryan:
I finished updating everything which relies on configuration. Now that `MDBook` no longer has dozens of builder methods for updating the configuration, it should allow us to make the crate a lot less decoupled and stateful.

src/config.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /// Convenience method for getting the html renderer's configuration.
+    ///
+    /// # Note
+    ///
+    /// This is for compatibility only. It will be removed completely once the
+    /// rendering and plugin system is established.
+    pub fn html_config(&self) -> Option<HtmlConfig> {
+        self.try_get_output("html").ok()
```
Phaiax: This is not cached, so it will be deserialized a few times, but I think we do not need to optimize this.
And editing a variable in code is not as straight forward as with the `BookConfig`. Maybe you should add a line in the documentation on how to change a html config variable.

src/config.rs
```diff
@@ -0,0 +1,215 @@
...
+
+/// Convenience function to load a value from some table then deserialize it.
+fn get_deserialized<'de, T: Deserialize<'de>, S: AsRef<str>>(name: S, table: &BTreeMap<String, Value>) -> Result<T> {
+    match table.get(name.as_ref()) {
+        Some(output) => output
+            .clone()
+            .try_into()
+            .chain_err(|| "Couldn't deserialize the value"),
+        None => bail!("Key Not Found"),
+    }
```
Phaiax: Can these two match arms produce a more helpful error message?
It is only helpful for library users, but still.

Phaiax:
I almost wanted to ask if I can refactor the config, but then I found this nice PR.

I did a simple review and I vote for merging this.

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -283,50 +295,52 @@ impl Renderer for HtmlHandlebars {
...
+    data.insert("description".to_owned(), json!(config.book.description.clone().unwrap_or_default()));
     data.insert("favicon".to_owned(), json!("favicon.png"));
-    if let Some(livereload) = book.get_livereload() {
+    if let Some(ref livereload) = book.livereload {
         data.insert("livereload".to_owned(), json!(livereload));
     }
 
     // Add google analytics tag
-    if let Some(ref ga) = book.get_google_analytics_id() {
+    if let Some(ref ga) = book.config.html_config().and_then(|html| html.google_analytics) {
```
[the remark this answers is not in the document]
Michael-F-Bryan: Good point! I think I was blindly trying to fix the compile errors and not thinking of the bigger picture, but this should really use the `config` argument.

That way, when we start restructuring the renderer/plugin interface you can pass in just the `Book` (representation of a book on disk) and the `Config`. Avoiding passing around the small kitchen sink which is `MDBook` 😜

src/config.rs
Michael-F-Bryan: Hmm... To be honest, I hadn't thought about letting people mutate the configuration after it has been loaded. 

I think I'll rewrite the config to internally hold the raw `Value`, with getters and setters for accessing things like an `output` table (e.g. `HtmlConfig`) by key. The `toml` crate has a [get_mut()] method which I could expose to people so they can get something by index (e.g. `output.html`) and possibly mutate it.

[get_mut()]: https://docs.rs/toml/0.4.5/toml/value/enum.Value.html#method.get_mut

src/config.rs
Michael-F-Bryan: Yeah good point. Although the deserializing error is chained, meaning if you iterate through the causes you can see the root reason for why deserializing failed.  Giving you a high level "deserializing failed" message, plus a more low level error if you need more information.

I'm not sure what to do in the `None` case though. I really want to avoid returning an `Option<Result<T>>` if I can, that just feels ugly.

src/config.rs
Michael-F-Bryan: I just experimented with `Config` holding a raw `Value` and then providing getters and setters for the various components (book settings, outputs, etc) and it doesn't end up being overly ergonomic. The issue is that every time you want to access something from the `BookConfig` you need to deserialize it again, which requires dealing with errors appropriately. 

I think if I leave the internals `pub` that should be enough to let people update and inspect the config if they need to.

Michael-F-Bryan:
I've just fixed up the things @Phaiax raised. @azerupi, do you reckon this is ready to merge?

src/bin/build.rs
```diff
@@ -16,10 +17,6 @@ pub fn make_subcommand<'a, 'b>() -> App<'a, 'b> {
             "--no-create 'Will not create non-existent files linked from SUMMARY.md'",
         )
         .arg_from_usage(
-            "--curly-quotes 'Convert straight quotes to curly quotes, except for those \
-             that occur in code blocks and code spans'",
-        )
-        .arg_from_usage(
             "[dir] 'A directory for your book{n}(Defaults to Current Directory \
```
azerupi: Why is the "curly quote" option removed?

Michael-F-Bryan: I think I removed it because something like that belongs more in the config file than as a command line argument. My thoughts were that options to alter the generated output and which usually stay the same belong to go in the config file, while options that alter runtime behaviour (e.g. `--open` or `--no-create`) are command line arguments... Does that sound logical?

src/config.rs
Phaiax: I also tried a little but, but it did not end well, (see comment in line 57).
https://play.rust-lang.org/?gist=29a32500c428bfae20593421ec7bc4ab&version=stable

I would propose two other ways:
1. Do not use toml::Value at all and just use `HtmlConfig` as the only struct in `output` with direct deserialization. Add an API to get the original config string (raw file contents), so that the plugin can deserialize for itself.
2. Move the `HtmlConfig` struct into the `HtmlHandlebars`. Let the `HtmlHandlebars` deserialize from a `toml::Value` and store it's own deserialized config.

src/bin/build.rs
Phaiax: That sounds reasonable to me, but I am not sure about the intent behind the `curly quote` option.

src/config.rs
Michael-F-Bryan: In the long run I want to go for number 2 because that'll completely separate the handlebars renderer from the book. I think I marked the `HtmlConfig` getter as deprecated to indicate that.

Phaiax:
ping @azerupi

Michael-F-Bryan:
I'm going to go back and do another review of this PR to check that it works as we want. @Phaiax, I know you need access to configuration as part of #51, so when I'm happy would you like to give this another look over and see if it suits your needs?

At the moment I'm thinking of giving people access to the `book.toml` config file's contents with some `config.get("search")` method that returns an arbitrary [`toml::Value`] which will be specific to their plugin/feature/use case. You can also convert a `Value` into any arbitrary serializable thing using [`Value::try_into()`].

[`toml::Value`]: https://docs.rs/toml/0.4.5/toml/value/enum.Value.html 
[`Value::try_into()`]: https://docs.rs/toml/0.4.5/toml/value/enum.Value.html#method.try_from

Michael-F-Bryan:
I think I've found quite a nice API for the `Config`. It's probably easiest to explain by showing a couple of the tests:

```rust
#[test]
fn load_arbitrary_output_type() {
    #[derive(Debug, Deserialize, PartialEq)]
    struct RandomOutput {
        foo: u32,
        bar: String,
        baz: Vec<bool>,
    }

    let src = r#"
    [output.random]
    foo = 5
    bar = "Hello World"
    baz = [true, true, false]
    "#;

    let should_be = RandomOutput {
        foo: 5,
        bar: String::from("Hello World"),
        baz: vec![true, true, false],
    };

    let cfg = Config::from_str(src).unwrap();
    let got: RandomOutput = cfg.get_deserialized("output.random").unwrap();

    assert_eq!(got, should_be);

    let baz: Vec<bool> = cfg.get_deserialized("output.random.baz").unwrap();
    let baz_should_be = vec![true, true, false];

    assert_eq!(baz, baz_should_be);
}
```
(`get_deserialized()` is just a helper which will `get()` a key then use serde to deserialize it automagically)

And I've also added a `get_mut()` function so you can mutate elements in the `Config`.

```rust
#[test]
fn mutate_some_stuff() {
    // really this is just a sanity check to make sure the borrow checker
    // is happy...
    let src = COMPLEX_CONFIG;
    let mut config = Config::from_str(src).unwrap();
    let key = "output.html.playpen.editable";

    assert_eq!(config.get(key).unwrap(), &Value::Boolean(true));
    *config.get_mut(key).unwrap() = Value::Boolean(false);
    assert_eq!(config.get(key).unwrap(), &Value::Boolean(false));
}
```

@budziq if you get a moment, can you have a look at this and let me know what you think?

Phaiax:
This is a good design. 👍 

I only looked over the config.rs and that seems good to me.

Michael-F-Bryan:
Something that's just occurred to me is that this will be a breaking change. At the moment configuration things specific to the book are present as top level items, e.g.

```toml
title = "my book"
authors = ["Michael-F-Bryan"]

[output.html]
theme = "./path/to/theme"
```

Whereas with this configuration I've taken some inspiration from `Cargo.toml` and put the "metadata" stuff under their own section. 

Meaning the previous example should now look like this:

```toml
[book]     #   <-- new section
title = "my book"
authors = ["Michael-F-Bryan"]

[output.html]
theme = "./path/to/theme"
```

We're not `1.0` so *technically* we can do whatever we want, but at the same time I'd prefer not to break people's stuff without letting them know how to fix it. I think I'll add a compatibility thing to `Config`'s `Deserialize` impl which will check for the previous format and load that, emitting a warning to tell the user to upgrade (with a link to the configuration page in the docs).

book-example/src/format/config.md
```diff
@@ -24,66 +20,118 @@ additional-css = ["custom.css"]
...
 **book.toml**
 ```toml
+[book]
 title = "Example book"
 authors = ["John Doe", "Jane Doe"]
 description = "The example book covers examples."
-
-source = "my-src"  # the source files will be found in `root/my-src` instead of `root/src`
+src = "my-src"  # the source files will be found in `root/my-src` instead of `root/src`
+build-dir = "build"
```
Phaiax: You moved the `build-dir` into the common configuration. Hmmm. If I have multiple output generators, I cannot put them into different directorys. I think the other way was better. What are the arguments for doing it this way?

src/bin/build.rs
```diff
@@ -28,21 +25,16 @@ pub fn make_subcommand<'a, 'b>() -> App<'a, 'b> {
...
 pub fn execute(args: &ArgMatches) -> Result<()> {
     let book_dir = get_book_dir(args);
-    let book = MDBook::new(&book_dir).read_config()?;
+    let mut book = MDBook::new(&book_dir).read_config()?;
 
-    let mut book = match args.value_of("dest-dir") {
-        Some(dest_dir) => book.with_destination(dest_dir),
-        None => book,
-    };
+    if let Some(dest_dir) = args.value_of("dest-dir") {
```
Phaiax: .. did you move the `destination` because this argument must be independent of the backend?

src/config.rs
```diff
@@ -0,0 +1,385 @@
...
+            .and_then(|html| html.as_table_mut())
+            .and_then(|html| html.remove("destination"))
+            .and_then(|dest| dest.try_into().ok());
+
+        if let Some(dest) = destination {
+            cfg.book.build_dir = dest;
+        }
+
+        cfg.rest = table;
+        cfg
```
Phaiax: LGTM

src/config.rs
```diff
@@ -0,0 +1,385 @@
...
+                return Err(D::Error::custom(
+                    "A config file should always be a toml table",
+                ));
+            }
+        };
+
+        if is_legacy_format(&table) {
+            warn!("It looks like you are using the legacy book.toml format.");
+            warn!("We'll parse it for now, but you should probably convert to the new format.");
+            warn!("See the mdbook documentation for more details");
```
Phaiax: The explicit differences are mentioned neither here nor in the docs. I think it would be a nice gesture to just say:

> In short words: Just move all top level configuration entries like `title`, `author` and `description` into a group called `[book]`.

src/config.rs
```diff
@@ -0,0 +1,385 @@
...
+            warn!("http://azerupi.github.io/mdBook/format/config.html");
```
Phaiax: Just moved into rust-lang-nursery.

book-example/src/format/config.md
Michael-F-Bryan: My thoughts were that (in the long term) you would have a single `build/` directory where *all* renderers would put their rendered output. During the rendering stage a particular renderer is simply told the directory they should put things in, instead of each renderer doing their own thing and deciding it themselves. 

That way `mdbook` would ensure you have a standardized directory layout looking something like this:

- my_book
  - build
    - html
    - pdf
    - some_other_renderer
  - src
  - book.toml

With the special case being that if you're only using one renderer, `mdbook` will put tell a renderer to put its stuff directly in the root (avoiding the redundant extra level of nesting from `build/html/`).

src/bin/build.rs
Michael-F-Bryan: Yeah. I was thinking a renderer would be told to put its stuff in the `$dest-dir/renderer_name/` directory, or `$dest-dir/` if there's only one renderer.

That way renderers are "sandboxed" and should only be touching stuff in the directory they're told to write output to.

book-example/src/format/config.md
Phaiax: ok, that sounds fine :)

src/bin/build.rs
Phaiax: kk

Phaiax:
I vote for merging this finally. :dancer: 
Any bugs or changes can be fixed later, we don't release immediately anyway.

src/config.rs
Michael-F-Bryan: Just updated the link and it seems to work.

src/config.rs
Michael-F-Bryan: Good point. I've seen `rustc` and other tools emit messages along the same line of "this is going to stop working soon so update your stuff. If you do x, y, and z then that should fix 99% of your problems" and found it a really nice gesture.

Michael-F-Bryan:
> Any bugs or changes can be fixed later, we don't release immediately anyway.

Awesome! I'll wait for travis to give the green light before hitting the merge button.

steveklabnik:
So, this is now causing the nomicon to fail building: https://travis-ci.org/rust-lang-nursery/nomicon/builds/313641706?utm_source=email&utm_medium=notification

is `book.toml` required now?

Michael-F-Bryan:
It looks like that was an accidental regression, when loading the config we should fall back to `Config::default()` if `book.toml` doesn't exist. I've made a PR (#506) to fix this.

PR #461, opened 2017-10-04
Custom JS belongs at the bottom

Michael-F-Bryan:
It's standard practice to put CSS in the `<head>` and any JavaScript imports go at the bottom of the body section. I've shuffled around the template to place custom JS at the end of the document.

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -51,11 +51,6 @@
         <!-- Fetch store.js from local - TODO add CDN when 2.x.x is available on cdnjs -->
         <script src="store.js"></script>
 
-        <!-- Custom JS script -->
-        {{#each additional_js}}
-        <script type="text/javascript" src="{{this}}"></script>
-        {{/each}}
-
     </head>
     <body class="light">
         <!-- Set the theme before any content is loaded, prevents flash -->
@@ -168,5 +163,11 @@
 
         <script src="highlight.js"></script>
         <script src="book.js"></script>
+
+        <!-- Custom JS script -->
+        {{#each additional_js}}
+        <script type="text/javascript" src="{{this}}"></script>
+        {{/each}}
+
     </body>
 </html>


budziq:
I'm no webdev but if it's a common practice then maybe we should move all of the other js includes to the bottom as well?

Michael-F-Bryan:
Good point. From memory, in the head section the browser will wait to download and run any javascript tags before moving on. Whereas if you move it to the bottom, the browser can start rendering while it's waiting. 

I've moved everything but the the `store.js` and jquery to the bottom. `store.js` is related to themes, so it's best to leave them at the top to prevent a flash of unstyled content, while JQuery seems is used by pretty much everything and is probably required to make the page viewable.

I'm no web dev either, so we may need a third opinion on this one...

azerupi:
Indeed, it is common practice to put the JS files at the bottom to avoid delaying the page display while the JS files are fetched. In theory, JQuery and store.js could be moved to the bottom too as long as they are loaded before everything else. But I am not a professional web developer either, so I don't know the exact implications of doing so.

jacwah:
Searched around a bit and found [this StackOverflow post](https://stackoverflow.com/q/436411/2073469).

> The old approach to solving this problem was to put `<script>` tags at the bottom of your `<body>`, because this ensures the parser isn't blocked until the very end. This approach has its own problem: the browser cannot start downloading the scripts until the entire document is parsed. 

Bart recommends having `<script>` in `<head>`, and using `async` or `defer`, depending on if order of execution is important.

> The current state-of-the-art is to put scripts in the <head> tag and use the async or defer attributes. This allows your scripts to be downloaded asap without blocking your browser.

Michael-F-Bryan:
I think I'll roll back to the first commit, where I simply moved the custom JS to the bottom so it is loaded last. Given we vendor pretty much everything locally there doesn't seem to be much point moving javascript tags around for "performance".

Having custom JS be loaded last means users can rely on certain things like Jquery and `highlight.js` already being loaded.

Should we merge this more conservative PR?

azerupi:
Sounds good for now :)
We can still revisit this later if we want to do more.

Thanks!

Issue #458, opened 2017-10-01
Spaghetti Time!


Michael-F-Bryan:
As mdbook has grown and gained features over time, a couple bits and pieces have suffered from rightward drift or turned into spaghetti code. I'm noting down a list of problem areas which need a little TLC.

This is just standard refactoring and code cleanup, so would be great as a first contribution. 
If you want to help with the cleanup, make a comment here so we can assign it to you. I'm also available to mentor if needed.

Individual modules which could do with some cleanup:

- [ ] [mdbook::renderer::html_handlebars::helpers::navigation] - several 80+ line functions and lots of rightward drift
- [ ] [mdbook::renderer::html_handlebars::helpers::toc] - the entire module is one long 100+ line method
- [ ] [mdbook::renderer::html_handlebars::hbs_renderer] - several long functions which could be broken up into smaller steps/methods
- [ ] [mdbook::bin::serve] - argument parsing and execution should be broken out into two separate steps
- [ ] [mdbook::bin::watch] - `trigger_on_change()` is doing a bit too much and should be broken out into smaller functions

There is already a PR which runs rustfmt over the repo (#438, cheers @prertik!), however [clippy] still complains.

There are also existing PRs for the configuration (#457)  and book (#409) modules.

[clippy]: https://github.com/rust-lang-nursery/rust-clippy
[mdbook::bin::watch]: https://github.com/azerupi/mdBook/blob/a6d4881e00527290d8d3843376b137842c5b2833/src/bin/watch.rs#L52-L113
[mdbook::bin::serve]: https://github.com/azerupi/mdBook/blob/a6d4881e00527290d8d3843376b137842c5b2833/src/bin/serve.rs#L31-L105
[mdbook::renderer::html_handlebars::hbs_renderer]: https://github.com/azerupi/mdBook/blob/a6d4881e00527290d8d3843376b137842c5b2833/src/renderer/html_handlebars/hbs_renderer.rs#L253-L426
[mdbook::renderer::html_handlebars::helpers::toc]: https://github.com/azerupi/mdBook/blob/a6d4881e00527290d8d3843376b137842c5b2833/src/renderer/html_handlebars/helpers/toc.rs#L13-L139
[mdbook::renderer::html_handlebars::helpers::navigation]: https://github.com/azerupi/mdBook/blob/a6d4881e00527290d8d3843376b137842c5b2833/src/renderer/html_handlebars/helpers/navigation.rs#L10-L160

[...]

PR #465, opened 2017-10-06
Refactor navigation helpers

jacwah:
This is my attempt at a piece of #458.

- Code duplication between `previous()` and `next()` have been removed.
- Finding a chapter and rendering it have been split to separate functions.

There are probably tons of potential improvements, but it still feels a lot cleaner than before :blush:.

### DIFF
--- a/src/renderer/html_handlebars/helpers/navigation.rs
+++ b/src/renderer/html_handlebars/helpers/navigation.rs
@@ -135,3 +135,42 @@ pub fn next(_h: &Helper, r: &Handlebars, rc: &mut RenderContext) -> Result<(), R
 
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+   use super::*;
+
+   #[test]
+   fn test_next_previous() {
+      let template = "
+         {{#previous}}{{title}}: {{link}}{{/previous}}
+         {{#next}}{{title}}: {{link}}{{/next}}";
+
+      let data = json!({
+         "name": "two",
+         "path": "two.path",
+         "chapters": [
+            {
+               "name": "one",
+               "path": "one.path"
+            },
+            {
+               "name": "two",
+               "path": "two.path",
+            },
+            {
+               "name": "three",
+               "path": "three.path"
+            }
+         ]
+      });
+
+      let mut h = Handlebars::new();
+      h.register_helper("previous", Box::new(previous));
+      h.register_helper("next", Box::new(next));
+
+      assert_eq!(h.template_render(template, &data).unwrap(), "
+         one: one.html
+         three: three.html");
+   }
+}
--- a/src/renderer/html_handlebars/helpers/navigation.rs
+++ b/src/renderer/html_handlebars/helpers/navigation.rs
@@ -140,12 +140,11 @@ pub fn next(_h: &Helper, r: &Handlebars, rc: &mut RenderContext) -> Result<(), R
 mod tests {
    use super::*;
 
+   static TEMPLATE: &'static str =
+      "{{#previous}}{{title}}: {{link}}{{/previous}}|{{#next}}{{title}}: {{link}}{{/next}}";
+
    #[test]
    fn test_next_previous() {
-      let template = "
-         {{#previous}}{{title}}: {{link}}{{/previous}}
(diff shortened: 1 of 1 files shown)


azerupi:
The changes look good at first glance. I am going to do a more in-depth review soon.
Thanks for the PR! :)

src/renderer/html_handlebars/helpers/navigation.rs
```diff
@@ -5,165 +5,133 @@ use serde_json;
...
-                        debug!("[*]: Render template");
-                        // Render template
-                        _h.template()
-                          .ok_or_else(|| RenderError::new("Error with the handlebars template"))
-                          .and_then(|t| {
-                            let mut local_rc = rc.with_context(Context::wraps(&previous_chapter)?);
-                            t.render(r, &mut local_rc)
-                        })?;
+                if let Some(previous) = previous {
+                    if let Some(item) = target.find(&base_path, &path, &item, previous)? {
```
Michael-F-Bryan: :+1: This looks so much nicer than the massive block we had before!

src/renderer/html_handlebars/helpers/navigation.rs
```diff
@@ -5,165 +5,133 @@ use serde_json;
...
 
+   Ok(None)
+}
+
+fn render(
+    _h: &Helper,
+    r: &Handlebars,
+    rc: &mut RenderContext,
+    chapter: &StringMap,
+) -> Result<(), RenderError> {
```
Michael-F-Bryan: How hard would it be to write a test which checks this handlebars helper? Would it be possible to mock up a quick template then run the handlebars helper and assert that the rendered output looks like we expect it to?

src/renderer/html_handlebars/helpers/navigation.rs
```diff
@@ -5,165 +5,133 @@ use serde_json;
...
+    Next,
+}
+
+impl Target {
+    /// Returns target if found.
+    fn find(&self,
+            base_path: &String,
+            current_path: &String,
+            current_item: &StringMap,
+            previous_item: StringMap,
```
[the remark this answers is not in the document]
jacwah: I was probably influenced by the fact that the call site already has `previous_item` available. This way we avoid a second clone. When looking at `Target::find` in isolation, that doesn't make a lot of though, as you point out.

jacwah:
@Michael-F-Bryan I changed the signature of `Target::find` and added some tests for the helpers. Do you have any other types of tests in mind?

src/renderer/html_handlebars/helpers/navigation.rs
```diff
@@ -169,8 +168,67 @@ mod tests {
...
+               "path": "three.path"
+            }
+         ]
+      });
+
+      let mut h = Handlebars::new();
+      h.register_helper("previous", Box::new(previous));
+      h.register_helper("next", Box::new(next));
+
+      assert_eq!(
```
Michael-F-Bryan: I usually break this out into a couple lines. So at the top of the test I'll assign what I expect to a `should_be` variable, then assign what we actually get to `got`. Then at the bottom you can write `assert_eq!(got, should_be);` (e.g. [config.rs]).

It's definitely a personal taste thing, but it's nice when you can look at the top and figure out what the expected output should be at a glance without needing to look for all `assert_eq!()` calls. 

[config.rs]: https://github.com/rust-lang-nursery/mdBook/blob/fb99276f528307238dc5c1c66f17e6524efcf4ad/src/config.rs#L296-L327

Michael-F-Bryan:
> Do you have any other types of tests in mind?

The tests you've written look good to me. They're clear, concise, and test most common scenarios :+1:

I restarted the Mac travis build. It looks like it was some spurious Travis issue that failed the build and not anything to do with you. Once travis gives the green light I reckon this is ready to merge.

Michael-F-Bryan:
The failed build for OSX/beta is completely unrelated to this PR and seems to be some sort of permissions issue with our `.travis.yml`. I encountered the exact same issue when trying to merge #422 and have opened an issue (#490) to figure out the root cause.

In the meantime I think it's fine to merge this PR. Nice work @jacwah!

Issue #459, opened 2017-10-01
bug: can't back to chapter after toggle the content on mobiles


huangjj27:
as is shown in the image follow, when I toggle the content, I can't back to the chapter that I was reading:
![image](https://user-images.githubusercontent.com/8068443/31051746-67a5b41a-a6a3-11e7-815a-f86f0308354e.png)

also, the printer icon also _shouldn't appear in the content.

please fix the bug soon!

jacwah:
Can reproduce in Chrome devtools with small screens. The reason is that sidebar has a fixed width of 300px, which is too wide for e.g. an iPhone 5 with screen width of 320px.

https://github.com/azerupi/mdBook/blob/c64384abc3df85354f524332d91d3d286bf8d4c3/src/theme/stylus/variables.styl#L1

budziq:
Thanks for reporting this issue @huangjj27!

This looks like a regression, we used to have this **exact** usecase under control.

Untill the issue is resolved you can either:
- Select onenof the chapters to close the sidebar
- Scroll horizontally to the right exposing the hamburger menu

Probably we need to have sidebar width bounded between 300px and screen width percent.

Would you be interested in looking into this problem further @jacwah?

huangjj27:
> Probably we need to have sidebar width bounded between 300px and screen width percent.
@budziq I think you could make it responsive and just make the toc button shown in the right edge, as the gitbooks does.

budziq:
Good call! We are very low on webdev contributors (I am certainly not one ;) so the idea did not even occur to me)

huangjj27:
@budziq 
> We are very low on webdev contributors
It's not about the skill level of software development, but the real usage story of the product. I'm a user, as well as a tester, and thus I could more easily know what I really want.

jacwah:
I'll look into it!

@budziq Do you know a point when this was working?

budziq:
@jacwah I've just done a quick bisect. Unfortunately it is not a regression but a persisting bug. We have tested only with mobile screen sizes of iPhone6 and up (I just remembered that there was commit fixing the layout for screen sizes of iPhone6 and up).

The layout was always broken on lower screen sizes. So my bad for misinforming you.

jacwah:
I've looked at a responsive solution using `calc`. Due to my lacking experience with CSS and Stylus being completely new to me I haven't gotten it to work.

A quick fix is to just reduce the width. Something responsive seems like a more appropriate and long-term solution though.

[...]

PR #470, opened 2017-10-13
Improve experience on smaller screens

stgn:
At lower screen widths, when the sidebar is visible or transitioning to/from visible, the page wrapper will attempt to fit within the remaining screen width. On mobile devices, this leads to undesirable line wrapping and overflow. This change sets the width of the page wrapper to be equivalent to the screen width when the screen width is less than the sidebar width plus the maximum content width (1080px), simply causing the sidebar to push the page wrapper off-screen when it is visible.

Before and after:
mdbook-sidebar

I also changed some behavior regarding the sidebar's state when the page is loaded. If the screen width is less than 1080px, it will be hidden (so it doesn't load with the content pushed off-screen). Otherwise, it uses the stored value, defaulting to visible if there isn't one.

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -1,7 +1,7 @@
-html,
-body {
+html {
   font-family: "Open Sans", sans-serif;
   color: #333;
+  overflow-x: hidden;
 }
 body {
   margin: 0;
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -228,32 +228,41 @@ table thead td {
   -ms-transition: color 0.5s;
   transition: color 0.5s;
 }
-.mobile-nav-chapters {
-  display: none;
-}
 .nav-chapters:hover {
   text-decoration: none;
 }
+.nav-wrapper {
+  margin-top: 50px;
+  display: none;
+}
+.mobile-nav-chapters {
+  font-size: 2.5em;
+  text-align: center;
+  text-decoration: none;
+  width: 90px;
+  border-radius: 5px;
+}
 .previous {
-  left: 15px;
-  -webkit-transition: left 0.5s;
-  -moz-transition: left 0.5s;
-  -o-transition: left 0.5s;
-  -ms-transition: left 0.5s;
-  transition: left 0.5s;
+  float: left;
 }
 .next {
+  float: right;
   right: 15px;
 }
-@media only screen and (min-width: 1251px) {
-  .sidebar-visible .previous {
-    left: 315px;
+@media only screen and (max-width: 1080px) {
+  .nav-chapters {
+    display: none;
+  }
+  .nav-wrapper {
+    display: block;
   }
 }
-@media only screen and (max-width: 1060px) {
-  .nav-wrapper {
-    max-width: 750px;
-    margin: 0 auto;
+@media only screen and (max-width: 1380px) {
+  .sidebar-visible .nav-chapters {
+    display: none;
+  }
+  .sidebar-visible .nav-wrapper {
+    display: block;
   }
 }
 .theme-popup {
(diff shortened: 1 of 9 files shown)


Michael-F-Bryan:
@stgn this looks like a nice improvement :+1:

Wouldn't 1080px be a bit big if this is mainly aimed at mobile platforms? The chrome dev tools use 1024px as a "laptop" screen width, which means the average laptop would probably see part of the page pushed off the screen when the menu is open, whereas you'd typically expect it to wrap so you can see both the page content and the sidebar at the same time.

Michael-F-Bryan:
@stgn, what are your thoughts on the default screen size? From what I can see that's the only question stopping us from merging this PR.

stgn:
Without this change, 1080px is roughly the point where the content begins shrinking and wrapping. It's also the point where [the sidebar would hide itself if the sidebar state isn't set/saved](https://github.com/rust-lang-nursery/mdBook/pull/470/commits/888da5fb82954f9d48138bb597597c297694efd6#diff-b0df3ed87303a75cae6bc17d8fccde59L70). I just tried to match that.

1024px is unusually small for a laptop screen. I believe most laptops nowadays have resolutions with widths >= 1280px, with 1366px being the most common. The "Laptop with MDPI screen" preset in Chrome DevTools's device mode uses a width of 1280px.

Michael-F-Bryan:
> 1024px is unusually small for a laptop screen. I believe most laptops nowadays have resolutions with widths >= 1280px, with 1366px being the most common. The "Laptop with MDPI screen" preset in Chrome DevTools's device mode uses a width of 1280px.

That sounds quite reasonable.

There aren't any other outstanding questions, are there? @stgn, once the conflicts are resolved do you think this PR is ready to merge?

stgn:
That should do it. Please test it ([because I don't think that one PR was](https://i.imgur.com/AyJz8Bb.png)).

Michael-F-Bryan:
I tried this out on my laptop using chrome dev tools at both `320px` ("mobile") and `768px` ("tablet") and got some funny results for the smaller resolution. All larger window sizes are just fine though.

---

master (65acb355d74d2c26fd9c9d9946c31bbcf4cd4ef5)

mobile-master

---

This PR (31fb4435625e543d373cc71ae57b6ef05f15fa21)

mobile-stgn

---

On the second recording you can see me struggling to close the sidebar once it's been opened because you can no longer pan the page. I actually needed to cheat and increase the window size to close the sidebar. In theory you could "fix" this by making the sidebar smaller so that even on the smallest realistic screen you can still see the button for closing it, but that doesn't really address the underlying problem.

@stgn, what do you think we should do?

stgn:
The inability to pan horizontally [is intentional](https://github.com/rust-lang-nursery/mdBook/pull/470/commits/6e90e520d60dc7dad28f6e3976c4c386746bad1a#diff-5d584030e066ad197a54d8b5b15241aeR9). I originally thought about allowing it, but it's not very intuitive mobile design. There are a few ways to go about addressing this (in order of personal preference):

- Implement wipe gestures, which is already done in #474
- Anchor the hamburger button to the top-left of the viewport or page
- As mentioned, reduce the sidebar width until it fits
- Allow horizontal panning

Michael-F-Bryan:
I think the best long-term solution would be to implement swipe gestures. As you've mentioned allowing horizontal panning isn't ideal, and reducing the sidebar width until it fits is a bit of a hack and doesn't solve the underlying problem. 

I don't think we had any other outstanding questions or concerns. I'm happy to merge this PR given we've got a solution in the works for closing the menu at smaller resolutions (#474).

Michael-F-Bryan:
#474 has been merged, so the issue of not being able to close the menu on mobile (#459) should be fixed now :grinning:

Michael-F-Bryan:
I was testing #551 on my mobile and noticed that we may have accidentally regressed on the line wrapping/overflow behaviour this PR introduced. 

When you open/close the sidebar it's normally meant to push the page content off screen (as introduced in this PR). However on `master` (fa84da0856b4f47dfe2f1cef2c5ac1cbcd9bb4ad) it's reverted to wrapping lines to make room for the sidebar again. 

@sorin-davidoi, thoughts? It seems like a purely CSS thing and `src/theme/stylus/page.styl` hasn't been touched since this PR, so I'm thinking moving the `index.hbs` template around broke a CSS selector. Maybe somewhere near 61fad2786b?

sorin-davidoi:
I'll take a look later today.

PR #472, opened 2017-10-22
Search with Elasticlunr

Phaiax:
Belongs to Issue #51 , [Proof of concept of mdbooks book](https://phaiax.github.io/mdBook/cli/init.html) and [of the rustbook](https://phaiax.github.io/mdBook/rustbook/ch01-00-introduction.html)

This PR is not ready for merging, but I wanted to allow reviews.

Tracking:
- [x] Indexing not working properly (uses workaround, investigation needed)
- [x] Configuration (waiting on #457)
- [x] Use keys (up/down) for selecting a search result
- [x] Disabled search must disable search index generation, currently it is simply not written into the indexfile.
- [x] Measure slow down
- [x] Bug: Escape scrolls to top.
- Won't do: Improve teaser generation speed


Michael-F-Bryan:
@Phaiax, would you be able to add a link to your [proof-of-concept version][poc] in the description so people can see what the user-facing search will be like?

[poc]: https://phaiax.github.io/mdBook/cli/init.html

mattico:
@Phaiax I've updated elasticlunr-rs, and the generated indexes actually work now.

However, for search to not miss any results we need the [stemmer](https://en.wikipedia.org/wiki/Stemming) used to generate the index to be identical to the one used to search. The version of elasticlunr-rs on crates.io uses a stemmer from rust-stemmers, which doesn't quite have the same output. The version on master on the repo uses a port of the stemmer from elasticlunr.js, but there's a bug where it doesn't work on words ending in "ted" or "ting". Until I get that fixed there may be results missing if you use generated indexes.

Phaiax:
@mattico: Nice (I saw your comment in the reddit _What's everyone working on_ thread :) )

Today I added configurability. Tomorrow I will update the proof of concept and do the key up/down handler.

One problem appeared. My teaser algorithm is way too slow and needs a rework. In the above example, it did not matter, but it is causing lags in the official rust book.

Michael-F-Bryan:
@Phaiax, I just had a look at the configuration stuff and it seems pretty good so far. 

Something that occurred to me is whether we want to have search enabled by default or not (e.g. `enabled` is `true` in `Search::default()`). If it ends up being pretty slow or expensive then users may prefer to have search deactivated while developing.

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -650,28 +652,92 @@ pub fn normalize_id(content: &str) -> String {
...
+    struct SearchOptionsFields {
+        title: SearchOptionsField,
+        body: SearchOptionsField,
+        breadcrumbs: SearchOptionsField,
+    }
+
+    /// The searchoptions for elasticlunr.js
+    #[derive(Serialize)]
+    struct SearchOptions {
+        bool: String,
```
Michael-F-Bryan: Just out of curiosity, what does this guy do?

Phaiax: This mimics the structure of the `searchconfig` object, that elasticlunr expects. At first i serialized the `config::Search` struct, but that uses kebab-case which is not nice and would cause hidden breakage if modified.

Phaiax:
I updated the previews. Links in first post.

Phaiax:
I will measure the slowdown.

mattico:
Oh, and I renamed the package (but not the crate!) to "elasticlunr-rs".

src/renderer/html_handlebars/hbs_renderer.rs
Michael-F-Bryan: Ah fair enough, so it's a elasticlunr-specific thing then. I thought `bool` was an unusual choice for a variable in a Rust program, given it's also the name of a primitive type.

Phaiax:
Almost done. (Commit when elasticlunr-rs 0.2.1 is pushed to crates.io)

Timing with the rust book, second edition:

|Search    |  Debug  | Release |
|------------|-----------|-------------|
|Enabled |  52s        | 2.2s |
|Disabled |  24s       | 0.86s |

You decide if it should be enabled by default, but I think it should. For small books it does not make a difference and for large books such as rust book, they will probably tell their contributors about this option.

Phaiax:
@Michael-F-Bryan this is ready to merge.

Phaiax:
Added two more things, but now it should be ready.

Michael-F-Bryan:
@Phaiax, what are your thoughts on adding a couple tests to this? We'll want to ensure search index generation doesn't accidentally get broken if those sections get refactored later on. Would it make sense to add a couple integration tests for search under the `tests/` directory?

Michael-F-Bryan:
Sorry @Phaiax, it looks like merging #491 broke a fair amount of this PR due to conflicts :disappointed: 

We can probably remove most of the changes this PR makes to the book's internal representation because they will have (mostly) been dealt with in #491. So this PR "only" needs to worry about the html renderer and adding a `Search` struct to the `HtmlConfig`.

Phaiax:
@Michael-F-Bryan I'm still planning to do this, but I have limited time in the moment

Michael-F-Bryan:
> @Michael-F-Bryan I'm still planning to do this, but I have limited time in the moment

No stress :slightly_smiling_face: I'd found myself with a lot of spare time recently and was just going through any outstanding PRs and issues, seeing if there's anything I can do to help them along. 

I really like this PR and your proof-of-concept looks awesome, so if there's anything I can do to help don't hesitate to let me know!

mattico:
I'm working on updating this PR: https://github.com/mattico/mdBook/tree/search-eljs-rebase

It's mostly complete, but there are a few JS & styling bugs, and the index is missing some things.

cetra3:
@mattico, I've got a couple of css/js changes to bring it inline with the original.  I've pushed this to this branch here: https://github.com/cetra3/mdBook/tree/search-eljs-css

mattico:
@cetra3 Moving the searchbar into the header was deliberate, I thought it worked better with the sticky header that we have now. Of course, then the results shouldn't be stuck into the top of the document... Yeah I'll change it back to how it was, we can have style discussions after this is merged. Thanks for the fixes!

cetra3:
@mattico yeah it can be changed.  I was just having a hard time lining up the search input with the search text, as there is some indentation with the icons.

Michael-F-Bryan:
@mattico and @Phaiax this has been succeeded by #604, is it okay to close this PR?

mattico:
I'd say yes.

Issue #274, opened 2017-05-16
Support touch gestures on mobile to open and close the sidebar


azerupi:
Mobile users could use touch gestures like swipe right and left to open and close the sidebar. This would be way more ergonomic than the little hamburger menu they need to use.

This would fix issue #125

marti1125:
@azerupi I would like to work on this =D

azerupi:
Great! It's all yours, if you need any help let me know :)

marti1125:
@azerupi could you give some instructions how I can get started ? please

azerupi:
I have never done any touch development myself, but I would look into JavaScript librairies like  [JQuery Mobile](https://jquerymobile.com/) or [Hammerjs](http://hammerjs.github.io/).

Particularily, [this demo](http://demos.jquerymobile.com/1.3.2/examples/panels/panel-swipe-open.html) might be of interest. :)

Michael-F-Bryan:
Fixed in #474

PR #474, opened 2017-10-24
Basic swipe gesture functionality for opening/closing sidebar

stgn:
Implements #274. Not 100% certain about the thresholds or details of the behavior, but it seems intuitive enough to me. Suggestions are appreciated, however.

~~Recommended merging #470 (or similar) first. Otherwise, swiping will also scroll the page on smaller screens due to overflowing content.~~

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -82,6 +82,33 @@ $( document ).ready(function() {
         sidebar.scrollTop(activeSection.offset().top);
     }
 
+    var firstContact = null;
+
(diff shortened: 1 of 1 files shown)


Michael-F-Bryan:
I just tried this out on both my laptop (using chrome dev tools) and mobile. I think the thresholds are *slightly* off, but it's hard to pinpoint exactly what feels odd about them when experimenting on a mobile. 

I think this should be good to merge, if we need to we can always come back and tweak the  thresholds anyway.

PR #477, opened 2017-10-27
add tooltip to icons

### LINKED ISSUE #388: Add tooltips to code buttons
image

Buttons on code blocks do not currently have tooltips explaining what they do. It would be great to add tooltips with a very short explanation of what the buttons do (couple of words) when hovering over them.

mockersf:
fixes #388
added titles to play/copy/reset, expand/collapse and sidebar/theme buttons

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -186,15 +186,17 @@ $( document ).ready(function() {
         if(!lines_hidden) { return; }
 
         // add expand button
-        pre_block.prepend("<div class=\"buttons\"><i class=\"fa fa-expand\"></i></div>");
+        pre_block.prepend("<div class=\"buttons\"><i class=\"fa fa-expand\" title=\"Show hidden lines\"></i></div>");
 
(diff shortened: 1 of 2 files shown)


budziq:
@mockersf Very cool! Could you also add tooltips to the next and prev buttons at the bottom of the page?

mockersf:
Oh, didn't see those as there are already tooltips for them in browser, they were only missing in mobile.
I didn't copy the text as they mention using the arrow keys to navigate which may not be available on mobile.

budziq:
back. Hmm it looks like the links in the navigation buttons are duplicated and the tooltips are set on the second pair of links which leaves the buttons without tooltips on chromium and firefox (both on linux)

mockersf:
Yes, the mobile-nav-chapters class is displayed instead of nav-chapters if screen is less than 1250px wide. Both sets of links should have the title now.

budziq:
Thanks @mockersf !

Issue #451, opened 2017-09-21
Add support for playground tests 


Zengor:
Discovered in https://github.com/rust-lang/rust-by-example/issues/593. PR #450 is the first step to fixing this, but before going on I wanted to get comments on how to go about it. 

This could be achieved this by doing [the same thing the playground frontend does](https://github.com/integer32llc/rust-playground/blob/bc42662817a1c160574fa1d72384add0a15f6906/ui/frontend/selectors/index.js), but I was wondering if it would be better to provide this functionality in a more direct way, such as having some extra annotation along with the language:
```
```rust,editable,test
```
to indicate that the code should be run with the test configuration. This would be nice for writing out explicitly what a code fragment should be run under, and could also be used to specify other things like `beta`, `nightly`, `lib`, `release`. However, adding this as a feature does raise issues with adding other playpens in the future, as they might not support all of these. Particularly in the case of other languages, as these keywords wouldn't likely be useful outside of rust. If it'll be specific to the integer32 playground anyway, might was well just use the same logic their frontend uses without providing this as a direct feature until more general playpen configuration is worked out.

Personally I think taking the same approach as the playground frontend is good enough in this case, but would like to hear what others think.

azerupi:
I read through this and the linked issue on RBE, I think I understand the issue. But it is not clear to me how it should work in practice. In the example linked by steve, should the test run or not?

Adding extra annotations could be an option. I don't think that other future playgrounds for other languages should be an issue here. Each language will have it's own playground and eventually expose different options. It should suffice to only pass the appropriate language blocks to the correct "playground handler".

budziq:
Hmm, at the moment we are searching the snippet code for "feature" to switch to nightly compiler. We could do exactly the same with "#[test]".

marioidival:
mdBook supports tests?

steveklabnik:
mdbook test will work similarly to doc tests for cargo test, yes.

PR #478, opened 2017-10-28
can run tests of code block

mockersf:
for #451
added annotation "testable"
when annotation is present, a new button is added to run tests

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -23,6 +23,9 @@ code {
 .play-button.hidden {
   display: none;
 }
+.test-button.hidden {
+  display: none;
+}
 h2,
 h3 {
   margin-top: 2.5em;
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -211,16 +211,21 @@ $( document ).ready(function() {
             pre_block.prepend("<div class=\"buttons\"></div>");
             buttons = pre_block.find(".buttons");
         }
-        buttons.prepend("<i class=\"fa fa-play play-button hidden\" title=\"Run this code\"></i>");
-        buttons.prepend("<i class=\"fa fa-copy clip-button\" title=\"Copy to clipboard\"><i class=\"tooltiptext\"></i></i>");
-
         let code_block = pre_block.find("code").first();
+        buttons.prepend("<i class=\"fa fa-play play-button hidden\" title=\"Run this code\"></i>");
+        if (code_block.hasClass("testable")) {
+            buttons.prepend("<i class=\"fa fa-check test-button hidden\" title=\"Run tests\"></i>");
+        }
+        buttons.prepend("<i class=\"fa fa-copy clip-button\" title=\"Copy to clipboard\"><i class=\"tooltiptext\"></i></i>");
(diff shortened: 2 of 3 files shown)


budziq:
Hi @mockersf sorry for taking so long. Thanks for the PR!

I'm not sure if we would like to add another button to the already crowded snippet button bar.
At the moment we are searching the snippet code for "feature" to switch to nightly compiler. How about we would do exactly the same, searching for "#[test]" making the switch to testing environment implicit? But I'm open to discussion.

mockersf:
Hey @budziq 

I think it's interesting to be able to run the same code block as normal and as test, and have the output of each, so I decided to add a new button.
Otherwise, to show both, you would have to have the same code block twice, and they could get out of sync if they are user editable.

budziq:
@mockersf do you have a usecase requiring to run both code (`fn main`) and tests in a single snippet that could not be achieved otherwise?

mockersf:
For the example in https://rustbyexample.com/meta/test.html
For me the goal of this example was to show the different behaviour between `cargo run` and `cargo test`

Michael-F-Bryan:
Other than that one Rust By Example case, I don't think there's ever going to be a time when you'll want to run a snippet as *both* a test or an executable, especially seeing as unless you do `#[cfg(not(test))] fn main() { ... }` you're going to get a compilation error. So to me it doesn't feel like adding a `testable` attribute and another button will pull its weight.

I'd be more inclined to updating the JS to act the same as the playground instead of adding another button to the already crowded button bar. So the PR could be reduced to something like this:

```diff
 function run_rust_code(code_block) {
      var result_block = code_block.find(".result");
      if(result_block.length === 0) {
          code_block.append("<code class=\"result hljs language-bash\"></code>");
         result_block = code_block.find(".result");
     }
 
     let text = playpen_text(code_block);
     
     var params = {
 	channel: "stable",
 	mode: "debug",
 	crateType: "bin",
+	tests: false,
 	code: text,
     }
 
     if(text.indexOf("#![feature") !== -1) {
          params.channel = "nightly";
      }
+
+    if(text.indexOf("#![test]") !== -1) {
+         params.test = true;
+     }
  
     result_block.text("Running...");

  ...  
}
```

jhenninger:
I think this would be extremely helpful for the [The Book](https://doc.rust-lang.org/book/second-edition/index.html), especially the [Testing chapter](https://doc.rust-lang.org/book/second-edition/ch11-00-testing.html).

Right now mdbook [wraps code examples without a `fn main()` in its own main function block](https://github.com/rust-lang-nursery/mdBook/blob/master/src/renderer/html_handlebars/hbs_renderer.rs#L546-L561), which breaks the tests.

Some examples in The Book contain a "hidden" and empty main function before the test module, but not all do. 

Best approach would probably be to not wrap the example if the code block has the "testable" CSS class and remove the empty main functions from the book.

simonhdickson:
I was just about to create a PR for this when I spotted one was open. Is this PR still active?

Dylan-DPC:
I'm closing this PR based on the feedback received and inactivity. Thanks for contributing.

PR #484, opened 2017-11-12
Fix URL in Cargo.toml

steveklabnik:


steveklabnik:
Oh yeah, I mostly wanted to see that Travis is still running :)

Michael-F-Bryan:
I'm pretty sure travis should still run because the git user details and `GH_TOKEN` being used in `ci/deploy.sh` stayed the same. I'll create another issue to deal with updating links and stuff.

Michael-F-Bryan:
@steveklabnik, it looks like travis and GitHub Pages are working so do we still need this PR?

steveklabnik:
We don't; we should fix all of these links at once like in the other PR.

Sorry for the late response here, I've been on vacation! Back now though.

PR #489, opened 2017-11-17
updated links from azerupi to rust-lang-nursery

anna-liao:
fixes #485 . The only link I didn't change was for appveyor.

```
$ rg azerupi
README.md
13:            <a href="https://ci.appveyor.com/project/azerupi/mdbook/"><img src="https://ci.appveyor.com/api/projects/status/o38racsnbcospyc8/branch/master?svg=true"></a>
```

I also clicked through the links on the updated documents to make sure there are no dead links.

PR #493, opened 2017-11-22
rename paths with azerupi to rust-lang-nursery

lehmacdj:
Since mdBook was migrated to rust-lang-nursery some of the links,
particularly those to the documentation broke because github is no
longer redirecting those links to their actual location with rust lang
nursery in the path. Modify all of those instances so that they now
point to the propper link.

I checked all the links that I modified to make sure that they all work as expected.


lehmacdj:
Oops, didn't notice there was a different pull request that already makes these changes. #489 should be used instead of this one, it is better and has already had more reviewing done to it.

PR #497, opened 2017-11-27
 WIP: Select Default Theme

DeltaManiac:
I basically created a placeholder to hold the default theme specified.
Thinking if we should write it into the book.toml file on init and read from there in subsequent builds ?
@Michael-F-Bryan 
Thoughts ?

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -202,6 +202,8 @@ pub struct BookConfig {
     pub build_dir: PathBuf,
     /// Does this book support more than one language?
     pub multilingual: bool,
+    /// Default theme to be used
+    pub default_theme: Option<String>,
 }
 
 impl Default for BookConfig {
@@ -213,6 +215,7 @@ impl Default for BookConfig {
             src: PathBuf::from("src"),
             build_dir: PathBuf::from("book"),
             multilingual: false,
+            default_theme: None,
         }
     }
 }


Michael-F-Bryan:
Hi @DeltaManiac, sorry for the delay. I thought I'd responded to this but it looks like I forgot to actually hit send!

I'd say the default theme option belongs as part of the `HtmlConfig`. The `BookConfig` struct contains metadata about your book (title, author, etc), whereas the theme is something that's specific to the HTML renderer.

DeltaManiac:
@Michael-F-Bryan 
So where do you think i should start modifying, Like how do we ask  a user to input  a default them to the HTMLConfig struct ? From the cmd?

Michael-F-Bryan:
You should be able to add an optional "default theme" field to `HtmlConfig` (e.g. `default_theme: Option<String>`).

In the rendered book you can use javascript to programatically change the current theme. I was thinking if the `default_theme` is not `None`, we can add it to the bunch of variables passed to the handlebars template engine (somewhere around [here][1]) and then in the `index.hbs` template if the `default_theme` is present we tell the page to use that as the default theme ([here]).

So it really boils down to:

- Add an optional `default_theme` field to `HtmlConfig`
- Let the template engine know about our `default_theme` if it exists
- Update the `index.hbs` template to use this theme, falling back to `light` if not present

[here]: https://github.com/rust-lang-nursery/mdBook/blob/396426662de5ff6a9f12b8892ef89eba17b3bba2/src/theme/index.hbs#L59
[1]: https://github.com/rust-lang-nursery/mdBook/blob/396426662de5ff6a9f12b8892ef89eba17b3bba2/src/renderer/html_handlebars/hbs_renderer.rs#L67-L72

Dylan-DPC:
I am closing this since this is in WIP and changes are trivial, so we can restart it if needed. Thanks

PR #498, opened 2017-11-28
Add support for the --no-create flag to the serve subcommand.

cspiegel:
The `build` subcommand already understands `--no-create`, but `serve` does not.  This adds support for `--no-create` to `serve`.

### DIFF
--- a/src/bin/serve.rs
+++ b/src/bin/serve.rs
@@ -29,6 +29,9 @@ pub fn make_subcommand<'a, 'b>() -> App<'a, 'b> {
             "-d, --dest-dir=[dest-dir] 'The output directory for \
              your book{n}(Defaults to ./book when omitted)'",
         )
+        .arg_from_usage(
+            "--no-create 'Will not create non-existent files linked from SUMMARY.md'",
+        )
         .arg_from_usage("-p, --port=[port] 'Use another port{n}(Defaults to 3000)'")
         .arg_from_usage(
             "-w, --websocket-port=[ws-port] 'Use another port for the \
@@ -53,6 +56,10 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
 
     if let Some(dest_dir) = args.value_of("dest-dir") {
         book.config.book.build_dir = PathBuf::from(dest_dir);
+    }
+
+    if args.is_present("no-create") {
+        book.create_missing = false;
     }
 
     let port = args.value_of("port").unwrap_or("3000");


Michael-F-Bryan:
Hmm... This option is probably going to be applicable to more than just `build` and `serve`, isn't it? 

If that's the case it may be an idea to move it to the top level argument parser so it's available to all commands. Maybe somewhere around [here]. What are your thoughts?

[here]: https://github.com/rust-lang-nursery/mdBook/blob/1aa9c92ac165ce812314574f2c75a8234b8d83bd/src/bin/mdbook.rs#L33-L39

cspiegel:
@Michael-F-Bryan I can see the utility of making it more widely available, although it won't necessarily be applicable to all subcommands.  What do you think about having it globally available, but just ignored by subcommands for which it makes no sense?

Michael-F-Bryan:
> What do you think about having it globally available, but just ignored by subcommands for which it makes no sense?

Sounds good to me. I think the only command which you wouldn't necessarily need `--no-create` is the `mdbook clean`. @budziq, do you have any opinions on this?

Ideally it could be something you put in your `book.toml` as part of the general configuration system so we're not having to thread some `no_create` variable through to every call. But that can be work for a later PR if we still feel it's necessary.

Michael-F-Bryan:
If you're using a makefile then that's a good indicator that some stuff can be moved from command-line arguments to the configuration file. My opinion is that command-line arguments are for once-off things, so if you need to specify the same arguments over and over they should be part of `book.toml`. What other arguments do you find yourself passing in over and over? If it's easy enough to do, we could probably kill two birds with one stone.

In this case we could do it in a couple ways. We could add an extra `build` table to `book.toml` which specifies various build options and leaves room for expansion later, or we can add the `no-create` value (although it may be better to call it `create-missing` with a default of `true` because negation is unintuitive) to the top-level `book` table. 

I'm happy to mentor on this if you want. I gave the config system a bit cleanup a while back so I'm pretty familiar with the code. In this case *if* we were to tweak how building is done, I think you'd need to:

- Create a new `Build` struct in `src/config.rs` and add it as part of the overall `Config` struct
- Update the `Deserialize` impl for `Config` so that it removes a `"build"` table and deserializes it (essentially a copy/paste of [this])
- Update the tests for configuration to make sure we're correctly handling the new format
- Update the places in code where we were using `no_create` to decide whether to create missing files and replace them with something like `cfg.build.create_missing`.
- Update the `book.toml` documentation (in `book-example/`) to mention this new `"build"` table
- Maybe add a couple integration tests which makes sure we get the expected generated stuff when building with and without this flag set?

It sounds like a lot of work, but I doubt it'd be more than 100 lines of changes all up. 

[this]: https://github.com/rust-lang-nursery/mdBook/blob/1aa9c92ac165ce812314574f2c75a8234b8d83bd/src/config.rs#L171-L173

cspiegel:
I've created a new merge request, #500, to implement this option in `book.toml`, so I'm closing this one.

PR #500, opened 2017-11-30
WIP: Add a create-missing option to book.toml.

cspiegel:


cspiegel:
@Michael-F-Bryan This implements the `create-missing` option as discussed in #498.  The only part of your recommended approach that I haven't done is the integration testing.  I'm not sure of the proper approach for it in this project, so I would appreciate your input on how to proceed with it.

src/bin/build.rs
```diff
@@ -32,7 +32,7 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
     }
 
     if args.is_present("no-create") {
-        book.create_missing = false;
+        book.create_missing = Some(false);
```
Michael-F-Bryan: Do you reckon it's worth adding a deprecation warning here as well?

src/book/mod.rs
```diff
@@ -25,8 +25,10 @@ pub struct MDBook {
     pub livereload: Option<String>,
 
     /// Should `mdbook build` create files referenced from SUMMARY.md if they
-    /// don't exist
-    pub create_missing: bool,
+    /// don't exist? If `None`, use the setting specified in `book.toml`,
+    /// otherwise use this.
+    /// This is for backward compatibility only.
+    pub create_missing: Option<bool>,
```
Michael-F-Bryan: You could probably get away with removing this completely. We're already changing the type from `bool` to `Option<bool>` so any 3rd parties using it will get compile errors. If that's the case, we may as well remove it completely so `create_missing` is only defined in one place (the `Build` struct under `Config`).

src/config.rs
```diff
@@ -217,6 +224,23 @@ impl Default for BookConfig {
     }
 }
 
+/// Configuration for the build procedure.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(default, rename_all = "kebab-case")]
+pub struct BuildConfig {
```
Michael-F-Bryan: Now we've got a `BuildConfig`, what are your thoughts on removing `build_dir` from the `BookConfig` and putting it here? I think that'd make sense seeing as `build_dir` is more of a build option than metadata about your book.

src/config.rs
```diff
@@ -290,6 +320,7 @@ mod tests {
         let got = Config::from_str(src).unwrap();
 
         assert_eq!(got.book, book_should_be);
+        assert_eq!(got.build, build_should_be);
```
Michael-F-Bryan: Nice. This is exactly what I had in mind when I mentioned testing the new config format :+1:

book-example/src/format/config.md
```diff
@@ -45,6 +48,20 @@ src = "my-src"  # the source files will be found in `root/my-src` instead of `ro
 build-dir = "build"
 ```
 
+### Build options
+
+This controls the build process of your book.
+
+- **create-missing:** By default, any missing files specified in `SUMMARY.md`
+  will be created when the book is built. If this is `false` then the build
```
Michael-F-Bryan: Would it be worth explicitly saying that this defaults to `true`? So something like *"... created when the book is built (i.e. `create-missing = true`)"*.

Michael-F-Bryan:
> The only part of your recommended approach that I haven't done is the integration testing. I'm not sure of the proper approach for it in this project

You may want to check out [`tests/init.rs`][1]. That file contains a couple tests which create a new `MDBook` and then run `build()`, making sure that the expected files were generated.

 You could create a test which adds a couple non-existent chapters to `SUMMARY.md` and then test that running `build()` with `create_missing` set to true will create those chapters. Then add another one which tests when `create_missing` is false.

You'd probably want to put your tests in either `tests/init.rs` or `tests/rendered_output.rs`. The former contains integration tests for creating a new book directory while the latter builds a book and checks its output. I'll leave the decision up to you seeing as both sets of integration tests are relevant.

[1]: https://github.com/rust-lang-nursery/mdBook/blob/master/tests/init.rs

Michael-F-Bryan:
Also, don't forget to add yourself to the [contributors list][cl] if you haven't already!

[cl]: https://github.com/rust-lang-nursery/mdBook/blob/master/book-example/src/misc/contributors.md

book-example/src/format/config.md
cspiegel: Yes, that makes sense to me.

src/bin/build.rs
cspiegel: Yes, I think that's a good idea.

src/book/mod.rs
cspiegel: Agreed, that's a better approach.  I missed the fact that the book config was available there!

src/config.rs
cspiegel: That seems logical.  I've implemented this as well as updating some of the documentation surrounding the build directory.

cspiegel:
I've implemented all code review changes.  The one remaining item is the integration tests which I hope to get to today.

Michael-F-Bryan:
I just had another look through the diff and it looks like you were really thorough in making sure all references to the previous location of `build-dir` were updated. That's really nice to see! :slightly_smiling_face:

cspiegel:
@Michael-F-Bryan I've added tests for `create-missing`; please take a look when you have a moment.

Michael-F-Bryan:
@cspiegel, I'm pretty happy with it :slightly_smiling_face: 

Do you think there's anything else that needs to be done, or should I hit the merge button?

cspiegel:
@Michael-F-Bryan I think it's good to go.  Thanks for helping get it into shape!

Michael-F-Bryan:
:tada: Yay! Thanks so much for working on this @cspiegel, it's really really appreciated.

PR #504, opened 2017-12-07
WIP: Add option to disable section numbers in the table of contents

cspiegel:
This is as much a request for comments as anything else.  I've added the ability to not display section numbers in the table of contents.  Nesting is still allowed: the only difference is that the section numbers themselves are not displayed.

The one large issue with this is that, as it stands, mdBook is somewhat centered on the idea of numbered sections.  By removing the section numbers, prefix and suffix chapters become redundant.  And the documentation refers to the "main" section of the book as numbered chapters.

Obviously there's no harm in keeping prefix and suffix chapters, as this patch does: you can either use or ignore them.  It may be somewhat confusing to keep them around, documentation-wise, but if they're removed when numbered sections are turned off, then the book building (code wise) becomes a lot more complex; plus you can't test what a book with prefix/suffix chapters would look like without section numbers without some surgery to `SUMMARY.md`.

Anyway, as I said, this is mainly to get some comments about this idea, whether it's at all desirable.

### DIFF
--- a/book-example/src/format/config.md
+++ b/book-example/src/format/config.md
@@ -35,6 +35,8 @@ This is general information about your book.
 - **src:** By default, the source directory is found in the directory named
   `src` directly under the root folder. But this is configurable with the `src`
   key in the configuration file.
+- **section-numbers:** By default, sections are numbered in the table of
+  contents. Setting this to `false` disables numbering.
 
 **book.toml**
 ```toml
@@ -43,6 +45,7 @@ title = "Example book"
 authors = ["John Doe", "Jane Doe"]
 description = "The example book covers examples."
 src = "my-src"  # the source files will be found in `root/my-src` instead of `root/src`
+section-numbers = false
 ```
 
 ### Build options
--- a/src/book/bookitem.rs
+++ b/src/book/bookitem.rs
@@ -2,10 +2,11 @@ use serde::{Serialize, Serializer};
 use serde::ser::SerializeStruct;
 use std::path::PathBuf;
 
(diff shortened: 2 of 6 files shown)


Michael-F-Bryan:
I think it'd be quite possible to tell `mdbook` not to *show* section numbers, but I don't think that's what you're talking about. 

Correct me if I'm mistaken, but you're thinking of changing the way `mdbook` represents a book internally so the concept of section numbers isn't necessary. How would one indicate whether they want section numbers or not, would it be an all-or-nothing thing (via a config option), or would you instead have to manually annotate the chapters with/without numbers (presumably in your `SUMMARY.md`)?

cspiegel:
I think it depends on the ultimate goal.  Initially I just wanted the section numbers to not display.  It didn't matter to me particularly whether they were kept track of internally or not; and I just didn't use prefix or suffix chapters, so that was a non-issue.

At the current time, it looks like section numbers are *only* used for display in the Table of Contents, so the hacky solution of just not displaying them works fine for that purpose.  The downside is that if section numbering starts to be used for more things, it would quickly become ugly, adding these config checks everywhere.  At that point, hacvng the internal representation understand numbers vs. non-numbers would be better.

Really, I'm not sure what the ideal approach would be, overall.  I think I'd be comfortable just closing this request, and I can add a new feature request which can hash out some of the details, including whether this feature is even worth it.

Michael-F-Bryan:
I think merging #533 made this PR redundant so I'll close it. @cspiegel, let me know if you think we should revisit it.

PR #515, opened 2017-12-21
Implement playpen defaults

jridgewell:
Avoids issues where we enable `editable` but forget to specify `editor` (eg, https://github.com/rust-lang/rust-by-example/issues/963). 
Since we already include Ace editor, seems like we can just treat it as the default.


Michael-F-Bryan:
Using `Ace` seems like a more sane default than just an empty string :+1:

 We should probably also emit a warning if the editor specified isn't valid (e.g. it doesn't exist because it's `""`), that can be a separate issue though.

Issue #490, opened 2017-11-18
Travis builds failing for OSX & beta


Michael-F-Bryan:
For some reason we're encountering a permissions error when running the OSX/beta travis build. An [example] is trying to merge #465 or #422.

This is the error message:

![screenshot_2017-11-16_150757](https://user-images.githubusercontent.com/17380079/32878216-10054bd8-cae0-11e7-864d-c77597c8c49a.png)

@budziq or @steveklabnik, have you got any idea what may be causing this?

[example]: https://travis-ci.org/rust-lang-nursery/mdBook/builds/301638661

budziq:
@Michael-F-Bryan on first glance it looks like https://github.com/rust-lang/cargo/issues/4699

Michael-F-Bryan:
I added a couple debug commands to see what's going on in that directory and that seems to be the issue.

![screenshot_2017-11-19_000033](https://user-images.githubusercontent.com/17380079/32982271-ce8cad3e-ccbc-11e7-9a08-f0ba14571ba7.png)

I'm thinking the easiest solution is to temporarily remove the OSX job until rust-lang/cargo#4699 is solved. @budziq, can you think of any other workaround?

budziq:
We can either disable osx (or just beta) temporarily or just wait doing nothing untill cargo gets fixed. I'd probably just disable the beta for now as the least valuable check.

Michael-F-Bryan:
I think I'll disable `beta` because you tend to only ever use `stable` or `nightly`.

budziq:
Looking at the 1.22.1 release notes, the problem with cargo on MacOS beta should be fixed.

sorin-davidoi:
@Michael-F-Bryan I guess this can be closed?

Michael-F-Bryan:
Yep.

PR #516, opened 2017-12-21
Update travis CI

Michael-F-Bryan:
The last several releases haven't uploaded binaries. I've updated `.travis.yml` to use japaric/trust to configure travis, plus created a new Personal Access Token so releases should hopefully work again.

Testing on the `beta` channel was previously deactivated because there was a compiler issue (see #490 and #492). It should be working again so I've re-enabled `beta` testing.

Because I've added a bunch more targets the CI build times may be adversely affected... I'll see how it goes, but may end up removing a couple targets if it becomes a problem.

---

I'm not sure what we're going to do about Appveyor. It's been broken ever since we moved mdbook to the nursery, meaning we're not currently testing or generating release binaries for windows. The easiest way to deal with it would be to tell Appveyor that I can look after `rust-lang-nursery/mdBook` and I can deal with it from there, but @budziq [mentioned] that this may not be an overly straightforward process.

[mentioned]: https://github.com/rust-lang-nursery/mdBook/issues/482#issuecomment-345529809

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ matrix:
   - env: TARGET=aarch64-unknown-linux-gnu
   - env: TARGET=arm-unknown-linux-gnueabi
   - env: TARGET=i686-unknown-linux-gnu
-  - env: TARGET=i686-unknown-linux-musl
   - env: TARGET=x86_64-unknown-linux-gnu
   - env: TARGET=x86_64-unknown-linux-musl
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,42 +1,92 @@
+# Based on the "trust" template v0.1.1
+# https://github.com/japaric/trust/tree/v0.1.1
+
+dist: trusty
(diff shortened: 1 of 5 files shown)


Michael-F-Bryan:
As a datapoint, the [most recent `master` build][1] ran for 10 minutes and 1 second (wall time). Note that this was done with a warm cache, whereas the first one or two builds generated by this PR will have to recompile *everything*.

[1]: https://travis-ci.org/rust-lang-nursery/mdBook/builds/319562543

Michael-F-Bryan:
I think I'll merge this as-is. It *should* fix the deploy issues with GitHub Releases and make sure we update GitHub Pages every time a PR is merged into `master`. So the travis side of CI should be alright.

The appveyor issue is still outstanding, but I think we'll need help from the infrastructure team to fix that.

I'll CC in the @rust-lang-nursery, hopefully someone there will be able to help us out.

PR #520, opened 2018-01-02
Implement partial include of source files

boxdot:
The macro `{{include some_file}}` accepts now optional line number arguments, s.t. the specified line range is included. The following forms are supported:

* `{{include some_file::}}` is equivalent to `{{include some_file}}`
* `{{include some_file:from:}}` includes lines [from, infinity)
* `{{include some_file::to}}` includes lines [0, to]
* `{{include some_file:from:to}}` includes lines [from, to]

This is useful when including files containing full source code.

### DIFF
--- a/src/preprocess/links.rs
+++ b/src/preprocess/links.rs
@@ -1,6 +1,8 @@
+use std::ops::{Range, RangeFrom, RangeTo, RangeFull};
 use std::path::{Path, PathBuf};
 use regex::{CaptureMatches, Captures, Regex};
 use utils::fs::file_to_string;
+use utils::string::take_lines;
 use errors::*;
 
 const ESCAPE_CHAR: char = '\\';
@@ -22,15 +24,38 @@ pub fn replace_all<P: AsRef<Path>>(s: &str, path: P) -> Result<String> {
     Ok(replaced)
 }
 
-#[derive(PartialOrd, PartialEq, Debug, Clone)]
+#[derive(PartialEq, Debug, Clone)]
 enum LinkType<'a> {
     Escaped,
-    IncludeRangeFrom(PathBuf, usize),
-    IncludeRange(PathBuf, usize, usize),
+    IncludeRange(PathBuf, Range<usize>),
+    IncludeRangeFrom(PathBuf, RangeFrom<usize>),
+    IncludeRangeTo(PathBuf, RangeTo<usize>),
+    IncludeRangeFull(PathBuf, RangeFull),
     Playpen(PathBuf, Vec<&'a str>),
 }
 
-#[derive(PartialOrd, PartialEq, Debug, Clone)]
+fn parse_include_path(path: &str) -> LinkType<'static> {
+    let mut parts = path.split(':');
+    let path = parts.next().unwrap().into();
+    let start = parts.next().and_then(|s| s.parse::<usize>().ok());
+    let end = parts.next().and_then(|s| s.parse::<usize>().ok());
+    match start {
+        Some(start) => {
+            match end {
+                Some(end) => LinkType::IncludeRange(path, Range{ start: start, end: end}),
+                None => LinkType::IncludeRangeFrom(path, RangeFrom{ start: start }),
+            }
+        }
+        None => {
+            match end {
+                Some(end) => LinkType::IncludeRangeTo(path, RangeTo{ end: end }),
+                None => LinkType::IncludeRangeFull(path, RangeFull),
+            }
+        }
+    }
+}
+
+#[derive(PartialEq, Debug, Clone)]
 struct Link<'a> {
     start_index: usize,
     end_index: usize,
@@ -39,39 +64,16 @@ struct Link<'a> {
 }
 
 impl<'a> Link<'a> {
-    fn parse_include_path(path: PathBuf) -> Option<LinkType<'a>> {
-        path.to_str().map(|p| {
-            let mut parts = p.split(':');
-            let p = parts.next().unwrap();
-            let from = parts.next().and_then(|s| s.parse::<usize>().ok());
-            let to = parts.next().and_then(|s| s.parse::<usize>().ok());
-            match from {
-                Some(from) => {
-                    match to {
-                        Some(to) => LinkType::IncludeRange(p.into(), from, to),
-                        None => LinkType::IncludeRangeFrom(p.into(), from),
-                    }
-                }
-                None => {
-                    match to {
-                        Some(to) => LinkType::IncludeRange(p.into(), 0, to),
-                        None => LinkType::IncludeRangeFrom(p.into(), 0),
-                    }
-                }
-            }
-        })
-    }
-
     fn from_capture(cap: Captures<'a>) -> Option<Link<'a>> {
         let link_type = match (cap.get(0), cap.get(1), cap.get(2)) {
             (_, Some(typ), Some(rest)) => {
                 let mut path_props = rest.as_str().split_whitespace();
-                let file_path = path_props.next().map(PathBuf::from);
+                let file_arg = path_props.next();
                 let props: Vec<&str> = path_props.collect();
 
-                match (typ.as_str(), file_path) {
-                    ("include", Some(pth)) => Self::parse_include_path(pth),
-                    ("playpen", Some(pth)) => Some(LinkType::Playpen(pth, props)),
+                match (typ.as_str(), file_arg) {
+                    ("include", Some(pth)) => Some(parse_include_path(pth)),
+                    ("playpen", Some(pth)) => Some(LinkType::Playpen(pth.into(), props)),
                     _ => None,
                 }
             }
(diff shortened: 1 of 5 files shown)


src/preprocess/links.rs
```diff
@@ -46,50 +70,73 @@ impl<'a> Link<'a> {
...
+                    start_index: mat.start(),
+                    end_index: mat.end(),
+                    link: lnk,
+                    link_text: mat.as_str(),
+                }
+            })
         })
     }
 
+    fn join_lines<'b, I: IntoIterator<Item = &'b str>>(lines: I) -> String {
```
boxdot: Most likely, better to use itertools::intersperse, however, I am not sure if it is ok to add another dependency to the crate.

src/preprocess/links.rs
```diff
@@ -180,6 +227,70 @@ fn test_find_links_simple_link() {
...
+            Link {
+                start_index: 22,
+                end_index: 46,
+                link: LinkType::IncludeRange(PathBuf::from("file.rs"), 0, 20),
+                link_text: "{{#include file.rs::20}}",
+            },
+        ]
+    );
+
+    let s = "Some random text with {{#include file.rs::}}...";
```
Michael-F-Bryan: I'm probably going to mention this in the main discussion area as well, but I think it might be better to drop the `::` when you're wanting to include the entire file's contents. Then the "intuitive" way of writing this (`{{#include some_file.rs}}`) *Just Works*.

Michael-F-Bryan:
Thanks for the PR @boxdot, this is definitely a feature I'd like to see in `mdbook` :grinning: 

It looks like this is not really documented in the user guide at the moment, so we can use this as an opportunity to make the feature more discoverable. I think the best thing to do would be to add a "mdbook-specific markdown" page to the "Format" chapter of the example book.

boxdot:
@Michael-F-Bryan, thank you for the review and elaborate comments. I will go through them and fix/implement proposals.

src/preprocess/links.rs
boxdot: Added itertools.

src/preprocess/links.rs
```diff
@@ -25,7 +25,8 @@ pub fn replace_all<P: AsRef<Path>>(s: &str, path: P) -> Result<String> {
 #[derive(PartialOrd, PartialEq, Debug, Clone)]
 enum LinkType<'a> {
     Escaped,
-    Include(PathBuf),
+    IncludeRangeFrom(PathBuf, usize),
```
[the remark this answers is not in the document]
boxdot: Done.

src/preprocess/links.rs
```diff
@@ -38,6 +39,29 @@ struct Link<'a> {
 }
 
 impl<'a> Link<'a> {
+    fn parse_include_path(path: PathBuf) -> Option<LinkType<'a>> {
```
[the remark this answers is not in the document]
boxdot: Done.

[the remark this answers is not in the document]
boxdot: Done.

src/preprocess/links.rs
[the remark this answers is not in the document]
boxdot: For the vec you need an intermediate vector of string, which basically duplicates the file content. Even worse, for each line, there would be a memory allocation, which we can avoid if we use the iterator of string slices directly.

src/preprocess/links.rs
```diff
@@ -46,50 +70,73 @@ impl<'a> Link<'a> {
...
     }
 
+    fn join_lines<'b, I: IntoIterator<Item = &'b str>>(lines: I) -> String {
+        let mut lines = lines.into_iter();
+        let first = String::from(lines.next().unwrap_or(""));
+        lines.fold(first, |acc, s| acc + "\n" + s)
+    }
+
+    fn take_lines(s: String, from: usize, to_inclusive: Option<usize>) -> String {
+        let lines = s.split('\n').skip(from);
```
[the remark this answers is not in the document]
boxdot: Fixed.

src/preprocess/links.rs
```diff
@@ -180,6 +227,70 @@ fn test_find_links_simple_link() {
                     }]);
 }
 
+#[test]
+fn test_find_links_with_range() {
```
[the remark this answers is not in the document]
boxdot: Indeed. Fixed.

src/preprocess/links.rs
boxdot: There is no strong reason for this. I included this case because of the symmetry. The only (minor) reason, I have in mind, would be the case, when you are trying to find the right line numbers and are playing with them, ie. you are editing your include to see how the full file looks like, to go back to partial include after that.

Now, after describing the use case to you, I have a feeling, that it should be removed. :) What do you think?

boxdot:
Still todo:
* [ ] Extend documentation.

I have a feeling, that it would make sense to use polymorphic object for the range i.e. `Include(PathBuf, Box<RangeArgument<usize>>`. This would reduce the number of cases and also the duplicated code. I will try it out tomorrow.

src/preprocess/links.rs
```diff
@@ -22,14 +24,38 @@ pub fn replace_all<P: AsRef<Path>>(s: &str, path: P) -> Result<String> {
     Ok(replaced)
 }
 
-#[derive(PartialOrd, PartialEq, Debug, Clone)]
```
boxdot: Note: Ranges are not implementing this. However, we are also not using it anywhere.

src/utils/string.rs
```diff
@@ -0,0 +1,69 @@
...
+        Some(&self.end)
+    }
+}
+
+/// Take a range of lines from a string.
+pub fn take_lines<R: RangeArgument<usize>>(s: &str, range: R) -> String {
+    let start = *range.start().unwrap_or(&0);
+    let mut lines = s.lines().skip(start);
+    match range.end() {
+        Some(&end) => lines.take(end).join("\n"),
```
boxdot: It would be better to have a proper separator for windows here. Could not find any definition in the stdlib.

Michael-F-Bryan: Are you looking for [std::path::MAIN_SEPARATOR](https://doc.rust-lang.org/std/path/constant.MAIN_SEPARATOR.html)?

src/preprocess/links.rs
Michael-F-Bryan: Seems fine to me, `PartialOrd` isn't overly useful anyway.

src/utils/string.rs
```diff
@@ -0,0 +1,69 @@
+use std::ops::{Range, RangeFrom, RangeFull, RangeTo};
+use itertools::Itertools;
+
+// This trait is already contained in the standard lib, however it is unstable.
+// When the feature will stabilize, pls remove!
```
Michael-F-Bryan: It may be a good idea to write this second line as a `TODO`, linking to the relevant RFC/PR in the Rust repo.

```
// TODO: Remove when the `collections_range` feature stabilises
// (https://github.com/rust-lang/rust/issues/30877)
```

Writing it that way tends to make it easier to `grep` for and will immediately pop out as something which needs to be done, instead of being just another comment.

src/utils/mod.rs
```diff
@@ -1,4 +1,5 @@
 pub mod fs;
+pub mod string;
```
Michael-F-Bryan: It may be easier to make the `string` module private and then re-export `RangeArgument` and `take_lines()`. `mdbook` has a bad habit of creating deeply nested modules, which makes it annoying when you're looking for things in the docs or have to import something.

src/preprocess/links.rs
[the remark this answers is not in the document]
Michael-F-Bryan: I don't think that's exactly correct. `some_string.lines()` gives you an iterator of string slices (`&str`, essentially just a raw pointer and a `usize` to indicate length). This means `content.lines().collect()` would give you a `Vec<&str>` which only takes up `num_lines * 2 * sizeof(usize)` bytes in memory. Most importantly, this meaning you *don't duplicate* the file's contents, and the only allocations are when the underlying `Vec` needs to grow.

([naive playground example](https://play.rust-lang.org/?gist=89d40993a1cf554b99c52eaf43823d18&version=stable&mode=release))

Either way, if we're using `itertools` we may as well take advantage of [`Itertools::join()`] for readability. 

[`Itertools::join()`]: https://docs.rs/itertools/0.7.4/itertools/trait.Itertools.html#method.join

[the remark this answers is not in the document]
boxdot: You are totally right. Somehow, I completely missed the point that it is a `Vec<&str>` and not `Vec<String>`.

src/utils/string.rs
boxdot: No, newline separator. As far as I know, on Windows `\n` also should work, except in some exotic editors like Notepad.

src/utils/string.rs
boxdot: Makes sense. Thanks. Done.

src/utils/mod.rs
boxdot: Done.

Michael-F-Bryan:
I believe this PR is pretty much good to merge. The newline separator thing shouldn't be a problem because the main HTML renderer uses `pulldown_cmark` to convert markdown into HTML, and I'm assuming `pulldown_cmark` won't care whether it gets a `\n` or `\r\n` for new lines.

@boxdot, do you think there's anything left to do? I'm thinking we'll make documenting the `{{include some_file:from:to}}` feature a separate PR.

boxdot:
@Michael-F-Bryan, from my side we can merge. I prepare then documentation in a separate review.

PR #523, opened 2018-01-03
Use <button> for menu bar

marionebl:
* **hbs**: Use `button[type=button]` for menu icons
* **styl**: Introduce `.icon-button` class to deal with browser default styling for `<button>`
* Move `title` attribute to interactive elements (`button`, `a`)

## Motivation
* Make important controls focusable via keyboard
* Ensure `title` attribute values show up on hover

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -78,15 +78,19 @@
                 {{> header}}
                 <div id="menu-bar" class="menu-bar">
                     <div class="left-buttons">
-                        <i id="sidebar-toggle" class="fa fa-bars" title="Toggle sidebar"></i>
-                        <i id="theme-toggle" class="fa fa-paint-brush" title="Change theme"></i>
+                        <button id="sidebar-toggle" class="icon-button" type="button" title="Toggle sidebar">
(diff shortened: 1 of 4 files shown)


src/theme/index.hbs
```diff
@@ -78,15 +78,19 @@
...
+                            <i class="fa fa-paint-brush"></i>
+                        </button>
                     </div>
 
                     <h1 class="menu-title">{{ book_title }}</h1>
 
                     <div class="right-buttons">
-                        <a href="print.html">
-                            <i id="print-button" class="fa fa-print" title="Print this book"></i>
+                        <a href="print.html" title="Print this book">
```
Michael-F-Bryan: Should the "print" button also be a `<button>`?

marionebl: I wondered about that, too. I decided on leaving it alone because it has a valid `href`, so printing can be deep linked currently.

Michael-F-Bryan: Ah good point. I think it would be best to leave it as-is because `print.html` is a link to the actual print page, while the other buttons are used by javascript.

PR #524, opened 2018-01-04
Prohibit 'print.md' files

### LINKED ISSUE #258: Article named 'print' triggers printing dialog
https://doc.rust-lang.org/nightly/unstable-book/library-features/print.html

JanLikar:
Fix #258 by emmiting an error whenever an mdBook contains a `print.md` file in its root.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -128,6 +128,8 @@ pub use renderer::Renderer;
 
 /// The error types used through out this crate.
 pub mod errors {
+    use std::path::PathBuf;
+
     error_chain!{
         foreign_links {
             Io(::std::io::Error);
@@ -147,9 +149,9 @@ pub mod errors {
                 display("Error at line {}, column {}: {}", line, col, message)
             }
 
-            ReservedFilenameError(message: String, filename: String) {
-                description("A reserved filename error")
-                display("Error while processing {}: {}", filename, message)
+            ReservedFilenameError(filename: PathBuf) {
+                description("Reserved Filename")
+                display("{} is reserved for internal use", filename.display())
             }
         }
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -146,6 +146,11 @@ pub mod errors {
                 description("A SUMMARY.md parsing error")
                 display("Error at line {}, column {}: {}", line, col, message)
             }
+
+            ReservedFilenameError(message: String, filename: String) {
+                description("A reserved filename error")
+                display("Error while processing {}: {}", filename, message)
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -54,6 +54,14 @@ impl HtmlHandlebars {
                                                                            to str")
                                                        })?;
 
+                // "print.html" is used for the print page.
+                if path == "print.md" {
```
Michael-F-Bryan: What happens if the path is `./print.md` instead?

src/lib.rs
```diff
@@ -146,6 +146,11 @@ pub mod errors {
                 description("A SUMMARY.md parsing error")
                 display("Error at line {}, column {}: {}", line, col, message)
             }
+
+            ReservedFilenameError(message: String, filename: String) {
```
Michael-F-Bryan: You can probably simplify this by removing the `message` parameter. If it's a reserved filename then the message is almost always going to be "{} is reserved for internal use".

i.e. something like this:

```rust
ReservedFilenameError(filename: PathBuf) {
  description("Reserved Filename")
  display("{} is reserved for internal use", filename.display())
}
```

src/renderer/html_handlebars/hbs_renderer.rs
Michael-F-Bryan: What happens if the path is `./print.md` instead? It might be better to do `if ch.path == Path::new("print.md")` to avoid those sorts of ambiguities.

Michael-F-Bryan:
Would it be worth adding a test to make sure rendering a chapter called `print.md` is an error?

JanLikar:
I applied the suggested fixes.

As far as testing goes, I don't think there is an easy way to test this, but maybe I am wrong. Do you have any suggestions?

Michael-F-Bryan:
I think the easiest way would be to add it as an integration test. That way you don't need to try and manually create all the internal state the HTML renderer uses. 

What I would do is add a new test to `tests/rendered_output.rs` which creates a directory tree something like this in a temporary directory:

- `TempDir::new("mdbook")`
  - src/
    - SUMMARY.md
    - print.md

(making sure to add a `print.md` entry to your `SUMMARY.md` file)

You'd then call [`MDBook::load()`] to load this book into memory, and assert that calling `build()` fails.

[`MDBook::load()`]: http://rust-lang-nursery.github.io/mdBook/mdbook/book/struct.MDBook.html#method.load

JanLikar:
Ok, the test is done, so this should be ready to be merged.

Michael-F-Bryan:
> Ok, the test is done, so this should be ready to be merged.

Awesome work! :+1:

PR #532, opened 2018-01-07
implementing preprocessors

JaimeValdemoros:


src/preprocess/links.rs
```diff
@@ -5,9 +5,35 @@ use utils::fs::file_to_string;
...
+pub struct ReplaceAllPreprocessor {
+    pub src_dir: PathBuf
+}
+
+impl Preprocessor for ReplaceAllPreprocessor {
+    fn run(&self, book: &mut Book) -> Result<()> {
+        for section in &mut book.sections {
+            match *section {
+                BookItem::Chapter(ref mut ch) => {
+                    let content = ::std::mem::replace(&mut ch.content, String::new());
```
Michael-F-Bryan: I don't think you need to use `mem::replace()` here. You could store the replaced text in a temporary (`let replaced = replace_all(&ch.content, base)`) then update `ch.content` on the next line.

JaimeValdemoros: Good point, thanks

Michael-F-Bryan:
@JaimeValdemoros I just had another skim through what you've done so far and it looks pretty good :+1:

You may want to add a couple tests for the preprocessor discovery, because I feel like there are a couple edge cases we haven't accounted for. For example, what happens if the user passes in an unknown preprocessor, if you specify a preprocessor twice, or if one of the elements in the `preprocessor` array is actually a table?

The easiest way to write the test is probably to create a bunch of dummy configs (just raw strings), run `Config::from_str()` to load it, then run `determine_preprocessors()` and make sure we got the preprocessors we expected. You may also want to add a `name()` method to the `Preprocessor` trait so we can log something like `"running the \"{}\" preprocessor"`, and it also helps figure out which preprocessors were found while testing.

Michael-F-Bryan:
Hi @JaimeValdemoros, this is just a friendly ping to see how you're going with everything :grinning: How far through this issue would you say you are? And is there anything I can do to help?

JaimeValdemoros:
Hey - sorry about the delay, I've been away and haven't been able to get a
free evening to get it finished. Will have it done tomorrow afternoon if
that's ok.

On 14 Jan 2018 10:34 am, "Michael Bryan" <notifications@github.com> wrote:

Hi @JaimeValdemoros <https://github.com/jaimevaldemoros>, this is just a
friendly ping to see how you're going with everything 😀 How far through
this issue would you say you are? And is there anything I can do to help?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/rust-lang-nursery/mdBook/pull/532#issuecomment-357499236>,
or mute the thread

JaimeValdemoros:
@Michael-F-Bryan I've made a couple of small changes that felt a bit more sensible:
- Moved the "preprocess" field under the "build" header rather than standalone (as an optional field)
- Report an error if the preprocessor isn't recognised
- Allow explicitly disabling the default option by setting an empty list
- Added a 'name' field to the Preprocessor trait, matching the Renderer trait
- Added some tests

Is there anything you think is still missing?

src/book/mod.rs
```diff
@@ -332,25 +332,26 @@ fn determine_renderers(config: &Config) -> Vec<Box<Renderer>> {
...
 
 /// Look at the `MDBook` and try to figure out what preprocessors to run.
-fn determine_preprocessors(config: &Config) -> Vec<Box<Preprocessor>> {
+fn determine_preprocessors(config: &Config) -> Result<Vec<Box<Preprocessor>>> {
+
+    let preprocess_list = match config.build.preprocess {
+        Some(ref p) => p,
+        // If no preprocessor field is set, default to the LinkPreprocessor. This allows you
+        // to disable the LinkPreprocessor by setting "preprocess" to an empty list.
+        None => return Ok(vec![Box::new(LinkPreprocessor::new())])
```
Michael-F-Bryan: I reckon we may want to pull this bit out into its own `default_preprocessors()` function, where we immediately return `vec![Box::new(LinkPreprocessor::new())]`.  I imagine that'll make things easier to update in the future, and it's immediately obvious what the default preprocessors are.

src/book/mod.rs
```diff
@@ -332,25 +332,26 @@ fn determine_renderers(config: &Config) -> Vec<Box<Renderer>> {
...
-    if let Some(preprocess_array) = config.get("preprocess").and_then(|o| o.as_array()) {
-        for key in preprocess_array.iter() {
-            match key.as_str() {
-                Some(key) if key == "links" => {
-                    preprocessors.push(Box::new(LinkPreprocessor::new()))
-                }
-                _ => {}
-            }
+    for key in preprocess_list {
+        if key == "links" {
```
Michael-F-Bryan: What are your thoughts on turning this `if key == "links" {...} else { bail!() }` into a match statement? Now that preprocessors are easier to work with, I imagine we'll be adding a bunch more in the future.

src/book/mod.rs
```diff
@@ -149,14 +154,23 @@ impl MDBook {
     pub fn build(&self) -> Result<()> {
         debug!("[fn]: build");
 
+        let mut preprocessed_book = self.book.clone();
```
Michael-F-Bryan: I like that we're not mutating the original book here and instead preprocessing its clone :+1:

src/book/mod.rs
```diff
@@ -190,6 +204,13 @@ impl MDBook {
         self
     }
 
+    /// You can add a new preprocessor by using this method.
+    /// The only requirement is for your renderer to implement the Preprocessor trait.
```
Michael-F-Bryan: This doc-comment should probably be reworded in the "imperative" tense to describe what it does. So something like "Register a \[`Preprocessor`\] to be used when rendering the book", with the word "Preprocessor" being a link to the relevant trait so people can see the required methods and what types already implement `Preprocessor`.

src/preprocess/mod.rs
```diff
@@ -1 +1,17 @@
...
+
+use std::path::PathBuf;
+
+pub struct PreprocessorContext {
+    pub src_dir: PathBuf
+}
+
+pub trait Preprocessor {
+    fn name(&self) -> &str;
+    fn run(&self, ctx: &PreprocessorContext, book: &mut Book) -> Result<()>;
```
Michael-F-Bryan: I'm tossing up whether to include the `Book` as part of the `PreprocessorContext`. Do you think the book to be processed should be part of the context or should we keep it as a second argument?

src/preprocess/mod.rs
```diff
@@ -1 +1,17 @@
...
+
+mod links;
+
+use book::Book;
+use errors::*;
+
+use std::path::PathBuf;
+
+pub struct PreprocessorContext {
+    pub src_dir: PathBuf
```
Michael-F-Bryan: It's probably better for `PreprocessorContext` to contain a reference (or clone, if you don't want to deal with lifetimes) to the `Config` and the book's root. That way a preprocessor can hook into `book.toml` for configuration, and the `src_dir` can be discovered by combining `config.book.src` and the book root.

So something like this:

```rust
pub struct PreprocessorContext {
  pub config: Config,
  pub root: PathBuf,
  // maybe also include the `Book` here... I dunno
}
```

Michael-F-Bryan:
@JaimeValdemoros, I've made a temporary fix (#549) to CI so travis won't fail and erroneously say that this PR is broken. If you want, you can `git rebase` on top of the latest master to pick up the fix and make CI work normally for this PR.

src/book/mod.rs
JaimeValdemoros: Done - you're right, makes it a bit cleaner

JaimeValdemoros:
Thanks @Michael-F-Bryan - this is the first time I've rebased onto a fork so hopefully I've done it right

src/preprocess/mod.rs
JaimeValdemoros: I think it's helpful to distinguish the fact that you only expect the `Book` itself to be mutated, and that the `PreprocessorContext` is additional information it can use to help in that mutation. Having a single parameter `&mut PreprocessorContext` obscures that slightly and has fewer guarantees about what the `Preprocessor` is allowed to do to the context.

Michael-F-Bryan: Ah yeah that's a good point. I completely forgot the `Book` is `&mut` while we pass an immutable reference for the `PreprocessorContext`.

tests/testing.rs
```diff
@@ -21,3 +29,33 @@ fn mdbook_detects_book_with_failing_tests() {
 
     assert!(md.test(vec![]).is_err());
 }
+
+#[test]
+fn mdbook_runs_preprocessors() {
+
+    lazy_static! {
+        static ref HAS_RUN: Mutex<bool> = Mutex::new(false);
```
Michael-F-Bryan: I don't think we need to use `lazy_static!()` here, an `Arc<Mutex<bool>>` should be enough.

Michael-F-Bryan:
> this is the first time I've rebased onto a fork so hopefully I've done it right

Looks like everything's okay :+1:

src/preprocess/mod.rs
JaimeValdemoros: That makes sense, given the other fields of the `MDBook` shouldn't be needed by the preprocessor. I've made it a clone for now since a borrow would be problematic here:

https://github.com/rust-lang-nursery/mdBook/blob/bf093e2f5f0e8c7f455b6e021cb72adb974836d4/src/book/mod.rs#L205

since we'd have a reference to an inner field and additionally trying to take a reference to itself.

tests/testing.rs
JaimeValdemoros: You're right, I was trying to have a `dummy` option in the Config and have the book add the `DummyPreprocessor` to the list when it found the `dummy` option, which did require the `lazy_static` since the test wasn't constructing it directly. I should've rethought the design when I switched to using `with_preprocessor` and building the `DummyPreprocessor` directly.

JaimeValdemoros:
Just to list what I've done since the last checklist:
- Factor out explicit `default_preprocessors` method
- Match on preprocessor options instead of nested if's
- Test to make sure preprocessors are being run
- Use rustdoc link to Preprocessor trait (this worked on my machine but the Renderer link *didn't* work as it stood on my machine so probably needs checking to see if my machine does something odd)
- Log to debug as preprocessors are being run

Michael-F-Bryan:
I just had another look through the code and I'm happy with it. I think I'll merge this now and then we can do a follow-up PR to document how preprocessors work in the user guide.

Awesome work @JaimeValdemoros, I'm really happy with this!

PR #533, opened 2018-01-07
Add option to disable section label in html

vabc3:
Allow user to disable section label in the html.

User may have custom rules for naming their sections in 'SUMMARY.md', e.g. "Chapter 0", "Chapter 1.1". By default, it would show up as "1. Chapter 0", and "2.1 Chapter 1.1". In this case, we may provide an option to hide the auto generated labels.

Just saw #504 for similar goal. The change here is a bit lightweighted, which controls html output only.

### DIFF
--- a/book-example/src/format/config.md
+++ b/book-example/src/format/config.md
@@ -84,6 +84,9 @@ The following configuration options are available:
   removing the current behaviour, you can specify a set of javascript files
   that will be loaded alongside the default one.
 - **playpen:** A subtable for configuring various playpen settings.
+- **no-section-label**: mdBook by defaults adds section label in table of
+  contents column. For example, "1.", "2.1". Set this option to true to
+  disable those labels. Defaults to `false`.
 
 **book.toml**
 ```toml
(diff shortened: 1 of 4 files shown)


Michael-F-Bryan:
As @cspiegel said in #504: 

> Initially I just wanted the section numbers to not display. It didn't matter to me particularly whether they were kept track of internally or not; and I just didn't use prefix or suffix chapters, so that was a non-issue.

Adding an option to skip rendering the section numbers seems to be the best option for now. If it doesn't matter to the user whether we keep track of section numbers under the hood then that shouldn't make a difference.

What are your thoughts on having the key be `label-sections` with a default of `true`, instead of the negative `no-section-label`? I usually like to think of configuration options as enabling behaviour instead of disabling it. But again, that's totally personal preference.

vabc3:
For naming I agree that 'positive' ones sound good.  As for this case, currently `HtmlConfig` uses  'derive(Default)', which means bool options default to false. To provide a  'true' default value, will require a reimplemention of default trait. I think here it's fine to consider 'labeling section' as default behaviour, and then providing a 'no-*' option to disable it. Just like we might also see 'no*' in fstab/linux-kernel parameters. I'm fine to update it if you think it's needed.

Michael-F-Bryan:
>  think here it's fine to consider 'labeling section' as default behaviour, and then providing a 'no-' option to disable it. Just like we might also see 'no' in fstab/linux-kernel parameters.

I think I agree with you now you've put it that way :+1:

PR #535, opened 2018-01-11
Improve accessibility

sorin-davidoi:
Most commits have a body explaining the reasoning behind the changes.

### DIFF
--- a/src/renderer/html_handlebars/helpers/toc.rs
+++ b/src/renderer/html_handlebars/helpers/toc.rs
@@ -25,7 +25,7 @@ impl HelperDef for RenderToc {
                         .ok_or_else(|| RenderError::new("Type error for `path`, string expected"))?
                         .replace("\"", "");
 
-        rc.writer.write_all(b"<ul class=\"chapter\">")?;
+        rc.writer.write_all(b"<ol class=\"chapter\">")?;
 
         let mut current_level = 1;
 
@@ -45,13 +45,13 @@ impl HelperDef for RenderToc {
             if level > current_level {
                 while level > current_level {
                     rc.writer.write_all(b"<li>")?;
-                    rc.writer.write_all(b"<ul class=\"section\">")?;
+                    rc.writer.write_all(b"<ol class=\"section\">")?;
                     current_level += 1;
                 }
                 rc.writer.write_all(b"<li>")?;
             } else if level < current_level {
                 while level < current_level {
-                    rc.writer.write_all(b"</ul>")?;
+                    rc.writer.write_all(b"</ol>")?;
                     rc.writer.write_all(b"</li>")?;
                     current_level -= 1;
                 }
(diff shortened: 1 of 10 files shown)


Michael-F-Bryan:
Looking through the commit messages it seems like a lot of the changes you are proposing are quite logical and will help out a lot with accessibility :+1: 

Unfortunately I'm not a massive expert on CSS and making accessible websites, so I don't know if I'm the best person to review this PR. @projektir mentioned he's done some front-end work in the past.

---

Something I noticed is that we *may* have an issue in 29dbcdcf211861bee8a6ebe06013f2287e532d87 where you're changing from using unordered lists to ordered ones. A book is actually composed of "front matter" chapters (e.g. "introduction"), the numbered chapters themselves, and some "back matter" chapters (where you'd put "conclusion" and appendices). The front and back matter chapters are meant to be un-numbered, I *think* we were using a `<ul>` so we could do the chapter numbering manually. 

I think the best fix for this would be to change the underlying `Book` struct to separate front-matter, back-matter, and the numbered chapters into their own fields. That way your template can be simplified a lot (it just turns into a `<ul>` followed by a `<ol>` and another `<ul>`) and we get proper numbering for free, meaning we can remove the `section_number` field from each `Chapter`. It'd be a good idea to do this regardless, so I'll probably make a PR for it some time this weekend.

projektir:
I don't know much about accessibility, unfortunately. I can't figure out how to reach the sidebar with keys alone, though.

I've pulled this down and looked at RBE and everything looks fine. I recall there was a book that didn't use proper number so maybe we should look at that to make sure `<ul>` -> `<ol>` didn't break anything.

sorin-davidoi:
> I can't figure out how to reach the sidebar with keys alone, though.

You can press `Tab` until the toggle button is focused, then click `Enter` to trigger it. If expanded, pressing `Shift-Tab` will take you through all the chapters. If not expanded, the chapters are not focusable (it is generally wrong for non-visible elements to be focusable). Attaching a recording (all navigation is via keyboard).

md-book-a11y-recording

projektir:
Oh, yeah, I got it to work. Seems like this works in the current RBE as well. Anyway, LGTM!

sorin-davidoi:
> Seems like this works in the current RBE as well.

Yes, but the chapters are focusable even when the sidebar is hidden.

> There's a Travis failure, though, but I am not sure what's up with it.

Looks like a timeout issue to me?

Michael-F-Bryan:
> There's a Travis failure, though, but I am not sure what's up with it. 

Yeah I've noticed that all the Mac jobs are "waiting to be queued" indefinitely, until travis must kill it for taking too long. I think it's alright to merge these types of PRs without the Mac builds seeing as all the changes are purely browser-based and should be platform agnostic.

Michael-F-Bryan:
Oh wow, I just checked out your branch and it really makes a difference! Just from a normal user's perspective, having keyboard navigation is pretty sweet, with the theme selector being a really nice touch. 

Overall that's a big :+1: from me! Is there anything else you'd like to add to the PR, or should we think about merging it?

sorin-davidoi:
No, I think this is it for now - I might follow up later with other fixes.

Michael-F-Bryan:
Merged :tada:

Issue #475, opened 2017-10-25
too heavy on mobile


jaroslaw-weber:
mdbook has lot of features but its getting heavy on mobile. i was able to make a static website with most of the functionality and go down to 70kb (cookbook is 250, rust book is over 300) so i think there is a room for improvement. 

what can we do:
- remove jquery
- create custom icon pack
- ... ?

sorin-davidoi:
Would a PR that removes jQuery (uses native DOM APIs instead) be accepted?

Michael-F-Bryan:
It'd be really nice to get rid of the JQuery dependency. Besides the fact that it makes the generated website smaller it'll also make packaging `mdbook` a lot easier on platforms where installers aren't allowed to access the internet. 

In particular @tjis (https://github.com/rust-lang-nursery/mdBook/issues/271#issuecomment-302269128) and @infinity0 have had issues trying to package Rust on Debian because *The Book* uses `mdbook`.

sorin-davidoi:
Wouldn't that be solved by ditching the CDN links and including the resources in the repository?

steveklabnik:
No, as that goes against distro policy.

That said, we removed jquery in rustdoc as well.

sorin-davidoi:
^ Is in response to my question (why don't bundle jQuery locally) which I shouldn't have deleted.

infinity0:
That depends on the packager and the specific situation. I for one would reuse an existing polyfill that's already packaged by Debian, e.g. es6-promise, instead of packaging a new Promise polyfill. Then I'd add a Debian-specific patch to make that package use es6-promise. Usually it's just a few lines in package.json. (I've done this a few times already.)

steveklabnik:
@sorin-davidoi oops, hehe, I replied via email so I didn't see it.

Michael-F-Bryan:
Looking through the comments made from @steveklabnik and @infinity0 it seems like my earlier thought of "remove all the JS dependencies" was a bit too wide-sweeping. If it's easy enough for Debian to do a `ln -s` to override the vendored JQuery before `mdbook` gets compiled, then there's no longer any reason to remove it.

@sorin-davidoi, what are your thoughts?

sorin-davidoi:
I am not familiar with the packaging process - my view on this is that removing jQuery is good for the performance of the page (less JS to load, parse and keep in memory).

PR #538, opened 2018-01-12
Remove jQuery

sorin-davidoi:
~~This is based on #535 for two reasons:~~
- ~~I assumed that #535 is likely to be (eventually) merged, so I can rebase this later against `master` in that case~~
- ~~Make it easier for reviewers to test both of them at the same time~~

~~Depending on which browsers we need to support (couldn't find them listed), we might need the following polyfills~~:
- ~~[Array.from](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#Browser_compatibility) (no IE)~~
- ~~[String.trim](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim#Browser_compatibility) (no IE 8)~~
- ~~[Element.closest](https://caniuse.com/#feat=element-closest) (no IE)~~
- ~~[Promise](https://caniuse.com/#feat=promises) (no IE)~~
- ~~[fetch](https://caniuse.com/#feat=fetch) (no IE)~~

~~I couldn't test the changes in `editor.js`.~~

Reduces JavaScript payload size from 229.88 KB to 197.91 KB (measured on `book-example`). Helps address #475.

### DIFF
--- a/src/theme/playpen_editor/editor.js
+++ b/src/theme/playpen_editor/editor.js
@@ -4,8 +4,8 @@ window.editors = [];
         return;
     }
 
-    $(".editable").each(function() {
-        let editor = ace.edit(this);
+    Array.from(document.querySelectorAll('.editable')).forEach(function(editable) {
+        let editor = ace.edit(editable);
             editor.setOptions({
             highlightActiveLine: false,
             showPrintMargin: false,
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -431,6 +431,12 @@ table thead td {
 .light pre > .buttons i {
   margin-left: 8px;
 }
+.light pre > .buttons button {
+  color: inherit;
+  background: transparent;
+  border: none;
+  cursor: inherit;
+}
 .light pre > .result {
   margin-top: 10px;
 }
@@ -566,6 +572,12 @@ table thead td {
 }
 .coal pre > .buttons i {
   margin-left: 8px;
+}
+.coal pre > .buttons button {
+  color: inherit;
+  background: transparent;
+  border: none;
+  cursor: inherit;
 }
 .coal pre > .result {
   margin-top: 10px;
@@ -703,6 +715,12 @@ table thead td {
 .navy pre > .buttons i {
   margin-left: 8px;
 }
+.navy pre > .buttons button {
+  color: inherit;
+  background: transparent;
+  border: none;
+  cursor: inherit;
+}
 .navy pre > .result {
   margin-top: 10px;
 }
@@ -839,6 +857,12 @@ table thead td {
 .rust pre > .buttons i {
   margin-left: 8px;
 }
+.rust pre > .buttons button {
+  color: inherit;
+  background: transparent;
+  border: none;
+  cursor: inherit;
+}
 .rust pre > .result {
   margin-top: 10px;
 }
@@ -974,6 +998,12 @@ table thead td {
 }
 .ayu pre > .buttons i {
   margin-left: 8px;
+}
+.ayu pre > .buttons button {
+  color: inherit;
+  background: transparent;
+  border: none;
+  cursor: inherit;
 }
 .ayu pre > .result {
   margin-top: 10px;
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -1,4 +1,6 @@
-$( document ).ready(function() {
+var html, sidebar, sidebarLinks, sidebarToggleButton;
+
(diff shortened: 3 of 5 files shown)


Michael-F-Bryan:
> Depending on which browsers we need to support (couldn't find them listed)

I don't believe we've got a proper list of supported browsers. My perception is that the vast majority of our users are going to be developers or computer-savvy people, and as a very broad rule those sorts of people tend to use a fairly up-to-date version of Chrome or Firefox. If we can support IE then that's awesome, but if it requires large amounts of effort or hacky band-aid solutions to do then I'll leave the decision up to the developer...

sorin-davidoi:
Well, ideally we would only load the polyfills for the browsers that need it (feature-detection) - but that is non-trivial without a build system (e.g. webpack). A simpler solution would be to just https://polyfill.io/v2/docs/. What do you think?

sorin-davidoi:
I see that crates.io uses Google Analytics. Assuming there is a significant overlap between the audiences, maybe @carols10cents could share some insights about browser usage to guide this decision?

sorin-davidoi:
Actually, crates.io [does not support IE](https://github.com/rust-lang/crates.io/blob/master/config/targets.js).

steveklabnik:
I came to this thread to post RFC 1985.

sorin-davidoi:
Rebased against master.

src/theme/playpen_editor/editor.js
```diff
@@ -2,10 +2,10 @@ window.editors = [];
 (function(editors) {
     if (typeof(ace) === 'undefined' || !ace) {
         return;
-    }
-
-    $(".editable").each(function() {
-        let editor = ace.edit(this);
+    
+    
+    Arom.from(document.querySelectorAll('.editable')).forEach(function(editable) {
```
projektir: What is `Arom`? Is this supposed to be `Array`?

sorin-davidoi: Yes, fixed 👍

projektir:
Have you tested this code? Specifically, did you test that the Ace Editor boxes still work?

sorin-davidoi:
I mentioned in the description that the changes in `editor.js` have not been tested (couldn't manage to trigger the codepath - it does not even load the file in the browser). What variables need to be toggled to enable the functionality?

sorin-davidoi:
Apparently, I've missed https://github.com/rust-lang-nursery/mdBook/issues/521 - just tested and it works fine from what I can tell.

projektir:
Yes, the Ace editor is not yet documented, which is my bad! Some settings need to be enabled for it to work.

Generally, I test by running mdBook on the Rust by Example book, since that's what Ace editor was added for originally in the first place, so it's all setup there. I've just tested it with your branch and it seems to be working fine. :blush:

Michael-F-Bryan:
I just checked it out on my machine and I think there's an issue with the code expansion/hiding feature.

For example, on [this page](https://rust-lang-nursery.github.io/mdBook/format/rust.html), here's what it looks like collapsed (the default):

screenshot_2018-01-16_193604

Then hit run and you can see what it prints out:

screenshot_2018-01-16_193619

Expanded:

screenshot_2018-01-16_193627

---

This is what it's actually meant to look like when expanded:

screenshot_2018-01-16_193914

sorin-davidoi:
@Michael-F-Bryan Thanks, fixed! Sidenote, I really hate how you never know whether you are operating on a single element or on an array in jQuery (in this case, I assumed single element, so only the first hidden line was toggled).

projektir:
Ah, blast, I didn't check code expansion, good catch @Michael-F-Bryan.

Michael-F-Bryan:
I don't think there were any other noticeable bugs/regression. @sorin-davidoi, do you think this is ready to merge?

sorin-davidoi:
Yes, I would say so.

sorin-davidoi:
Is jQuery part of the official `mdBook` API? I'm thinking that this (and possibly #550) may be breaking scripts people placed in `additional-js`.

Michael-F-Bryan:
> Is jQuery part of the official mdBook API? 

I don't think we've ever said it's guaranteed to be loaded, it just happens to be because we use it internally. I'm going to say that people should make sure they explicitly load jQuery if they need it, plus this PR will be part of a major semver bump, so people shouldn't be surprised if they need to update their `additional-js`... Does that sound reasonable?

PR #541, opened 2018-01-13
Override configuration using environment variables

### LINKED ISSUE #540: If SUMMARY.md doesn't contain a file, mdbook doesn't give an error
Suppose you changed a line in SUMMARY.md of the example book from:

```markdown
    - [watch](cli/watch.md)
```
to
```markdown
    - [watch](cli/wa_tch.md)
```
The file `wa_tch.md` doesn't exist. If you ask mdbook to create the book, it doesn't give you any error or complaint that the file couldn't be found. Instead it generates an empty page. This has led to an issue with Rust's documentation staying undiscovered: https://github.com/rust-lang/rust/issues/47394 .
Therefore mdbook should fail generation of the book if it encounters a file it can't find.

Michael-F-Bryan:
This lets you override config options by setting the appropriate environment variables. This is essentially just a wrapper around the `Config::set()` method, with the original "index" being provided by the environment variable's key (after a couple transforms).

closes #540, closes #291.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ tempdir = "0.3.4"
 itertools = "0.7.4"
 tempfile = "2.2.0"
 shlex = "0.1.1"
+toml-query = "0.6"
 
 # Watch feature
 notify = { version = "4.0", optional = true }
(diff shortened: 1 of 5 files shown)


Michael-F-Bryan:
Looks like travis are still having issues with their mac builds, all 4 mac jobs are still "waiting to be queued". I think it should still be safe to merge this PR though.

PR #543, opened 2018-01-14
Increase Documentation Coverage

Michael-F-Bryan:
I've incrementally been adding `#![deny(missing_docs)]` to the various modules in `mdbook` to make sure they get documented. This adds a couple more modules to the list, hopefully adding in-depth examples and stuff.

This also adds a pretty in-depth tutorial on how to create a new backend ([rendered]).

[rendered]: https://github.com/Michael-F-Bryan/mdBook/blob/moar-docs/book-example/src/for_developers/backends.md


Michael-F-Bryan:
The preprocessor section can probably be fleshed out a bit, but other than that the docs are pretty much ready to merge. The main thing was making sure we have as much documentation coverage as possible in the API docs. The user guide can easily be updated after the 0.1 release.

PR #547, opened 2018-01-16
Meta theme color

sorin-davidoi:
Makes the browser chrome match the selected theme (to avoid having a bright white toolbar when you chose a dark theme - also looks nice). Currently only supported in Chrome for Android (left is before, right is with this patch).

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -161,6 +161,10 @@ $( document ).ready(function() {
             ace_theme = "ace/theme/dawn";
         }
 
+        setTimeout(function() {
+            document.querySelector('meta[name="theme-color"]').content = getComputedStyle(document.body).backgroundColor;
+        }, 1);
+
         if (window.ace && window.editors) {
             window.editors.forEach(function(editor) {
                 editor.setTheme(ace_theme);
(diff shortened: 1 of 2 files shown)


src/theme/book.js
```diff
@@ -161,6 +161,10 @@ $( document ).ready(function() {
             ace_theme = "ace/theme/dawn";
         }
 
+        setTimeout(function() {
```
sorin-davidoi: The timeout is needed to allow the browser to apply the stylesheet before extracting the background color.

src/theme/index.hbs
```diff
@@ -6,6 +6,7 @@
         <meta content="text/html; charset=utf-8" http-equiv="Content-Type">
         <meta name="description" content="{{ description }}">
         <meta name="viewport" content="width=device-width, initial-scale=1">
+        <meta name="theme-color" content="#ffffff" />
```
sorin-davidoi: `#ffffff` is the default value, but putting it in the markup avoids having to manage two code paths (adding the tag and updating it).

src/theme/book.js
Michael-F-Bryan: Would this mean it's possible for the menu bar to (very quickly) flash white, kinda like the flash-of-unstyled-text issue you sometimes get with CSS?

sorin-davidoi: Possible, but Chrome applies a subtle transition effect and I have not found it to be an issue.

sorin-davidoi: So it might be white for a very brief period when the page is first loaded. After that in changes instantly.

Michael-F-Bryan:
LGTM :+1:

PR #550, opened 2018-01-16
refactor: Remove store.js (use localStorage)

### LINKED ISSUE #548: Remove store.js
It is 17.25 KB, only used in three places, and we can just use `localStorage` instead (wrapped in a `try/catch` for safety). Should help to address #475.

sorin-davidoi:
Shouldn't cause any issues with the browsers listed in [RFC 1985](https://github.com/rust-lang/rfcs/blob/master/text/1985-tiered-browser-support.md#supported-browsers). The only gotcha is Safari, where `localStorage` methods throw errors in Incognito Mode (this will fail gracefully, but the changes won't be saved - I would say this is not an issue since it is Incognito Mode after all).

Closes #548.


sorin-davidoi:
I've tested this in:
- [x] Firefox (including Private Mode)
- [x] Chrome (including Incognito)
- [x] Epiphany (including Private browsing) - WebKit based, closed thing to Safari I have to test

Michael-F-Bryan:
One of the tests references `Theme`'s `store_js` field which no longer exists, hence why travis is complaining.

Michael-F-Bryan:
It looks like there are merge conflicts in `book.js`. Would you be able to resolve them, @sorin-davidoi?

sorin-davidoi:
@Michael-F-Bryan, yes, just did 👍

Michael-F-Bryan:
I think this may have broken themes the first time you load a book because local storage is handled differently. Once you change the theme local storage will be updated and you'll never see the missing styles again.

I've created an issue for this, #573.

PR #552, opened 2018-01-17
Documenting the editor #521

### LINKED ISSUE #521: Document the Ace Editor
#338 added the ace editor but we didn't really document it in the [user guide]. There are currently two parts to making code snippets editable, first you need to enable editing in your `book.toml`, then you need to add an `editable` annotation to your code blocks.

The config file:

```toml
[output.html.playpen]
editor = "./path/to/ace/editor"  # defaults to using the ace editor
editable = false # defaults to false
```

Making a specific code snippet editable

    ```rust,editable
    use std::fs::File;
    File::open("foo.txt").unwrap();
    ```

[user guide]: https://rust-lang-nursery.github.io/mdBook/

projektir:
Fixes #521.

### DIFF
--- a/book-example/src/format/theme/editor.md
+++ b/book-example/src/format/theme/editor.md
@@ -1,0 +1,40 @@
+# Editor
+
+In addition to providing runnable code playpens, mdBook optionally allows them to be editable. In order to enable editable code blocks, the following needs to be added to the ***book.toml***:
+
+```toml
+[output.html.playpen]
+editable = true
+```
+
+To make a specific block available for editing, the attribute `editable` needs to be added to it:
+
+<pre><code class="language-markdown">```rust,editable
+fn main() {
+    let number = 5;
+    print!("{}", number);
+}
+```</code></pre>
+
+The above will result in this editable playpen:
+
+```rust,editable
+fn main() {
+    let number = 5;
+    print!("{}", number);
+}
+```
+
+Note the new `Undo Changes` button in the editable playpens.
+
+## Customizing the Editor
+
+By default, the editor is the [Ace](https://ace.c9.io/) editor, but, if desired, the functionality may be overriden by providing a different folder:
+
+```toml
+[output.html.playpen]
+editable = true
+editor = "/path/to/editor"
+```
+
+Note that for the editor changes to function correctly, the `book.js` inside of the `theme` folder will need to be overriden as it has some couplings with the default Ace editor.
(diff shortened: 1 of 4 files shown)


book-example/src/format/theme/editor.md
```diff
@@ -0,0 +1,40 @@
...
+    let number = 5;
+    print!("{}", number);
+}
+```
+
+Note the new `Undo Changes` button in the editable playpens.
+
+## Customizing the Editor
+
+By default, the editor is the [Ace](https://ace.c9.io/) editor, but, if desired, the functionality may be overriden by providing a different folder:
```
projektir: It was originally intended that the playpen editor was also customizable in the same way the theme is: by providing a folder with files that are named the same. But I am a bit confused because the [theme](https://rust-lang-nursery.github.io/mdBook/format/theme/theme.html) page says that the new `theme` folder needs to be _next to_ `src`, while I always thought it was _inside_ `src`, and when I added a folder next to `src` rather than inside, mdBook did not report that it was copying anything.

I generally had trouble getting this functionality working regardless; I only ever recall theme/playpen replacement working if I manually specified the folder in the .toml. Not sure if bug, or I'm using it wrong, or... but since it's not functionality I can replicate I decided to not document it.

Michael-F-Bryan: I'm not exactly sure how it works either :( 

I'm planning to give the HTML renderer a massive makeover some time soon and in the process I'll be able to document how it actually does rendering. 

That kind of functionality should be easy to test. We just drop a `themes/` directory in the dummy book's `src/` (or next to it) and then make sure `mdbook` picks it up.

book-example/src/format/theme/editor.md
```diff
@@ -0,0 +1,40 @@
...
+
+```toml
+[output.html.playpen]
+editable = true
+editor = "/path/to/editor"
+```
+
+Note that for the editor changes to function correctly, the `book.js` inside of the `theme` folder will need to be overriden as it has some couplings with the default Ace editor.
```
Michael-F-Bryan: So what this is saying is that *technically* you can override the editor, but in doing so you also need to maintain your own patched version of `book.js`? That sounds... not ideal 😜

projektir: It is not, but it is the truth. At least at the time that I worked on this, I didn't see any feasible way of making `book.js` decoupled. That would require quite a bit of refactoring.

Michael-F-Bryan:
Looks good to me :+1:

Regarding the ability to swap out editors... I think it's probably worth removing the `editor = ...` config option seeing as, by its very nature, the editor is going to be pretty coupled to the rest of `book.js`. If people ask for that level of customisability it's not too hard to put back in again.

PR #556, opened 2018-01-18
refactor(theme/book): Split into modules

### LINKED ISSUE #559: Error in console
`themePopup is undefined` - happens when you scroll immediately after loading the page. Introduced by #551, since `themePopup` is defined after `DOMContentLoaded` is fired, but `hideTheme` is called on scroll, which may happen before the event is fired.

Does not affect the functionality of the page.

sorin-davidoi:
Advantages:
 - Easier to reason about
 - Can easily disable some modules when debugging
 - Shared dependencies are explicit (playpen_text)
 - Enables some fancier things later one (e.g. run `codeSnippets` slightly later, to avoid blocking the page)

I'm aware that `codeSnippets` should be split into the highlighter and the editor, but I'm not sure I understand exactly how they interact so I've left it as it is for now.

~Also replaced some occurences of `let` with `var` (for consistency).~

Closes #559.

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -1,18 +1,115 @@
-var html, sidebar, sidebarLinks, sidebarToggleButton, themeToggleButton, themePopup;
+// Fix back button cache problem
+window.onunload = function () { };
 
-document.addEventListener('DOMContentLoaded', function() {
+// Global variable, shared between modules
+function playpen_text(playpen) {
+    let code_block = playpen.querySelector("code");
 
-    // url
-    var url = window.location.pathname;
+    if (window.ace && code_block.classList.contains("editable")) {
+        let editor = window.ace.edit(code_block);
+        return editor.getValue();
+    } else {
+        return code_block.textContent;
+    }
+}
 
-    // Fix back button cache problem
-    window.onunload = function(){};
+(function codeSnippets() {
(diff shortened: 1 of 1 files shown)


projektir:
This looks great! Ran this on RBE, not seeing issues so far...

> Also replaced some occurences of `let` with `var` (for consistency).

This is a bit confusing (maybe even alarming) to me. `let` and `var` do different things, there's no need to make them consistent, and, if anything, `let` should be preferred... changing this could, in theory, change functionality, by making something previously only available to a block available in broader scope.

Probably not the case for this particular code file but I disagree with the reasoning, if anything, we should be moving towards using `let` more.

sorin-davidoi:
There was no need for them to be let in the first place, since they are constants. Should we make the switch they other way around - only use const and let?

projektir:
Yes, we can fix all that up in a different PR. LGTM, would probably be best if another person also pulls this down and checks for issues, though.

Havvy:
All of the empty lines have an extra space on them.

Michael-F-Bryan:
@sorin-davidoi, #557 is fixed now so if you rebase onto `master` all the CI checks should pass.

This looks like it's ready to merge, although you'll probably want to clean up the whitespace issue @Havvy mentioned first.

Issue #558, opened 2018-01-19
0.1 Release Prerequisites


Michael-F-Bryan:
As @projektir mentioned (https://github.com/rust-lang-nursery/mdBook/issues/482#issuecomment-358526332), it's about time to do a major version bump. Technically this should have been done when the configuration format was updated (#457), but I didn't realise that PR was backwards incompatible until after the fact :disappointed: 

Before we bump from `0.0.x` to `0.1` we'll probably want to resolve the following issues, seeing as they directly affect the user.

- [x] Make sure Appveyor works again so we get compiled binaries (#557) 
- [x] Document as much of the crate as possible (#543)
- [x] Error in console (#559)

(fixes #481, fixes #368)

sorin-davidoi:
I'd say #559 should be fixed as well.

Michael-F-Bryan:
Good idea. I don't know if there are any other issues we *need* to resolve before making a release, but let me know if you can think of anything.

sorin-davidoi:
Including a Rust file seems broken - taking a look (https://rust-lang-nursery.github.io/mdBook/format/rust.html).

Michael-F-Bryan:
Looks like we've checked everything off for this so I'm (finally) going to make another release :tada:

Michael-F-Bryan:
0.1.0 was just published.

PR #561, opened 2018-01-19
Added a windows-specific `tee` equivalent

### LINKED ISSUE #557: Appveyor builds are broken
One of the integration tests for alternate backends [calls out to `tee`][t] to write `stdin` to a file. Because the `tee` program doesn't exist on Windows, appveyor builds are failing.

[t]: https://github.com/rust-lang-nursery/mdBook/blob/80f42675d6b72b64447d71fc140784ea96b6bba7/tests/alternate_backends.rs#L37

Michael-F-Bryan:
This currently breaks appveyor builds because we're assuming the `tee` command will be available when testing alternate backends and the `CmdRenderer`.

(fixes #557)

### DIFF
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -151,7 +151,7 @@ impl Renderer for CmdRenderer {
     }
 
     fn render(&self, ctx: &RenderContext) -> Result<()> {
-        info!("Invoking the \"{}\" renderer", self.cmd);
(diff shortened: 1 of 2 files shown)


Michael-F-Bryan:
I've temporarily disabled the `backends_receive_render_context_via_stdin` test in `tests/alternate_backends.rs` so appveyor passes again. Once we release `0.1` (#558) I'll be able to come back and fix this properly.

PR #563, opened 2018-01-21
feat(src/theme): Scrollbar theme

sorin-davidoi:
Unfortunately, it only works in WebKit browsers for the moment. There is a [new spec](https://www.w3.org/wiki/Css-scrollbars) being written which will hopefully gain more traction.

# Before
image

# After
image

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -340,6 +340,12 @@ html:not(.sidebar-visible) #menu-bar:not(:hover).folded > #menu-bar-sticky-conta
   background-color: #fafafa;
   color: #364149;
 }
+.light .sidebar::-webkit-scrollbar {
+  background: #fafafa;
+}
+.light .sidebar::-webkit-scrollbar-thumb {
+  background: #364149;
+}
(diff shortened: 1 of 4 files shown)


Michael-F-Bryan:
Is the purpose of this just to make the scrollbar more aesthetic? It sounds like this isn't really a standardised feature just yet, more a browser-specific extension to CSS... Is it wise to include something not standardised as part of the UI?

I'll go back over that RFC about Rust's online content guidelines again and see what they say there.

Michael-F-Bryan:
@sorin-davidoi is the scroll bar meant to be this dark when using the "light" theme? I would have thought it should have a grey colour, similar to the rest of the theme.

screenshot_2018-01-26_125739

sorin-davidoi:
I gave it the color of the text, to avoid contrast issues. Should we may it less obvious?

Michael-F-Bryan:
Yeah I think so, for the light theme it's kinda jarring. At first I thought it was some sort of glitch in chrome.

PR #565, opened 2018-01-22
Reverted #549

Michael-F-Bryan:
This reverts #549 so we are running mac builds in CI again. Needed for #558.


Michael-F-Bryan:
Looks like mac builds are still queued indefinitely. I'll leave this open and see how long it takes to complete, seeing as we're going to need the mac builds to run when the 0.1 release comes out.

Michael-F-Bryan:
Mac builds are working again!

PR #568, opened 2018-01-22
fix(theme): Use aria-label alongside title

sorin-davidoi:
Tested this on macOS with VoiceOver, and it does not pick up the title as the text of the button. Kind of makes sense, since `title` and `aria-label` are not the same. This will make sure that the buttons and links are labeled properly.

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -172,7 +172,7 @@ function playpen_text(playpen) {
 
         var buttons = document.createElement('div');
         buttons.className = 'buttons';
-        buttons.innerHTML = "<i class=\"fa fa-expand\" title=\"Show hidden lines\"></i>";
+        buttons.innerHTML = "<button class=\"fa fa-expand\" title=\"Show hidden lines\" aria-label=\"Show hidden lines\"></button>";
 
         // add expand button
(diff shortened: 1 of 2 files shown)


src/theme/book.js
```diff
@@ -172,7 +172,7 @@ function playpen_text(playpen) {
 
         var buttons = document.createElement('div');
         buttons.className = 'buttons';
-        buttons.innerHTML = "<i class=\"fa fa-expand\" title=\"Show hidden lines\"></i>";
+        buttons.innerHTML = "<button class=\"fa fa-expand\" title=\"Show hidden lines\" aria-label=\"Show hidden lines\"></button>";
```
Michael-F-Bryan: Doesn't using a proper `<button>` here instead of an `<i>` mean it becomes something clickable with the usual 3D depression effect?

sorin-davidoi: No, it looks the same but it becomes focusable. The other buttons were changed in the [remove jQuery](https://github.com/rust-lang-nursery/mdBook/pull/538/files#diff-1fda0023a3432a0ba6f67574daa70c28R267) pull request, but I've somehow missed this one.

PR #570, opened 2018-01-22
fix(doc): make installation from git work again

flosse:
cargo install --git https://github.com/rust-lang-nursery/mdBook.git
leads to error: multiple packages with binaries found: mdbook, mdbook-wordcount
Fix: append name of binary (`mdbook`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ j
    the git version of mdBook yourself. Cargo makes this ***super easy***!
 
    ```
-   cargo install --git https://github.com/rust-lang-nursery/mdBook.git
+   cargo install --git https://github.com/rust-lang-nursery/mdBook.git mdbook
    ```
 
    Again, make sure to add the Cargo bin directory to your `PATH`.


Michael-F-Bryan:
I just tried this out locally and it looks like I was mistaken that removing the `[workspace]` would fix things. According to the `cargo install` RFC:

> ### Sources with multiple crates
>
> Sources like git repositories and paths can have multiple crates inside them, and Cargo needs a way to figure out which one is being installed. If there is more than one crate in a repo (or path), then Cargo will apply the following heuristics to select a crate, in order:
>
> 1. If the -p argument is specified, use that crate.
> 2. If only one crate has binaries, use that crate.
> 3. If only one crate has examples, use that crate.
> 4. Print an error suggesting the -p flag.

It looks like cargo is detecting the presence of both `mdbook` and the `mdbook-wordcount` example, so the correct way to fix it is to update your `cargo install --git ...` command as this PR proposes.

Issue #46, opened 2015-09-05
Local fallback for CDN hosted ressources


azerupi:
- [ ] Font Awesome
- [ ] Google Fonts (Open Sans)
- [ ] JQuery

Are loaded from CDN's but should have a local fallback for offline use and for when the CDN is down.

azerupi:
MathJax seems like a lot of files, currently not easily embeddable into the binary (as far as I know)..

JQuery and Font Awesome where the most important ones. I will therefore remove this issue from 0.0.2 milestone. Leave it like this until a better solution is found.

Rufflewind:
FYI: [MathJax CDN is shutting down](https://www.mathjax.org/cdn-shutting-down/).  Might want to switch to CDNJS in the meantime.

infinity0:
It would be good to add an option to disable remote loading completely. We have all of this stuff in Debian and would ideally just point to the local files. At the moment we have to carry patches in rustc instead.

infinity0:
One primary benefit of loading this stuff locally, is that my ISP really doesn't need to know when I'm reading rust documentation. :/

azerupi:
@infinity0 You mean when you are reading the docs locally?
Indeed the CDN thing is only an improvement when loading from a remote source. I am not sure how I could differentiate the two use cases though?

infinity0:
Perhaps it could read a `mdbook.toml` file that lets you define alternate local URLs for these resources? Or, get them from an environment variable.

azerupi:
Do you want to deactivate the CDN completely everywhere?

infinity0:
Yeah, that would be nice. We have them available locally in various subdirectories of `/usr/share/fonts` and `/usr/share/javascript`, though I can't speak for other distros.

azerupi:
I have made a new issue #271. Feel free to comment there :)

infinity0:
Great, thanks!

anderejd:
Thanks for working on this issue!

Is there a setting that allows me to deactivate CDN usage? Can I host all dependencies in the `build` folder?

reillysiemens:
I ran into an issue with a poor WiFi connection on the train the other day. The Rust Book was slow to load pages in what I thought was offline mode using `rustup doc --book`. The cause was the need to load the Google fonts which I did not have cached. When they were downloaded things improved, but the experience before that was frustrating.

[...]

PR #571, opened 2018-01-22
WIP: Make available offline using Service Worker

### LINKED ISSUE #546: Offline support
We could leverage the Service Worker API to make the book work offline.

Brief outline:
- Off by default
- Add option to make available offline
- Cache all JavaScript and HTML files in the Service Worker
- Handle offline state in the UI
- Handle update logic

The main question is - would this be worth it? If I understand correctly, The Rust Book is bundled with Rust (or packaged in some distributions), so most of the people interested in it you have it locally. But then again, it might be nice to provide this regardless.

sorin-davidoi:
This PR introduces a Service Worker in order to dynamically cache the assets (only what the page requests, so respecting all the flags) and statically precache the chapters. The end result is that the resulting book would be available offline in supported browsers - Chrome (desktop and Android), Opera and Firefox (desktop and Android), for now, Edge and Safari later. As a bonus, if the book is used while offline, Google Analytics events are also cached and sent out when the user is back online.

Things intentionally left unaddressed (waiting for feedback):
 - user notification/configuration - I personally think caching should be done in the background, with a small toast notification informing the user that the book is available offline.

A small caveat is that the book is not available offline after the first-page load, but only after the second one. This is because on first load we precache the chapters, but the assets will be cached only when they are requested the second time.

Alongside supporting offline usage, this will also introduce a massive performance boost on second visits, since there is no need to hit the network for any of the assets.

The Service Worker is not installed during development (same procedure as for Google Analytics).

If you want to test this:
 - Add `offline-support = true` to the `[output.html]` section
 - Disable the check for localhost [at the end of `book.js`](https://github.com/rust-lang-nursery/mdBook/pull/571/files#diff-1fda0023a3432a0ba6f67574daa70c28R539)

Closes #546.

### DIFF
--- a/src/theme/sw.js
+++ b/src/theme/sw.js
@@ -1,0 +1,29 @@
+importScripts('https://unpkg.com/workbox-sw@2.0.3/build/importScripts/workbox-sw.dev.v2.0.3.js');
+
+// clientsClaims tells the Service Worker to take control as soon as it's activated
+const workbox = new WorkboxSW({ clientsClaim: true });
+
+// https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#stale-while-revalidate
+// TLDR: If there's a cached version available, use it, but fetch an update for next time.
+const staleWhileRevalidate = workbox.strategies.staleWhileRevalidate();
+
+// Remote fonts and JavaScript libraries
+workbox.router.registerRoute(new RegExp('https:\/\/fonts\.googleapis\.com\/css'), staleWhileRevalidate);
+workbox.router.registerRoute(new RegExp('https:\/\/fonts\.gstatic\.com'), staleWhileRevalidate);
+workbox.router.registerRoute(new RegExp('https:\/\/maxcdn\.bootstrapcdn\.com\/font-awesome'), staleWhileRevalidate);
+workbox.router.registerRoute(new RegExp('https:\/\/cdnjs\.cloudflare\.com\/ajax\/libs\/mathjax'), staleWhileRevalidate);
+workbox.router.registerRoute(new RegExp('https:\/\/cdn\.jsdelivr\.net\/clipboard\.js'), staleWhileRevalidate);
+
+// Local resources
+workbox.router.registerRoute(new RegExp('\.js$'), staleWhileRevalidate);
+workbox.router.registerRoute(new RegExp('\.css$'), staleWhileRevalidate);
+
+// Here hbs_renderer.rs will inject the chapters, making sure they are precached.
+//  
+// const chapters = [
+//     { url: '/', revision: '11120' },
+//     { url: 'cli/cli-tool.html', revision: '12722' },
+//     { url: 'cli/init.html', revision: '12801' },
+// ];
+//   
+//   workbox.precaching.precacheAndRoute(chapters);
--- a/book-example/src/format/config.md
+++ b/book-example/src/format/config.md
@@ -72,6 +72,8 @@ The following configuration options are available:
 - **theme:** mdBook comes with a default theme and all the resource files
   needed for it. But if this option is set, mdBook will selectively overwrite
   the theme files with the ones found in the specified folder.
+- **offline-support** Precache the chapters so that users can view the book
+  while offline. Available in [browsers supporting Service Worker](https://caniuse.com/#feat=serviceworkers).
 - **curly-quotes:** Convert straight quotes to curly quotes, except for
   those that occur in code blocks and code spans. Defaults to `false`.
 - **google-analytics:** If you use Google Analytics, this option lets you
--- a/src/config.rs
+++ b/src/config.rs
@@ -406,6 +406,8 @@ pub struct HtmlConfig {
     pub curly_quotes: bool,
     /// Should mathjax be enabled?
     pub mathjax_support: bool,
+    /// Cache chapters for offline viewing
+    pub offline_support: bool,
     /// An optional google analytics code.
     pub google_analytics: Option<String>,
     /// Additional CSS stylesheets to include in the rendered page's `<head>`.
@@ -498,6 +500,7 @@ mod tests {
         theme = "./themedir"
         curly-quotes = true
         google-analytics = "123456"
+        offline-support = true
         additional-css = ["./foo/bar/baz.css"]
 
         [output.html.playpen]
@@ -529,6 +532,7 @@ mod tests {
         };
         let html_should_be = HtmlConfig {
             curly_quotes: true,
+            offline_support: true,
             google_analytics: Some(String::from("123456")),
             additional_css: vec![PathBuf::from("./foo/bar/baz.css")],
             theme: Some(PathBuf::from("./themedir")),
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -9,14 +9,22 @@ use regex::{Captures, Regex};
 
 #[allow(unused_imports)] use std::ascii::AsciiExt;
 use std::path::{Path, PathBuf};
-use std::fs::{self, File};
+use std::fs::{self, File, OpenOptions};
 use std::io::{Read, Write};
 use std::collections::BTreeMap;
 use std::collections::HashMap;
+use std::collections::hash_map::DefaultHasher;
+use std::hash::Hasher;
 
 use handlebars::Handlebars;
 
 use serde_json;
+
+#[derive(Default)]
+pub struct ChapterFile {
+    pub path: String,
+    pub revision: u64,
+}
 
 #[derive(Default)]
 pub struct HtmlHandlebars;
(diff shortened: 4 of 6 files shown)


sorin-davidoi:
This also opens up the possibility of implementing search in a different way than #472 does it. Since all the pages are in the cache, we can implement it client-side only, though we would have to see what impact that would have on the memory and CPU usage.

sorin-davidoi:
I've looked into how much space will the chapters take in cache:
- mdBook: 329 KiB
- The Rust Programming Language, 1st edition: 2.6 MiB
- The Rust Programming Language, 2nd edition: 4.6 MiB

projektir:
This is great!

The only thing I've found is the clipboard seems broken in offline mode.

projektir:
Oh, another thing I am seeing is on this branch, every time I click on a sidebar chapter, the sidebar closes.

Michael-F-Bryan:
> Alongside supporting offline usage, this will also introduce a massive performance boost on second visits, since there is no need to hit the network for any of the assets.

I have a couple questions regarding caching and how long the service worker persists for:

- How do we figure out when to invalidate the cache? For example, say I visited *The Book* last week and they've made an update since then, will I see the changed content? 
- Does the service worker/cache persist across browser sessions?

sorin-davidoi:
> How do we figure out when to invalidate the cache? For example, say I visited The Book last week and they've made an update since then, will I see the changed content?

When you visit again, you will receive a new `sw.js` which will re-fetch the chapters that have changed. But the first page you see will be from cache, only subsequent visits (when you navigate to a new page) will serve the new content. The common pattern for this is to show a small toast notification (e.g. "Rust by example has been updated! Refresh to see the new version."), like https://www.chromestatus.com/ is doing.

> Does the service worker/cache persist across browser sessions?

Yes.

I've just published this over at https://sorin-davidoi.github.io/mdBook/ for easier testing. Load the page, refresh and then go offline (Chrome Dev Tools -> Network tab -> Offline). You should be able to see all the content.

Michael-F-Bryan:
I just had a look at Chrome's audit tool and it mentioned something about the "service worker does not serve the manifest's start_url". Other than that it seems to work quite well.

screenshot_2018-01-25_012525

---

> Oh, another thing I am seeing is on this branch, every time I click on a sidebar chapter, the sidebar closes.

@sorin-davidoi is @projektir's problem something to do with local storage not persisting the sidebar's state properly?

sorin-davidoi:
> I just had a look at Chrome's audit tool and it mentioned something about the "service worker does not serve the manifest's start_url"

Yes, we should also have a `manifest.json`. I want to add it, but not in this PR.

> Oh, another thing I am seeing is on this branch, every time I click on a sidebar chapter, the sidebar closes.

Not sure, since I can't replicate it. Maybe his browser window [was not wide enough](https://github.com/rust-lang-nursery/mdBook/blob/master/src/theme/index.hbs#L73)?

Michael-F-Bryan:
> What is your opinion on not relying on CDNs and serving the assets locally?

Yes please! 

A lot of people have asked about serving all assets locally and making it possible to view when not connected to the internet. Making the entire thing self-contained is also pretty important for people who want to package `mdbook` or anything that depends on it (like Rust). In particular the Debian guys had a lot of trouble because there's no way to verify the javaScript we use, and it's frowned upon for applications to reach out to the internet without the user expressly asking it to (e.g. serving locally but still fetching stuff from a CDN).

I really want to go through the HTML renderer and rewrite large chunks of the internals. Over time that section of the codebase has "grown organically" as new features get added and different people work on it, and the technical debt is starting to make things brittle and hard to maintain. A better dependency and asset management system will probably just "fall out" when we go through and restructure everything.

sorin-davidoi:
@Michael-F-Bryan Perfect, what about I split this PR into three?
 - A pull request to remove the CDNs
 - A pull request to move the asset paths in Rust (as in the code sample above)
 - This PR, which should be much cleaner and easier to review once the previous two are merged

Michael-F-Bryan:
> A pull request to move the asset paths in Rust (as in the code sample above)

How were you planning to move the assets? As well as the standard use case when someone doesn't have access to the internet, you may want to skim through #46 and #271 to give you a better idea for what Debian (and other organisations with similar requirements) are looking for.

In particular, @tjis [mentioned](https://github.com/rust-lang-nursery/mdBook/issues/271#issuecomment-302269128):

> Debian packages aren't supposed to make network connections without the user's consent. Also, packages aren't supposed to contain embedded resources that could be provided by other packages instead. Therefore, both the CDN and the fallback are problematic for Debian.
>
> We are running into this issue while packaging rust, which uses mdBook to build the various rust books. In order to make the generated books comply with debian policy, we strip the embedded resources from mdBook and alter the templates so that the generated documentation complies.
>
> But this is far from ideal. Not only might future mdBook changes break our present patch, it also means we have to ship a patched copy of mdBook with the rust source package, rather than using a pristine packaged version of mdBook and build-depend on that.

sorin-davidoi:
So, our external dependencies are (as far as I can tell):
 - clipboard.js
 - MathJax
 - Google Analytics
 - Google Fonts
 - Font Awesome
 - highlight.js
 - Ace Editor

My first thought was to hard-code the paths, but it seems that reading them from the environment (falling back to the local assets) might be a better solution?

Michael-F-Bryan:
> My first thought was to hard-code the paths, but it seems that reading them from the environment (falling back to the local assets) might be a better solution?

I had something similar in mind. Cargo lets a build script set environment variables during the compilation process, which you can pair with the `env!()` macro to pass an arbitrary string from the build script to the compiled crate. The idea is we'd use something like `const HIGHLIGHT_JS: &[u8] = include_bytes!(env!("HIGHLIGHT_JS"));` to embed a copy of `highlight.js` in the binary at *compile time*. 

This approach lets the people packaging `mdbook` override the file to use while compiling and installing, for example using `highlight.js` from the debian repositories instead of our vendored version.

Ideally all dependencies should be embedded in the `mdbook` executable, that way we don't need to call out to a CDN.

sorin-davidoi:
Closing for now. Will probably open PRs with the issues mentioned above before taking another shot at this.

jasonwilliams:
> This also opens up the possibility of implementing search in a different way than #472 does it. Since all the pages are in the cache, we can implement it client-side only, though we would have to see what impact that would have on the memory and CPU usage.
> 
> The rough idea would be to run the following steps lazily (e.g. when the search field is focused, after the chapters are precached, inside a `requestIdleCallback` call):
> 
> * Fetch chapters from the Service Worker cache
> * (Re)Generate the search index using Elisticlunr inside a Web Worker (off the main thread)
> * Cache the search index in `localStorage` or some other place

This sounds awesome, although we should keep the scope small for now and do this in a separate PR?

jasonwilliams:
> > My first thought was to hard-code the paths, but it seems that reading them from the environment (falling back to the local assets) might be a better solution?
> 
> I had something similar in mind. Cargo lets a build script set environment variables during the compilation process, which you can pair with the `env!()` macro to pass an arbitrary string from the build script to the compiled crate. The idea is we'd use something like `const HIGHLIGHT_JS: &[u8] = include_bytes!(env!("HIGHLIGHT_JS"));` to embed a copy of `highlight.js` in the binary at _compile time_.
> 
> This approach lets the people packaging `mdbook` override the file to use while compiling and installing, for example using `highlight.js` from the debian repositories instead of our vendored version.
> 
> Ideally all dependencies should be embedded in the `mdbook` executable, that way we don't need to call out to a CDN.

@Michael-F-Bryan Does this mean that the build script pulls all those external deps from the CDNs when building? Or do we manually fetch the assets and put them in some folder? 
Would like to pick up this bit but unsure on the implementation

src/theme/sw.js
```diff
@@ -0,0 +1,29 @@
...
+workbox.router.registerRoute(new RegExp('https:\/\/cdn\.jsdelivr\.net\/clipboard\.js'), staleWhileRevalidate);
+
+// Local resources
+workbox.router.registerRoute(new RegExp('\.js$'), staleWhileRevalidate);
+workbox.router.registerRoute(new RegExp('\.css$'), staleWhileRevalidate);
+
+// Here hbs_renderer.rs will inject the chapters, making sure they are precached.
+//  
+// const chapters = [
+//     { url: '/', revision: '11120' },
```
jasonwilliams: @sorin-davidoi how do you plan to calculate the revision here?

jasonwilliams:
> A small caveat is that the book is not available offline after the first-page load, but only after the second one. This is because on first load we precache the chapters, but the assets will be cached only when they are requested the second time.

@sorin-davidoi is this because you need to actually hit the (chapter) page for those assets to load and go into the cache? So you won't actually cache a chapter until you visit it?

sorin-davidoi:
To be honest I don't remember :disappointed:

jasonwilliams:
@sorin-davidoi how were you testing the `mdBook` binary against the book, were you cloning the `rust` repo to get the rust book markdown files?

Michael-F-Bryan:
> Does this mean that the build script pulls all those external deps from the CDNs when building? Or do we manually fetch the assets and put them in some folder?

I believe the current state of affairs is that we've [vendored a copy](https://github.com/rust-lang-nursery/mdBook/tree/master/src/theme) of the files in the `mdbook` repository itself. So it's not pulled from a CDN or `npm` when building the crate.

> @sorin-davidoi is this because you need to actually hit the (chapter) page for those assets to load and go into the cache? So you won't actually cache a chapter until you visit it?

That sounds about right. It's been a while since the original PR though, so there's a good chance the story around service workers has changed or this could be implemented differently.

peaceshi:
nice.
I did something like this by using CDN.
https://peaceshi.github.io/GameProgrammingPatterns/index.html

jasonwilliams:
There’s still a couple of bugs with https://github.com/rust-lang/mdBook/pull/1000 but I believe it’s getting there

jasonwilliams:
Help Needed!

https://github.com/rust-lang/mdBook/pull/1000 seems to be working pretty well from what i can see so far, I just need some feedback now. The PR allows the rust book to be read offline, new changes will still take affect.

You can just navigate to https://jason-williams.co.uk/book/ and try it out, then leave any feedback in the issue above.

PR #572, opened 2018-01-22
fix(theme): Sticky header support in Safari

sorin-davidoi:
Safari scrolls on body, not on the html tag. It also needs sticky to be prefixed.


Michael-F-Bryan:
LGTM :+1:

Issue #573, opened 2018-01-23
Themes are broken the first time a page is loaded


Michael-F-Bryan:
I noticed that on `master` when you open up a book you've previously visited it'll be rendered without any CSS styling. This fixes itself when you change the themes and will never show itself again for that domain. 

I've seen something similar in the past where we'd changed the theme name being stored in local storage (because of a typo it went from `light` to `"light"`, or vice versa). I believe the underlying issue is that when loading the theme we don't properly handle the case where the theme specified in local storage doesn't exist.

I noticed this when rendering *The Book* as part of my final checks before pushing the 0.1 release. Unfortunately it blocks the release because styling will be broken for every single person who visits *The Book* (or any other document generated by `mdbook`) for the first time after this release and has been there before. 

@sorin-davidoi we may need to have a look at #550, possibly reverting it if there isn't an easy fix.

sorin-davidoi:
I can't seem to be able to replicate this.

Michael-F-Bryan:
When we had this issue a while back it was something to do with how the `index.hbs` template was being rendered. Chances are we may have accidentally reverted that original fix when updating the templates, so me mentioning #550 was probably a red herring. 

We may want to insert a check in `book.js` to make sure local storage gives us a valid theme, even if there wasn't one before #550. That way we can fall back to the default theme if we try to load an invalid one for whatever reason.

Michael-F-Bryan:
Fixed :tada:

PR #574, opened 2018-01-23
fix(theme/index): Use localStorage, not store

sorin-davidoi:
~This still doesn't explain #573.~ Added a commit to fix it.

We could solve this in a few ways:
 - maintain the convention (in 4/5 places) that values are stored in `localStorage` wrapped in quotes
 - introduce theme validation in two separate places (`index.html` and `book.js`)
 - unwrap items in `localStorage` in one place

For this PR I went with the last option.

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -46,7 +46,7 @@
         <!-- Set the theme before any content is loaded, prevents flash -->
         <script type="text/javascript">
             var theme;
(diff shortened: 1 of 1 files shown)


Michael-F-Bryan:
I believe this fixes the issue. Are we ready to merge, @sorin-davidoi?

sorin-davidoi:
Yes, I would say this is ready to go.

PR #575, opened 2018-01-23
fix(theme/book): Use passive listeners for touchstart, touchmove

sorin-davidoi:
["Passive event listeners are a new feature in the DOM spec that enable developers to opt-in to better scroll performance by eliminating the need for scrolling to block on touch and wheel event listeners."](https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md)

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -422,7 +422,7 @@ function playpen_text(playpen) {
             x: e.touches[0].clientX,
             time: Date.now()
         };
-    });
+    }, { passive: true });
 
     document.addEventListener('touchmove', function (e) {
         if (!firstContact)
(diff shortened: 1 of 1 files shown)


Michael-F-Bryan:
Do you know what older browsers will do if you pass in a `{passive: true}`? It'd be annoying if an exception was thrown while registering scroll callbacks.

sorin-davidoi:
[The support matrix](https://caniuse.com/#feat=passive-event-listener) covers everything in the RFC. Unfortunately, I don't have any of the browsers that don't support it at hand, but looking at [how the detector](https://github.com/rafrex/detect-passive-events/blob/master/src/index.js) works, they will just ignore it.

PR #583, opened 2018-01-25
add `clean` subcommand

bmusin:
Hi,
Was working on adding `clean`'.
However, don't know how to correctly return Err in case of error.

### DIFF
--- a/src/bin/clean.rs
+++ b/src/bin/clean.rs
@@ -1,0 +1,30 @@
+use std::fs;
+use std::path::PathBuf;
+use clap::{App, ArgMatches, SubCommand};
+use mdbook::MDBook;
+use mdbook::errors::*;
+use get_book_dir;
+
+// Create clap subcommand arguments
+pub fn make_subcommand<'a, 'b>() -> App<'a, 'b> {
+    SubCommand::with_name("clean")
+        .about("Delete built book")
+        .arg_from_usage(
+            "-d, --dest-dir=[dest-dir] 'The directory of built book{n}(Defaults to ./book when \
+             omitted)'",
+        )
+}
+
+// Clean command implementation
+pub fn execute(args: &ArgMatches) -> ::mdbook::errors::Result<()> {
+    let book_dir = get_book_dir(args);
+    let book = MDBook::load(&book_dir)?;
+
+    let dir_to_remove = match args.value_of("dest-dir") {
+        Some(dest_dir) => PathBuf::from(dest_dir),
+        None => book.root.join(&book.config.build.build_dir),
+    };
+    fs::remove_dir_all(&dir_to_remove).chain_err(|| "Unable to remove the build directory")?;
+
+    Ok(())
+}
--- a/src/bin/mdbook.rs
+++ b/src/bin/mdbook.rs
@@ -19,6 +19,7 @@ use env_logger::Builder;
 use mdbook::utils;
 
 pub mod build;
+pub mod clean;
 pub mod init;
 pub mod test;
 #[cfg(feature = "serve")]
@@ -44,7 +45,8 @@ fn main() {
                              at: https://github.com/rust-lang-nursery/mdBook")
                 .subcommand(init::make_subcommand())
                 .subcommand(build::make_subcommand())
-                .subcommand(test::make_subcommand());
+                .subcommand(test::make_subcommand())
+                .subcommand(clean::make_subcommand());
 
     #[cfg(feature = "watch")]
     let app = app.subcommand(watch::make_subcommand());
@@ -55,6 +57,7 @@ fn main() {
     let res = match app.get_matches().subcommand() {
         ("init", Some(sub_matches)) => init::execute(sub_matches),
         ("build", Some(sub_matches)) => build::execute(sub_matches),
+        ("clean", Some(sub_matches)) => clean::execute(sub_matches),
         #[cfg(feature = "watch")]
         ("watch", Some(sub_matches)) => watch::execute(sub_matches),
         #[cfg(feature = "serve")]


src/bin/clean.rs
```diff
@@ -0,0 +1,31 @@
...
+// Init command implementation
+pub fn execute(args: &ArgMatches) -> Result<()> {
+    let mut book_dir = get_book_dir(args);
+    book_dir.push("book");
+
+    match fs::remove_dir_all(book_dir.to_str().unwrap()) {
+        Ok(_) => (),
+        m @ _ => {
+            error!("Can't delete built book");
+            return Err("error message")
```
Michael-F-Bryan: You aren't returning a `Result<(), Error>` here, it's just a string. If you change it to `Err(Error::from("error message"))` it'll compile.

src/bin/clean.rs
```diff
@@ -0,0 +1,31 @@
...
+}
+
+// Init command implementation
+pub fn execute(args: &ArgMatches) -> Result<()> {
+    let mut book_dir = get_book_dir(args);
+    book_dir.push("book");
+
+    match fs::remove_dir_all(book_dir.to_str().unwrap()) {
+        Ok(_) => (),
+        m @ _ => {
```
Michael-F-Bryan: What's the point of the `m @ _` here? It doesn't look like you're using the `m` anywhere...

bmusin: I thought about using it, passing somehow to Err.

src/bin/clean.rs
bmusin: @Michael-F-Bryan, i rewrote as:
```
    let mut book_dir = get_book_dir(args);
    book_dir.push("book");

    match fs::remove_dir_all(book_dir.to_str().unwrap()) {
        Ok(_) => Ok(()),
        Err(_) => {
            error!("Can't delete built book");
            return Err(Error::from("Can't delete built book"))
        },
    }
}
```
Still it gives error:

src/bin/clean.rs
Michael-F-Bryan: It'd be much better to do something like this:

```rust
fs::remove_dir_all(&book_dir).chain_err(|| "Unable to remove the build directory")?;
```

The `chain_err()` method lets you add extra context to an error ([docs]), which we then use to print something like this to the user:

```
Error: Unable to remove build directory
  Caused By: File not found
```

It also helps with your `return Err("message")` issue later on. You probably want to read the [error handling] chapter in *The Book* to get a better understanding of what's happening.

[docs]: https://docs.rs/error-chain/0.11.0/error_chain/#chaining-errors
[error handling]: https://doc.rust-lang.org/book/second-edition/ch09-02-recoverable-errors-with-result.html#propagating-errors

bmusin:
@Michael-F-Bryan, I fixed.

src/bin/clean.rs
```diff
@@ -0,0 +1,23 @@
...
+use get_book_dir;
+
+// Create clap subcommand arguments
+pub fn make_subcommand<'a, 'b>() -> App<'a, 'b> {
+    SubCommand::with_name("clean").about("Delete built book")
+}
+
+// Init command implementation
+pub fn execute(args: &ArgMatches) -> Result<()> {
+    use mdbook::errors::*;
```
Michael-F-Bryan: It probably doesn't make sense to do a glob import here. Why don't you move it to the top of the file and replace the `use mdbook::errors::Result`?

src/bin/clean.rs
```diff
@@ -0,0 +1,23 @@
...
+    let mut book_dir = get_book_dir(args);
+    book_dir.push("book");
```
Michael-F-Bryan: I don't think this respects when people use a different build directory. For example, you can set it [in your `book.toml`][1] or override it using [the `--dest-dir` flag][2]. Instead, you probably want to join the book root with either the [`config.build.build_dir`][3] or the `--dest-dir` argument (if manually specified by the user).

[1]: https://rust-lang-nursery.github.io/mdBook/format/config.html#build-options
[2]: https://rust-lang-nursery.github.io/mdBook/cli/build.html#--dest-dir
[3]: http://rust-lang-nursery.github.io/mdBook/mdbook/config/struct.BuildConfig.html#structfield.build_dir

src/bin/clean.rs
```diff
@@ -0,0 +1,23 @@
...
+use std::io::Write;
+use std::fs;
+use clap::{App, ArgMatches, SubCommand};
+use mdbook::MDBook;
+use mdbook::errors::Result;
+use get_book_dir;
+
+// Create clap subcommand arguments
+pub fn make_subcommand<'a, 'b>() -> App<'a, 'b> {
+    SubCommand::with_name("clean").about("Delete built book")
```
Michael-F-Bryan: Should this accept a `--dest-dir` argument? (like `mdbook build` and friends)

bmusin: Fixed.

bmusin: WIP: did.

src/bin/clean.rs
bmusin: WIP: did.

src/bin/clean.rs
bmusin: @Michael-F-Bryan, what do you mean by "join book root"?

bmusin: @Michael-F-Bryan, I tried my best, now it should work. Please test.

src/bin/clean.rs
```diff
@@ -0,0 +1,31 @@
...
+             omitted)'",
+        )
+}
+
+// Clean command implementation
+pub fn execute(args: &ArgMatches) -> ::mdbook::errors::Result<()> {
+    let book_dir = get_book_dir(args);
+    let book = MDBook::load(&book_dir)?;
+
+    let dir_to_remove = if let Some(dest_dir) = args.value_of("dest-dir") {
```
Michael-F-Bryan: This `if let` should probably swapped with a normal `match` statement to make it a bit more readable.

src/bin/clean.rs
```diff
@@ -0,0 +1,31 @@
...
+        PathBuf::from(dest_dir)
+    } else {
+        book.config.build.build_dir
```
Michael-F-Bryan: This is *almost* right. The problem is that `build_dir` is relative to a **book's root directory**, so when calling `mdbook clean` from a different directory (e.g. `mdbook clean /path/to/book`) it'll delete `build_dir` relative to the **current directory**.

This should really be `book.root.join(&book.config.build.build_dir)`, it will join the `book.root` path and `build_dir` to get the full path to the build directory.

src/bin/clean.rs
bmusin: WIP: did.

src/bin/clean.rs
bmusin: Did.

bmusin:
@Michael-F-Bryan, I fixed.

Michael-F-Bryan:
@bmusin awesome, this looks ready to merge. We'll need to mention how to use `mdbook clean` in the user guide and the README, but that can be done in a separate commit.

PR #590, opened 2018-01-27
Updated the light theme to have a lighter scrollbar

Michael-F-Bryan:
The current scrollbar colour for the light theme is a little jarring and doesn't quite feel like it fits the rest of the theme. This just pulls the scrollbar colour out into its own variable which defaults to the `$sidebar-fg`, overriding it to `#ccc` for the light theme.

---

Before:

screenshot_2018-01-26_125739

After:

screenshot_2018-01-27_105044


Michael-F-Bryan:
@sorin-davidoi, can you have a look at this and make sure it looks alright?

PR #592, opened 2018-01-27
WIP: Add support for virtual chapters

### LINKED ISSUE #483: Allow for non-link parent items
I'd like to have a layout like this:

1. A chapter about foo
  - Section 1
  - Section 2
2. A chapter about bar
  - Section 1
  - Section 2

where the top-level chapter items are *not* links, but rather, each chapter begins with the first subsection.

### LINKED ISSUE #576: mdBook forbids "empty chapters" since 0.1.0
Until version 0.0.28, it was possible to have a `SUMMARY.md` like this:

```markdown
# Summary

- [Real chapter](./real.md)
- [Empty chapter]()
```

While the "real chapter" was rendered and given a link in the summary on the left-hand side, the "empty chapter" was just given an entry in the summary without a link.

This behaviour was very helpful in lots of cases, e.g.
- when the book was not yet complete or
- to give the summary additional structure.

Since version 0.1.0, empty chapters are not longer supported. They result in the following error:

```
$ mdbook build
2018-01-24 16:08:20 [ERROR] (mdbook::utils): Error: Summary parsing failed
2018-01-24 16:08:20 [ERROR] (mdbook::utils): Caused By: There was an error parsing the numbered chapters
2018-01-24 16:08:20 [ERROR] (mdbook::utils): Caused By: Error at line 4, column 16: You can't have an empty link.
```

Is the new behaviour intended? If yes, is it possible to revert that decision since the old behaviour has practical benefits?

teiesti:
**Beware:** This is currently WIP.

This pull request introduces virtual chapters which do not correspondent to a file on disk. Virtual chapters can be used for namespacing purposes since they have an entry in the `SUMMARY.md` and the rendered navigation on the left side.

Work to be done:
- [X] Add a variant to `BookItem` and `SummaryItem`.
- [ ] Modify the parser.
  - [X] Syntax A: `- [Some virtual chapter]()`
  - [ ] Syntax B: `- Some virtual chapter` if desired (**Design decision pending!**)
- [ ] Modify the renderer.
- [ ] Add test cases.
- [ ] Adjust the crate documentation.
- [ ] Write a passage for the user guide. 

Closes #483.
Closes #576.

### DIFF
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -133,6 +133,9 @@ pub enum BookItem {
     VirtualChapter(VirtualChapter),
     /// A section separator.
     Separator,
+    // To make sure clients have a `_ =>` case
+    #[doc(hidden)]
+    __NonExhaustive,
 }
 
 impl From<Chapter> for BookItem {
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -136,13 +136,13 @@ pub enum BookItem {
 }
 
 impl From<Chapter> for BookItem {
-    fn from(other: Chapter) -> BookItem {
+    fn from(other: Chapter) -> Self {
         BookItem::Chapter(other)
     }
 }
 
 impl From<VirtualChapter> for BookItem {
-    fn from(other: VirtualChapter) -> BookItem {
+    fn from(other: VirtualChapter) -> Self {
         BookItem::VirtualChapter(other)
     }
 }
@@ -199,8 +199,8 @@ pub struct VirtualChapter {
 
 impl VirtualChapter {
     /// Create a new chapter with the provided content.
-    pub fn new(name: &str, content: String) -> VirtualChapter {
-        VirtualChapter {
+    pub fn new(name: &str, content: String) -> Self {
+        Self {
             name: name.to_string(),
             content: content,
             ..Default::default()
@@ -246,6 +246,7 @@ fn load_summary_item<P: AsRef<Path>>(
             load_chapter(link, src_dir, parent_names).map(|c| BookItem::Chapter(c))
         }
         SummaryItem::VirtualLink(ref link) => unimplemented!(),
+        SummaryItem::Separator => Ok(BookItem::Separator),
     }
 }
 
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -245,6 +245,7 @@ fn load_summary_item<P: AsRef<Path>>(
         SummaryItem::Link(ref link) => {
             load_chapter(link, src_dir, parent_names).map(|c| BookItem::Chapter(c))
         }
+        SummaryItem::VirtualLink(ref link) => unimplemented!(),
     }
 }
 
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -129,6 +129,8 @@ where
 pub enum BookItem {
     /// A nested chapter.
     Chapter(Chapter),
+    /// A nested virtual chapter.
+    VirtualChapter(VirtualChapter),
     /// A section separator.
     Separator,
 }
@@ -136,6 +138,12 @@ pub enum BookItem {
 impl From<Chapter> for BookItem {
     fn from(other: Chapter) -> BookItem {
         BookItem::Chapter(other)
+    }
+}
+
+impl From<VirtualChapter> for BookItem {
+    fn from(other: VirtualChapter) -> BookItem {
+        BookItem::VirtualChapter(other)
     }
 }
 
@@ -170,6 +178,31 @@ impl Chapter {
             content: content,
             path: path.into(),
             parent_names: parent_names,
+            ..Default::default()
+        }
+    }
+}
+
+/// The representation of a "virtual chapter", available for namespacing
+/// purposes and not mapping to a file on disk.
+#[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize)]
+pub struct VirtualChapter {
+    /// The chapter's name.
+    pub name: String,
+    /// The chapter's contents.
+    pub content: String,
+    /// The chapter's section number, if it has one.
+    pub number: Option<SectionNumber>,
+    /// Nested items.
+    pub sub_items: Vec<BookItem>,
+}
+
+impl VirtualChapter {
+    /// Create a new chapter with the provided content.
+    pub fn new(name: &str, content: String) -> VirtualChapter {
+        VirtualChapter {
+            name: name.to_string(),
+            content: content,
             ..Default::default()
         }
     }
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -301,10 +301,6 @@ fn load_virtual_chapter<P: AsRef<Path>>(link: &VirtualLink, src_dir: P) -> Resul
     ch.sub_items = sub_items;
 
     Ok(ch)
-        /* TODO Open question: Do we really want to return a `Result<_>` here?
-         * Pro: Function would have almost signature as load_chapter()
-         * Con: We do not use it!
-         */
 }
 
 /// A depth-first iterator over the items in a book.
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -4,7 +4,7 @@ use std::collections::VecDeque;
 use std::fs::{self, File};
 use std::io::{Read, Write};
 
-use super::summary::{parse_summary, Link, SectionNumber, Summary, SummaryItem};
+use super::summary::{parse_summary, Link, SectionNumber, Summary, SummaryItem, VirtualLink};
 use config::BuildConfig;
 use errors::*;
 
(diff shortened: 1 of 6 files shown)


src/book/book.rs
```diff
@@ -166,6 +174,31 @@ impl Chapter {
...
 }
 
+/// The representation of a "virtual chapter", available for namespacing
+/// purposes and not mapping to a file on disk.
+#[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize)]
+pub struct VirtualChapter {
+    /// The chapter's name.
+    pub name: String,
+    /// The chapter's contents.
+    pub content: String,
```
azerupi: Is there any content in a virtual chapter?

src/book/book.rs
```diff
@@ -231,6 +267,31 @@ fn load_chapter<P: AsRef<Path>>(link: &Link, src_dir: P) -> Result<Chapter> {
...
+        .iter()
+        .map(|i| load_summary_item(i, src_dir))
+        .collect::<Result<Vec<_>>>()?;
+
+    ch.sub_items = sub_items;
+
+    Ok(ch)
+        /* TODO Open question: Do we really want to return a `Result<_>` here?
+         * Pro: Function would have almost signature as load_chapter()
+         * Con: We do not use it!
```
azerupi: But we do use it. When you call `load_summary_item` recursively on the sub-items you return an error if any of the sub-items error.

src/book/book.rs
Michael-F-Bryan: I don't believe so. A virtual chapter just has a name, maybe a number, and a bunch of sub-items.

src/book/summary.rs
```diff
@@ -101,11 +100,48 @@ impl Default for Link {
     }
 }
 
-/// An item in `SUMMARY.md` which could be either a separator or a `Link`.
+/// A linked virtual chapter in the `SUMMARY.md`, possibly with nested entries.
+///
+/// This is roughly the equivalent of `- Some virtual section`.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct VirtualLink {
```
Michael-F-Bryan: What would the difference be between a normal link and a virtual one? My understanding is you'd write a normal link as `- [Some Chapter](./chapter.md)` and a virtual link as `- [Some Chapter]()` (i.e. virtual chapters have no link in `SUMMARY.md`).

src/book/book.rs
teiesti: You're right. I overlooked the `?`-operator in line 284. Thanks!

src/book/summary.rs
teiesti: I think, that is an important design decision to make!

I propose that a user can write

- `- [Some Chapter](./chapter.md)` for a link and
- both `- Some virtual chapter` and `- [Some virtual chapter]()` for a virtual link.

Maybe, we could add a lint for `- [Some virtual chapter]()` but I would rather not do that.

src/book/book.rs
teiesti: > Is there any content in a virtual chapter?

This was actually a question somewhere else in a `// TODO`. I will remove the field.

src/book/summary.rs
teiesti: For the moment, I've only implemented parsing `- [Some virtual chapter]()`. `- Some virtual chapter` seems to be more tricky, therefore I would like to have a decision before going on... ;-)

teiesti:
The parser seems to work -- at least for syntax A. Unfortunately, I have no idea where to start with the renderer.  Can someone mentor me through?

src/book/summary.rs
```diff
@@ -373,15 +421,31 @@ impl<'a> SummaryParser<'a> {
...
+                                .number
+                                .as_ref()
+                                .expect("All numbered chapters have numbers");
+
+                            let sub_items = self.parse_nested_numbered(last_item_number)?;
+
+                            last_link.nested_items = sub_items;
+                        },
+                        SummaryItem::Separator => unreachable!(),
+                    };
```
teiesti: I don't really like this match because of the code duplication. Some ideas on how to make it nicer?

src/book/summary.rs
```diff
@@ -401,20 +465,35 @@ impl<'a> SummaryParser<'a> {
...
+                            trace!(
+                                "Found virtual chapter: {} {}",
+                                number,
+                                link.name,
+                            );
+
+                            link.number = Some(number);
+                        },
+                        SummaryItem::Separator => panic!(),
+                    }
```
teiesti: Another match. Same problem.

Michael-F-Bryan:
You mentioned there are a lot of redundant `match` statements where we only care about dealing with a `SummaryItem::VirtualLink` or a `SummaryItem::Link`... What are your thoughts on removing the `SummaryItem::VirtualLink` variant and converting the `path` on a link to `Option<PathBuf>`. As far as I can see, the only difference between a virtual link and a normal one is the presence of a concrete path.

Michael-F-Bryan:
I believe the main thing we need to do on the renderer side is update the `toc` and `next`/`previous` handlebars helpers. When the toc helper encounters a virtual chapter it should just emit a normal `<li>` with no link and just the chapter's name as the tag body (maybe also adding a CSS class?). 

I think the navigation helpers should skip over a virtual chapter when you hit next/previous, although I'd have to have another look at the code to see how you are meant to do that.

teiesti:
> What are your thoughts on removing the SummaryItem::VirtualLink variant and converting the path on a link to Option<PathBuf>. As far as I can see, the only difference between a virtual link and a normal one is the presence of a concrete path.

My thoughts are:

1. If we go for an `Option<PathBuf`>, we can also go for a `PathBuf` allowed to be empty. (This is, as I understand, the old solution.)
2. There is currently a correspondence between `VirtualLink` and `VirtualChapter` and `Link` and `Chapter`. I think, the solution you propose is less intuitive in this regard.

I am really unsure which solution is better. @azerupi What do you think?

azerupi:
Yeah I think we should definitely have a separate item in the enum to represent those empty chapters. I am not up to date with the summary loading code however, so I don't have an opinion there.

src/book/book.rs
```diff
@@ -128,16 +128,24 @@ where
 pub enum BookItem {
     /// A nested chapter.
     Chapter(Chapter),
+    /// A nested virtual chapter.
+    VirtualChapter(VirtualChapter),
     /// A section separator.
     Separator,
```
mattico: While we're making breaking changes, it may be a good idea to add a variant:
```rust
// To make sure clients have a `_ =>` case
#[doc(hidden)]
__NonExhaustive,
```
to ensure client libraries can handle new enum variants.

teiesti: Good idea. I will do that!

Michael-F-Bryan:
Hey @teiesti, were you wanting any help with the rendering side of things?

teiesti:
@Michael-F-Bryan: Yes.

By the way: Sorry for being inactive. I was busy learning for some university exams.

suhr:
Status of this PR?

(If it stays inactive, I'll try to complete it myself. This feature is important)

teiesti:
I am still waiting for help on the rendering side. If you want to give it a try, don't hesitate. Your work is appreciated... :-)

Michael-F-Bryan:
@teiesti, what do you think you'll need help with from the rendering side? I believe it should be enough to update the `make_data()` function in `src/renderer/html_handlebars/hbs_renderer.rs` (creates the global  context for template rendering) and then change the TOC helper so it'll render a virtual chapter correctly.

You'll probably want to write a test or two in either the TOC helper or as part of the integration test suite to make sure virtual chapters get added to the TOC but don't contain links to anywhere.

teiesti:
I was thinking about this pull request again, and it seems that I won't have time for it soon. So, if anyone wants to take over, your welcome. Otherwise, I'll keep this open and try to complete my work when time is less rare. But this won't be in near future.

Sorry...

steveklabnik:
> (If it stays inactive, I'll try to complete it myself. This feature is important)

Agreed, though I'm also busy. @suhr do you want to give it a shot, as per @teiesti's last comment?

@teiesti thanks so much for all your work on this, I really appreciate it

suhr:
Hm. It finds out, I have to deal with that 100 lines long `toc.rs` function. Which should have been fixed by https://github.com/rust-lang-nursery/mdBook/pull/467, but still isn't.

How unfortunate.

azerupi:
I rebased this PR onto the latest master. The renderer actually still had support for "virtual chapters" from before. So I just added virtual chapters as chapters without a path when making the context for handlebars and it just worked as before.

I am sure it can be improved to make it more robust, but I propose that we merge this as is and improve it in further PRs.

azerupi:
I fixed the two remaining bugs, I think it is working correctly now.

image

azerupi:
@Michael-F-Bryan could you review this when you have some time?
It should be mostly done for a first PR, there is maybe some polishing left to do. But I would like to merge this functionality soon, because one of my personal projetcs could use this. :)

azerupi:
Thanks for the review!
I will try address the comments later today.

>have you tried using this branch for your project?

I have, it works great on the books I tried it on :)

Dylan-DPC:
@teiesti do you want to continue with this? else it will be a good idea to close it and open a new PR when you feel like doing it again. Thanks for contributing

teiesti:
@Dylan-DPC I think, it was way overdue to close this PR, cause it was laying around bitrotting. I still want to finish this work but in fact I won't have time anytime soon.

PR #593, opened 2018-01-27
Add docs for mdBook specific include feature

boxdot:
Also:
* Fix bug in `take_lines` taking `end`-many lines instead of `end-start` many.
* Handle special case `include:number` as including a single line.
* Start counting lines at 1 and not 0.


Michael-F-Bryan:
Should we merge this new chapter of the user guide with the `Rust code specific features` page? I feel like they both fall under the same umbrella because they're mdbook-specific helpers for including things in a document.

boxdot:
> Should we merge this new chapter of the user guide with the `Rust code specific features` page?

Good idea. I merged them into a single page.

PR #600, opened 2018-01-29
Inline footnotes

### LINKED ISSUE #210: Footnotes have no styles
A simple

```css
div.footnote-definition p {
       display: inline;
}
```

makes them look kind of okay : 

theme

But I wonder if we can do better.

boxdot:
Footnotes are inlined and made slightly smaller. Also, the first non-consecutive footnote has a bigger margin to the previous element. Fixes #210.

Rendered example:
example

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -47,6 +47,17 @@ table td {
 table thead td {
   font-weight: 700;
 }
+:not(.footnote-definition) + .footnote-definition,
+.footnote-definition + :not(.footnote-definition) {
+  margin-top: 2em;
+}
+.footnote-definition {
+  font-size: 0.9em;
+  margin: 0.5em 0;
+}
+.footnote-definition p {
+  display: inline;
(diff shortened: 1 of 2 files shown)


Michael-F-Bryan:
I think you'll want to make these changes in the `*.styl` files instead of editing `book.css` directly. The main `book.css` is actually generated by processing all the stylus files. Have a look at the [making changes to the style][1] section in `CONTRIBUTING.md`.

[1]: https://github.com/rust-lang-nursery/mdBook/blob/master/CONTRIBUTING.md#making-changes-to-the-style

boxdot:
Ah, sure. I also added a margin to the last footnote (more precise, the element after it; cf. the updated rendered image).

ashutoshmjain:
Hi All - Wanted to bring attention of this team to issue #1927 

tldr - Footnotes style looks good but the backlinks are not working.

PR #604, opened 2018-02-02
Search with Elasticlunr, updated

### LINKED ISSUE #51: Have a search bar

mattico:
@Phaiax's #472, manually rebased, plus a few improvements.

Major Changes from #472:
 - Breadcrumbs are made by adding `Chapter.parent_names`, rather than modifying the `BookItems` iterator
 - Markdown renderer renders footnotes, and also inserts spaces for certain tags to ensure words don't get merged together in the search results
 - `SearchDocument` removed, `render_item` renders directly into `elasticlunr::Index`
 - CSS modifications moved into stylus files
 - Removed JQuery from JS
 - Added integration test
 - Added aria attributes to search elements
 - Added search cargo feature:
   - Search index creation moved into module
   - Search JS moved into theme module for override/disable
 - HTML textification uses ammonia sanitizer

Closes #472 
Fixes #51


Cargo.toml
```diff
@@ -59,12 +62,13 @@ pretty_assertions = "0.4"
...
 
 [features]
-default = ["output", "watch", "serve"]
+default = ["output", "watch", "serve", "search"]
 debug = []
 output = []
 regenerate-css = []
 watch = ["notify", "time", "crossbeam"]
 serve = ["iron", "staticfile", "ws"]
+search = ["elasticlunr-rs"]
```
mattico: Forgot that I added this feature, it doesn't currently do anything. If desired, I can figure out how to extract the search functionality from hbs_renderer so it can be disabled.

mattico:
@Michael-F-Bryan 

If there's anything I can do to make this easier to review, let me know. I'd be happy to break it into smaller commits.

No pressure, though. Wouldn't want a good maintainer to get burned out.

src/config.rs
```diff
@@ -425,29 +425,84 @@ pub struct HtmlConfig {
...
         }
     }
 }
 
+/// Configuration of the search functionality of the HTML renderer.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(default, rename_all = "kebab-case")]
+pub struct Search {
+    /// Enable in browser searching. Default: `true` (if `search` feature is enabled).
+    pub enable: bool,
```
Michael-F-Bryan: Is it necessary to have an `enabled` flag? I would have thought we can see search is enabled because a `output.html.search` is either present or not.

So an alternative to the `enabled` flag would be to change the `search` field in `HtmlConfig` to be `Option<Search>`... Thoughts?

src/config.rs
```diff
@@ -425,29 +425,84 @@ pub struct HtmlConfig {
...
+    /// The path to the searcher to use. If not specified, will use the searcher code included in
+    /// MDBook. Default: `""`
+    pub searcher: PathBuf,
```
Michael-F-Bryan: Hmm, I think we may end up having a similar issue to themes here. By adding this to the config file we're *theoretically* allowing people to swap in their own implementations, but because there is no well defined (and documented) interface to our `searcher`, people probably won't be able to make use of this knob in practice.

What would one gain by being able to use their own `searcher` implementations?

src/config.rs
```diff
@@ -425,29 +425,84 @@ pub struct HtmlConfig {
...
+    /// Documents are split into smaller parts, seperated by headings. This defines, until which
+    /// level of heading documents should be split. Default: `3`. (`### This is a level 3 heading`)
+    pub heading_split_level: u8,
+}
+
+impl Default for Search {
+    fn default() -> Search {
+        // Please update the documentation of `Search` when changing values!
+        Search {
+            enable: cfg!(feature = "search"),
```
Michael-F-Bryan: Nice use of the `cfg!()` macro! :+1:

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -415,6 +413,14 @@ fn make_data(root: &Path, book: &Book, config: &Config, html_config: &HtmlConfig
                     json!("theme-tomorrow_night.js"));
     }
 
+    if html.search.enable {
+        if cfg!(feature = "search") {
+            data.insert("search_enabled".to_owned(), json!(true));
+        } else {
+            warn!("mdBook compiled without search support, ignoring `output.html.search.enable`");
```
Michael-F-Bryan: Good idea with adding a warning! Should we also say something like "please reinstall with `cargo install mdbook --force --features search` to use the search feature"?

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -690,9 +647,9 @@ mod tests {
 
     #[test]
     fn anchor_generation() {
-        assert_eq!(id_from_content("## `--passes`: add more rustdoc passes"),
+        assert_eq!(utils::id_from_content("## `--passes`: add more rustdoc passes"),
```
Michael-F-Bryan: Now that the `id_from_content()` function has been moved to the `utils` module it may make sense to relocate these tests as well. I'm also curious to see how the ID generation will go when your header includes emojis, bold, or bits of code.

src/renderer/html_handlebars/search.rs
```diff
@@ -0,0 +1,221 @@
...
+use book::BookItem;
+use config::Search;
+use errors::*;
+use renderer::RenderContext;
+use utils;
+use theme::searcher::Searcher;
+
+/// Creates all files required for search.
+pub fn create_files(ctx: &RenderContext) -> Result<()> {
+    let search_config = ctx.config.html_config().unwrap_or_default().search;
```
Michael-F-Bryan: Is there any chance we can pass in a `&HtmlConfig` here? It seems such a waste to be re-deserializing it when the caller should already have one lying around.

src/renderer/html_handlebars/search.rs
```diff
@@ -0,0 +1,221 @@
...
+            | Event::End(Tag::Paragraph)
+            | Event::Start(Tag::Rule)
+            | Event::End(Tag::TableHead)
+            | Event::End(Tag::TableCell)
+            | Event::Start(Tag::List(_))
+            | Event::End(Tag::Item)
+            | Event::Start(Tag::Image(..))
+            | Event::End(Tag::Image(..))
+            | Event::SoftBreak
+            | Event::HardBreak => {
```
Michael-F-Bryan: This match statement looks like it was painful and error-prone to write out. Unfortunately I can't think of a better way of expressing this than what you've already got :disappointed:

src/theme/searcher/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+pub static JS: &'static [u8] = include_bytes!("searcher.js");
+pub static MARK_JS: &'static [u8] = include_bytes!("mark.min.js");
+pub static ELASTICLUNR_JS: &'static [u8] = include_bytes!("elasticlunr.min.js");
+
+/// The `Searcher` struct should be used instead of the static variables because
+/// the `new()` method will look if the user has a searcher directory in their source folder and use
+/// the user's searcher instead of the default.
+///
+/// You should only use the static variables if you need the default searcher even if the
+/// user has specified another searcher.
```
Michael-F-Bryan: It seems like a better way of expressing this warning would be to make the statics private and then implement `Default` for `Searcher` so people still have a way of accessing the defaults. 

Also, I don't think the doc-comment mentions what this struct is actually for... Although from context it's obviously a bag of bytes containing the JS assets, with a nice constructor for letting users provide their own versions.

src/utils/fs.rs
```diff
@@ -16,6 +16,27 @@ pub fn file_to_string<P: AsRef<Path>>(path: P) -> Result<String> {
     Ok(content)
 }
 
+/// Changes path to use standard UNIX/URL separators
+pub fn normalize_path(path: &str) -> String {
```
Michael-F-Bryan: This is just a thought, but would it make sense to use a `Path` or `Url` here to make it more obvious what these strings are?

src/utils/fs.rs
```diff
@@ -16,6 +16,27 @@ pub fn file_to_string<P: AsRef<Path>>(path: P) -> Result<String> {
...
+    use std::path::is_separator;
+    path.chars()
+        .map(|ch| if is_separator(ch) { '/' } else { ch })
+        .collect::<String>()
+}
+
+/// Write the given data to a file, creating it first if necessary
+pub fn write_file<P: AsRef<Path>>(
```
Michael-F-Bryan: **Thank you** for removing this from the HTML renderer! It actually started life as a helper method on `MDBook` and got moved out because it coupled the entire HTML renderer to `MDBook`'s internals. At the time I was refactoring a different part of the system, so didn't want to go off on a tangent by also fixing `write_file()`.

src/utils/mod.rs
```diff
@@ -3,13 +3,60 @@
...
 
 use pulldown_cmark::{html, Event, Options, Parser, Tag, OPTION_ENABLE_FOOTNOTES,
                      OPTION_ENABLE_TABLES};
+
 use std::borrow::Cow;
 
 pub use self::string::{RangeArgument, take_lines};
 
+pub fn remove_html_tags<'a>(text: &'a str) -> Cow<'a, str> {
+    let regex = Regex::new(r"<[^>]*?>").unwrap();
```
Michael-F-Bryan: Is it a good idea to use a regex here instead of something like a HTML sanitizer? I'm not 100% familiar with how the `pulldown-cmark` markdown renderer deals with angle brackets, but if we're not careful when calling it, we could accidentally clobber the wrong stuff (e.g. an innocent paragraph like "if \`x < 0\` and \`x > -5\`, blah blah blah").

Of course, that approach could also be overkill and 95% of the time the regex will work just fine...

src/renderer/html_handlebars/search.rs
```diff
@@ -0,0 +1,221 @@
...
+fn make_doc_ref(anchor_base: &str, section_id: Option<String>) -> Cow<str> {
+    if let Some(id) = section_id {
+        format!("{}#{}", anchor_base, id).into()
+    } else {
+        anchor_base.into()
+    }
+}
+
+/// Renders markdown into flat unformatted text and adds it to the search index.
+fn render_item(index: &mut Index, search_config: &Search, item: &BookItem) -> Result<()> {
```
Michael-F-Bryan: Is it worth creating a bunch of tests for this so we can be sure the text gets stripped as intended?

src/renderer/html_handlebars/search.rs
```diff
@@ -0,0 +1,221 @@
...
+        let doc_ref = make_doc_ref(&anchor_base, section_id);
+
+        index.add_doc(&doc_ref, &[&heading, &body, &breadcrumbs.join(" » ")]);
+    }
+
+    Ok(())
+}
+
+/// Converts the index and search options to a JSON string
+fn write_to_json(index: Index, search_config: &Search) -> Result<String> {
```
Michael-F-Bryan: Do you have access to any "known good" versions of the generated JSON? It's probably a good idea to add a test or two to make sure we're able to generate valid JSON, that way we can also prevent accidental regressions in the future if someone needs to come back and change things.

src/theme/searcher/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+pub struct Searcher {
+    pub js: Vec<u8>,
+    pub mark_js: Vec<u8>,
+    pub elasticlunr_js: Vec<u8>,
+}
+
+impl Searcher {
+    pub fn new(src: &Path) -> Self {
```
Michael-F-Bryan: Should we create a test for this to make sure the file loading logic is sound and people are able to override the scripts?

src/config.rs
mattico: Agreed. Especially now that the JS is conditionally included, the flag is not very useful.

src/config.rs
mattico: I agree that the way it's setup is not ideal, I just copied the setup used for playpen_editor.

While we give no guarantees about the stability or interfaces for the searcher, the current system can be useful in practice for making small changes to search behavior (remove highlighting, change keyboard shortcuts, etc.). I'll agree, however, that this feature is unlikely to be used much if at all. In general I don't think many users will make significant theme changes out-of-tree until we provide a better way to do that.

So what change can we make to this PR? Here's what I'm thinking:

1. Remove the `editor` and `searcher` settings and the corresponding file override behavior.
2. Add `copy-js` settings for both. If set to false, the JS code for the feature will not be copied to the output directory. This is more flexible since users can then include whatever JS they want with `additional-js`, and it reduces the complexity of this little-used feature to almost zero.

src/config.rs
mattico: Hm, that would change search to be disabled by default. Is that your intention?

src/renderer/html_handlebars/search.rs
mattico: It was trial-and-error, yes, but the good thing is that if it's wrong it doesn't matter that much. Now that I think about it the match could be replaced with `Event::Start(_) | Event::End(_) | Event::SoftBreak | Event::HardBreak` and the only difference would be a few more spurious spaces in the search result teasers. Hmm, and we could collapse multiple spaces down to one.

src/utils/fs.rs
mattico: The function operates on a string but `Path` needs to be converted into a string, which can throw an error. Both places this function is used already converted the path to a string. I don't think it's worth the trouble.

src/utils/mod.rs
mattico: Using a Regex on HTML does feel bad, but I didn't want to pull in a whole sanitizer. Angle brackets already have to be escaped in HTML blocks so it could clobber some data, but I blame the user. Worst-case scenario is that some pathological text inside of an HTML tag isn't in the search index.

src/renderer/html_handlebars/search.rs
mattico: It's difficult to have any significant test for the entire search index. Tiny changes in the source material change the index drastically, and it's so large that it's difficult to determine if any change matters at all. If you have suggestions, I'm all ears.

src/config.rs
Michael-F-Bryan: I think that sounds quite reasonable :+1:

Besides, it's not hard to put back in if we find people actually want the feature.

src/config.rs
Michael-F-Bryan: I dunno, to be honest it's one of those features you'll always want for your document. 

It should probably be always turned on unless you manually deactivate the `search` feature flag when installing `mdbook` (e.g. when packaging for Debian).

src/renderer/html_handlebars/search.rs
Michael-F-Bryan: Hmm, you make a good point... What I usually do is generate a small example and store that as a file, giving it a quick look through to make sure it seems alright. Then every time after that you can have a test which uses the same input and check the rendered document is the same as our known good copy.

I have a similar issue at work where it's not really possible to manually construct an object and check each of its elements are the same every time. For that I've got a `create_fixture()` function which looks somewhat like this:

```rust
/// Regenerates the test fixture (pre-nested list of cutlines).
///
/// # Note
///
/// You'll need to run `cargo test` manually here, if tests are being run via
/// `cargo watch` and you try to regenerate the test fixture we'll get midway
/// through writing out the fixture before `cargo watch` restarts the process.
/// Leaving a mangled fixture file.
const GENERATE_FIXTURE: bool = false;


// loads of tests


fn get_fixture() -> Vec<Cutline> {
    if cfg!(windows) && GENERATE_FIXTURE {
        let inputs = create_known_dummy_inputs();
        let thing = generate_output(&inputs).unwrap();

        let fixture_path =
            Path::new(env!("CARGO_MANIFEST_DIR")).join("tests/data/fixture.json");
        let mut f = ::std::fs::File::create(&fixture_path).unwrap();
        serde_json::to_writer_pretty(&mut f, &thing).unwrap();

        nested
    } else {
        let jason = include_str!("data/fixture.json");
        serde_json::from_str(jason).expect("Unable to deserialize the fixture")
    }
}
```

mattico: Ah that's a good solution! One of the issues I was running into is that rendering dummy_book is different when you're in a test (due to env vars?) so it's annoying to make fixtures.

mattico:
I think that addresses everything, except for the JS dependency question. I would like to address that, I'm working out a solution right now. I think we should leave that for another PR, though.

Phaiax:
I just wanted to say thank you for taking this over. :+1:

(I wished I had enough time for this but nope, got a new job and more...)

src/book/book.rs
```diff
@@ -152,15 +152,20 @@ pub struct Chapter {
     pub sub_items: Vec<BookItem>,
     /// The chapter's location, relative to the `SUMMARY.md` file.
     pub path: PathBuf,
+    /// An ordered list of the names of each chapter above this one, in the hierarchy.
+    pub parent_names: Vec<String>,
```
Michael-F-Bryan: Seeing that we need to manually store the parent names and keep them in sync makes me question whether it may be better to represent a `Book` as containing a "doubly-linked" tree of chapters, or even turn it into a graph structure. That way a `Chapter` can contain a pointer back to its parent and you can use that to walk up and down the levels of the tree. 

It would also hopefully simplify tracking section numbers and parsing. I might play around with this idea on a private branch and see how it goes.

src/config.rs
```diff
@@ -425,29 +425,79 @@ pub struct HtmlConfig {
     pub livereload_url: Option<String>,
     /// Should section labels be rendered?
     pub no_section_label: bool,
+    /// Search settings. If `None`, the default will be used.
```
Michael-F-Bryan: > If `None`, the default will be used.

I believe this is already sorted out by the `#[serde(default)]` annotation on `HtmlConfig`. That annotation tells serde to use `Search::default()` if it can't deserialize the settings.

To me, when I see `Option<Search>` I'd think this indicates that searching is optional and when no `Search` is found in my `book.toml` this means my book won't have a search bar... Am I interpreting it wrong, or is this comment out of sync with the implementation?

src/config.rs
```diff
@@ -425,29 +425,79 @@ pub struct HtmlConfig {
...
-    /// The path to the editor to use. Defaults to the [Ace Editor].
-    /// 
-    /// [Ace Editor]: https://ace.c9.io/
-    pub editor: PathBuf,
-    /// Should playpen snippets be editable? Defaults to `false`.
+    /// Should playpen snippets be editable? Default: `false`.
     pub editable: bool,
+    /// Copy JavaScript files for the editor to the output directory?
+    /// Default: `true`.
+    pub copy_js: bool,
```
Michael-F-Bryan: This probably belongs in its own PR. Changing how the playpen works and whether we copy across JavaScript files is a non-trivial change.

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -307,15 +296,20 @@ impl Renderer for HtmlHandlebars {
         fs::create_dir_all(&destination)
             .chain_err(|| "Unexpected error when constructing destination path")?;
 
-        for (i, item) in book.iter().enumerate() {
+        let mut depthfirstiterator = book.iter();
+        let mut is_index = true;
+        while let Some(item) = depthfirstiterator.next() {
```
Michael-F-Bryan: Shouldn't this just be a plain old `for` loop?

tests/rendered_output.rs
```diff
@@ -339,3 +339,101 @@ fn book_with_a_reserved_filename_does_not_build() {
...
+
+    // So you've broken the test. If you changed dummy_book, it's probably
+    // safe to regenerate the fixture. If you haven't then make sure that the
+    // search index still works. Run `cargo run -- serve tests/dummy_book`
+    // and try some searches. Are you getting results? Do the teasers look OK?
+    // Are there new errors in the JS console?
+    //
+    // If you're pretty sure you haven't broken anything, change `GENERATE_FIXTURE`
+    // above to `true`, and run `cargo test` to generate a new fixture. Then
+    // change it back to `false`.
```
Michael-F-Bryan: I got a nice laugh out of this block of comments, yet it's still surprisingly useful for people who see it because of a failing test. Thank you!

Michael-F-Bryan:
I just tried this out on my laptop. One thing I've noticed is the search index is only available when running mdbook with `mdbook serve` which is a bit of a pain when testing with `mdbook build --open` locally. 

The browser console shows the following error.

```
searcher.js:260 Failed to load file:///home/michael/Documents/mdBook/book-example/book/searchindex.json: Cross origin requests are only supported for protocol schemes: http, data, chrome, chrome-extension, https.
init @ searcher.js:260
```

After a brief search it seems like this is a [Chrome-specific issue][1] and there isn't really any way around it other than invoking chrome from the command-line with the `--allow-file-access-from-files` flag. @sorin-davidoi, have you encountered this before?

[1]: http://ahmed.amayem.com/solution-to-chrome-cross-origin-requests-are-only-supported-for-protocol-schemes-http-data-chrome-chrome-extension-https-chrome-extension-resource/

src/renderer/html_handlebars/hbs_renderer.rs
mattico: Yeah, I'm not sure why this was changed. This is from the original PR.

sorin-davidoi:
Don't think the issue is Chrome-specific - you just can't read arbitrary files from the filesystem. A (nasty) workaround could be to make a `searchindex.js` file which contains the entire JSON and assigns it to `window` - you can then add the file as a script tag, wait for it to load and execute, and proceed with reading the JSON from `window`.

src/config.rs
mattico: > I believe this is already sorted out by the #[serde(default)] annotation on HtmlConfig. That annotation tells serde to use Search::default() if it can't deserialize the settings.

The default it'll use is `None`, since that's how `Default::default::<Option>()` is implemented: https://doc.rust-lang.org/src/core/option.rs.html#900-906, and we're using `#[derive(Default)]`. The html renderer will then manually use `unwrap_or_default` to get the default configuration. The reason to do this is so we can print the warning message if the user is using mdbook compiled without `feature = "search"`, but does have an `[output.html.search]` table in their `book.toml`.

src/config.rs
mattico: I suppose. Should I revert the changes to `editor`, but leave the `output.html.search.copy-js` setting?

mattico:
I looked into HTML sanitization. `ammonia` isn't that big of a dependency, and it could work with [a small change](https://github.com/notriddle/ammonia/issues/89).

src/config.rs
mattico: It may be less confusing overall to just bring back the `[output.html.search.enabled]` field.

mattico:
Probably wait for https://github.com/notriddle/ammonia/pull/90 now.

src/config.rs
Michael-F-Bryan: Mmm, don't worry about it for now, it's probably not worth the effort. As long as it's noted in the PR description it should be fine.

Michael-F-Bryan:
@mattico, I just looked through the PR again and I'm pretty happy with it :+1:

On a side note, it looks like `html5ever` emits a bunch of `unimplemented` warnings when generating the search index, do you know what may be causing them? The `stop_parsing` operation sounds like more of an optimisation to stop parsing HTML early so I doubt it's actually going to have any negative impact on `mdbook`, but the user experience isn't the best. 

We can always hack around this by updating the `init_logging()` function in `src/bin/mdbook.rs` to use `builder.filter(Some("mdbook"), LevelFilter::Info)` when the `RUST_LOG` environment variable isn't set, although I'd be interested to see if there's a way to fix the underlying `html5ever` issue.

For posterity's sake, this is what I see when running `mdbook build` locally:

```text
$ cargo run -- build --open book-example                                     
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/mdbook build --open book-example`
2018-03-05 16:15:36 [INFO] (mdbook::book): Book building has started
2018-03-05 16:15:36 [INFO] (mdbook::book): Running the html backend
2018-03-05 16:15:37 [WARN] (html5ever::tree_builder): stop_parsing not implemented, full speed ahead!
2018-03-05 16:15:37 [WARN] (html5ever::tree_builder): stop_parsing not implemented, full speed ahead!
2018-03-05 16:15:37 [WARN] (html5ever::tree_builder): stop_parsing not implemented, full speed ahead!
2018-03-05 16:15:37 [WARN] (html5ever::tree_builder): stop_parsing not implemented, full speed ahead!
2018-03-05 16:15:37 [WARN] (html5ever::tree_builder): stop_parsing not implemented, full speed ahead!
2018-03-05 16:15:37 [WARN] (html5ever::tree_builder): stop_parsing not implemented, full speed ahead!
2018-03-05 16:15:37 [WARN] (html5ever::tree_builder): stop_parsing not implemented, full speed ahead!
2018-03-05 16:15:37 [WARN] (html5ever::tree_builder): stop_parsing not implemented, full speed ahead!
Created new window in existing browser session.
```

Other than the annoying `html5ever` warnings, I think this is about ready to merge.

mattico:
Ah, should've run more than cargo test. I'll look into it.

mattico:
Fixed.

Michael-F-Bryan:
And merged :tada:

mattico:
4 months ago I looked at the search issue and thought "how hard can it be"?. 🎉🎊🎖️📯

Michael-F-Bryan:
I've been using `master` locally for a couple days now and can't see any issues with search, so I'll probably to do another release either tomorrow or the day after.

Should I also make a post to [/r/rust]? Is there anything in particular you'd like me to say about `mdbook`'s new search functionality?

[/r/rust]: https://www.reddit.com/r/rust/

mattico:
> Should I also make a post to /r/rust?

That's a great idea! People should know to rebuild their docs.

> Is there anything in particular you'd like me to say about mdbook's new search functionality?

Hmm. I'd mention that if anyone notices something which won't show up in search results that they should file an issue on here. There's possibly still some bugs in the index preprocessing or elasticlunr-rs.

If you're going to have a "thanks to" section make sure to mention phaiax for writing most of this ^^ and notriddle for maintaining ammonia.

There are a lot of questions on /r/rust like "what's a project I can contribute to to learn Rust". I think _mdbook_ works well for that, the code is easy to understand and its fun to see the results in the browser. On the other hand sometimes working on it is more HTML than Rust, and a lot of the stuff that needs to be worked on is more architecture than self-contained improvements.

Michael-F-Bryan:
@mattico I think he have a [small CSS issue](https://rust-lang-nursery.github.io/mdBook/?search=command) with the search results 😜 
 
image

Do you know what may have caused this? I'm using Firefox on Windows, but I remember testing the search feature before our `0.1.4` release (Chrome on Arch Linux) and the search results would get inserted above the page contents just fine.

@sorin-davidoi, this should be a pretty trivial CSS fix, shouldn't it?

PR #608, opened 2018-02-05
WIP: add support for j/k - vi-style navigation

bmusin:
I want to add suport for h/l (vi-like), F9 (or something else)
to open/close left navigation panel.

What tags/ids of DOM elments should I manipulate to open/close side panel?

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -455,6 +455,26 @@ function playpen_text(playpen) {
         }
     }, { passive: true });
 
+    document.addEventListener('keydown', function (e) {
+        if (e.altKey || e.ctrlKey || e.metaKey || e.shiftKey) { return; }
+
+        switch (e.key) {
+            case 'F9':
+                if (html.classList.contains("sidebar-hidden")) {
+                    showSidebar();
+                } else if (html.classList.contains("sidebar-visible")) {
+                    hideSidebar();
+                } else {
+                    if (getComputedStyle(sidebar)['transform'] === 'none') {
+                        hideSidebar();
+                    } else {
+                        showSidebar();
+                    }
+                }
+                break;
+        }
+    });
+
     // Scroll sidebar to current active section
     var activeSection = sidebar.querySelector(".active");
     if (activeSection) {
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -415,8 +415,7 @@ function playpen_text(playpen) {
         try { localStorage.setItem('mdbook-sidebar', 'hidden'); } catch (e) { }
     }
 
-    // Toggle sidebar
-    sidebarToggleButton.addEventListener('click', function sidebarToggle() {
+    function toggleSidebar() {
         if (html.classList.contains("sidebar-hidden")) {
             showSidebar();
         } else if (html.classList.contains("sidebar-visible")) {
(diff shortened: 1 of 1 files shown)


sorin-davidoi:
> What tags/ids of DOM elments should I manipulate to open/close side panel?

You should not manipulate the DOM directly. I would add another `keydow` event listener in [the sidebar](https://github.com/rust-lang-nursery/mdBook/blob/master/src/theme/book.js#L389) component for this functionality and use the `showSidebar` and `hideSidebar` functions instead.

sorin-davidoi:
For the `j/k` shortcuts, you should look into how to define multiple keys in the `aria-keyshortcuts` attribute and update it in `index.hbs` (it appears in four places).

bmusin:
@sorin-davidoi, `sidebar()` doesn't add callbacks for handling key events.
Only `chapterNavigation()`.

sorin-davidoi:
If it toggles the sidebar, I think it belongs to the sidebar component. There is no harm in adding a second event listener there.

bmusin:
What about j/k and potentially h/l? Can I handle them in chapterNavigation?

sorin-davidoi:
Yeah, those should be fine there 👍

bmusin:
@sorin-davidoi, how should I implement going between chapters, e.g. go from 3.3 to 2.0 or 4.0?

src/theme/book.js
```diff
@@ -455,6 +455,26 @@ function playpen_text(playpen) {
         }
     }, { passive: true });
 
+    document.addEventListener('keydown', function (e) {
+        if (e.altKey || e.ctrlKey || e.metaKey || e.shiftKey) { return; }
+
+        switch (e.key) {
+            case 'F9':
+                if (html.classList.contains("sidebar-hidden")) {
```
sorin-davidoi: Could you extract this into a separate function (since it is the same logic as in the `click` handler)? We could then just call it from the switch and the `click` handler would be

```js
sidebarToggleButton.addEventListener('click', toggleSidebar);
```

bmusin: Did.

bmusin:
@sorin-davidoi, how should I implement going between chapters, e.g. go from 3.3 to 2.0 or 4.0?
Should I manually parse href?

pickfire:
Maybe we should focus to `<main>` by default (maybe with `tabindex="0"`?) so that users can navigate up and down after navigating to another page with <kbd>←</kbd> and <kbd>→</kbd>.

But it's a nice idea to have vi keys navigation. :+1:

sassman:
The vi navigation bindings for left / right (j/k) are super handy. What about focusing here only on that, since the code is already there are any objections of merging this? 

IMO the chapter navigation could be handled as a follow-up PR.

Any opinions?

src/theme/book.js
```diff
@@ -468,18 +481,28 @@ function playpen_text(playpen) {
...
                 e.preventDefault();
-                var previousButton = document.querySelector('.nav-chapters.previous');
-                if (previousButton) {
-                    window.location.href = previousButton.href;
-                }
+                // TODO: add code
+                break;
+            case 'l':
+                e.preventDefault();
+                // TODO: add code
```
pickfire, line 505: TODO?

sassman: seems that the author abandoned this PR, I created #1408 to follow up on this.

GuillaumeGomez:
We'd prefer avoid adding more shortcuts as it might conflict with the browser's. More detail here: https://github.com/rust-lang/mdBook/issues/2404.

Thanks for working on this in any case!

PR #611, opened 2018-02-06
Accessibility improvements

sorin-davidoi:
Follows the [menu ARIA guideline](https://www.w3.org/TR/wai-aria-practices-1.1/#menu).


Michael-F-Bryan:
Looking through the aria guidelines this seems to be quite reasonable and should make keyboard navigation a lot nicer :+1:

PR #613, opened 2018-02-07
Scroll to top when clicking the page title

sorin-davidoi:
Common pattern, especially on mobile devices where the page can be quite long.


Michael-F-Bryan:
Seems like a good idea to me :+1:

PR #614, opened 2018-02-07
Handle some cases when JavaScript is disabled

sorin-davidoi:


Michael-F-Bryan:
Does disabling javascript affect anything other than the sidebar? I would have thought themes would be broken because they get loaded dynamically.

sorin-davidoi:
You cannot change the default theme and syntax highlighting does not work. But currently the layout is broken (as shown below) - this fixes that.

image

Michael-F-Bryan:
Oh gawd, that looks awful. Thanks for the fix!

Issue #617, opened 2018-02-11
`mdbook` without arguments should show something useful


Havvy:
Here's what it currently shows:

```
$ mdbook
error: 'mdbook' requires a subcommand, but one was not provided

USAGE:
    mdbook <SUBCOMMAND>

For more information try --help
```

It should either just default to showing the help message after the error or list what the actual subcommands are.

Michael-F-Bryan:
I believe this is the default behaviour for `clap`. There is a [setting] to display the help text if no arguments (or subcommands) are provided though. It should be trivial to update our `msbook` binary to enable that setting.

[setting]: https://docs.rs/clap/2.29.4/clap/enum.AppSettings.html#variant.ArgRequiredElseHelp

Michael-F-Bryan:
Fixed in #624.

PR #624, opened 2018-02-17
Changing clap settings

maccoda:
Really simple fix for #617 as was described in the issue. A few formatting changes sorry, let me know if rustfmt wasn't meant to do this...


Michael-F-Bryan:
This seems to be exactly what we needed to fix #617. Thanks a lot @maccoda!

Issue #331, opened 2017-06-14
`mdbook init` should create a stub book.toml based on args


budziq:
This is connected with discussion in https://github.com/azerupi/mdBook/issues/313

I currently find 3 problems in this area:

- [ ] I guess that book `init` should create book.toml filled with defaults (possibly from gitconfiig) and updated with info from flags such as `--theme`
- [ ] Currently options such as `mdbook init --theme` do nothing if book.toml is not present.
Namely the book `init` tries to read the book.toml and does not find it, then `copy_theme` starts and has nothing to do as the `themedir` is not set. This will be confusing to the user.
- [ ] `--theme` option always triggers confirmation prompt (unless `--force`)? There seams to be no reason to prompt if the target `theme` directory does not exists yet.

maccoda:
Hi,
I was wanting to have a go at this one. A few quick questions before I go down the wrong path.
1. > create book.toml filled with defaults

    In regards to this should it be the entire set of defaults or only a subset? The main reason I am asking is that elements like the title, author, and description are all optional, hence writing these in there will just result in empty elements. Of course following your point of looking at the gitconfig we should be able to at least populate the author but for the other two I am not too sure how we should handle that.

1. In regards to using the gitconfig, what would be your preferred method? I looked at how cargo handles it and it uses the `git2` crate I believe. Did we want to follow this example or is that a bit of overhead considering we will be wanting one line perhaps? Also whilst looking at cargo I saw that they will support several other vcs (most of which I am not too familiar with...) did we want to support those if gitconfig was not found?

I will make a start on this just using file reading for the gitconfig and publishing the full set of defaults for the configuration for now but let me know if you would like to change it up 😄

PR #627, opened 2018-02-18
Adding for content to book.toml on init

maccoda:
Hopefully this gets addresses some of the points requested in #331. I haven't used this with custom themes myself so wasn't entirely certain this was the config behaviour desired. Let me know, more than happy to change it! 👍


src/bin/init.rs
```diff
@@ -20,9 +23,11 @@ pub fn make_subcommand<'a, 'b>() -> App<'a, 'b> {
 pub fn execute(args: &ArgMatches) -> Result<()> {
     let book_dir = get_book_dir(args);
     let mut builder = MDBook::init(&book_dir);
+    let mut config = config::Config::default();
 
     // If flag `--theme` is present, copy theme to src
     if args.is_present("theme") {
+        config.set("output.html.theme", "src/theme")?;
```
Michael-F-Bryan: I think this actually conflicts with the HTML config's `theme_dir()` function. By default that will take the book root and append `theme` to it (i.e. you get `/path/to/book/theme`) whereas this will put the themes directory in `src/` (i.e. you get `/path/to/book/src/theme`)... Was that intentional?

I know the user guide says the `theme` directory should be at the book root:

src/bin/init.rs
```diff
@@ -47,12 +54,37 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
         builder.create_gitignore(true);
     }
 
+    if let Some(author) = get_author_name() {
+        debug!("Obtained user name from gitconfig: {:?}", author);
+        config.book.authors.push(author);
```
Michael-F-Bryan: While we're adding the author's name, should we think about giving the book a default title (or even prompting for one) for the user?

src/bin/init.rs
maccoda: So the reason I chose this one was because I ran `cargo run -- init --theme test_book` to determine where the theme was initially copied to when this argument was given and I found that it was copied under `mdBook/test_book/src/theme`. Unfortunately I haven't fully wrapped my head around this so I am not sure if I have used this correctly... Is that where it should be copied to initially? More than happy to change it to "theme" if that is the correct manner. 👍

src/bin/init.rs
maccoda: Can definitely add this one in there! 😄

Michael-F-Bryan:
LGTM :+1:

I don't think the theme directory's location is a massive issue. As long as it's set in the config and that config ends up in `book.toml`, it shouldn't matter.

PR #636, opened 2018-03-05
Switch from Stylus to CSS

### LINKED ISSUE #616: Discussion: Should we migrate from Stylus, and to what?
I've seen this brought up a number of times, most recently in #610. #351 mentions it in passing. I want to start this discussion, and hopefully build a consensus so that I can work on the migration soon.

As far as I can tell, this is what we use stylus for:
- Variables like sidebar-width and content-max-width
- Variables and unquote for themes
- Convenience features like nested selectors and arithmetic without calc()
- Automatic vendor prefixing with Nib

## Option 1: SASS

Unlike Stylus, there is a C library for SASS, and corresponding Rust bindings. This would be a good path to eliminate our node dependency. SASS has many of the same features as Stylus, so it should be a relatively easy transition.

## Option 2: CSS with Variables

CSS has native support for variables now. Support for it [is very good](https://caniuse.com/#feat=css-variables), with the notable exceptions of IE, old Android Browser, and Samsung Internet 4. While this surpasses the requirements listed in the [Rust browser support RFC](https://github.com/rust-lang/rfcs/blob/master/text/1985-tiered-browser-support.md), I think its worthwhile to put a bit of effort into supporting IE11 at least, with a conditional comment solution.

This solution is missing some of the convenience features offered by the preprocessors. Personally, I don't find those features particularly useful. I find them a bit frustrating, in fact, if I have to look up what they mean.

The next thing to consider is vendor prefixes. Browsers are [moving away from](https://webkit.org/blog/6131/updating-our-prefixing-policy/) using these, but they can still matter for the time being. I scanned through the prefixed properties in book.css to determine how useful autoprefixing tools are for us. I found two cases where prefixes are necessary if we limit the scope to browsers that already support CSS variables:

1. Safari requires `position: sticky` to be prefixed. It's only used for `#menu-bar`.
2. Certain proprietary features like `-webkit-scrollbar-thumb` don't exist without their prefix (and don't benefit from autoprefixing tools)

Based on this, I don't think that autoprefixing tools are necessary if we adopt standard CSS.

-------
While migrating to CSS would be more difficult, it's the better option in my opinion. It removes a build step and a dependency. It also removes the confusion for [users](https://github.com/rust-lang-nursery/mdBook/issues/605) and contributors who don't notice that book.css isn't the real source for styling. It also makes it easy for users to write themes in CSS while still taking advantage of our defined variables.

mattico:
This is a big change in direction, so it deserves some consensus. I'm pretty happy with the end result, though, so I decided to just submit a PR.

Moving from Stylus has been discussed before, but to summarize:
- Removes a build step and several dependenices.
- Removes the confusion for users and contributors who don't notice that `book.css` isn't the real source for styling, or who don't understand Stylus syntax.
-  Makes it easy for users to write themes in CSS.

That last point is really cool because by just using CSS+variables, it's really easy to add custom theme support. The only change we'd need to make is add a config option for themes (just to make it easy to add the theme to the theme menu), and then include their CSS file which defines the same variables that the other themes do.

Here's what this PR does:
1. Convert Stylus to CSS (manually).
2. Remove Stylus supporting code and documentation.
3. Merge the many CSS files into 4 files.
   Having lots of small files is annoying to deal with in our current theme implementation (fixing *that* is
   another discussion), and it's bad for page load times. While we could use a build step to concatenate
   it's much simpler to *not*, and we don't actually have enough CSS to need many files. We could easily
   just have one stylesheet, but I liked having a little bit of organization.

Closes #616.


Michael-F-Bryan:
> The only change we'd need to make is add a config option for themes (just to make it easy to add the theme to the theme menu), and then include their CSS file which defines the same variables that the other themes do.

I doubt this will be very hard. We need to figure out what themes are available anyway (probably via some `additional-themes` key in `book.toml`) in order to copy them to the output directory, so we can easily generate the themes selector from that.

mattico:
Looks like I pushed e300d48 to the wrong branch... but it's fine being here. There aren't any changes since elasticlunr 0.2.2, I'm just following the semver guide: 
>If your software is being used in production, it should probably already be 1.0.0.

mattico:
I looked into implementing theme support with this. It gets a bit more complicated since themes will want to be able to set the theme used by `highlight.js` and `ace` and they'll probably want to be able to add their own highlight/ace themes. That is a solvable problem but it starts to get really hacky. I think we should postpone that idea until we figure out what to do about themes in general.

mattico:
I should cc @sorin-davidoi @azerupi @budziq for your opinions.

Michael-F-Bryan:
> I think we should postpone that idea until we figure out what to do about themes in general.

I agree. There are definitely ways we can hack in setting the theme used by `highlight.js` and the `ace` editor but they'll most probably be just that... hacks. 

Either way, that probably comes under the larger umbrella of defining some sort of public interface so people can tweak themes. It's kinda related, but I'd say that issue is orthogonal to whether we use CSS or Stylus or whatever.

PR #638, opened 2018-03-06
Improve css so anchor don't go under sidebar

GuillaumeGomez:
When clicking on anchors, they don't go under the top navbar now.

### DIFF
--- a/src/theme/stylus/page.styl
+++ b/src/theme/stylus/page.styl
@@ -2,20 +2,18 @@
 
 .page-wrapper {
     box-sizing: border-box
+    left: 0
+    position: absolute
+    right: 0
+    top: 0
+    bottom: 0
 
     // Animation: slide away
-    transition: padding-left 0.5s, margin-left 0.5s
+    transition: padding-left 0.5s, margin-left 0.5s, left 0.5s
 }
 
 .sidebar-visible .page-wrapper {
-    padding-left: $sidebar-width
-}
-
-@media only screen and (max-width: $page-plus-sidebar-width - 1) {
-    .sidebar-visible .page-wrapper {
-        padding-left: 0
-        margin-left: $sidebar-width
-    }
+    left: $sidebar-width
 }
 
 .page {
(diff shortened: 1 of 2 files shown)


sorin-davidoi:
Hi, thanks for opening this! While this certainly fixed the issue, it slightly breaks other functionality:
 - The header no longer hides when scrolling down
 - The sidebar no longer animates properly

The changes also must go into the Stylus files, otherwise the changes will be overwritten during the build step (see https://github.com/rust-lang-nursery/mdBook/blob/master/CONTRIBUTING.md#making-changes-to-the-style for a guide).

GuillaumeGomez:
The sidebar never hid when scrolling down for me... I'll fix the other issues.

sorin-davidoi:
I think it only hides when the sidebar is hidden.

GuillaumeGomez:
Ah that's why! Then it's an easy fix.

src/theme/stylus/page.styl
```diff
@@ -24,15 +22,30 @@
...
         text-decoration: none;
         &:hover { text-decoration: underline; }
     }
 
     img { max-width: 100%; }
 }
+
+.sidebar-visible .content {
+    position: absolute
+    top: 52px
```
sorin-davidoi: Not sure how well this will scale - the header may have a different size depending on font settings (it's `51px` on my machine).

src/theme/stylus/page.styl
```diff
@@ -2,20 +2,18 @@
 
 .page-wrapper {
     box-sizing: border-box
+    left: 0
+    position: absolute
```
sorin-davidoi: Don't really dig having `position: absolute`, but can't think of another way to do it either.

src/theme/stylus/page.styl
GuillaumeGomez: It's 52 on mine haha.

GuillaumeGomez:
Ok, should be fine.

src/theme/stylus/page.styl
Michael-F-Bryan: Can't you define it as something like `3 em` to make sure the size scales appropriately? I imagine hard-coding the size is probably going to make a difference when people view things on mobile.

PR #642, opened 2018-03-14
Exposed the sections inside a book

Michael-F-Bryan:
This exposes a `Book`'s `sections` field so other people can traverse a book recursively without the ugly contortions we had in #629 (no offense intended @Byron!).


Byron:
Haha 😂, none taken, and indeed a very good description of the workaround!

PR #645, opened 2018-03-15
Fix search

### LINKED ISSUE #643: Search results drawn on top of the page content
I think a recent update to CSS (presumably #638) changed how the search results are placed. If you do a search on the [User Guide] you'll see the search results being rendered directly on top of the page contents. 

I'm assuming this issue crept in because I'd looked at #638 locally in the past, but then merged #604 without checking to make sure #638 was still okay.
 
image

I'm not a CSS expert, but I *believe* the issue is because the search results (`id=searchresults-outer`) is `display: block`, while the page contents (`id=content`) is positioned using absolute coordinates. If you close the sidebar, the page gets rendered just fine again.

CC: @GuillaumeGomez, @mattico 

---

We're lucky I noticed this before making a big announcement on Reddit. Looking at the crates.io stats, it seems like only 65 people have downloaded `v0.1.4` so far, two of which are from me.

[User Guide]: https://rust-lang-nursery.github.io/mdBook/?search=command

GuillaumeGomez:
Fixes #643.

### DIFF
--- a/src/theme/searcher/searcher.js
+++ b/src/theme/searcher/searcher.js
@@ -351,8 +351,10 @@ window.search = window.search || {};
     function showSearch(yes) {
         if (yes) {
             searchbar_outer.style.display = 'block';
+            content.style.display = 'none';
             searchicon.setAttribute('aria-expanded', 'true');
         } else {
+            content.style.display = 'block';
             searchbar_outer.style.display = 'none';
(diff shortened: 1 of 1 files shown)


Michael-F-Bryan:
I just checked this out locally and the search seems to be rendered properly :+1:

Thanks a lot for coming up with a fix so quickly @GuillaumeGomez!

PR #646, opened 2018-03-19
Update elasticlunr-rs

mattico:
... and Iron.

The one notable change is that el-rs 2.0 includes search configuration structs, so we don't need to define them in mdBook. It also adds support for different languages, but I'll leave implementing that for another time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -54,7 +54,7 @@ name = "backtrace-sys"
 version = "0.1.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "cc 1.0.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cc 1.0.8 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


Michael-F-Bryan:
I was originally holding off upgrading `iron` because `staticfile` used a different version of `iron`, meaning you get annoying error messages saying the static file `Handler` doesn't implement iron's `Handler` trait.

I'll check things out locally, but if it compiles and serves properly this should be good to merge :+1:

Michael-F-Bryan:
LGTM :slightly_smiling_face:

greysteil:
Would you be up for automatically generating dependency update PRs like this one (well, without all the work to handle breaking changes)? I built [Dependabot](https://dependabot.com) to try and help people keep their dependencies up-to-date, and just added Rust support. When I ran it on my fork it generated [this PRs](https://github.com/greysteil/mdBook/pulls).

It's totally free for open source, and always will be - I'm just looking for beta testers and advice. :octocat:

PR #648, opened 2018-03-19
Fix several theme issues

### LINKED ISSUE #647: Page content overlaps search bar after search term is erased
If you search for something then erase the search term the content reappears and overlaps the search bar.

image

mattico:
Fixes #647

See the commit messages.

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -81,6 +81,8 @@ table thead td {
   box-sizing: border-box;
   -webkit-overflow-scrolling: touch;
   overscroll-behavior-y: contain;
+}
+.js .sidebar {
   -webkit-transition: -webkit-transform 0.3s;
   -moz-transition: -moz-transform 0.3s;
   -o-transition: -o-transform 0.3s;
@@ -140,6 +142,8 @@ table thead td {
   -webkit-box-sizing: border-box;
   -moz-box-sizing: border-box;
   box-sizing: border-box;
+}
+.js .page-wrapper {
   -webkit-transition: margin-left 0.3s ease, -webkit-transform 0.3s ease;
   -moz-transition: margin-left 0.3s ease, -moz-transform 0.3s ease;
   -o-transition: margin-left 0.3s ease, -o-transform 0.3s ease;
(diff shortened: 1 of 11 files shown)


Michael-F-Bryan:
Hmm, my concern here is it could be more of a band-aid solution than a real fix. I would have thought it'd be easy enough to use CSS to make sure the search bar doesn't overlap the content.

The root problem seems to be that we're using absolute positioning for the page content (so it'll leave space below the header bar) and that means other things can't flow/reposition themselves around it properly. If that's the case, the proper fix would probably be to remove the absolute positioning and `display: block` and position the page content differently... Would that work?

@sorin-davidoi, do you know of a way to make sure the search results+bar, page content, and top header bar can be positioned nicely using CSS?

sorin-davidoi:
Don't think there is a nice way as long as absolute positioning is used.

Michael-F-Bryan:
Do we need to use absolute positioning in the first place?

mattico:
I agree re: bandaid solutions. This is what, the 3rd PR for a similar issue?

>I would have thought it'd be easy enough to use CSS to make sure the search bar doesn't overlap the content.

Possibly. My intent was just to fix the current behavior, but I suppose it'd be nice to show page content under the search.

mattico:
Okay, you nerd-sniped me there :) There were a few small CSS issues that I've wanted to fix, so I fixed them. Before merging, I should go through the git-blame and make sure I'm not regressing anything by removing some CSS hacks. I'll leave that for later though since I've already spent too much time on this for today.

mattico:
I think this is good to go. I tested it pretty thoroughly in FF, Chrome, Edge, and couldn't get anything to break.

mattico:
Looks like rust 1.25 broke netbsd.... *sigh*

mattico:
I'm going to see if a hello world fails to build on netbsd, report it upstream, then disable netbsd on travis.

mattico:
There, rebased and working.

Michael-F-Bryan:
And merged :tada:

steveklabnik:
When is a release going to be made with this in it? It seems like this affects all of the books in the bookshelf 😅

Michael-F-Bryan:
@steveklabnik I just pushed out a [new release]. Looking at the example book locally it seems like our previous styling problems are fixed :crossed_fingers: 

[new release]: https://crates.io/crates/mdbook/0.1.6

PR #649, opened 2018-03-25
Use `git config` to get author name in `mdbook init`

DivineGod:
I was getting panics while trying to run `mdbook init` and located the issue being that I don't store my global git config in `~/.gitconfig` so I changed this function to call `git config` instead.

`git config` will automatically find the correct config file using it's preferred method (https://git-scm.com/docs/git-config#FILES)

### DIFF
--- a/src/bin/init.rs
+++ b/src/bin/init.rs
@@ -72,15 +72,13 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
 fn get_author_name() -> Option<String> {
     let output = Command::new("git")
         .args(&["config", "--get", "user.name"])
-        .output();
+        .output()
+        .ok()?;
 
-    match output {
-        Ok(output) => if output.status.success() {
-            Some(String::from_utf8_lossy(&output.stdout).trim().to_owned())
-        } else {
-            None
-        },
-        _ => return None,
+    if output.status.success() {
+        Some(String::from_utf8_lossy(&output.stdout).trim().to_owned())
+    } else {
+        None
     }
 }
 
--- a/src/bin/init.rs
+++ b/src/bin/init.rs
@@ -68,17 +68,19 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
     Ok(())
 }
 
-/// Obtains author name from git config file by running the `git config` command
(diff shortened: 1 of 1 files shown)


src/bin/init.rs
```diff
@@ -68,20 +68,15 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
...
-            .map(|x| x.trim_left())
-            .filter(|x| x.starts_with("name"))
-            .next();
-        user_name
-            .and_then(|x| x.rsplit("=").next())
-            .map(|x| x.trim().to_owned())
+    let output = Command::new("git")
+        .args(&["config", "--get", "user.name"])
+        .output()
+        .expect("failed to get user.name from git config");
```
Michael-F-Bryan: Can we change this `expect()` call to something like `ok()?`? That way if there was an error executing `git` (i.e. git isn't installed) we'll just say no author was found instead of panicking.

src/bin/init.rs
```diff
@@ -68,20 +68,15 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
     Ok(())
 }
 
-/// Obtains author name from git config file if it can be located.
+/// Obtains author name from git config file by running the `git config` command
```
Michael-F-Bryan: Trailing full stop.

Michael-F-Bryan:
Thanks for tracking this down @DivineGod! I guess nobody ever noticed that fetching the author's name will fail because we all have a `~/.gitconfig` file. 

I'm pretty happy with this PR. It should hopefully be more flexible than parsing `~/.gitconfig` manually, plus it allows git to do the usual config resolution and use fallbacks/overrides.

Once we replace the panic with something that returns early this should be good to merge :grin:

Michael-F-Bryan:
yWith the last couple versions of the compiler you are able to use `?` with `Option<T>`. So You should be able to do `.output().ok()?` to return a `None` early, avoiding the verbose `match` statement.

DivineGod:
Oh! I was looking through the documentation to find that. Must’ve missed it. Will update with that change!

Michael-F-Bryan:
Nice work @DivineGod!

PR #650, opened 2018-03-26
Replace tempdir by tempfile

Eijebong:
The former has been deprecated in favor of the latter

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -496,7 +496,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "mdbook"
-version = "0.1.6-alpha.0"
+version = "0.1.6"
 dependencies = [
  "ammonia 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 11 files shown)


Eijebong:
I didn't include a version bump in there because I don't know what the right thing to do is (since it's in an alpha version and all :p).
It would be nice to have this released quickly as it's one of the last thing that's keeping winapi 0.2 in rustc

Michael-F-Bryan:
Thanks for the PR @Eijebong, I never noticed that `tempdir` has been deprecated.

> I didn't include a version bump in there because I don't know what the right thing to do is 

The `tempdir` crate is just an internal implementation detail of `mdbook` and not exposed anywhere, so there's no need for a version bump.

> It would be nice to have this released quickly as it's one of the last thing that's keeping winapi 0.2 in rustc

Works for me. It looks like you forgot to update a bunch of things in the `tests/` directory, but as soon as all the tests pass again we should be good to merge.

Eijebong:
Yup, I'm on it ^_^ forgot to check that cargo test was compiling

PR #653, opened 2018-03-29
Prevent search from triggering when editing code

nilsnh:
Heyo 👋 

I was playing with the interactive code editor in Rust By Example: https://rustbyexample.com/primitives/tuples.html when I noticed that typing `s` would bring up search and prevent me from coding with the letter s.

So, I added a check that exits the search event handler if the event originated in a `<textarea />`. I could have checked for class `ace_text-input` but that might have made the code more brittle if someone would happen to change the class for styling purposes.

### DIFF
--- a/src/theme/searcher/searcher.js
+++ b/src/theme/searcher/searcher.js
@@ -297,7 +297,7 @@ window.search = window.search || {};
     
     // Eventhandler for keyevents on `document`
     function globalKeyHandler(e) {
-        if (e.altKey || e.ctrlKey || e.metaKey || e.shiftKey) { return; }
+        if (e.altKey || e.ctrlKey || e.metaKey || e.shiftKey || e.target.type === 'textarea') { return; }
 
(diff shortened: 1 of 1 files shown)


Michael-F-Bryan:
I just tried this out on my laptop and I don't think this *quite* fixes the issue. 

I believe what I'm seeing is when you hit `s` the first time it'll release focus on the text area, then when you hit it again the search box will pop up.

nilsnh:
Thank you for the reply @Michael-F-Bryan .

I tested building an executable based on this PR, symlink it into `/usr/local/bin` and building Rust by Example locally with it. When I tested in Safari and Chrome the fix seems to work correctly. With this fix focus is not released when typing s within an editable code area.

Or have I missed something? 🤔

Michael-F-Bryan:
Hmm, if it's working fine for you it may have been a caching issue on my side then. I'll clear caches and try again.

PR #660, opened 2018-04-03
Add a note in MathJax documentation

THeK3nger:
There is a "hard to google" issue with some advanced math commands that uses double backslashes. I added a note in the documentation so that it is clear how to solve it.


azerupi:
Thanks! 

MathJax handling is currently not optimal, because it is parsed by the markdown parser. But until we refactor that part of the code, this note will probably help people who want to use it.

PR #661, opened 2018-04-03
Downgrade NetBSD's Rust version

mattico:
to workaround a build failure
https://github.com/rust-lang/rust/issues/49627


Michael-F-Bryan:
I was actually thinking we should cut out a lot of CI builds and go back to running on the raw VM. Due to the way `cross` is set up, we essentially have to re-download and re-compile *everything* every time a build is run, I don't believe it lets you use a cached `target/` directory. 

This means each build takes upwards of 20 minutes, even if there was a trivial change (e.g. updated styles or README).

Thoughts?

mattico:
That sounds good to me. We're a pretty simple project. If something doesn't work, it's probably a bug in rustc or a dependency.

PR #664, opened 2018-04-06
Remove unnecessary travis jobs

Michael-F-Bryan:
We're currently using `cross` to test over a dozen platforms on every commit and PR. This wastes a lot of time and, because `cross` only has read access to the host filesystem, we can't take advantage of caching to speed things up.

This PR removes all the extra CI stuff and makes sure dependencies are up-to-date.

Supersedes #661


Michael-F-Bryan:
Wow... Travis build times dropped from 30+ minutes down to 4 :open_mouth:

PR #665, opened 2018-04-09
Replace his with their in reference to reader

glfmn:
It's good practice to not assume the gender of the reader or a hypothetical developer.  The simple fix is to replace instances of `his` with `their` and `him` with `them` or `the user`.

I did a simple search with `ripgrep` so I may have missed some instances, but  `his` only appears twice as far as I could tell, so I replaced both instances with `their`.


Michael-F-Bryan:
Thanks for taking the time to make this PR @glfmn! I can't find any more instances of "his/her", so I think this is good to merge :+1:

PR #669, opened 2018-04-17
Put the search bar into an HTML form

mbrubeck:
This enables "Add a keyword for this search" in the contex menu for the
search field, in Firefox and other browsers.

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -116,9 +116,9 @@
 
                 {{#if search_enabled}}
                 <div id="search-wrapper" class="hidden">
-                    <div id="searchbar-outer" class="searchbar-outer">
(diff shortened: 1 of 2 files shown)


mattico:
This also causes the browser to show the form submission history pop-up, which seems good, and causes the browser to submit/reload when you hit enter, which seems fine. LGTM

mbrubeck:
It probably makes sense to prevent the form submission behavior, since it causes an unnecessary reload.

mbrubeck:
> It probably makes sense to prevent the form submission behavior, since it causes an unnecessary reload.

Done.  Note, this is also what rustdoc does for its search form.

Michael-F-Bryan:
I just tried this out locally, looks good to me :+1:

PR #672, opened 2018-04-22
Update deps

mattico:
`elasticlunr-rs` 2.0 added support for different languages using `rust-stemmers`. We didn't use this functionality since we only support English at the moment, but it's a small dependency with no new transitive dependencies so it didn't matter much. However, when trying to update the MDbook used by the compiler, [tidy complained that `rust-stemmers` had an incompatible license](https://github.com/rust-lang/rust/pull/50147#issuecomment-383395998).

So I released `elasticlunr-rs` 2.2 which makes the language support optional. This PR updates to that version, removing our dependency on `rust-stemmers`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,7 +29,7 @@ dependencies = [
 
 [[package]]
 name = "atty"
-version = "0.2.8"
+version = "0.2.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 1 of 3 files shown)


Michael-F-Bryan:
I just made another patch release which includes this change. @mattico `tidy` should be happy now.

mattico:
@Michael-F-Bryan thanks :)

Issue #671, opened 2018-04-20
cargo test fail on failing_alternate_backend


kazuk:
I fork and clone this repo.

cargo build is successfull with 1 warning.
cargo test fail on failing_alternate_backend.

this test pass normaly? 

```
PS C:\dev\mdBook> cargo test --test alternate_backends
    Finished dev [unoptimized + debuginfo] target(s) in 0.62 secs
     Running target\debug\deps\alternate_backends-e3b04d53d0b18fff.exe

running 4 tests
test missing_backends_arent_fatal ... ok
test alternate_backend_with_arguments ... ok
test passing_alternate_backend ... ok
test failing_alternate_backend ... FAILED

failures:

---- failing_alternate_backend stdout ----
        thread 'failing_alternate_backend' panicked at 'called `Result::unwrap_err()` on an `Ok` value: ()', src\libcore\result.rs:916:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::windows::backtrace::unwind_backtrace
             at C:\projects\rust\src\libstd\sys\windows\backtrace\mod.rs:65
   1: std::sys_common::backtrace::print
             at C:\projects\rust\src\libstd\sys_common\backtrace.rs:56
   2: std::panicking::default_hook::{{closure}}
             at C:\projects\rust\src\libstd\panicking.rs:381
   3: std::panicking::default_hook
             at C:\projects\rust\src\libstd\panicking.rs:391
   4: std::panicking::rust_panic_with_hook
             at C:\projects\rust\src\libstd\panicking.rs:577
   5: std::panicking::begin_panic<alloc::string::String>
             at C:\projects\rust\src\libstd\panicking.rs:538
   6: std::panicking::begin_panic_fmt
             at C:\projects\rust\src\libstd\panicking.rs:522
   7: std::panicking::rust_begin_panic
             at C:\projects\rust\src\libstd\panicking.rs:498
   8: core::panicking::panic_fmt
             at C:\projects\rust\src\libcore\panicking.rs:71
   9: core::result::unwrap_failed<()>
             at C:\projects\rust\src\libcore\macros.rs:23
  10: core::result::Result<(), mdbook::errors::Error>::unwrap_err<(),mdbook::errors::Error>
             at C:\projects\rust\src\libcore\result.rs:841
  11: alternate_backends::failing_alternate_backend
             at .\tests\alternate_backends.rs:23
[...]

PR #673, opened 2018-04-22
Ignore backend tests which require bash commands on windows

mattico:
except on Appveyor.

cc #671


mattico:
The other option is to write some bat scripts into the current directory before running these tests (true.bat: `@%COMSPEC% /C exit 1 >nul`), but this is less-hacky. I think it's most important for tests like this to run on Appveyor anyway.

Michael-F-Bryan:
@mattico are you sure we should be using cargo features to disable bash tests on windows? I thought features were mainly for enabling/disabling user-facing bits of code.

mattico:
I agree that it's not very elegant, but it's the best solution I could come up with.

PR #680, opened 2018-04-27
Improve print output

mattico:
Since #648, the content doesn't use `left`/`padding-left` to make room for the sidebar, instead using `transform`/`margin-left`. The print CSS wasn't updated for this change, so if the sidebar was visible it would push content to the right when printing.

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -1322,12 +1322,14 @@ html:not(.sidebar-visible) #menu-bar:not(:hover).folded > #menu-bar-sticky-conta
   .mobile-nav-chapters {
     display: none;
   }
-  #page-wrapper {
-    left: 0;
-    overflow-y: initial;
-  }
   #page-wrapper.page-wrapper {
-    padding-left: 0px;
+    -webkit-transform: none;
+    -moz-transform: none;
+    -o-transform: none;
+    -ms-transform: none;
+    transform: none;
+    margin-left: 0px;
+    overflow-y: initial;
   }
   #content {
     max-width: none;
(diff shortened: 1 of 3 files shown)


mattico:
Really need a checklist or something for making CSS changes.

mattico:
Hmm. Probably the best thing to do is wait for mathjax rendering to finish before opening the print dialog.

Michael-F-Bryan:
I tried running this on the user guide and got some pretty funny results. Using `master` everything gets pushed right by the sidebar, exactly as mentioned in the PR description.

[master.pdf](https://github.com/rust-lang-nursery/mdBook/files/1958506/master.pdf)

But then when I tried saving the example book as PDF (via the print button) from the `fix-print-sidebar` branch the output went all funny and I don't think it contains the actual content any more.

[with-sidebar-fix.pdf](https://github.com/rust-lang-nursery/mdBook/files/1958507/with-sidebar-fix.pdf)

---

Before

After

mattico:
@Michael-F-Bryan I can't reproduce that issue with either Chrome or Firefox and "Microsoft Print to PDF", can you give more specifics?

mattico:
Ah, OK. If I open the html in Chrome, I see the same issue. Doesn't happen with `serve` so probably another CORS issue.

mattico:
The issue doesn't happen in Firefox, and Chrome doesn't print any messages to the console. The print CSS doesn't reference any external files. I'm at a loss why this only happens on Chrome+Print+File and not any other situation...

mattico:
Unconditionally enabled the print CSS and the page renders fine, but the problem persists 😠.

mattico:
Alright, adding a 100ms delay seems to solve the issue.

Michael-F-Bryan:
@mattico I just checked out your latest changes and that seems to have fixed the issue. Chrome+File and Chrome+Serve is now rendering the print page correctly :tada:

PR #682, opened 2018-04-28
Recursively apply preprocessor

### LINKED ISSUE #678: Include a file that includes a file renders the handlebars
So I wanted to build rust Cookbook's table of contents dynamically.  I was creating each chapter's table and then including each of those chapter's tables into a master table.

Each chapter includes the relevant badges for crates (which are images) and category links.  In fact most of the cookbook's pages have at the bottom {{#include links.md}}.  

When I render the master table, including each of the chapter tables, they a separated by a rendered {{#include links.md}}  I'm looking for an easy solution to either:

 1. Include this file from the included file
 2. remove the {{#include }} from the output with a preprocessor

The only solution I can think of is creating each of the chapter's tables without the include at the bottom, and include that in a file that has:

{{#include algorithms/algorithms.md}}
{{#include links.md}}

and doing:
{{#include algorithms/algorithms.md}} in the master table.  

The challenge is that will require an extra file for every chapter, just to include the master table, and I'm afraid the project will begin to look less like a book and more like a template.  

Before I move that direction, I wanted to make sure this wasn't just something I was doing wrong.  There's no errors output by mdbook in the rendering.  When I see {{#include link.md}} somewhere, I usually see an error in the cli saying the file wasn't found (to that effect).

AndyGauge:
Sends the content inserted through the links processor.

fixes #678

### DIFF
--- a/src/preprocess/links.rs
+++ b/src/preprocess/links.rs
@@ -9,6 +9,7 @@ use super::{Preprocessor, PreprocessorContext};
 use book::{Book, BookItem};
 
 const ESCAPE_CHAR: char = '\\';
+const MAX_LINK_NESTED_DEPTH: usize = 10;
 
 /// A preprocessor for expanding the `{{# playpen}}` and `{{# include}}`
 /// helpers in a chapter.
@@ -36,7 +37,7 @@ impl Preprocessor for LinkPreprocessor {
                     .map(|dir| src_dir.join(dir))
                     .expect("All book items have a parent");
 
-                let content = replace_all(&ch.content, base);
+                let content = replace_all(&ch.content, base, &ch.path, 0);
                 ch.content = content;
             }
         });
@@ -45,11 +46,12 @@ impl Preprocessor for LinkPreprocessor {
     }
 }
 
-fn replace_all<P: AsRef<Path>>(s: &str, path: P) -> String {
+fn replace_all<P: AsRef<Path>>(s: &str, path: P, source: &P, depth: usize) -> String {
     // When replacing one thing in a string by something with a different length,
     // the indices after that will not correspond,
     // we therefore have to store the difference to correct this
     let path = path.as_ref();
+    let source = source.as_ref();
     let mut previous_end_index = 0;
     let mut replaced = String::new();
 
@@ -58,7 +60,15 @@ fn replace_all<P: AsRef<Path>>(s: &str, path: P) -> String {
 
         match playpen.render_with_path(&path) {
             Ok(new_content) => {
-                replaced.push_str(&new_content);
+                if depth < MAX_LINK_NESTED_DEPTH {
+                    if let Some(rel_path) = playpen.link.relative_path(path) {
+                        replaced.push_str(&replace_all(&new_content, rel_path, &source.to_path_buf(), depth + 1));
+                    }
+                }
+                else {
+                    error!("Stack depth exceeded in {}. Check for cyclic includes",
+                            source.display());
+                }
                 previous_end_index = playpen.end_index;
             }
             Err(e) => {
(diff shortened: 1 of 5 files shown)


Michael-F-Bryan:
This seems fairly reasonable. If I'm wanting to include a document it makes sense to also expand any `{#include ...}` statements inside it.

Should we bother to guard against accidental include-cycles? It's certainly an edge case of an edge case, but we'd end up with infinite recursion and a crash due to stack overflow. Which would be kinda annoying.

AndyGauge:
I created a book that in chapter_1.md, I {{#include chapter_1.md}} and this was the message:

```
2018-05-01 08:24:58 [INFO] (mdbook::book): Book building has started

thread 'main' has overflowed its stack
fatal runtime error: stack overflow
Aborted (core dumped)
```

What this doesn't tell me is what file caused the overflow, so maybe some messaging around it could curtail issues.

AndyGauge:
```
2018-05-01 11:06:58 [INFO] (mdbook::book): Book building has started                                                      
2018-05-01 11:06:58 [ERROR] (mdbook::preprocess::links): Stack depth exceeded in chapter_1.md. Check for cyclic includes  
2018-05-01 11:06:58 [INFO] (mdbook::book): Running the html backend                                                       
2018-05-01 11:06:58 [INFO] (mdbook::serve): Serving on: http://localhost:3000                                             
2018-05-01 11:06:58 [INFO] (ws): Listening for new connections on 127.0.0.1:3001.                                         
2018-05-01 11:06:58 [INFO] (mdbook::watch): Listening for changes...                                                      
```

src/preprocess/links.rs
```diff
@@ -58,8 +59,18 @@ fn replace_all<P: AsRef<Path>>(s: &str, path: P) -> String {
...
             Ok(new_content) => {
-                replaced.push_str(&new_content);
-                previous_end_index = playpen.end_index;
+                if depth < 200 {
+                    replaced.push_str(&replace_all(&new_content, path, &source, depth + 1));
+                    previous_end_index = playpen.end_index;
+                }
+                else {
+                    if let Some(source_path) = source.as_os_str().to_str() {
+                        error!("Stack depth exceeded in {}. Check for cyclic includes",
```
Michael-F-Bryan: I like this error message, it's quite helpful and user friendly.

src/preprocess/links.rs
```diff
@@ -58,8 +59,18 @@ fn replace_all<P: AsRef<Path>>(s: &str, path: P) -> String {
...
         match playpen.render_with_path(&path) {
             Ok(new_content) => {
-                replaced.push_str(&new_content);
-                previous_end_index = playpen.end_index;
+                if depth < 200 {
+                    replaced.push_str(&replace_all(&new_content, path, &source, depth + 1));
+                    previous_end_index = playpen.end_index;
+                }
+                else {
+                    if let Some(source_path) = source.as_os_str().to_str() {
```
Michael-F-Bryan: You can actually remove the need for the `if let Some(...)` and two branches by using `source.display()` in the `error!()` message.

src/preprocess/links.rs
```diff
@@ -58,8 +59,18 @@ fn replace_all<P: AsRef<Path>>(s: &str, path: P) -> String {
 
         match playpen.render_with_path(&path) {
             Ok(new_content) => {
-                replaced.push_str(&new_content);
-                previous_end_index = playpen.end_index;
+                if depth < 200 {
```
Michael-F-Bryan: This can probably be pulled out into its own private `const` (e.g. `MAX_NESTED_INCLUDES_DEPTH`) so it's not a magic number. 

I highly doubt anyone's going to have more than 2 or 3 nested includes, so we can probably cap it off at 10 instead of 200. It'd be annoying if we actually overflowed the stack before hitting the 200 depth limit.

tests/dummy_book/src/first/recursive.md
```diff
@@ -0,0 +1,2 @@
+Around the world, around the world
```
AndyGauge: https://genius.com/Daft-punk-around-the-world-lyrics

AndyGauge:
So I assume the `unwraps()` have to go, but I wanted to get some feedback on this last change.  What was happening is I included a file like {{#include data/adata-file.md}} then adata-file.md had {{#include ../links.md}} and the relative path was parsing from the original file, not from the context of the included file's path.

Michael-F-Bryan:
> If I looked at the output from the file vs the output of the file when it was included, they would be different--one would have a broken include and one would do it right.

Ah I can see how that would happen. As you mentioned, the "current directory" changes when trying to include a link from a different directory. I think how you've done it (new source path = old current directory + new link) is a good solution and should hopefully be intuitive for people.

Once the unwraps are swapped with proper error handling I think this PR will be good to merge. I'm assuming hitting one of the unwraps while rendering `recursive.md` is what's causing travis to complain.

Michael-F-Bryan:
Oops, I didn't notice you'd added a commit to resolve those last couple nits. This should be good to merge now.

AndyGauge:
Will this make it into v0.1.8?

Thanks!

PR #684, opened 2018-04-30
Add `rewrite_to_dir` config to rewrite index file path

weihanglo:
## The PR is for

Rewrite specified filename of URLs to directory index `/`.

## Intent

Need cleans URL for sharing article and chapters.  File paths of `index.html` are redundant.

## Details

Add a new config field `rewrite-to-dir` under `[output.html]`. This field accepts `Vec<String>` of filename. Take an example:

```toml
[output.html]
mathjax-support = true
no-section-label = true
rewrite-to-dir = ["index.html"]
```

This configuration only affects TOC and navigations without touch any contents.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -424,6 +424,8 @@ pub struct HtmlConfig {
     pub livereload_url: Option<String>,
     /// Should section labels be rendered?
     pub no_section_label: bool,
+    /// Filenames listed here would be rewritten to directory index `/`.
+    pub rewrite_to_dir: Vec<String>,
     /// Search settings. If `None`, the default will be used.
     pub search: Option<Search>,
 }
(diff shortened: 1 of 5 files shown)


Michael-F-Bryan:
This sounds like a pretty good idea. I usually write the "main" chapter for each level as `index.md` just so it'll be accessible at `/`. 

How should this interact with #685? Ideally I should be able to write the top level item in each directory as `README.md` and then have it accessible as `/`.

Also something to keep in mind is this `blah/index.html => blah/` behaviour only works when you're viewing the file on a server. If you read the book from file (e.g. `rustup doc --book`, `mdbook build --open`) clicking on a `blah/` link will present you with the contents of the `blah` directory instead of a rendered `blah/README.md` file...

PR #685, opened 2018-05-01
Add index preprocessor

weihanglo:
## The PR is for

Add new preprocessor to convert **README.md** to **index.md**.

## Intent

**README.md** is a de facto index file in markdown-based documentation. Many version control service such as GitHub, Gitlab and Bitbucket render README.md as index files while browsering source code on their web client. Hence, we may respect to README.md and convert it into index.md which would then convert into index.html.

I think this behavior is what people expect nowadays. It would be nice to include this feature.


src/preprocess/index.rs
```diff
@@ -0,0 +1,34 @@
...
+
+impl Preprocessor for IndexPreprocessor {
+    fn name(&self) -> &str {
+        "index"
+    }
+
+    fn run(&self, _ctx: &PreprocessorContext, book: &mut Book) -> Result<()> {
+        book.for_each_mut(|section: &mut BookItem| {
+            if let BookItem::Chapter(ref mut ch) = *section {
+                if ch.path.file_name().unwrap_or_default() == "README.md" {
```
Michael-F-Bryan: Just out of curiosity, do you think it's worth also checking for something like `readme.md` or `README`? So using the file stem instead and a case insensitive match.

weihanglo:
I have some concerns with opt-in this feature by default:

1. What if a markdown file named `index.md` co-exists with `README.md`?
1. mdbook picks the first prefix chapter item to generate the root `index.html`. How to solve the conflicts with another `README.md`?

Should we need to inspect this issue? Or just leave the decision for users. Let users opt-out the preprocessor or reorganize their book structure manually.

We can also check out [how gitbook did on this issue](https://github.com/GitbookIO/gitbook/issues/392). I tried and it seems that if both `README.md` and `index.md` exist, Gitbook will generate the last one listed in `SUMMARY.md` as the only `index.html`.

weihanglo:
- Integration tests are added at 009ebdc. 
- Case-insensitive matching is implemented at 92487a3.

As for file name conflict issue, I decided to convert all README.md and index.md without regarding any pitfall. Just as the same behavior you metioned above, it only emit some warnings to inform users to fix it.

Michael-F-Bryan:
I like the integration tests and the check for matching different variants of `README.md` :+1:

This PR looks about ready to merge. If you've got time, would you be able to make a follow-up PR which documents the `index` preprocessor and what it does in the user guide?

weihanglo:
@Michael-F-Bryan 
Thanks for you review! I am willing to document some notes about the `index` preprocessor.

Issue #688, opened 2018-05-09
Warnings when using rustfmt


gnzlbg:
rustfmt spits a bunch of warnings on mdbook:

```
Warning: Unknown configuration option `array_layout`
Warning: Unknown configuration option `chain_indent`
Warning: Unknown configuration option `fn_args_layout`
Warning: Unknown configuration option `fn_call_style`
Warning: Unknown configuration option `generics_indent`
Warning: Unknown configuration option `array_layout`
Warning: Unknown configuration option `chain_indent`
Warning: Unknown configuration option `fn_args_layout`
Warning: Unknown configuration option `fn_call_style`
Warning: Unknown configuration option `generics_indent`
Warning: Unknown configuration option `array_layout`
Warning: Unknown configuration option `chain_indent`
Warning: Unknown configuration option `fn_args_layout`
Warning: Unknown configuration option `fn_call_style`
Warning: Unknown configuration option `generics_indent`
Warning: Unknown configuration option `array_layout`
Warning: Unknown configuration option `chain_indent`
Warning: Unknown configuration option `fn_args_layout`
Warning: Unknown configuration option `fn_call_style`
Warning: Unknown configuration option `generics_indent`
Warning: Unknown configuration option `array_layout`
Warning: Unknown configuration option `chain_indent`
Warning: Unknown configuration option `fn_args_layout`
Warning: Unknown configuration option `fn_call_style`
Warning: Unknown configuration option `generics_indent`
Warning: Unknown configuration option `array_layout`
Warning: Unknown configuration option `chain_indent`
Warning: Unknown configuration option `fn_args_layout`
Warning: Unknown configuration option `fn_call_style`
Warning: Unknown configuration option `generics_indent`
Warning: Unknown configuration option `array_layout`
Warning: Unknown configuration option `chain_indent`
Warning: Unknown configuration option `fn_args_layout`
Warning: Unknown configuration option `fn_call_style`
Warning: Unknown configuration option `generics_indent`
Warning: Unknown configuration option `array_layout`
Warning: Unknown configuration option `chain_indent`
Warning: Unknown configuration option `fn_args_layout`
Warning: Unknown configuration option `fn_call_style`
Warning: Unknown configuration option `generics_indent`
Warning: Unknown configuration option `array_layout`
Warning: Unknown configuration option `chain_indent`
Warning: Unknown configuration option `fn_args_layout`
Warning: Unknown configuration option `fn_call_style`
[...]

PR #690, opened 2018-05-11
Remove rustfmt.toml

mattico:
Removes `rustfmt.toml` as discussed in #688.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -24,7 +24,6 @@ mod execs {
     }
 }
 
-
 error_chain!{
  foreign_links {
         Io(std::io::Error);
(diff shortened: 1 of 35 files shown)


azerupi:
Thanks!
Did you run rustfmt without config on the code base?
I there are any changes, it might be worth including them in this PR to avoid noise in the next PR that runs rustmft.

mattico:
I didn't do that in this PR because it would cause rebase churn for other branches/PRs. I suppose we have to do it at some point, though, might as well do it all at once.

Michael-F-Bryan:
I'd be happy with removing `rustfmt.toml` and just falling back to the defaults. They're pretty good now.

@mattico, is this ready to merge?

mattico:
I feel like there should be a merge conflict by now, but GitHub seems to think not. That's my only concern with merging.

PR #691, opened 2018-05-13
Update Regex to 1.0.0

phansch:
See https://github.com/rust-lang/regex/releases/tag/1.0.0

This would make Rust 1.20 the minimum required version.


azerupi:
>This would make Rust 1.20 the minimum required version.

I would be Ok with that, but let's check with @Michael-F-Bryan first :)
Is there a policy that has been setup about updating the minimum Rust version?

Michael-F-Bryan:
I *believe* this should be okay because the vast majority of Rust users tend to be on `stable`, or at most a couple releases behind.

I'm probably not a good guide to go by though, seeing as I routinely use `nightly` for the improved compile times or the recently stabilised `impl Trait`.

> Is there a policy that has been setup about updating the minimum Rust version?

I don't believe anything's been set up for `mdbook`. Do you know if there are any accepted community guidelines on the topic?

mattico:
https://github.com/rust-lang-nursery/api-guidelines/issues/123#issuecomment-342470263 is a pretty good summary.

Since mdbook is almost always (?) used as a binary, I'd think it's much less important that we work on old compilers compared to other library projects.

phansch:
If I'm reading https://tracker.debian.org/pkg/rustc correctly, then debian stable ships with Rust 1.24.1 by default.

Michael-F-Bryan:
> Since mdbook is almost always (?) used as a binary

Crates.io says there are [only 3 published crates][1] which depend on `mdbook`, two of which I own, so I'd say most people just install the binary from source.

> If I'm reading https://tracker.debian.org/pkg/rustc correctly, debian stable is currently on Rust 1.14 :/

Wow, that's ancient! We didn't even have custom derives back then...

@phansch I can't remember where I saw it exactly, but I do remember hearing one of the Debian maintainers doing work to update the debian version to around `1.20`. We should also be generating release binaries for Linux, so Debian users have the option of using them if they don't want to update their `rustc`.

[1]: https://crates.io/crates/mdbook/reverse_dependencies

azerupi:
Arbitrarily, I would vote to support at least the last three stable releases and only update the required Rust version when we have a (good) reason. I would argue that updating a dependency to 1.0 is a good reason. 😉

Michael-F-Bryan:
> I would argue that updating a dependency to 1.0 is a good reason. 

Couldn't agree more :+1:

@phansch can you also add a line saying something like "requires at least Rust 1.20" under the "*From Crates.io*" section on the README? Once there's a note on the minimum compiler version I think this should be good to merge.

phansch:
@Michael-F-Bryan done :)

azerupi:
Thanks!

PR #696, opened 2018-05-14
Fix default theme dir selection

### LINKED ISSUE #695: Theme directory behavior does not match documentation
[The guide says](https://rust-lang-nursery.github.io/mdBook/format/config.html):

However, the actual implementation looks for a theme directory inside of src:

https://github.com/rust-lang-nursery/mdBook/blob/b9ca108fcae2b738ca3f0c88c84ae5dc5a6f843f/src/renderer/html_handlebars/hbs_renderer.rs#L274-L277

Which was changed in #491, and we just haven't noticed yet!?

mattico:
Fixes #695 

Another option would be to accept both `/theme` and `/src/theme`, since some may have begun relying on the undocumented behavior.

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -1,9 +1,9 @@
 use book::{Book, BookItem, Chapter};
 use config::{Config, HtmlConfig, Playpen};
 use errors::*;
-use renderer::{RenderContext, Renderer};
 use renderer::html_handlebars::helpers;
-use theme::{self, Theme, playpen_editor};
+use renderer::{RenderContext, Renderer};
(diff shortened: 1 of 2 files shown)


azerupi:
>Another option would be to accept both /theme and /src/theme, since some may have begun relying on the undocumented behavior.

I would avoid this if we can. But if this bug has existed for long, we might not have much of a choice. If we go this route, I would add a deprecation warning so that we can remove it sometime in the future.

mattico:
It was changed in #491, merged in December 2017. At the same time, though, fixing this is really easy. Maybe we could detect a /src/theme directory and warn?

azerupi:
I agree with you. The warning should suffice :)

PR #710, opened 2018-06-19
Fix different font sizes in editor and code

### LINKED ISSUE #705: Editable code fragments use a font size too small to be easily readable
As per request on rust-lang/rust-by-example#1087 that flagged this as upstream issue.

On [rust-by-example](https://doc.rust-lang.org/stable/rust-by-example/hello.html) the editable code example has a very small font size (for me):
foo
While [book/second-edition](https://doc.rust-lang.org/book/second-edition/ch01-02-hello-world.html#writing-and-running-a-rust-program) uses the same font size for text and code, which is preferable in my opinion.

Eyenseo:
Fixes #705

### DIFF
--- a/src/theme/playpen_editor/editor.js
+++ b/src/theme/playpen_editor/editor.js
@@ -12,7 +12,8 @@ window.editors = [];
             showPrintMargin: false,
             showLineNumbers: false,
             showGutter: false,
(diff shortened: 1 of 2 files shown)


mattico:
Thanks!

PR #715, opened 2018-06-22
user guide: Add instructions for running `mdbook` in CI

### LINKED ISSUE #714: Add instructions on how to run `mdbook` in Travis CI
Both for testing and deploying to gh-pages.

I'm installing with `cargo install` right now, but compiles are killing my CI turnaround times. I see that tarballs exist in the releases tab on github, but it would be super awesome if there were instructions on how to download and use them from Travis CI. I doubt I'm the only person who would be interested in such instructions :)

fitzgen:
This adds instructions for building and testing books in CI on every PR and push, as well as instructions for how to automatically deploy to gh-pages on successful CI runs on `master`.

Fixes #714

r? @steveklabnik

### DIFF
--- a/book-example/src/SUMMARY.md
+++ b/book-example/src/SUMMARY.md
@@ -17,6 +17,7 @@
         - [Editor](format/theme/editor.md)
     - [MathJax Support](format/mathjax.md)
     - [mdBook specific features](format/mdbook.md)
+- [Continuous Integration](continuous-integration.md)
 - [For Developers](for_developers/README.md)
     - [Preprocessors](for_developers/preprocessors.md)
     - [Alternate Backends](for_developers/backends.md)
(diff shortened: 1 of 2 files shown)


steveklabnik:
I cannot re-run appveyor, but I guess it's not a big deal; only nightly was failing and it was due to rust infra issues.

fitzgen:
Thanks @steveklabnik :)

repi:
Is it the best to recommend using Cargo install and relying on Travis caching the build results? It feels a bit wasteful for everyone to build mdbook from source every time and have had some mixed experiences with the Travis caching.

We just get the latest pre-built binary from the repo instead which is always fast and doesn't use any storage for caching:

```
before_script:
  - wget -qO- https://github.com/rust-lang-nursery/mdBook/releases/download/v0.1.5/mdbook-v0.1.5-x86_64-unknown-linux-gnu.tar.gz | tar xvz  
```

Just a thought, both approaches work

PR #716, opened 2018-06-24
Use a set of default system font instead of webfont

ishitatsuyuki:
The reason is simple: I don't like Open Sans. The font set used in this PR is identical to GitHub's one.

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -1,5 +1,5 @@
 html {
-  font-family: "Open Sans", sans-serif;
+  font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Helvetica, Arial, sans-serif;
   color: #333;
 }
 body {
@@ -8,7 +8,7 @@ body {
   overflow-x: hidden;
 }
 code {
-  font-family: "Source Code Pro", Consolas, "Ubuntu Mono", Menlo, "DejaVu Sans Mono", monospace, monospace;
+  font-family: "SFMono-Regular", Consolas, "Liberation Mono", Menlo, Courier, monospace;
   font-size: 0.875em;
 }
 .left {
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -12,8 +12,6 @@
         <base href="{{ path_to_root }}">
 
         <link rel="stylesheet" href="book.css">
-        <link href="https://fonts.googleapis.com/css?family=Open+Sans:300italic,400italic,600italic,700italic,800italic,400,300,600,700,800" rel="stylesheet" type="text/css">
-        <link href="https://fonts.googleapis.com/css?family=Source+Code+Pro:500" rel="stylesheet" type="text/css">
 
         <link rel="shortcut icon" href="{{ favicon }}">
 
--- a/src/theme/stylus/general.styl
+++ b/src/theme/stylus/general.styl
@@ -1,6 +1,6 @@
 html {
-    font-family: "Open Sans", sans-serif
-    color: #333
+    font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Helvetica, Arial, sans-serif;
+    color: #333;
 }
 
 body {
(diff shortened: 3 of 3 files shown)


mattico:
👍 Been meaning to do this for a long time.

clarcharr:
IMHO using a webfont locally hosted instead of Google-hosted would be better.

Is there a reason why you'd rather not use Open Sans?

ishitatsuyuki:
I prefer Helvetica over Open Sans, because my eyes are more used to it. I don't like Open Sans's round-ish glyphs.

clarcharr:
I understand that, although I'm not sure if that alone is enough to justify the change. Having a consistent font is nice for this sort of thing.

Personally, I'm not sure why mdbook uses Open Sans when it could be using Source Sans Pro. I personally much prefer its look, and it's open source.

ishitatsuyuki:
> The font set used in this PR is identical to GitHub's one.

Does this qualify as "consistent"?

frewsxcv:
> > The font set used in this PR is identical to GitHub's one.
> 
> Does this qualify as "consistent"?

I think @clarcharr might be referring to the incorporation of `-apple-system` in this pull request, which means mdbooks will look different across different operating systems.

ishitatsuyuki:
> which means mdbooks will look different across different operating systems.

Makes sense. Though, my argument on this is that this set of system fonts provides a *native* looking, instead of aligning the same font everywhere.

clarcharr:
Yes, my definition of "consistent" was "same on every device" and not "similar to other services."

While I feel that it doesn't matter a lot for Rust docs, one thing that web fonts solve is ensuring that websites aren't *too* disturbed by low-quality system fonts. Additionally, adding that uniformity means that you're not going to be too surprised by what something looks like on another system.

The fact that Arial is at the end of this list makes me cringe, considering how despite the fact that it did its job very well at the time (looking well on low-pixel-density displays), it looks absolutely atrocious on modern systems. On my system, the only reason why this setup would look good at all is because I manually have Helvetica set to use Source Sans Pro instead, because the alternatives look absolutely atrocious to my eyes. Taking a good font (Open Sans) and replacing it with a Helvetica so that it looks good on your system isn't an excuse for making it look terrible on others.

"I don't like Open Sans" is not really a compelling reason to get rid of it. If there's a genuine design thing, like making it look more readable, I'm all for it. But all things considered, Open Sans is a solid font, and this suggests switching to potentially *terrible* system alternatives.

mattico:
Since mdBook is used for Rust's official documentation I think it's a good idea to be fairly conservative in general, and removing external resources like these fonts would help with some edge cases:

1. If fonts are [slow to load](https://github.com/rust-lang-nursery/mdBook/issues/46#issuecomment-388579733) (high latency connection, misconfigured DNS/proxy, etc.) they can block page load. This could potentially be worked around with `rel="preload"` on webkit or some JS, but that would still cause a rerender when the font finally loads.
2. Certain users don't want to depend on external CDN resources for privacy reasons.
3. Packagers (Debian et. al.) need to patch these to point to local packages.
3. Removing these fonts completely would cut down on requests and data transferred. These font files are not very big (<100KiB) so it's not a huge issue but it would still be an improvement.

Vendoring these fonts would be straightforward, and would solve 2 and 3. I'm still leaning towards removal since it would be simpler and the motivation to keep them is weak, IMO.

It isn't important that mdBook uses the exact same fonts across devices. We don't have a strict brand identity we need to uphold, we just want to use some decent readable fonts. Reasonable people can disagree but most systems should (and do) have decent readable fonts at this point. You mentioned Arial which is not ideal but Segoe UI was introduced in Windows Vista so it shouldn't get used much. Arial could perhaps be removed entirely, since it likely isn't any better than the browser's default `sans-serif` which the user can usually change easily.

ehuss:
I'm going to close this, as it is pretty old. Also, the `copy-fonts` option was added so that you can control whether or not the other fonts are used.

PR #717, opened 2018-06-24
Font size and padding tuning

ishitatsuyuki:
These changes are quite opinionated, so feedbacks are welcome. [Docusaurus](https://github.com/facebook/Docusaurus/blob/1094dbd35217f057e521aab76d877ec64c0ba141/lib/static/css/main.css) is used as a reference.

### DIFF
--- a/src/theme/book.css
+++ b/src/theme/book.css
@@ -5,7 +5,7 @@ html {
 body {
   margin: 0;
   font-size: 1rem;
-  overflow-x: hidden;
+  line-height: 1.5;
 }
 code {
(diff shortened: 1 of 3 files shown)


stappersg:
I think, and say, that it about time to close this pull request.
* CI checks have failed from the begin
* No one cared about this p.r. for two years
* We don't know much human energy this p.r. did suck
* We should prevent that this p.r. drains more developer resource

Please close is pull request.

By the way:  Happy Birthday

Dylan-DPC:
yes forgot this was open :D 

closing it as it's too inactive and it might have been fixed or considered pointless now. 

Thanks :)

PR #718, opened 2018-06-25
Use NPM packages for executing stylus

ishitatsuyuki:
This allows:

- Not polluting global /usr/bin
- Managing dependencies with proper version requirements so things don't break


ishitatsuyuki:
Hmm, command execution doesn't work correctly on Windows. I'll debug later.

ishitatsuyuki:
Addressed. Can this get a review?

frewsxcv:
Any reason why need to use yarn instead of just using [npm's lockfile](https://docs.npmjs.com/files/package-lock.json)?

ishitatsuyuki:
It's just a matter of preference, but in many cases, yarn lockfile has a more stable output than NPM's one. In the JS community, I've seen more and more repos opting out of NPM's lockfile.

mattico:
I'd prefer to remove stylus altogether (#636), and plan to do that after the next release. If we do end up sticking with stylus, however, this PR looks good to me.

ishitatsuyuki:
This PR is also a groundwork for dealing with NPM packages, which you will probably end up using as long as web is concerned.

Even if we got rid of stylus, it's likely that we will be using something like PostCSS as compatibility (on top of bleeding edge) is always a thing to care about.

Dylan-DPC:
@ishitatsuyuki can you resolve the conflicts?

ishitatsuyuki:
@Dylan-DPC I think someone have already ditched stylus so this is obsolete.

Dylan-DPC:
Ah didn't realise that. Thanks for pointing that. Closing this

PR #726, opened 2018-07-19
Update to elasticlunr 2.3.

waywardmonkeys:
This removes one of the remaining dependencies upon pre-1.0 regex.


frewsxcv:
For posterity, here's the diff: https://github.com/mattico/elasticlunr-rs/compare/v2.2.0...v2.3.3

PR #728, opened 2018-07-21
- Fixed token type.

kggp1995:
OAuth token type doesn't work. Have to change it to PAT.

---
This change is [Reviewable](https://reviewable.io/reviews/rust-lang-nursery/mdbook/728)


mattico:
Thanks!

Issue #721, opened 2018-07-06
Optionally add an always-visible link to the repository of an mdBook


gnzlbg:
Finding issues in mdBooks while reading them is extremely common, yet it is currently extremely hard to report them and fix them. For example, googling for ["cargo book github"](https://www.google.de/search?q=cargo+book+github&oq=cargo+book+github&aqs=chrome.0.69i59j69i60.5137j1j7&sourceid=chrome&ie=UTF-8) doesn't report many meaningful links to the book source code.

Reporting issues about the book should be trivial. Fixing issues in the book should be trivial too.

It would already be a great improvement if the bar would have a "Found an issue? Please report it or fix it!" link somewhere that is always visible, and that links to the book's repository.

This way, if I find an issue while reading a book, I can just click there, and either open a github issue or send a PR. 

FWIW, while I think this link and the text should be optional, I think that such a link is so important for the evolution of a book that it should actually be opt-out.

mattico:
Related but somewhat more general proposal: https://github.com/rust-lang-nursery/mdBook/issues/694

This feature is probably useful to enough mdbook users that it's worth adding a specific easy feature for it, rather than relying a more general "add a button to the header" interface.

I think adding a repository field to Book.toml is a good way to go:

```toml
[book]
title = "Example book"
author = "John Doe"
description = "The example book covers examples."
repository = "https://github.com/jdoe/example-book" #< add this book.toml key
```

Which would then render as either:
- A button + text in the header
- A link at the bottom of the left nav
- A link in the page footer

I'm leaning to a link at the bottom of the nav, something like:

> Book created using [mdBook](https://github.com/rust-lang-nursery/mdBook) | Contribute to this book {on Github,at our repository}

gnzlbg:
Sounds good! @steveklabnik what do you think about such a message for Rust's books ?

gnzlbg:
I just spent 10 minutes searching for the cargo book repository on google, again, without _any_ luck.

ehuss:
I believe this is fixed by #802.  As mentioned, see also #694 and #810 for making extra links more configurable.

As for the rust-lang-specific books, I plan to migrate more books to 0.2 in the near future, I'll try to add the `git-repository-url` to more rust-lang books as I do.

PR #734, opened 2018-07-24
Adjust header icon margins

mattico:
Before:

image

image

After:

image

image

I think an ideal solution would move these controls into the sidebar for very narrow devices, but this at least shows a few more characters of the title and allows the sidebar to be closed (barely).


mattico:
Adding an edit button (#721) will probably make this simple solution insufficient.

Issue #674, opened 2018-04-22
It is the Rust Playground, not the Playpen


shepmaster:
"Playpen" was a very old name for it.

pavelrappo:
It would definitely help if someone could describe the relationship between _playground_ and _playpen_.  When reading mdBook documentation it's not clear what the difference is between `{{#playpen}}` and fenced runnable code blocks. Both these types seem to be compiled and executed the same way by the Rust Playground.

ehuss:
`playpen` is just an old name for `playground`, they are one in the same.

`{{#playpen file.rs}}` is identical to:

```markdown
    ```rust
    {{#include file.rs}}
    ```
```

It looks like `{{#playpen}}` predates `{{#include}}`.  Probably not a super-strong reason to have both.

pavelrappo:
I do like it as a metaphor, though. Kids go outside of their house to get to a  playground. Toddlers stay inside and play in a playpen.

`mdBook` sends the snippets for compilation and execution, receives the results, and shows them in-place. Whereas `rustdoc` opens a new tab with your code copied to the playground.

Hope that makes sense.

shepmaster:
My original idea when opening this issue is that anything public-facing in mdBook be updated to use the term "playground". This *probably* would have backwards-compatibility effects, so I'd expect that `playpen` in code would stay around as aliases to the newly-renamed thing. Thus `#playpen` and `#playground` would do the same, but in all the docs it would only say "playground".

It's a low-priority change, but one that might be more valuable before more people are introduced to mdBook.

pavelrappo:
Right. Being consistent and expanding vocabulary only if absolutely necessary are good goals. Rustdoc seems to be using the word "playground", so one more reason for mdBook to use that instead of "playpen".

Issue #602, opened 2018-02-02
Document mdBook playpen attributes


projektir:
There does not appear to be documentation for the various playpen attributes in the UserGuide. I.e.:

- `editable`
- `ignore`
- `mdbook-runnable`

uint:
Looks like `editable` is already documented by now.

Correct me where I'm wrong! As far as I can understand, `ignore` and `noplaypen` do the exact same thing and remove the run button. `mdbook-runnable` overwrites both and makes the thing runnable again, though I'm not sure what use case that would have?

The logic I found for this is in the [specific renderers](https://github.com/rust-lang/mdBook/blob/master/src/renderer/html_handlebars/hbs_renderer.rs#L612-L613). Is there more that I've missed?

ehuss:
There is a subtle distinction from `ignore` and `noplaypen` where `noplaypen` blocks will still be tested.  It might be good to include the rustdoc attrs since they are relevant to how things work.  And some examples would be good, too.

- `editable` — Allows a code block to be editable, requires `output.html.playpen.editable=true` in config.
- `noplaypen` — Will not show playground link, but will be tested. (Need to also specify `rust`.)
- `mdbook-runnable` — Forces the playground link to be shown. This is intended to be combined with the `ignore` attr for examples that you don't want to test, but want to allow the reader to run.
- `ignore` — Skips Rust tests, will not show playground link.
- `should_panic` — Rust code should panic when run. Shows playground link.
- `no_run` — Rust code should build, but not be run during a test. Shows playground link.
- `compile_fail` — Rust code should fail to compile. Shows playground link.
- `edition2015` or `edition2018` — Specifies the Rust edition.

ISSOtm:
> UPDATE: See #1249 where some of these have changed.

The PR ID looks wrong, I'm not sure which changes this is mentioning. Otherwise, I'd be ready to tackle this.

In the meantime, could the OP be updated with a task list of the attributes that need to be documented?

ehuss:
I think #1249 is correct.  `no_run` was changed so that it does not show the playground "play" button.  I'm not sure if that changed any of the others (offhand I don't think it did, but it would be good to verify).

The list in https://github.com/rust-lang/mdBook/issues/602#issuecomment-561874933 should be mostly complete with one change that `noplayground` has been renamed `noplaypen`.  It might also be nice to link to the rustdoc docs [here](https://doc.rust-lang.org/rustdoc/documentation-tests.html#attributes).

[...]

PR #744, opened 2018-07-26
Add a "noplaypen" class for rust code samples.

eminence:
This class will supress the "play" button in the html backend (which you can also do with the "ignore" class), but it will still let the code be tested via `mdbook test` (which is not possible with the "ignore" class).

This is useful for code examples that don't really do much (and so the user doesn't gain much from running them), but as an author you still want to test them to guard against syntax errors and typos and the like.

Notes:  this new class is not documented (see #602), and also this uses the term "playpen", which is prevalent in the mdbook source code, but is not the preferred name (see #674)


mattico:
Sounds good to me! This should definitely be documented better, but I suppose that can happen separately.

PR #755, opened 2018-08-01
Add directory argument to `mdbook test`

### LINKED ISSUE #711: can you support the specified path to run mdbook test ?
Now mdbook just support for -h, -V, and -L parameters. It can support -P to specifiy a path to run test.

mattico:
Fixes #711.


mattico:
I don't especially like needing to do `--` after a `-L` argument, but this is consistent with every other command. In 0.2.0 we can change -L to take a single/comma-delimited argument.

PR #757, opened 2018-08-02
Improve command-line argument parsing

mattico:
Mostly improves wording and formatting. Includes breaking changes to two subcommands:
- **serve**: some arguments have been renamed to make more sense
- **test**: the `library-path` argument now requires a comma-delimiter so that ` -- ` isn't required to give a `dir` argument


mattico:
Travis is having some difficulty...

PR #758, opened 2018-08-03
Add path slash for additional css files in handlebars template

chadmorrow-fd:
This change ensures that additional css files are correctly sourced and apply to all pages, regardless of their location in the book's folder structure.

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -27,7 +27,7 @@
 
         <!-- Custom theme stylesheets -->
         {{#each additional_css}}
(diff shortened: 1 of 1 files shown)


chadmorrow-fd:
Weird. If I take the slash out it stops working again. I'll poke around at it some more and see if I can figure out what's up.

chadmorrow-fd:
path_to_root was actually evaluating to nothing which means it was just adding that slash before `{{ this }}` that caused the css to work properly again.

chadmorrow-fd:
Probably isn't the best way to fix this issue.

PR #765, opened 2018-08-11
Fix table and quote related color

weihanglo:
`darken` and `lighten` functions only live in CSS preprocessors. They should be replaced with actual colors 😃.

## Before (incorrect styles)

screen shot 2018-08-12 at 11 37 01

## After

screen shot 2018-08-12 at 11 36 31


mattico:
Thank you for this! I know I was going to miss something 😉. Using HSL is a nice touch.

Issue #780, opened 2018-09-06
mdBook does not produce any output


gnzlbg:
I was setting uploading the simd perf-guide to the nursery in packed_simd, and executing the following command in the root directory `packed_simd/`:

```
mkdir -p target/doc
 mdbook build perf-guide --dest-dir target/doc/
```

produces the following output

```
2018-09-06 10:44:34 [INFO] (mdbook::book): Book building has started
2018-09-06 10:44:34 [INFO] (mdbook::book): Running the html backend
```

but the destination directory is empty:

```
ll target/doc
total 0
drwxr-xr-x  2 gnzlbg  staff    64B Sep  6 10:37 .
drwxr-xr-x  5 gnzlbg  staff   160B Sep  6 10:37 ..
```

I haven't been able to find the output book anywhere after running this command (there doesn't seem to be a `doc/` or `book/` directory anywhere).

So what I've ended up doing is:

```
cd perf-guide
mdbook build
cd -
cp -r perf-guide/book target/doc/perf-guide
```

instead. IIUC `--dest-dir target/doc` or `--dest-dir target/doc/perf-guide` should do what I want, but they appear to not be working.

mattico:
If the dest-dir is a relative path, it's interpreted relative to the book's root directory. I agree that this is confusing (see https://github.com/rust-lang-nursery/mdBook/issues/698).

I believe this is a duplicate of that issue, but I've been wrong before. Feel free to reopen.

PR #782, opened 2018-09-06
Document dest-dir relative path behavior

mattico:
CC #780. Even if we release 0.3 soon and 'fix' this, we should have a release on the 0.2 branch which documents this behavior.


mattico:
@Michael-F-Bryan, this should be good to go, AFAIK. Not sure why I forgot about it.

PR #785, opened 2018-09-08
Fix the theming stuff in browsers where the parser doesn't wait for sheet loads before executing scripts.

### LINKED ISSUE #784: Themes don't work in Chrome.
Hey, so I was taking a quick look at https://github.com/rust-lang-nursery/rust-bindgen/issues/1367, and realized that chrome rendered differently than Firefox the bindgen book.

Turns out that the theming stuff is not working in Chrome because chrome doesn't wait for stylesheets to load in order to execute script. You can see an example in:

```html
<!doctype html>
<link rel="stylesheet" href="foo.css" onload="console.log('onload')">
<body>
<script>
  console.log("script");
</script>
```

(with an existing `foo.css` file). `onload` doesn't reliably get logged before `script`. You can see the testcase online in https://crisal.io/tmp/load-sheet.html. For me https://rust-lang-nursery.github.io/mdBook/cli/watch.html renders the code blocks dark on chrome, for example.

The reason why they don't work is because `HTMLLinkElement.disabled`, when called via the WebIDL setter, it doesn't do anything if `HTMLLinkElement.sheet` is null (this bit is compatible across browsers, but for more fun related to `disabled` and such I invite you to https://github.com/whatwg/html/issues/3840 :)).

In any case, this should probably just use the `rel` attribute, using `alternate stylesheet` as the `rel` value for the disabled sheets. I _think_ that should work everywhere.

emilio:
Fixes #784

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -320,22 +320,37 @@ function playpen_text(playpen) {
     function set_theme(theme) {
         let ace_theme;
 
+        function setSheetEnabled(sheetName, enabled) {
+            let element = stylesheets[sheetName];
+            let value = enabled ? "stylesheet" : "alternate stylesheet";
+            if (element.getAttribute("rel") != value)
+                element.setAttribute("rel", value);
+        }
+
+        function enable(sheetName, disabled) {
+            setSheetEnabled(sheetName, true);
+        }
+
+        function disable(sheetName, disabled) {
+            setSheetEnabled(sheetName, false);
+        }
+
         if (theme == 'coal' || theme == 'navy') {
-            stylesheets.ayuHighlight.disabled = true;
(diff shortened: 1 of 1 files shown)


rbuckland:
Thanks Emilio. This looks good. I had often wondered that but never looked into it

ehuss:
I did a little testing on this, and I can't reproduce the problem.  With Chrome (v80), rendering and loading `<script>` files seemed to be blocked until after the CSS files are downloaded.  I tested with a server which added artificial delays to all `css` files.  Also tested on firefox and safari with similar results.

I'm inclined to close this unless someone has a reproduction.

ehuss:
Closing as I can't reproduce this now. Please comment on the issue if there is a way to reproduce this now.

PR #786, opened 2018-09-09
test scripts to compare mdbook version output

rbuckland:
This PR adds some BASH scripts that forcibly test 2 configurable branch versions of mdbook with all the rust-lang-nursery/* and rust-lang/* book.toml repos *(finds them via github API search).

The tests/mdbook_version_validator/README.md has a LOT more information.

but in short, these scripts are good for running two separate versions (build by the scripts) of mdbook, and compiling a list of books, and checking for any differences - and reporting.

The scripts are NOT run as part of cargo test, they are just BASH scripts to give some level of confidence for current (smart-preprocessor) and future changes.

This PR was created to helps support potentially breaking PRS like #658

TODO before this is ready to be done.

* [ ] - Add a config file, allowing users to add their "books" to the config for searching. (either org or direct repo with book.toml)
* [ ] - Add some documentation, making the feature aware so other can register their - dependency / interest

It would be good to rewrite the test as rust - (as an excercise). (later)

### DIFF
--- a/tests/mdbook_version_validator/README.md
+++ b/tests/mdbook_version_validator/README.md
@@ -1,0 +1,155 @@
+# Testing Upstream Books for Breakage from mdbook
+
+- Theory - when mdbook changes, we should be able to coarsely tell if building a book with the new version fails
+- In Practice - Build the book twice and compare
+
+
+    1. Build two binaries of mdbook, some "current" (master) and some "new" branch
+       (Building dependent binaries - mdbook-linkcheck for example from https://github.com/Michael-F-Bryan/mdbook-linkcheck - are not supported "per build"
+       but rather the same binary is executed for both. Should not be hard to change that)
+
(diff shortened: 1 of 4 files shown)


rbuckland:
From: @Michael-F-Bryan 

> This is pretty cool! I imagine they'd be useful when checking for regressions or trying to reproduce an issue.
> 
> What are your thoughts on putting the README.md in the user guide under the for developers section? Technically it's for mdbook developers and not developers of 3rd party plugins, but being in the user guide will probably make the scripts more discoverable.

Covered in a TODO above in this PR.

ehuss:
I'm going to close as this is quite old. We now have something similar in https://github.com/rust-lang/mdBook/pull/2842, and the testsuite has been extended quite a bit to try to catch breaking changes. Hopefully those will help with being able to view changes that happen between releases. Thanks for the PR, though!

Issue #771, opened 2018-08-21
Feature: Emojis


pouladpld:
### :zap: Feature Request  :smiley_cat: 

Emojis make books :dark_sunglasses: visually interesting :dark_sunglasses: . It would be nice :fire: to have them supported in mdBook :book: .

<https://www.webfx.com/tools/emoji-cheat-sheet/>

MihaZupan:
👍

Havvy:
Emoji are part of unicode. Since mdbook supports unicode, what are you actually requesting?

MihaZupan:
Parsing for emoji markdown like `:smile:`

rbuckland:
I just have a few more things to fix and then it is ready. 
I am aware that : 
- it needs to move into a plugin preprocessor architecture
- should be optional (in config)
- should use the sets as found here - https://github.com/markdown-it/markdown-it-emoji/tree/master/lib/data

rbuckland:
@MihaZupan  This is now ready for use. It is waiting on the PR for `smart-processor` from @Michael-F-Bryan #658

Once that smart-preprocessor is merged, I will create a PR for this.

toVersus:
@rbuckland Any update on this? I cannot find merged PR for supporting emoji markup.

Progdrasil:
Is there any news on this? seems like the branch has been stale for 2 years

shonfeder:
In case this is useful for anyone who ends up here before this is implemented inside the mdBook project:I hacked together a general purpose utility to emojify github names: https://github.com/shonfeder/emojitsu 

We've just dropped the prebuilt binary into our CI.

PR #787, opened 2018-09-09
Updated the documentation for new preprocessor format

rbuckland:
@Michael-F-Bryan  This is the documentation you require for the preprocessor change.

The link - `[preprocessor-docs]` is forward looking 
   * that is a little harder - we need a preprocessor for that :-D

### DIFF
--- a/book-example/src/format/config.md
+++ b/book-example/src/format/config.md
@@ -14,9 +14,9 @@ description = "The example book covers examples."
 build-dir = "my-example-book"
 create-missing = false
 
-[preprocess]
-index  = true
-linkes = true
+[preprocess.index]
+
+[preprocess.links]
 
 [output.html]
 additional-css = ["custom.css"]
@@ -62,17 +62,29 @@ This controls the build process of your book.
   will be created when the book is built (i.e. `create-missing = true`). If this
   is `false` then the build process will instead exit with an error if any files
   do not exist.
+- **use-default-preprocessors:** Disable the default preprocessors of (`links` &
+  `index`) by setting this option to `false`.
 
-### Configuring Preprocessors
+  If you have the same, and/or other preprocessors declared via their table
+  of configuration, they will run instead.
+
+  - For clarity, with no preprocessor configuration, the default `links` and 
+    `index` will run.
+  - Setting `use-default-preprocessors = false` will disable these
+    default preprocessors from running.
+  - Adding `[preprocessor.links]`, for example, will ensure, regardless of 
+    `use-default-preprocessors` that `links` it will run.
+
+## Configuring Preprocessors
 
 The following preprocessors are available and included by default:
 
-- `links`: Expand the `{{# playpen}}` and `{{# include}}` handlebars helpers in
-  a chapter.
+- `links`: Expand the `{{ #playpen }}` and `{{ #include }}` handlebars helpers in
+  a chapter to include the contents of a file.
 - `index`: Convert all chapter files named `README.md` into `index.md`. That is
   to say, all `README.md` would be rendered to an index file `index.html` in the
   rendered book.
-- `emoji`: Convert `:emoji:` text into Emojis. eg: `:smile:` to :smile:
+
 
 **book.toml**
 ```toml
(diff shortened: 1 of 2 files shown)


book-example/src/format/config.md
```diff
@@ -14,6 +14,10 @@ description = "The example book covers examples."
 build-dir = "my-example-book"
 create-missing = false
 
+[preprocess]
+index  = true
```
Michael-F-Bryan: I don't think this is correct. Each preprocessor gets its own table and you can enable/disable inserting the default preprocessors by setting `build.use-default-preprocessors` to `true`.

book-example/src/format/config.md
```diff
@@ -59,10 +62,8 @@ This controls the build process of your book.
   will be created when the book is built (i.e. `create-missing = true`). If this
   is `false` then the build process will instead exit with an error if any files
   do not exist.
-- **preprocess:** Specify which preprocessors to be applied. Default is
```
Michael-F-Bryan: We'll need to mention the `use-default-preprocessors` flag here. It's used to enable/disable including the default preprocessors and defaults to `true`.

I've also added a couple tests in 18a36ec to make sure the `build.use-default-preprocessors` flag works as intended.

book-example/src/format/config.md
```diff
@@ -71,14 +72,39 @@ The following preprocessors are available and included by default:
 - `index`: Convert all chapter files named `README.md` into `index.md`. That is
   to say, all `README.md` would be rendered to an index file `index.html` in the
   rendered book.
-
+- `emoji`: Convert `:emoji:` text into Emojis. eg: `:smile:` to :smile:
```
Michael-F-Bryan: Ooh, there's an `emoji` preprocessors? That's really cool!

book-example/src/format/config.md
```diff
@@ -71,14 +72,39 @@ The following preprocessors are available and included by default:
...
+#### Locking a Preprocessor dependency to a renderer
+
+You can explicitly specify that a preprocessor should run for a renderer by binding the two together.
+
+```
+[preprocessor.mathjax]
+renderers = ["html"]  # mathjax only makes sense with the HTML renderer
+```
+
+#### Nested Configuration
```
Michael-F-Bryan: This section will probably need to mention that, like renderers, each preprocessor will always need to be given its own table (e.g. `[preprocessor.mathjax]`). From there, passing extra configuration to the preprocessor is just a case of adding key-value pairs to the table, so there isn't really a separate "*style of configuration*".

book-example/src/format/config.md
```diff
@@ -71,14 +72,39 @@ The following preprocessors are available and included by default:
...
+- `emoji`: Convert `:emoji:` text into Emojis. eg: `:smile:` to :smile:
 
 **book.toml**
 ```toml
 [build]
 build-dir = "build"
 create-missing = false
-preprocess = ["links", "index"]
+
+[preprocess]
```
Michael-F-Bryan: This will need to be changed to something like:

```toml
[preprocess.links]

[preprocess.index]

[preprocess.emoji]
```

Although those three preprocessors will be enabled by default, so it's kinda redundant to even mention them...

Michael-F-Bryan:
@rbuckland about the `[preprocessor-docs]` markdown link, turns out you can link to the `latest` version on `docs.rs` instead of writing an explicit version number.

https://docs.rs/mdbook/latest/mdbook/preprocess/trait.Preprocessor.html

rbuckland:
@Michael-F-Bryan wrote:
> Ooh, there's an emoji preprocessors? That's really cool!
oops - I didn't mean to include the text yet. Sorry.
Yes - I have completed it - and it is sitting here https://github.com/rbuckland/mdBook/tree/feature/emoji_v2

I will PR that once this change comes through. That is for issue #771

rbuckland:
I have updated that documentation - please take a read - and make sure I got it right :-D

Michael-F-Bryan:
LGTM :+1:

PR #788, opened 2018-09-09
Allow non alphabetic initial in heading anchor

weihanglo:
In HTML5 standard, there is no restriction on what forms an ID attribute. The only reason to append a alphabet in the front of original characters is for compatibility. 

As far as I know, some popular content generating services, including GitHub, GitBook and GitLab, accept a heading with non-alphabetic initial as a valid ID attribute. Maybe we can conform with the standards without appending an unexpected arbitrary character on ID. This feature will make writing much more ergonomic for non-English users.

Reference: https://www.w3.org/TR/html52/dom.html#element-attrdef-global-id

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -329,12 +329,32 @@ more text with spaces
         #[test]
         fn it_generates_anchors() {
             assert_eq!(
-                id_from_content("## `--passes`: add more rustdoc passes"),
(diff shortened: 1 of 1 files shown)


Michael-F-Bryan:
This seems pretty reasonable, I think we were just being conservative during the initial implementation.

PR #791, opened 2018-09-16
Export the mdbook version from the crate root

Michael-F-Bryan:
This allows custom preprocessors and renderers to use the `semver` crate to make sure they're compatible with whatever version of `mdbook` is calling them.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -114,6 +114,12 @@ pub mod renderer;
 pub mod theme;
 pub mod utils;
 
+/// The current version of `mdbook`.
+///
+/// This is provided as a way for custom preprocessors and renderers to do
+/// compatibility checks.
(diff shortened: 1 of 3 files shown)


Michael-F-Bryan:
As a forwards compatibility thing I've also added private `__non_exhaustive` elements to the renderer and preprocessor context structs to makes sure people can't construct or exhaustively destructure them. 

This will break any custom renderers/preprocessors who previously depended on this, but I'd argue it was a `mdbook` bug to allow this in the first place. People should only ever be constructing a context with the `new()` constructor anyway. As far as [I can tell], this would only affect myself, @Byron, and @JoshBrudnak.

[I can tell]: https://crates.io/crates/mdbook/reverse_dependencies

Michael-F-Bryan:
I'm closing this PR in favour of #792, which was already based on top of this branch.

Issue #762, opened 2018-08-07
provide mermaidjs rendering


rbuckland:
MermaidJS provides and excellent set of "features" for diagramming. I would like to provide this feature, similar to MathJAX into mdBook as an optional feature. 

I am after some guidance regarding modularity.

1. how do I cater for renderers such as epub or PDF ?
2. at a high level, which way is it preferred to be integrated. With my quick hacking, I suspect it is with the `hbs_renderer.rs` (adding html.mermaidjs_support) flags for example.

Guidance will be appreciated.

mattico:
I think this could be implemented as a [preprocessor](https://rust-lang-nursery.github.io/mdBook/for_developers/preprocessors.html). It could transform something like this:

    ```mermaidjs
    chart goes here
    ```

into this

    <div class="mermaid">
    chart goes here
    </div>

The mermaidjs script tag could be added with `output.html.additional-js`, or perhaps an additional-js facility could be added onto the preprocessor itself somehow.

badboy:
I built this as a preprocessor: https://github.com/badboy/mdbook-mermaid
Would this be considered to be included in mdbook directly?

Michael-F-Bryan:
I think this would be better written as a custom preprocessor, using a similar mechanism to [custom renderers][1] (e.g. the [`CmdRenderer`]). That way you can provide `mdbook-mermaid` as its own binary on crates.io and it won't be bound to `mdbook`'s release cycle.

... Of course users can't yet provide their own preprocessors for `mdbook` to run, so we'd need to implement. You can wrap `mdbook` like @badboy has done, but that's not as nice as `mdbook` being able to shell out to a `mdbook-mermaid` preprocessor and interact with it via stdin/stdout.

[1]: http://rust-lang-nursery.github.io/mdBook/for_developers/backends.html
[`CmdRenderer`]: https://docs.rs/mdbook/*/mdbook/renderer/struct.CmdRenderer.html

badboy:
That’s exactly the issue I’m facing right now.
I built 2 different preprocessors, but that also means I needed to build a mdbook clone that adds these 2 preprocessors.
Going forward it would be nice to either have a mechanism to use external preprocessors or to include more preprocessors in mdbook itself.

Michael-F-Bryan:
[...]

PR #792, opened 2018-09-16
WIP: Custom Preprocessors

Michael-F-Bryan:
This lets anyone provide their own custom preprocessor by shelling out to the provided command.

CC: #762, @badboy

### DIFF
--- a/book-example/src/for_developers/preprocessors.md
+++ b/book-example/src/for_developers/preprocessors.md
@@ -11,68 +11,71 @@ the book. Possible use cases are:
   mathjax equivalents
 
 
-## Implementing a Preprocessor
+## Hooking Into MDBook
 
-A preprocessor is represented by the `Preprocessor` trait.
+MDBook uses a fairly simple mechanism for discovering third party plugins.
+A new table is added to `book.toml` (e.g. `preprocessor.foo` for the `foo`
+preprocessor) and then `mdbook` will try to invoke the `mdbook-foo` program as
+part of the build process.
 
-```rust
-pub trait Preprocessor {
-    fn name(&self) -> &str;
-    fn run(&self, ctx: &PreprocessorContext, book: Book) -> Result<Book>;
-    fn supports_renderer(&self, _renderer: &str) -> bool {
-        true
-    }
-}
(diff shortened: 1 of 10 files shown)


Michael-F-Bryan:
I think I've got the basic idea implemented, and the example program works exactly as I'd expect. 

This will need a proper review and we may want to bikeshed on the nuances of how information is passed/received from the subcommand though. Plus I still need to create a new page in the user guide.

badboy:
One thing I'd like to see mentioned in the docs is the handling of multiple preprocessors: Are they executed in the specified order?

And another thing: Will the input/output format be documented somewhere? This would make writing preprocessor not based on the mdbook lib possible (e.g. in other languages)?

Michael-F-Bryan:
> Are they executed in the specified order?

In general, I believe it'll run default preprocessors first then all other preprocessors will be run in the order they were defined in `book.toml`. I'm tempted to say the order will be implementation defined, seeing as it's more a side-effect of [how preprocessors are discovered][1] and not due to any explicit plan.

> Will the input/output format be documented somewhere?

At the moment the format is a JSON array where the first element is `PreprocessorContext` and the second element is a `Book`, but the best reference would be the [`CmdPreprocessor::parse_input()`][2] helper function. 

In the long term it should probably be documented, but that would mean we also need to document (and stabilize) the entire internals for `Book` and `Config`. If possible, I'd like to defer that decision until there is a need for it (e.g. someone makes an issue wanting to create a preprocessor or backend in Python).

[1]: https://github.com/rust-lang-nursery/mdBook/blob/b1c7c54108e4eaeeb2a549debdef53db795f8d4e/src/book/mod.rs#L359-L386
[2]: https://github.com/rust-lang-nursery/mdBook/blob/b1c7c54108e4eaeeb2a549debdef53db795f8d4e/src/preprocess/cmd.rs#L44-L51

Michael-F-Bryan:
I think this implementation for custom preprocessors is a good starting point. I've merged it so people can start using it in real-life scenarios (like @badboy's) as of the next release.

badboy:
Thanks!

PR #795, opened 2018-09-18
Fix broken link to API pages

mwilbur:
Old link resulted in 404.  This PR make the link consistent with README.md in the book-example/src directory.


mwilbur:
Whoohoo!  First rust-lang-related contribution :)

PR #796, opened 2018-09-19
Don't strip relative path of additional javascript files

badboy:
Previously, additional JavaScript files inside a directory were
correctly copied (with their parent created), but the link to it was
stripped of that parent.
There's no need for that (and it was not done for CSS)


Dylan-DPC:
Thanks

PR #803, opened 2018-10-14
avoid using `cd` in example travis-ci script

xieyuheng:
I debugged 10 commits to figure out why the following failed
```
  - cd docs && mdbook build
  - cargo doc && cp -r target/doc docs/book/api
```
It should be
```
  - mdbook build docs
  - cargo doc && cp -r target/doc docs/book/api
```


Michael-F-Bryan:
Oops, sorry for the misleading documentation @xieyuheng!

Issue #713, opened 2018-06-20
Allow setting different default theme in book.toml


fitzgen:
I have no desire to write my own custom theme, but I particularly like the "rust" theme that comes bundled, and I would like to set that as the default for my book.

Perhaps something like

```toml
default-theme = "Rust"
```

?

ehuss:
Fixed by #804.

PR #804, opened 2018-10-16
Default theme option

### LINKED ISSUE #95: Select default theme
Hi. Is it possible to select a theme during init?

Bassetts:
Opening a PR to initiate discussion.

Everything is there and working, but there are a few things I dislike/am unsure of.

1. Calling this option `default-theme` is confusing to the user. The options are presented as 'themes' to the visitor in the theme popup, but from a user's perspective the theme is the whole theme inside `src\theme`. Maybe this should be `default-style` or `default-theme-style`? Maybe I just need to explain the option better in the documentation?
2. I created a new handlebars helper to append ` (default)` to the selected default theme in the theme popup. Originally I had done this in the [javascript that sets the theme on page load](https://github.com/rust-lang-nursery/mdBook/blob/396426662de5ff6a9f12b8892ef89eba17b3bba2/src/theme/index.hbs#L57-L61), but it seemed a bit hacky when we know which theme is the default during render. Not sure if this is the best way, or best implementation of a helper though.
3. I have left the ordering of the theme popup as it is, but this means the default may now not be the top item. Should we re-order the list of themes to always have the default as the first item?

Once complete this will resolve #95 and #713

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -214,6 +214,7 @@ impl HtmlHandlebars {
         );
         handlebars.register_helper("previous", Box::new(helpers::navigation::previous));
         handlebars.register_helper("next", Box::new(helpers::navigation::next));
+        handlebars.register_helper("theme_option", Box::new(helpers::theme::theme_option));
     }
 
     /// Copy across any additional CSS and JavaScript files which the book
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -395,6 +395,12 @@ fn make_data(
         data.insert("livereload".to_owned(), json!(livereload));
     }
 
+    let default_theme = match html_config.default_theme {
+        Some(ref theme) => theme,
+        None => "light",
+    };
+    data.insert("default_theme".to_owned(), json!(default_theme));
+
     // Add google analytics tag
     if let Some(ref ga) = config.html_config().and_then(|html| html.google_analytics) {
         data.insert("google_analytics".to_owned(), json!(ga));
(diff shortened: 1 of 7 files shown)


src/renderer/html_handlebars/helpers/theme.rs
```diff
@@ -0,0 +1,30 @@
+use handlebars::{Context, Handlebars, Helper, Output, RenderContext, RenderError};
+
+pub fn theme_option(
```
Michael-F-Bryan: I like the thinking behind using a `theme_option` helper append " (default)", it's much nicer than using javascript to add it after the fact :+1:

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -395,6 +396,12 @@ fn make_data(
         data.insert("livereload".to_owned(), json!(livereload));
     }
 
+    let default_theme = match html_config.default_theme {
```
Michael-F-Bryan: Just out of curiosity, what's going to happen if I set the `default-theme` value in my `book.toml` to something which isn't a theme?

Michael-F-Bryan:
> Calling this option `default-theme` is confusing to the user

I think `default-theme` should be okay. People will probably be more familiar with the drop-down than the `theme/` directory, and someone playing around with `theme/` should understand the difference. Maybe we could update the documentation to mention this only affects the colour schemes selected by default from the themes drop-down, and is totally separate from the `theme/` directory?

src/renderer/html_handlebars/hbs_renderer.rs
Bassetts: It will be accepted and will be used as the class on the body and the value for the global `default_theme` js variable. This will cause the book to not look quite right in certain places, though for the most part the default styling does a good job.

The reason for not validating this further is that someone could create their own theme with an option that the default theme doesn't have, so there is no real way to check the validity other than checking if a value is specified.

Bassetts:
@Michael-F-Bryan realised I forgot to comment. I have updated the documentation to specify this affects the themes drop-down.

Michael-F-Bryan:
Merged :tada: Thanks for all the hard work @Bassetts!

fitzgen:
Awesome! 🎉

PR #809, opened 2018-10-23
Improve documentation comments

hikarinotomadoi:
It is quite confusing that you should assign `true`, when the answer to the documentation comment is actually *"no"*.


Michael-F-Bryan:
I agree, that documentation is pretty confusing. Thanks for making it easier to understand @hikarinotomadoi!

PR #817, opened 2018-11-08
Change default code font-size to 0.8em

StefanieJaeger:
This resolves weird rendering behavior in Chrome and Safari on macOS.
This is most likely because of some floating point rounding error.

---
### Before
screenshot 2018-11-08 at 20 37 45

### After
screenshot 2018-11-08 at 19 18 45


StefanieJaeger:
Closed in favour of #818

Issue #383, opened 2017-07-28
serve: double reload on backup files deletion


dns2utf8:
The problem arises when working with an editor like [Kate](https://kate-editor.org/) which keeps your current edits in a swap file.

When saving the swapped contents to the real file an event occurs and triggers the rebuild of the book.
While the book is building another event occurs because the swap file is deleted. 

On my machine this creates a race. 
Sometimes the second thread is fast enough to delete the folder before the first finishes working and nothing happens.
But around 50% of the time the deletion happens right after the first build finished. This means the browser will receive a reload command at the same time as the second thread starts deleting the files. In that case the browser gets a 404 and the reload function is dead. A manual reload is required now.

I see a probable solution by triggering the rebuild with some delay on the server.
Delaying the client is also possible but the client would either have to poll the server for the file or delay a fixed amount of time.
What do you think?

budziq:
Hi @dns2utf8 thanks for the bug report.
Trying to fix race conditions by sleep tuning is universally a bad idea 😉  . We will most likely need to introduce some synchronization with a command queue

dns2utf8:
Another way to solve this would could be the 404 page. If it had code included for may displaying the index or detect the page beeing served (maybe again). 
This would cover the scenario where the user switched from one branch of the book to another or merges.

azerupi:
Maybe we could add a command line argument to blacklist certain patterns of files, would that help?

dns2utf8:
What if mdbook were to ignore all changes coming from files marked by the `.gitignore` (or similar) file?

budziq:
@dns2utf8 could you look on this PR https://github.com/azerupi/mdBook/pull/427? It solves multiple reloads on my system with vim (using notify crates builtin event aggregation logic which we have circumvented so far). It might not solve your case yet though.

azerupi:
Release 0.0.24 includes the fix from @budziq.
Could anyone try with the new version and see if it fixed the issues?

dns2utf8:
Happy new year everybody. With the latest 0.1.1 release this behavior has become more intrusive. Even though I have this ignore file the reloads get triggerd while I type.
```
book
.DS_Store
.directory
*.kate-swp
```

On my system the actual file does not change while typing, so I get no live preview. The edits go into the `.filename.kate-swp` and I get many rebuilds and reloads without any effect. As soon as I save I get a 404 because in the very moment it builds the original file is removed and then kate moves the swap file to the original location.

[...]

PR #831, opened 2018-11-24
Add -m arg to watch/serve to only watch .md files

9999years:
Sometimes, a change to a garbage file (in my case, Vim swap files) in the source directory will cause a book rebuild, which is unnecessary, particularly in `serve` mode where it triggers a visually noisy page refresh.

This PR adds a `-m` / `--markdown-only` switch to the `watch` and `serve` subcommands, which causes the file-watcher to ignore changes in files that don't end in `.md`.

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -12,6 +12,7 @@ use mdbook::errors::*;
 use mdbook::utils;
 use mdbook::MDBook;
 use std;
+use std::path::Path;
 use {get_book_dir, open};
 
 struct ErrorRecover;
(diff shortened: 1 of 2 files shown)


9999years:
This might close or partially close #383. We could also consider using a user-supplied regex for files to watch.

Michael-F-Bryan:
Instead of using a command-line flag (which you have to remember every time you run `mdbook`) that *only* whitelists markdown files, I'd prefer a more flexible solution where you can provide your own glob patterns to ignore.

For example,

```toml
[build]
# provide a list of globs to ignore
ignored = [
  "*.tmp",
  "*~",
]

# or use your gitignore file (e.g. with https://crates.io/crates/ignore)

[build]
ignored = "gitignore"
```

Also, should this alter *just* the behaviour of file watcher in `mdbook watch` and `mdbook serve`, or could it also be used to prevent things from accidentally being copied to the `book/` output directory?

9999years:
I would be fine with either of those, although I feel like ignoring editor-dependent things like swap files should be an end-user task, not bundled with the repository; I feel the same way about putting `.*.swp` in a `.gitignore`; they belong, in my opinion, in a global `.gitignore`.

I only have a narrow need here: excluding my Vim swap files. Any solution that can accomplish that is fine by me. If you’d like to decide on a more flexible system, I’ll see what I can do to help implement it.

ehuss:
I'm going to close as this is somewhat outdated.  Support for .gitignore has been added, and there is more discussion at #1187 and #1440 to configure what to ignore.  Thanks for the PR, though!

PR #834, opened 2018-11-29
Fixes #826

### LINKED ISSUE #826: Blue hover in sidebar does not work.
In the past, if you hovered over links in the sidebar, they would be highlighted in blue. This does not work currently. I bisected it, and this behavior change occurred somewhere between b3e0942bc957575bab6a47c792396774aad82a3c (old, has highlighting) and 028c8b0f752de275d89f7f64cc8f643437bc8a8e (new, no highlighting). I couldn't bisect further since the commits in between didn't build.

I'm not sure if this change was intentional or not. I believe it was not intentional and is thus a bug because:
- That range of commits seemed to be trying to refactor, and no commit messages in there indicated an intent to change a (fairly major) part of the UI experience.
- Without the highlighting it's pretty hard to tell and unintuitive that the sidebar has links, not just text.

However, I can't be sure that it's not intended behavior.

donald-pinckney:
I re-arranged a few CSS rules to fix the issue I pointed out in #826.


Dylan-DPC:
THanks

PR #837, opened 2018-12-04
Random acts of kindness

basbossink:
Since the CONTRIBUTING.md file suggests running rustfmt and clippy before submitting a pull request this pull request does only that. Since the start state for this pull request changed when running rustfmt and there were some easy to fix clippy warnings. I'm very new to rust so I didn't feel comfortable fixing all clippy warnings.

Question: When I run `cargo test` one test fails on windows but `cargo test`  succeed on the windows subsystem for linux:

```sh
test failing_alternate_backend ... FAILED
test passing_alternate_backend ... ok
test missing_backends_arent_fatal ... ok
test alternate_backend_with_arguments ... ok

failures:

---- failing_alternate_backend stdout ----
thread 'failing_alternate_backend' panicked at 'called `Result::unwrap_err()` on an `Ok` value: ()', libcore\result.rs:1009:5
```
Is that _normal_? The CONTRIBUTING.md does not mention anything about tests and/or how to run them and what to expect maybe it would be nice to add a section about testing. I noticed some tests are disabled on windows e.g. should I run the tests on multiple platforms?

### DIFF
--- a/examples/de-emphasize.rs
+++ b/examples/de-emphasize.rs
@@ -17,6 +17,7 @@ fn main() {
     panic!("This example is intended to be part of a library");
 }
 
+#[allow(dead_code)]
 struct Deemphasize;
 
 impl Preprocessor for Deemphasize {
--- a/examples/de-emphasize.rs
+++ b/examples/de-emphasize.rs
@@ -56,10 +56,7 @@ where
     Ok(())
 }
 
-fn remove_emphasis(
-    num_removed_items: &mut usize,
-    chapter: &mut Chapter,
-) -> Result<String> {
+fn remove_emphasis(num_removed_items: &mut usize, chapter: &mut Chapter) -> Result<String> {
     let mut buf = String::with_capacity(chapter.content.len());
 
     let events = Parser::new(&chapter.content).filter(|e| {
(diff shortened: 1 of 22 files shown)


Michael-F-Bryan:
It may be something to do with how the `false` command is handled on Windows machines. On WSL, the Linux kernel will look for a program named `false` on the `$PATH` and try to execute it, returning us an exit code of 1. I'm not sure how Windows works and whether `false` is an actual program or something built into `cmd.exe` or PowerShell.

A more portable way of doing things would be to copy the `tee_command()` function and replace the `"false"` in `let (md, _temp) = dummy_book_with_backend("failing", "false")` with a function that returns an OS-specific command that will always fail.

```rust
fn fail_cmd() -> &'static str {
  if cfg!(windows) {
    r#"cmd.exe /c "exit 1""#
  } else {
    "false"
  }
}

fn success_cmd() -> &'static str {
  if cfg!(windows) {
    r#"cmd.exe /c "exit 0""#
  } else {
    "true"
  }
}
```

basbossink:
I've updated this pull request with the suggestion and code from @Michael-F-Bryan, thanks for the suggestion 🙏.

PR #841, opened 2018-12-05
Fix issue 832

### LINKED ISSUE #832: Existing SUMMARY.md not taken into account
I'm trying to [bootstrap a book project](https://github.com/basbossink/bisd) however when running:
```sh
mdbook init 
```
in the root directory of the clone of the above mentioned repository the `SUMMARY.md` is overwritten and no additional files appart from the empty `chapter_1.md` are created in the `src` directory. While the documentation states: 

> Tip: Generate chapters from SUMMARY.md
> When a SUMMARY.md file already exists, the init command will first parse it and generate the missing files according to the paths used in the SUMMARY.md. This allows you to think and create the whole  structure of your book and then let mdBook generate it for you.

I've tried this on both Windows and the Windows subsystem for Linux. For the later the issue is also confirmed on a debug build of `master` with `005dfc55b` as the `HEAD`.

basbossink:
After a very brief inspection of `src/book/init.rs` I came to the conclusion it was not checking for an existing `SUMMARY.md` this PR adds an `if` and some tracing to check for an existing `SUMMARY.md` and only generate the stub files if no `SUMMARY.md` is currently present. 

In my humble opinion this fixes #832.


ehuss:
This looks good, though I think it would be good to have a test in [tests/init.rs](https://github.com/rust-lang-nursery/mdBook/blob/8b2e1c2daad74058dda408ae2aa9d5c8f6cea5da/tests/init.rs).

basbossink:
I've added the requested test.

PR #842, opened 2018-12-06
Resolve symlinks when copying files

rinon:
While symlinked directories may cause infinite recursion when copying files,
symlinked files shouldn't be a problem. This change copies the resolved
destination of the symlink in place of the symlink file itself, which will
preserve any references to the symlinked file in the rendered output.

### DIFF
--- a/src/utils/fs.rs
+++ b/src/utils/fs.rs
@@ -144,7 +144,13 @@ pub fn copy_files_except_ext(
                 true,
                 ext_blacklist,
             )?;
-        } else if metadata.is_file() {
+
+        // DirEntry.metadata() does not read through symlinks, so we have to
+        // use fs::metadata().
+        } else if fs::metadata(entry.path())?.is_file() {
+            // resolve symlinks when copying files
+            let realpath = fs::canonicalize(entry.path())?;
+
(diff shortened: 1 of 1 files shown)


rinon:
Is this change something people want? We're using this to build a manual with sections symlinked from their existing locations in the repository. Seems like this would be a useful thing for others as well (plus I'd love to not need custom tooling to build our manual :smiley:).

rinon:
ping?

sytsereitsma:
I solved this differently by adding an additional resources section to book.toml (nearly ready for filing a pull request, struggling with the serve command on this). The user can add zero or more sections like the following:

```toml
[[output.html.additional-resources]]
output-dir="img" # Relative to book output dir
src="../plantuml/*.png" # Relative to working dir/book.toml dir
```
It allows more granular control over the files you wish to import from external directories and it can be committed without having to check in symlinks.

Would that work for your case too?

Regards,

Sytse

rinon:
@sytsereitsma Unfortunately I don't think that would help me. I'm actually using symlinks to the markdown files themselves, not just HTML resources.

ehuss:
I think this is now resolved via #1323.

Issue #813, opened 2018-11-04
GitLab guide for continuous integration


kravemir:
I'm considering to use mdBook for documentation of my project hosted on Gitlab, in combination with Gitlab Runner and Gitlab Pages.

In [continuous-integration.md](https://github.com/rust-lang-nursery/mdBook/blob/877bf37d18fb4ab533927b7980ef3794cfc9313b/book-example/src/continuous-integration.md) there's example only for Travis CI and GitHub Pages.

It would be nice to have also guide showing how to setup CI with Gitlab (Runner, Pages).

Michael-F-Bryan:
Translating the Travis example to `gitlab-ci.yml` should be pretty straightforward. We actually use it at work for some internal documentation and the only pain point I had was getting caching right. 

I'll see if I can find my `gitlab-ci.yml` and add a simple GitLab CI example to the user guide.

kravemir:
So far, I had used following setting, which I had found somewhere else:

```
pages:
  # https://hub.docker.com/_/rust/
  image: "rust:1.30.1-stretch"
  stage: pages
  script:
    - mkdir public-tmp
    - cargo install mdbook --no-default-features
    - ./bin/copyJavadocToBook.sh
    - mdbook build docs
    - cp -r docs/book/* public-tmp
    - mv public-tmp public
    - find $PWD/public
  artifacts:
    paths:
      - public
```

I'll merge current with your example to get best results.

kravemir:
@Michael-F-Bryan, can you please take a look at #843?

IBISLabUMD:
This successfully worked for me: https://gitlab.com/nmg/ngspice-manual/raw/master/.gitlab-ci.yml

PR #843, opened 2018-12-08
Build Docker image

kravemir:
Useful for GitLab CI. Compared to job installing `mdBook`, runtime decreased from `8m 30s` to `1m 18s`, which saves lots of time.

For example, can be used with job declaration mentioned in #813:

```yaml
pages:
  # https://hub.docker.com/_/rust/
  image: "kravemir/mdbook:0.2.1"
  stage: pages
  script:
    - mkdir public-tmp
    - ./bin/copyJavadocToBook.sh
    - mdbook build docs
    - cp -r docs/book/* public-tmp
    - mv public-tmp public
    - find $PWD/public
  artifacts:
    paths:
      - public
```

I had built it for myself and pushed to https://hub.docker.com/r/kravemir/mdbook/, with command:

```bash
./bin/releaseDocker.sh kravemir 0.2.1
```

IMHO: there's lots of space for improvements, but this provides at least initial working image, which can be used to build mdBook GitLab Pages with GitLab Runner quite easily.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,5 @@
 FROM rust:1.30.1-stretch
 
-RUN cargo install mdbook --no-default-features
+ARG VERSION
+
+RUN cargo install --vers "$VERSION" mdbook --no-default-features
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,3 @@
+FROM rust:1.30.1-stretch
+
+RUN cargo install mdbook --no-default-features
--- a/bin/releaseDocker.sh
+++ b/bin/releaseDocker.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+# USAGE: ./bin/releaseDocker.sh <ACCOUNT_NAME> <VERSION>
+#
+# Example usage: ./bin/releaseDocker.sh dockeruser 0.2.1
+
+set -e
+
+SCRIPT_DIRECTORY="$(dirname "$0")"
+PROJECT_DIRECTORY="$(dirname "${SCRIPT_DIRECTORY}")"
+
+cd "${PROJECT_DIRECTORY}"
+
+docker build --no-cache  -t mdbook .
+docker tag mdbook "${1}/mdbook:${2}"
+
+echo "To run image:"
+echo docker run -it mdbook sh
+
+echo "To push (publish) image"
+echo docker push "${1}/mdbook:${2}"
+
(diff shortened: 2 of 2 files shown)


anisimovsergey:
I realy like the idea of using mdbook via Docker but the resulting image is quite big, aroung 1.6 Gb. Would it be better to use something like this:

```Dockerfile
FROM ekidd/rust-musl-builder AS builder

ARG VERSION

RUN cargo install mdbook --vers ${VERSION} --no-default-features

FROM alpine:latest

COPY --from=builder \
    /usr/local/bin/mdbook \
    /usr/local/bin/

CMD /usr/local/bin/mdbook
```

The resulting image is around 17Mb in this case.

Dylan-DPC:
@ehuss thoughts on this?

ehuss:
I'm not sure I understand the motivation.  It seems to me that it would be much easier to use a pre-built binary.  It should download and be usable much faster than getting a docker image.  I'm not familiar with GitLab Runner, but some more context and motivation would be helpful.

moikot:
The image produced is quite large - 557 Mb. Here is a [Dockerfile](https://github.com/moikot/mdbook/blob/master/Dockerfile) which is producing [an image](https://hub.docker.com/r/moikot/mdbook/tags) which is 6.65 Mb only.

ehuss:
I'm going to close this since there hasn't been much activity for a while (and the 500Mb image size sounds worrisome).  I recommend people to just download the pre-built binaries offered here on GitHub if possible.

AndreaGiardini:
@moikot I think it would be nice to have an official docker image released by the mdbook project.
This is extremely useful when running in CI (Gitlab, but not only).
Would you be willing to merge the Dockerfile into the main mdbook repo?

@ehuss what is your opinion about releasing a docker image together with the offical mdbook relases?

moikot:
@AndreaGiardini I can prepare a PR for the Dockerfile, it's not a problem.

ehuss:
@AndreaGiardini can you describe the use case or motivation for a docker image as opposed to just downloading a pre-built executable?  It seems like the latter should be much more lightweight. I would prefer not to add more maintenance burden to the release process.

AndreaGiardini:
Hi @ehuss 
Docker images make it much easier to run in container-based environments (e.g. Kubernetes) compared to the classic executable. Especially for CI/CD this could be a great addition.
If you are already using github actions setting up a pipeline for publishing a docker image at every release it's not that difficult and requires minimal effort. 

The advantage would be also to have one official image rather than dozens of unofficial ones ( https://hub.docker.com/search?q=mdbook&type=image ). As you can see from the number of people creating custom mdbook docker images, the request for something like that is definitely there!

PR #846, opened 2018-12-10
Move hiding of boring lines into static content

adrianheine:
This saves a lot of useless client-side computation and makes the output look nicer when no client-side JS is available.


Dylan-DPC:
Thanks. Sorry for the long delay in merging this

PR #857, opened 2018-12-20
src/theme/book.js: switch to "/execute" json interface for playpen

haraldh:
Somehow "/evaluate.json" gives a very different output than expected and
even errors out on some code.

Seems like an old compiler version is used there.

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -103,18 +103,19 @@ function playpen_text(playpen) {
         let text = playpen_text(code_block);
 
         var params = {
-            version: "stable",
-            optimize: "0",
+            channel: "stable",
+            mode: "debug",
+            edition: "2018",
             code: text
         };
 
         if (text.indexOf("#![feature") !== -1) {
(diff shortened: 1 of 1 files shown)


haraldh:
Example:
https://github.com/haraldh/chainerror

Rendered:
https://haraldh.github.io/chainerror/

haraldh:
Example of code not running with evaluate.json:
https://haraldh.github.io/chainerror/tutorial2.html

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c04674c7bf653aea6b34e4e31346622f

ehuss:
Thanks for the PR!  The link in the comment above seems to be working, so I'm not sure what was wrong at the time this was posted.  mdBook cannot use the other endpoint.  I have updated `/evalute.json` to support the edition and update mdBook in #946, if that was maybe the issue.

haraldh:
Yeah.. fixed it for my web site. This is why it worked for you

PR #860, opened 2018-12-25
Fix duplication with Cargo.toml

k-nasa:
If there is an overlap, I think that it will happen that the number of changes will increase, or forget to change either.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ authors = [
     "Michael-F-Bryan <michaelfbryan@gmail.com>",
     "Matt Ickstadt <mattico8@gmail.com>"
 ]
-description = "Create books from markdown files"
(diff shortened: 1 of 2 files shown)


Oslint:
I think we should write `Creates a book from markdown files`, the last written description in 'Cargo.toml'.
<https://github.com/rust-lang-nursery/mdBook/blob/b8f8e768998dbca1fb29ee09f8a1f737297556d3/src/main.rs#L31> was updated on 3 Aug 2018. <https://github.com/rust-lang-nursery/mdBook/blob/d48a27f94f03bb57990ed82f54fbe81fed081291/Cargo.toml#L5> was updated on 15 Mar 2018.

Dylan-DPC:
Thanks

PR #867, opened 2019-01-17
Fixed loss of focus when clicking the Copy button

ffissore:
When the Copy button is pressed, focus is lost and scrolling the page with the arrow keys stops working. Fixed by giving focus back to the `trigger` object (clipboard.js provided) which is the copy button itself


ffissore:
Steps to reproduce:
1. open the example book, open a page with a code block
2. try scrolling using your keyboard arrow keys, verify it works
3. click on any "copy" button on the right side of a code block
4. try again to scroll using the arrow keys, verify it's not working any more

ehuss:
This seems to work.

An alternative is to update the version of clipboard.js to the latest (2.0.4) which already has this fix (https://github.com/zenorocha/clipboard.js/pull/419).  AFAICT, the only change necessary to switch to 2.0 is to use the constructor `ClipboardJS` instead of `Clipboard` (details are in https://github.com/zenorocha/clipboard.js/releases).

ffissore:
Thanks for the hint. Delegating the fix to ClipboardJS makes much more sense. I've updated the PR accordingly

Dylan-DPC:
Thanks

PR #870, opened 2019-01-31
Group file changes and rebuild book only once

### LINKED ISSUE #679: Serve rebuilds for every file individually when a lot of files change
When a lot of source files change and we were running `mdbook serve`, it will rebuild the book for each file that changed. This is particularly painful when using git and changing branches. During those rebuilds, the served book is broken.

Simply quitting `mdbook` and relaunching the command solves the problem, but it is an annoying paper-cut.

cauebs:
Closes #679.

Any and all feedback is appreciated.


Dylan-DPC:
Thanks

PR #872, opened 2019-02-03
Fix deploy on Travis.

ehuss:
The deploy scripts require TARGET env var to be set, but it wasn't set anywhere, causing it to fail: https://travis-ci.com/rust-lang-nursery/mdBook/builds/97850452.

This also reduces the number of mac builders, since they aren't necessary, and capacity is limited.

The api key will likely need to be updated, too, since the transition to travis.com.

Appveyor also seems to be broken, but I suspect it is also a key issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,19 @@
 language: rust
 
-rust:
-  - stable
-  - beta
-  - nightly
+matrix:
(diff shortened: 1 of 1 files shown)


Dylan-DPC:
Thanks

PR #878, opened 2019-02-16
Cleanup build directory before preprocessors run to keep files genera…

shen390s:
I want to use preprocessor to executable some special code block to generate some diagram(such as using dot to generate diagrams for code block written to dot language), so I need to keep the generated files by preprocessors.


Dylan-DPC:
Thanks

PR #883, opened 2019-03-04
Expose API for building a book with a custom Summary.

anp:
This is useful for situations where you'd like to supplement or replace the existing Summary parsing with
custom filesystem traversal code or other similar changes.

In this context, a Summary is almost a "build rules" input for the book, so that an embedder can use their own layout while still benefiting from the actual file reading/parsing/etc provided by mdbook.


Dylan-DPC:
Thanks

PR #889, opened 2019-03-20
new "Book" theme

s3bk:
The idea is to add a theme that mimics traditional typesetting for books.

Preview:
[Rust Reference](https://rynx.org/sebk/rust-reference)
[Rust Book](https://rynx.org/sebk/rust-book)

### DIFF
--- a/src/theme/css/book.css
+++ b/src/theme/css/book.css
@@ -56,7 +56,7 @@ h4::before {
 }
 
 @media print {
-    .bottons {
+    .menu-bar, .buttons {
         display: none;
     }
     body {
--- a/src/theme/css/book.css
+++ b/src/theme/css/book.css
@@ -7,7 +7,16 @@ p {
 }
 code {
     font-family: "Source Code Pro", monospace;
+    padding: 0 !important;
+}
+code:not(.block) {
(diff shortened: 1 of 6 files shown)


src/theme/css/book.css
```diff
@@ -0,0 +1,65 @@
...
+    position: absolute;
+    right: 0;
+}
+    
+#menu-bar-sticky-container {
+    background-color: rgba(255, 255, 255, 0.8) !important;
+}
+
+@media print {
+    .bottons {
```
Havvy: `.buttons`?

s3bk: oops.

Dylan-DPC:
Thanks

ehuss:
This PR has broken formatting for all of the themes.  All fonts have changed to `Source Serif Pro`.  This also makes the font story more difficult by adding more fonts.

I can't really tell what the intention was, but I don't think the included `book.css` is a good idea.  It should probably use variables like all other themes.

Dylan-DPC:
Reverting this

Dylan-DPC:
Update: can't revert it because GH considers it might be changed.

s3bk:
Oops. I will try to look into that today.

PR #891, opened 2019-03-23
Write preprocessed content to file before testing with rustdoc

### LINKED ISSUE #855: Expected `include` to be processed before `mdbook test`
## What happened:

I'm experimenting with extracting code examples from TRPL into their own files for a variety of reasons. However, I'd like to retain the `mdbook test` functionality.

[I extracted an example that compiles and currently passes `mdbook test` into its own file in this commit](https://github.com/rust-lang/book/commit/99817d9078f1669cfcbbe51e40bb4d163923d35c).

I then ran `mdbook test`, and got this error (`mdbook --version` = v0.1.7):

```
failures:

---- /var/folders/qc/0ry2bjtd7wd9q8r1c7wylplr0000gn/T/mdbookmFaM3A/ch03-01-variables-and-mutability.md - _::Variables_and_Mutability (line 74) stdout ----
error: expected `[`, found `include`
 --> /var/folders/qc/0ry2bjtd7wd9q8r1c7wylplr0000gn/T/mdbookmFaM3A/ch03-01-variables-and-mutability.md:75:4
  |
3 | {{#include ../examples/ch03-01-variables-and-mutability/unnumbered-02/src/main.rs }}
  |    ^^^^^^^ expected `[`

thread '/var/folders/qc/0ry2bjtd7wd9q8r1c7wylplr0000gn/T/mdbookmFaM3A/ch03-01-variables-and-mutability.md - _::Variables_and_Mutability (line 74)' panicked at 'couldn't compile the test', librustdoc/test.rs:332:13
```

I just tried with mdbook 0.2.2 and got the same result. I'm using Rust 1.31.0.

## What I expected to happen:

I expected the `#include` preprocessing to happen before running doctests so that `mdbook test` would have the same behavior as it did before I extracted the file. It looks like that's not happening because rustdoc is trying to compile `{{#include`.

carols10cents:
Rather than writing the unpreprocessed content to the file for rustdoc to test with.

Fixes #855.

I was trying to figure out whether the current behavior was intentional or not, and if I might be breaking anything with this change, but I think the fix I have here is what was intended-- my first hint was [this comment](https://github.com/rust-lang-nursery/mdBook/blob/c068703028d19fb6c5a4690e904ae6f2bd74292f/src/book/mod.rs#L248) that described what I had hoped was happening but that wasn't what was actually happening.

As of [this commit](https://github.com/rust-lang-nursery/mdBook/commit/ddf31dcc086ffc45defebfd30aa0de8b318f7348), it seems like the behavior I was hoping for was happening and was intended ("now `mdbook test` does full link expansion to temp file prior to running" in the commit message), but [this line got lost in refactoring](https://github.com/rust-lang-nursery/mdBook/commit/cad76a9f6c#diff-dbfdf1ed62f03d8bae38d1b1177880c9L226) and [the preprocessors were being run](https://github.com/rust-lang-nursery/mdBook/commit/cad76a9f6c#diff-dbfdf1ed62f03d8bae38d1b1177880c9R222) but [this line](https://github.com/rust-lang-nursery/mdBook/commit/cad76a9f6c#diff-dbfdf1ed62f03d8bae38d1b1177880c9R229) read the unprocessed file from disk again and gave that content to rustdoc :(

If this fails CI I will fix it tomorrow ❤️

### DIFF
--- a/src/book/mod.rs
+++ b/src/book/mod.rs
@@ -242,13 +242,12 @@ impl MDBook {
             if let BookItem::Chapter(ref ch) = *item {
                 if !ch.path.as_os_str().is_empty() {
                     let path = self.source_dir().join(&ch.path);
-                    let content = utils::fs::file_to_string(&path)?;
                     info!("Testing file: {:?}", path);
 
                     // write preprocessed file to tempdir
                     let path = temp_dir.path().join(&ch.path);
                     let mut tmpf = utils::fs::create_file(&path)?;
-                    tmpf.write_all(content.as_bytes())?;
+                    tmpf.write_all(ch.content.as_bytes())?;
 
                     let output = Command::new("rustdoc")
                         .arg(&path)
--- a/tests/dummy_book/mod.rs
+++ b/tests/dummy_book/mod.rs
@@ -58,8 +58,11 @@ impl DummyBook {
         })?;
 
         let sub_pattern = if self.passing_test { "true" } else { "false" };
-        let file_containing_test = temp.path().join("src/first/nested.md");
-        replace_pattern_in_file(&file_containing_test, "$TEST_STATUS", sub_pattern)?;
+        let files_containing_tests = ["src/first/nested.md", "src/first/nested-test.rs"];
+        for file in &files_containing_tests {
+            let path_containing_tests = temp.path().join(file);
+            replace_pattern_in_file(&path_containing_tests, "$TEST_STATUS", sub_pattern)?;
+        }
 
         Ok(temp)
     }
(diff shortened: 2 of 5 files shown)


carols10cents:
Wooo ok it passed! I was second guessing myself about the tests after submitting this, so in case anyone has the same thoughts I did: it would appear that the `mdbook_can_correctly_test_a_passing_book` would pass even if `nested-test.rs` wasn't included, but it fails without the included code change because rustdoc can't parse `{{#include...` as Rust code.

The `mdbook_detects_book_with_failing_tests` would be better if it tested that there were 2 failures rather than just that the tests failed, but I'm not sure how to do that.

ehuss:
Thanks!  Sorry I lost track of this.

PR #892, opened 2019-03-23
Fix deprecation warnings for trim left/right matches

carols10cents:
There's still a deprecation warning for `Error::source` that [comes from error-chain](https://github.com/rust-lang-nursery/error-chain/pull/255), but this fixes warnings originating in mdBook's code, at least.


Dylan-DPC:
Thanks

PR #894, opened 2019-03-28
Use only relative font sizes

### LINKED ISSUE #203: Don't use "px" as font size unit
Displays have different resolutions, so it is not recommended using _px_ as font-size unit. It is better to use _em_ or _%_.
i looked at the new book version of [Rust Book](http://rust-lang.github.io/book/) and I could find one font-size declaration using _px_:

```CSS
.menu-title {
  display: inline-block;
  font-weight: 200;
  font-size: 20px;
  //...
}
```
Maybe this was made on purpose, maybe not. I'm not really the web-dev guy. :)

zbrox:
This replaces the only use of `px` for `font-size` by setting up a base
`rem` size on the root element in a way that is easy to calculate (`1 rem =
10px`) and scaling up according to browser settings.

It should address #203 

I have generated the book example with the changes and without and they seem to render exactly the same.

Fixes #203


weihanglo:
@zbrox Thanks!

@Dylan-DPC I've tested TRPL and it looks just the same. Shall we test more books?

Dylan-DPC:
@weihanglo nah. This should be fine.

PR #901, opened 2019-04-25
Refactor hbs renderer

sytsereitsma:
Some work  in issue #458, mainly cleaned up the render_item method of HtmlHandlebars
Added some unit tests for the parts that could be easily tested.

Note that these are the first lines of rust I've ever written, so any suggestions for improvement are welcome.


ehuss:
I'm going to close this as it is quite old. Substantial rework of the HTML renderer was done in https://github.com/rust-lang/mdBook/pull/2844. Thanks for the PR, though!

Issue #902, opened 2019-04-28
Library depends on `same-file`


FreeMasen:
Currently even the library associated with this project will require that it is compiled on either unix-like or windows target because of a transitive dependency. `same-file` is a dependency of `walkdir` which has 2 architecture dependent implementations of it's functionality. While the direct dependency on `walkdir` is in the `dev-dependencies` section, which would avoid any issues for users of the mdbook library, it happens to be a dependency of `handlebars` as well. The `handlebars` crate has a feature gate to remove this depenency (`no_dir_source`) which would remove this problem.  

According to the unit tests adopting this feature shouldn't break anything.

Dylan-DPC:
Fixed by PR linked (#903). Thanks

PR #903, opened 2019-04-28
remove walkdir from lib via handlebars

FreeMasen:
To address #902 

This would allow users of the library to compile on unknown architectures.


Dylan-DPC:
Thanks

PR #920, opened 2019-05-13
new "Book" theme (second attempt)

s3bk:
the same as #889, but this time it _should_ not mess up the fonts.
@Dylan-DPC, @ehuss can you confirm this?


Dylan-DPC:
leaving this to @ehuss

s3bk:
@ehuss wrote:
> This PR has broken formatting for all of the themes. All fonts have changed to Source Serif Pro. This also makes the font story more difficult by adding more fonts.

> I can't really tell what the intention was, but I don't think the included book.css is a good idea. It should probably use variables like all other themes.

The Intention was to make it easier to modify the book theme, by keeping in its own CSS file. That way no rules from the other styles *should* leak over, nor the other way around. For some reason last time, I didn't commit the code that disables the CSS file unless the Book theme is selected.

s3bk:
uhh. I think I forgot how to git. This PR makes no sense to me on github.

Dylan-DPC:
Uggh i think you were on the right track.

PR #912, opened 2019-05-06
Automatic deploy to gh-pages and gh release

lzutao:
You need to set up [GitHub person access token](https://github.com/settings/tokens)
image
then save it in [Travis settings](https://travis-ci.org/m4b/goblin/settings) under the name `GITHUB_OAUTH_TOKEN`.

The old key should be revoked or reused if needed.

Demo is at:
* https://github.com/lzutao/mdBook/releases/tag/v0.2.4
* https://lzutao.github.io/mdBook/

r? @Dylan-DPC 

Closes #904 .

PR #923, opened 2019-05-15
Fix automatic deploy.

ehuss:
This should fix deploying release artifacts and gh-pages for the documentation.
Secrets should now be configured in the UI for travis and appveyor.

This also removes some of the appveyor jobs, since they aren't really
necessary, and there are limited jobs on rust-lang-libs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,18 +51,26 @@ script:
     fi
 
 before_deploy:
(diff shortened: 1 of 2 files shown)


lzutao:
cc #912 , but I think I will close that one in favor of this one.

PR #933, opened 2019-05-24
Transition to 2018 edition

lzutao:
We were unaware that mdBook isn't in 2018 edition before.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -121,10 +121,10 @@ pub mod utils;
 /// compatibility checks.
 pub const MDBOOK_VERSION: &str = env!("CARGO_PKG_VERSION");
 
-pub use book::BookItem;
-pub use book::MDBook;
-pub use config::Config;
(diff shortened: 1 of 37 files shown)


src/lib.rs
```diff
@@ -85,24 +85,14 @@
...
 #[macro_use]
 extern crate lazy_static;
 #[macro_use]
 extern crate log;
-extern crate memchr;
-extern crate pulldown_cmark;
-extern crate regex;
-extern crate serde;
 #[macro_use]
 extern crate serde_derive;
```
Dylan-DPC: you can use macro imports and get rid of extern crates for these as well

lzutao: Thanks. But I think we should land it as is. Because we already transit
new edition. Also because I don't want to deal with missing macro importing
somewhere while trying to get rid of those `extern crate`.

Dylan-DPC: cool no issues

PR #936, opened 2019-05-25
Exclude unnecessary items from package

lzutao:
r? @Dylan-DPC

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,16 +8,13 @@ authors = [
 ]
 documentation = "http://rust-lang-nursery.github.io/mdBook/index.html"
 edition = "2018"
-exclude = [
-    "/book-example/*",
-    "/ci/*",
-    "/examples/*",
-    "/tests/*",
-    "/.travis.yml",
-    "/appveyor.yml",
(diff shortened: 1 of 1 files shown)


ehuss:
Is there a reason to change this?  It only saves about 3% of the package size.  It can be helpful to have these extra files (for example, crater can run tests against it).

lzutao:
The only reason I thought is that dependencies of mdbook don't need those 
exluded files.

I don't know that crater depend on tests of package. I thought crater cloned
those package and tested crates. If this is right, I think rust team should state
that somewhere, and this PR will not need to be opened.

lzutao:
Ping again. I talk a bit with infra team (pietro) and they think that
the crater stuff could be ignored.

ehuss:
It's still not clear to me why this needs to change.  It only saves a few % of the package size.  What benefit does it have?

Dylan-DPC:
@lzutao any updates on this

lzutao:
I think I don't have strong opinions on this. Also @ehuss didn't want this. 
In this case, I would close this.

PR #941, opened 2019-05-30
Change language attribute of the book to configurable 

rnitta:
Currently the language attribute of the book, which is used like `<html lang={{ language }}>`, is hardcoded as fixed value `en`.
I want to use mdBook in Japanese, so make it configurable.

With these changes, we will be able to set the language of the book like:

`book.toml`

```toml
[book]
title = "the title of the book"
authors = ["rnitta"]
src = "src"
language = "ja"
```

The default value of the language is still `en`, and no changes of `book.toml` is required if you want to set language `en` as same as usual.

### DIFF
--- a/book-example/book.toml
+++ b/book-example/book.toml
@@ -2,6 +2,7 @@
 title = "mdBook Documentation"
 description = "Create book from markdown files. Like Gitbook but implemented in Rust"
 authors = ["Mathieu David", "Michael-F-Bryan"]
+language = "en"
 
 [output.html]
 mathjax-support = true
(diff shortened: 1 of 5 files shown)


zzy:
> The default value of the language is still `en`, and no changes of `book.toml` is required if you want to set language `en` as same as usual.

If you init project with `--theme`, I think it's convenient to set the lang value `<html lang="zh" class="sidebar-visible no-js">` in index.hbx.

rnitta:
Thanks for comment,
I know that way to set html lang attr with modify index.hbs,
but not all of mdbook users may initialize with `--theme` option and
`BookConfig.language` is currently used only for `<html lang="{{ language }}" class="sidebar-visible no-js">`  in `index.hbs` by defaults, so making it configurable will not cause problem. It may increase usability. Maybe. I think.

Thoughts?

Dylan-DPC:
@ehuss thoughts on this? looks fine to me

ehuss:
Yep!  Thanks @rnitta!

PR #945, opened 2019-05-30
Update toml.

### LINKED ISSUE #938: Update toml and toml-query dependencies
Block on matthiasbeyer/toml-query#43

ehuss:
Just keeping up-to-date.

Closes #938.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -129,10 +129,6 @@ pub mod errors {
             SerdeJson(serde_json::Error) #[doc = "JSON conversion failed"];
         }
 
-        links {
-            TomlQuery(toml_query::error::Error, toml_query::error::ErrorKind) #[doc = "A TomlQuery error"];
-        }
-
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -151,6 +147,12 @@ pub mod errors {
                 description("Reserved Filename")
                 display("{} is reserved for internal use", filename.display())
             }
+
+            /// Error with a TOML file.
+            TomlQueryError(inner: ::toml_query::error::Error) {
```
lzutao: We don't need leading :: before crate name in 2018 edition.

lzutao:
This should close #938 .

src/lib.rs
ehuss: Oh, yea, I was just copy/pasting, and keeping the style. I noticed all the leading `::` and thought it was odd. I'll update after #944.

PR #948, opened 2019-06-03
Header elements wrap links

### LINKED ISSUE #896: Header links’ whitespace cause accidental activation
When you click to the right of a header it’s still considered part of the link and it scrolls you to that header.

There’s little value to having an overflowing link like that, most of those clicks are likely accidental clicks. It personally happens to me a lot, both on desktop and on mobile. It’s a constant source of annoyance.

The solution is to put the &lt;a&gt; element inside the &lt;h2&gt;/&lt;h3&gt; element instead of the opposite.

rnitta:
Fixes #896

now: 
> Jun-03-2019 15-48-34

with this change:
> Jun-03-2019 15-48-58


Dylan-DPC:
Thanks

Issue #712, opened 2018-06-20
Copy button copies incorrect code to clipboard 


mtn:
This issue originally came up [here](https://github.com/intermezzOS/book/issues/191).

It seems like in some cases, the "Copy to clipboard button" copies incorrect code to the clipboard. For example (from that issue), we get this:

```rust

#![allow(unused_variables)]
#![feature(panic_implementation)]
#![feature(core_intrinsics)]
#![no_std]
#![no_main]

fn main() {
use core::intrinsics;
use core::panic::PanicInfo;

#[panic_implementation]
#[no_mangle]
fn panic(_info: &PanicInfo) -> ! {
    unsafe { intrinsics::abort() }
}

#[no_mangle]
pub fn _start() -> ! {
    loop {}
}
}
```

where we'd expect this:

```rust
#![feature(panic_implementation)]
#![feature(core_intrinsics)]
#![no_std]
#![no_main]

use core::intrinsics;
use core::panic::PanicInfo;

#[panic_implementation]
#[no_mangle]
fn panic(_info: &PanicInfo) -> ! {
    unsafe { intrinsics::abort() }
}

#[no_mangle]
pub fn _start() -> ! {
    loop {}
}
```

iolivia:
would it be feasible to add a configuration option on the mdbook level to control whether the main gets added or not?

yoyomo:
fixed it here https://github.com/rust-lang/mdBook/pull/1911#issue-1409974442

Issue #618, opened 2018-02-12
Partial include should show full file instead of stub when showing hidden lines.


shingtaklam1324:
Currently, when parts of the code is included, the default stub is included.

Michael-F-Bryan:
Would you be able to show us an example of what you see versus what you expected to happen?

carols10cents:
Ok I think I understand this one because I think I'm having the same issue.

When I have an external file that, say, looks like this:

```
struct User {
    username: String,
    email: String,
    sign_in_count: u64,
    active: bool,
}

fn main() {
    let user1 = User {
        email: String::from("someone@example.com"),
        username: String::from("someusername123"),
        active: true,
        sign_in_count: 1,
    };
}
```

And I include part of the file like this:

``````
```rust
{{#include ../examples/ch05-using-structs-to-structure-related-data/listing-05-02/src/main.rs:9:14}}
```
``````

I correctly see the part of the file rendered:

<img width="764" alt="Screen Shot 2019-04-28 at 3 36 15 PM" src="https://user-images.githubusercontent.com/193874/56869198-709b1500-69cb-11e9-9c71-ef5afa16853e.png">

But when I click the "show hidden lines" link, I see this:

<img width="769" alt="Screen Shot 2019-04-28 at 3 37 32 PM" src="https://user-images.githubusercontent.com/193874/56869215-9b856900-69cb-11e9-989f-9933f5dd610e.png">

The example has been wrapped in `main` and the `allow(unused_variables)` has been added. I think that's what @shingtaklam1324 means by "default stub".

What would be nice is if clicking the "show hidden lines" link expanded to show the rest of the file instead, and only wrapping in `main` if the rest of the file didn't have `main`.

I'm working on extracting TRPL's examples into external files, and this would be really nice to have. I'll probably give this a try soon.

PR #955, opened 2019-06-14
stop auto-wrapping code snippets in fn main()

mikemorris:
I'm guessing this would be a semver breaking change, but would be happy to do the accompanying work of migrating code snippets in https://github.com/rust-lang/book to include `fn main() {}` wrapping explicitly (and `#![allow(unused_variables)]` where necessary) if this change is accepted (and the book migration could actually happen first because of the current `text.contains("fn main")` exclusion).

This seems to be a somewhat persistent issue in code snippet formatting in the Rust book, particularly for `src/lib.rs` files, with [some awkward fixes](https://github.com/rust-lang/book/pull/1339) (see expanded Listing 11-12 in https://doc.rust-lang.org/stable/book/ch11-03-test-organization.html for implications of this approach, extraneous code still added) without a particularly good solution (using the `ignore` class seems [undesirable](https://github.com/rust-lang-nursery/mdBook/issues/712#issuecomment-398902592)).

An alternative implementation could be adding support for a class like `nomain` or `nowrap` instead of changing this default behavior.

Refs #222 for initial implementation, #283 (`quick_main!` macro), #618, #712, https://github.com/rust-lang/book/issues/1206, https://github.com/rust-lang/book/issues/1874 (`editable` prevents `fn main()` wrapping), https://github.com/rust-lang/book/issues/1982

Examples of problematic code snippets in the Rust book:
- https://doc.rust-lang.org/stable/book/ch11-03-test-organization.html
- https://doc.rust-lang.org/stable/book/ch15-05-interior-mutability.html#a-use-case-for-interior-mutability-mock-objects
- https://doc.rust-lang.org/stable/book/ch17-02-trait-objects.html
- https://doc.rust-lang.org/stable/book/ch17-03-oo-design-patterns.html#defining-post-and-creating-a-new-instance-in-the-draft-state

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -579,21 +579,7 @@ fn add_playpen_pre(html: &str, playpen_config: &Playpen) -> String {
                 || classes.contains("mdbook-runnable")
             {
                 // wrap the contents in an external pre block
-                if playpen_config.editable && classes.contains("editable")
-                    || text.contains("fn main")
-                    || text.contains("quick_main!")
-                {
-                    format!("<pre class=\"playpen\">{}</pre>", text)
-                } else {
-                    // we need to inject our own main
-                    let (attrs, code) = partition_source(code);
-
-                    format!(
-                        "<pre class=\"playpen\"><code class=\"{}\">\n# \
-                         #![allow(unused_variables)]\n{}#fn main() {{\n{}#}}</code></pre>",
-                        classes, attrs, code
-                    )
-                }
+                format!("<pre class=\"playpen\">{}</pre>", text)
             } else {
                 // not language-rust, so no-op
                 text.to_owned()


carols10cents:
Note that this is only changing the rendering for the playground snippet presented in the HTML. [Rustdoc also does auto `main` wrapping](https://doc.rust-lang.org/stable/rustdoc/documentation-tests.html#pre-processing-examples), and `mdbook test` uses rustdoc's testing, so the awkward fixes in the book that allow them to be not ignored like in https://github.com/rust-lang/book/pull/1339 would still need to happen unless rustdoc gains the ability to specify in another way that the code shouldn't be wrapped in `main`.

Dylan-DPC:
The next release is most likely going to be a breaking change so that shouldn't matter

mikemorris:
@carols10cents Thanks for mentioning the wrapping in rustdoc, wasn't aware of that - I took a look into rustdoc and it looks like there's some pretty well-tested functionality around `dont_insert_main` and `already_has_main` in https://github.com/rust-lang/rust/blob/master/src/librustdoc/test.rs.

Should rustdoc add "contents contain `#[cfg(test)]` or `#[test]`" as an exclusion case for wrapping (is this technically allowed alongside a `main` function or not?) and/or would adding a rustdoc [attribute](https://doc.rust-lang.org/stable/rustdoc/documentation-tests.html#attributes) for skipping the wrapping behavior be a possibility?

Does rustdoc actually run test blocks like in the code snippet in https://github.com/rust-lang/book/pull/1339?

carols10cents:
You'd have to ask the rustdoc team those questions; I'd recommend opening an issue on rust-lang/rust and tagging rust-lang/rustdoc.

ehuss:
Yea, I don't think this can be done in this way.  Whatever the solution is will need to start with rustdoc. Some options:

- rustdoc could be smarter about what it wraps in main (seems very hard).
- tell rustdoc that a code block is a "test" and it should build and run it with the harness (seems tricky).
- add an annotation that something should compile (as a lib?), but not be run/tested.

There are probably better ideas out there.

Another question is if you expect the `#[test]` functions to actually run.  Currently they are only checked that they compile.

ehuss:
I think the solution for this will likely need to start with rustdoc.  When rustdoc has better `main` wrapping smarts, we can revisit updating mdbook.

PR #956, opened 2019-06-18
Release 0.3.0

ehuss:


remexre:
Could musl builds be provided? I'm building docs on alpine, where glibc isn't available.

ehuss:
At this time I'd prefer to stay with tier 1 platforms.

PR #961, opened 2019-06-20
Switch to the standard library's fs::read_to_string

carols10cents:
It was stabilized in 1.26, which is fine [because the current required Rust version is 1.34](https://github.com/rust-lang-nursery/mdBook/pull/960).


ehuss:
Thanks!

PR #962, opened 2019-06-21
Use stdlib RangeBounds

carols10cents:
Found this TODO; `RangeBounds` was stabilized in 1.28, so this is fine [because the current required Rust version is 1.34](https://github.com/rust-lang-nursery/mdBook/pull/960).


ehuss:
Nice!

PR #964, opened 2019-06-21
Don't keep gh-pages history.

ehuss:
This changes it so that Travis will squash the gh-pages branch each time a new release is done. The git repo is quite large (~100MB), and the hope is that this will significantly shrink it.  I don't think there is any real reason to keep the history, since it is generated from the source, and the history for that still exists.


lzutao:
I was about to do the same change. But I was afraid of the need for rolling back.
But whatever, I and you have the same argument.

PR #967, opened 2019-06-29
Fix memory leak and warning

ehuss:
This bumps the ammonia dependency to fix a memory leak and a warning log message (https://github.com/rust-ammonia/ammonia/pull/119).  This would happen if `<script>` or `<style>` tags were found in the book.


Dylan-DPC:
"Ammonia" "leak" 😛

PR #985, opened 2019-07-20
Allow backends to cache previous results

Michael-F-Bryan:
This is something that came up in my rewrite of `mdbook-linkcheck` (https://github.com/Michael-F-Bryan/mdbook-linkcheck/pull/11). I'd implemented a cache so we don't execute lots of slow web requests too frequently, and found that the cache file didn't exist every time the `mdbook-linkcheck` backend was invoked.

We probably want to let backends cache previous results. This PR removes the call to `utils::fs::remove_dir_content(&build_dir)` in `MDBook::build()` and makes sure the the HTML renderer manually cleans its destination directory.

I've also updated the user-docs to mention this.

### DIFF
--- a/book-example/src/for_developers/backends.md
+++ b/book-example/src/for_developers/backends.md
@@ -261,6 +261,10 @@ in [`RenderContext`].
 > **Note:** There is no guarantee that the destination directory exists or is
 > empty (`mdbook` may leave the previous contents to let backends do caching),
 > so it's always a good idea to create it with `fs::create_dir_all()`.
+>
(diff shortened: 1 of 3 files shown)


ehuss:
I *think* I understand this, and it does seem to make sense. It was already documented in the paragraph above that it may exist for caching, though emphasizing it twice doesn't hurt. 😉

sytsereitsma:
Deleting the book dir also removes any files written to the book output directory by preprocessors, which seems a bit too rigorous. See issue #1087

PR #989, opened 2019-07-23
ui: improve menu folding

WofWca:
Fold/unfold the menu bar just by the amount of scroll, not by its full width and remove animation.

Demo: https://wofwca.github.io/mdBook/format/config.html
### Before
This is noticeable on touchscreen devices. Suppose you scroll the page so that the line you're reading is the first line on the screen. Now, if you decide to scroll just one line up the menu bar unfolds and covers the part you were reading.

bad

### After
good

### Existing functionality checklist/TODO
- [x] Unfolds on hover
- [ ] Fold-unfold `transition`s (on hover, on sidebar show)
- [x] Folds-unfolds on scroll
- [x] Unfolded when the sidebar is visible
- [x] Always visible in case if JS is disabled

❓ Not sure what to do with animation. E.g. what do we do if an animation is playing and the user decides to scroll?


weihanglo:
Wow. I love this improvement.
Just one concern: Should we care about cross-platform compatibility? Though `position: sticky` is widely adopted about 90% of global usage, but only 71% use the unprefixed version.

WofWca:
> but only 71% use the unprefixed version

Here we manipulate the presence of the class named `sticky` that contains both prefixed and unprefixed versions.

weihanglo:
Looks good to me. Just a weird out-of-scope bug. 

On any mobile supports auto-bouncing effect when scrolling area exceeded. And then we may face "false positive" when doing a scroll-to-up action. This may be separated to another issue.

Still thanks for this improvement.

WofWca:
Ah, you must be talking about this, right?

Appetize.io

My mobile devices don't have this, I didn't know it was a thing.

Dylan-DPC:
@WofWca hey! if you can rebase this, i'm ready to review and merge it :) thanks

WofWca:
Will do within the next couple of days.

WofWca:
@Dylan-DPC Done.
Also, the overscroll problem (https://github.com/rust-lang/mdBook/pull/989#issuecomment-533774893) is gone, thanks to #1070

PR #991, opened 2019-07-31
Add `--path .` to cargo install commands.

lbeckman314:
Changing `cargo install` to `cargo install --path .` prevents the following error:

**error:** Using `cargo install` to install the binaries for the package in current working directory is no longer supported, use `cargo install --path .` instead. Use `cargo build` if you want to simply build the package.

### DIFF
--- a/book-example/src/for_developers/backends.md
+++ b/book-example/src/for_developers/backends.md
@@ -93,7 +93,7 @@ Now we've got the basics running, we want to actually use it. First, install the
 program.
 
 ```shell
-$ cargo install
+$ cargo install --path .
 ```
 
 Then `cd` to the particular book you'd like to count the words of and update its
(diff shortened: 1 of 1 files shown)


book-example/src/for_developers/backends.md
```diff
@@ -171,6 +171,16 @@ Now imagine you don't want to count the number of words on a particular chapter
...
 
+```diff
+  [output.wordcount]
++ ignores = [
++     "For Developers",
++     "Preprocessors",
++     "Alternative Backends",
++ ]
++ deny-odds = true
+```
```
steveklabnik: this change seems unrelated, was it intended?

lbeckman314: Ah, sorry! You're right — that was an unrelated commit. I have reset to the earlier commit. Thanks!

PR #993, opened 2019-08-03
ui: Fix the "Next chapter" link position

WofWca:
Remove the gap between it and the right edge of the page.

Demo: https://wofwca.github.io/mdBook/format/config.html

This reverts f1cd9f54c243ac1da136c3cbddf6b6cf622deae7 (That fixed /rust-lang/book/issues/29)
The page layout has been restructured and this workaround is no
longer required

### DIFF
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -162,7 +162,7 @@ html:not(.sidebar-visible) #menu-bar:not(:hover).folded > #menu-bar-sticky-conta
 
 .next {
     float: right;
-    right: var(--page-padding);
+    right: 0;
 }
 
 @media only screen and (max-width: 1080px) {


WofWca:
I don't think that my changes are the cause of the build failure.

ehuss:
> I don't think that my changes are the cause of the build failure.

Looks like an issue due to https://github.com/mozilla/sccache/issues/488.  cc @lzutao

ehuss:
Hm, so in testing this out this seems to cause some issues on firefox and chrome. If the scroll bar is not displayed, this prevents the scroll bar from appearing when the cursor is positioned over it.  You have to scroll a little bit to get it to appear, and then you can interact with it.  It seems a little awkward.  Is there something in particular this is trying to improve or fix?

WofWca:
Could you please provide more details on how to reproduce this? I've played around a bit and couldn't seem to experience this.

> Is there something in particular this is trying to improve or fix?

Yes. Sometimes you miss the button by clicking right on this gap. Also with the #994 change, this should make it look better (remove the gap).

ehuss:
Hm, maybe it is unique to Firefox on macos.  Firefox auto-hides the scroll bar, and you end up with this situation:

next_firefox_macos

Clicking in this region goes to the next page instead of scrolling.

What's weird is that on safari and chrome, the scroll bar never auto-hides.  That's not normal, so I guess it is related to some of the scrollbar css settings?

GuillaumeGomez:
I'll close this PR as the UI evolved a lot since then. Please open an issue/PR if the bug is still present nowadays and thanks for working on this!

WofWca:
Note that the problem that this fixes is still present.

GuillaumeGomez:
Do you mind opening an issue with screenshots? And if motivated enough, I'd love a PR to fix the bug with a `gui` test to prevent regressions.

PR #994, opened 2019-08-03
ui: Improve next/prev chapter links' style

WofWca:
Before: https://rust-lang-nursery.github.io/mdBook/format/config.html 
After: https://wofwca.github.io/mdBook/format/config.html

* Set on-hover background-color. This should prevent accidental clicks as the buttons were not very noticeable.
* Stretch them to full height

Works ok on all themes.

Related: #993 
Will conflict with #989


WofWca:
Same thing with the build as in #993

ehuss:
Thanks, this seems fine to me.

PR #998, opened 2019-08-06
Refactoring of some functionality in links.rs

carols10cents:
Hi, I have a new feature I'm working on that I could really use for the book (https://github.com/rust-lang-nursery/mdBook/issues/618), but these refactorings make it easier. I decided to submit these first separately from the new feature, to get feedback before basing the new feature on this.

There are 2 big chunks of refactorings in this PR, and I'm happy to split them into 2 PRs if anyone wants:

- The first commit extracts a data structure to hold the different kind of ranges, so that different kinds of file including mechanisms can use the same data structure. This reduces some kinds of repetition, but adds a bit more boilerplate.
- The rest of the commits improve the `parse_include_path` function: I think I've gotten test coverage around all the branches, and I think I've improved the code while keeping those tests passing. I think the code is clearer now, but it's certainly debatable, so I'd love to hear any thoughts.

Thanks! ❤️

### DIFF
--- a/src/preprocess/links.rs
+++ b/src/preprocess/links.rs
@@ -187,12 +187,10 @@ fn parse_include_path(path: &str) -> LinkType<'static> {
     let start = if let Some(value) = next_element.and_then(|s| s.parse::<usize>().ok()) {
         // subtract 1 since line numbers usually begin with 1
         Some(value.saturating_sub(1))
+    } else if let Some("") = next_element {
+        None
     } else if let Some(anchor) = next_element {
-        if anchor == "" {
-            None
-        } else {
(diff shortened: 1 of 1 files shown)


carols10cents:
Looks like the build failed because nightly can't build rand_core.... i'll try that one again tomorrow 🤷‍♀

ehuss:
> Looks like the build failed because nightly

Yea, there is an issue with sccache, I just merged a temporary fix.

carols10cents:
> Yea, there is an issue with sccache, I just merged a temporary fix.

Thank you, it passes now after a rebase!

ehuss:
Thanks, the `parse_include_path` is definitely easier to read.

PR #1000, opened 2019-08-12
Offline support fixes #546

### LINKED ISSUE #546: Offline support

jasonwilliams:
This PR is adding a service worker (via the use of [workbox.js](https://developers.google.com/web/tools/workbox/)) to the mdBook pages which allows for offline navigation and usage. We use the `StaleWhileRevalidate` mode from workbox

In the gif below the red you see in the network tab is the _StaleWhileRevalidate_ kicking in and trying to load a fresh version, this obviously fails offline so is expected.

Gif, click for animation:
swRust

I have updated the workbox library as it was very out of date. Not sure on the failing test though
@sorin-davidoi  @Michael-F-Bryan

**How does this work?**
When you visit the page the service worker will start to pre-cache all of the chapters within the book, it basically visits those urls and caches the response (but not the assets on those pages). So it doesn't matter which page you hit first, the service work will cache all pages in the book. 
When you visit a page in the book, the contents are served from the cache, if there is no matching content in the cache it is served from the network.

Below is an image of what the cache table looks like in Chrome.
I made a slight change to the appendix page, you can see workbox replaced the cache with the updated version, but kept the other pages as they are. (Time Cached is different to the other entries)
image

**When does the cache invalidate?**
The mode set is staleWhileRevalidate so it will always try to pull a fresh copy in the background (whilst serving you content from the cache). If the content is the same it is basically thrown away. If the content is new workbox replaces the cache with the up to date page and uses that instead.

There is no explicit expiry time set.

**Testing Instructions**
* Navigate to https://jason-williams.co.uk/book/
* Go offline, (open console, go to network tab, set preset to offline)
* Try clicking some links

**Testing Instructions (build)**
* Checkout https://github.com/rust-lang/book
* Add offline-support = true to the [output.html] section
* Disable the check for localhost at the end of book.js in mdBook repo
* Build mdBook (set the binary in your path `target/debug/mdBook`)
* call mdBook serve in the `book` repo

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -600,9 +600,15 @@ function playpen_text(playpen) {
   var isLocalhost =
     ["localhost", "127.0.0.1", ""].indexOf(document.location.hostname) !== -1;
 
-  if ("serviceWorker" in navigator && !isLocalhost) {
-    navigator.serviceWorker.register("sw.js").catch(function(error) {
-      console.error("Service worker registration failed:", error);
-    });
+  if (
+    window.sevice_worker_script_src &&
+    "serviceWorker" in navigator &&
+    !isLocalhost
+  ) {
+    navigator.serviceWorker
+      .register(window.sevice_worker_script_src)
+      .catch(function(error) {
+        console.error("Service worker registration failed:", error);
+      });
   }
 })();
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -594,3 +594,17 @@ function playpen_text(playpen) {
         previousScrollTop = document.scrollingElement.scrollTop;
     }, { passive: true });
 })();
+
+
+(function serviceWorker() {
+  var isLocalhost =
+    ["localhost", "127.0.0.1", ""].indexOf(document.location.hostname) !== -1;
+
+  if ("serviceWorker" in navigator && true) {
+    navigator.serviceWorker
+      .register(document.baseURI + "sw.js")
+      .catch(function(error) {
+        console.error("Service worker registration failed:", error);
+      });
+  }
+})();
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -602,7 +602,7 @@ function playpen_text(playpen) {
 
   if ("serviceWorker" in navigator && !isLocalhost) {
     navigator.serviceWorker
-      .register(document.baseURI + "sw.js")
+      .register(document.location.origin + "/sw.js")
       .catch(function(error) {
         console.error("Service worker registration failed:", error);
       });
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -601,10 +601,8 @@ function playpen_text(playpen) {
     ["localhost", "127.0.0.1", ""].indexOf(document.location.hostname) !== -1;
 
   if ("serviceWorker" in navigator && !isLocalhost) {
-    navigator.serviceWorker
-      .register(document.location.origin + "/sw.js")
-      .catch(function(error) {
-        console.error("Service worker registration failed:", error);
-      });
+    navigator.serviceWorker.register("sw.js").catch(function(error) {
+      console.error("Service worker registration failed:", error);
+    });
   }
 })();
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -600,7 +600,7 @@ function playpen_text(playpen) {
   var isLocalhost =
     ["localhost", "127.0.0.1", ""].indexOf(document.location.hostname) !== -1;
 
-  if ("serviceWorker" in navigator && true) {
+  if ("serviceWorker" in navigator && !isLocalhost) {
     navigator.serviceWorker
       .register(document.baseURI + "sw.js")
       .catch(function(error) {
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -601,12 +601,12 @@ function playpen_text(playpen) {
     ["localhost", "127.0.0.1", ""].indexOf(document.location.hostname) !== -1;
 
   if (
-    window.sevice_worker_script_src &&
+    window.service_worker_script_src &&
     "serviceWorker" in navigator &&
     !isLocalhost
   ) {
     navigator.serviceWorker
-      .register(window.sevice_worker_script_src)
+      .register(window.service_worker_script_src)
       .catch(function(error) {
         console.error("Service worker registration failed:", error);
       });
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -259,7 +259,7 @@
         {{#if offline_support}}
         <script type="text/javascript">
             // Must insert before loading book.js
-            window.sevice_worker_script_src = "{{ path_to_root }}sw.js";
+            window.service_worker_script_src = "{{ path_to_root }}sw.js";
         </script>
         {{/if}}
 
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -235,7 +235,7 @@
             window.playpen_line_numbers = true;
         </script>
         {{/if}}
-        
+
         {{#if playpen_copyable}}
         <script type="text/javascript">
             window.playpen_copyable = true;
@@ -254,6 +254,13 @@
         <script src="{{ path_to_root }}elasticlunr.min.js" type="text/javascript" charset="utf-8"></script>
         <script src="{{ path_to_root }}mark.min.js" type="text/javascript" charset="utf-8"></script>
         <script src="{{ path_to_root }}searcher.js" type="text/javascript" charset="utf-8"></script>
+        {{/if}}
+
+        {{#if offline_support}}
+        <script type="text/javascript">
+            // Must insert before loading book.js
+            window.sevice_worker_script_src = "{{ path_to_root }}sw.js";
+        </script>
         {{/if}}
 
         <script src="{{ path_to_root }}clipboard.min.js" type="text/javascript" charset="utf-8"></script>
--- a/src/theme/sw.js
+++ b/src/theme/sw.js
@@ -30,7 +30,7 @@ workbox.routing.registerRoute(
 
 // Local resources
 workbox.routing.registerRoute(
-  /\.(woff2?|ttf|css|js|json|png|svg)$/,
+  /\.(woff2?|ttf|css|js|json|png|svg)(\?v\=.*)?$/,
   staleWhileRevalidate
 );
 
--- a/src/theme/sw.js
+++ b/src/theme/sw.js
@@ -30,7 +30,7 @@ workbox.routing.registerRoute(
 
 // Local resources
 workbox.routing.registerRoute(
-  new RegExp(".woff2?|.ttf|.css|.js|.json"),
+  new RegExp(".woff2?|.ttf|.css|.js|.json|.png|.svg"),
   staleWhileRevalidate
 );
 
--- a/src/theme/sw.js
+++ b/src/theme/sw.js
@@ -30,7 +30,7 @@ workbox.routing.registerRoute(
 
 // Local resources
 workbox.routing.registerRoute(
-  new RegExp(".woff2?|.ttf|.css|.js|.json|.png|.svg"),
+  /\.(woff2?|ttf|css|js|json|png|svg)$/,
   staleWhileRevalidate
 );
 
--- a/src/theme/sw.js
+++ b/src/theme/sw.js
@@ -1,0 +1,46 @@
+importScripts(
+  "https://unpkg.com/workbox-sw@2.0.3/build/importScripts/workbox-sw.dev.v2.0.3.js"
+);
+
+// clientsClaims tells the Service Worker to take control as soon as it's activated
+const workbox = new WorkboxSW({ clientsClaim: true });
+
+// https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#stale-while-revalidate
+// TLDR: If there's a cached version available, use it, but fetch an update for next time.
+const staleWhileRevalidate = workbox.strategies.staleWhileRevalidate();
+
+// Remote fonts and JavaScript libraries
+workbox.router.registerRoute(
+  new RegExp("https://fonts.googleapis.com/css"),
+  staleWhileRevalidate
+);
+workbox.router.registerRoute(
+  new RegExp("https://fonts.gstatic.com"),
+  staleWhileRevalidate
+);
+workbox.router.registerRoute(
+  new RegExp("https://maxcdn.bootstrapcdn.com/font-awesome"),
+  staleWhileRevalidate
+);
+workbox.router.registerRoute(
+  new RegExp("https://cdnjs.cloudflare.com/ajax/libs/mathjax"),
+  staleWhileRevalidate
+);
+workbox.router.registerRoute(
+  new RegExp("https://cdn.jsdelivr.net/clipboard.js"),
+  staleWhileRevalidate
+);
+
+// Local resources
+workbox.router.registerRoute(new RegExp(".js$"), staleWhileRevalidate);
+workbox.router.registerRoute(new RegExp(".css$"), staleWhileRevalidate);
+
+// Here hbs_renderer.rs will inject the chapters, making sure they are precached.
+//
+// const chapters = [
+//     { url: '/', revision: '11120' },
+//     { url: 'cli/cli-tool.html', revision: '12722' },
+//     { url: 'cli/init.html', revision: '12801' },
+// ];
+//
+//   workbox.precaching.precacheAndRoute(chapters);
--- a/src/theme/sw.js
+++ b/src/theme/sw.js
@@ -29,8 +29,10 @@ workbox.routing.registerRoute(
 );
 
 // Local resources
-workbox.routing.registerRoute(new RegExp(".js$"), staleWhileRevalidate);
-workbox.routing.registerRoute(new RegExp(".css$"), staleWhileRevalidate);
+workbox.routing.registerRoute(
+  new RegExp(".woff2?|.ttf|.css|.js|.json"),
+  staleWhileRevalidate
+);
 
 // Here hbs_renderer.rs will inject the chapters, making sure they are precached.
 //
--- a/src/theme/sw.js
+++ b/src/theme/sw.js
@@ -1,39 +1,36 @@
 importScripts(
-  "https://unpkg.com/workbox-sw@2.0.3/build/importScripts/workbox-sw.dev.v2.0.3.js"
+  "https://storage.googleapis.com/workbox-cdn/releases/4.3.1/workbox-sw.js"
 );
-
-// clientsClaims tells the Service Worker to take control as soon as it's activated
-const workbox = new WorkboxSW({ clientsClaim: true });
 
 // https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#stale-while-revalidate
 // TLDR: If there's a cached version available, use it, but fetch an update for next time.
-const staleWhileRevalidate = workbox.strategies.staleWhileRevalidate();
+const staleWhileRevalidate = new workbox.strategies.StaleWhileRevalidate();
 
 // Remote fonts and JavaScript libraries
-workbox.router.registerRoute(
+workbox.routing.registerRoute(
   new RegExp("https://fonts.googleapis.com/css"),
   staleWhileRevalidate
 );
-workbox.router.registerRoute(
+workbox.routing.registerRoute(
   new RegExp("https://fonts.gstatic.com"),
   staleWhileRevalidate
 );
-workbox.router.registerRoute(
+workbox.routing.registerRoute(
   new RegExp("https://maxcdn.bootstrapcdn.com/font-awesome"),
   staleWhileRevalidate
 );
-workbox.router.registerRoute(
+workbox.routing.registerRoute(
   new RegExp("https://cdnjs.cloudflare.com/ajax/libs/mathjax"),
   staleWhileRevalidate
 );
-workbox.router.registerRoute(
+workbox.routing.registerRoute(
   new RegExp("https://cdn.jsdelivr.net/clipboard.js"),
   staleWhileRevalidate
 );
 
 // Local resources
-workbox.router.registerRoute(new RegExp(".js$"), staleWhileRevalidate);
-workbox.router.registerRoute(new RegExp(".css$"), staleWhileRevalidate);
+workbox.routing.registerRoute(new RegExp(".js$"), staleWhileRevalidate);
+workbox.routing.registerRoute(new RegExp(".css$"), staleWhileRevalidate);
 
 // Here hbs_renderer.rs will inject the chapters, making sure they are precached.
 //
--- a/src/theme/sw.js
+++ b/src/theme/sw.js
@@ -2,6 +2,11 @@ importScripts(
   "https://storage.googleapis.com/workbox-cdn/releases/4.3.1/workbox-sw.js"
 );
 
+self.addEventListener("install", event => {
+  // Take over old service worker immediately, should hopefully fix weird caching issues
+  self.skipWaiting();
+});
+
 // https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#stale-while-revalidate
 // TLDR: If there's a cached version available, use it, but fetch an update for next time.
 const staleWhileRevalidate = new workbox.strategies.StaleWhileRevalidate();
--- a/book-example/book.toml
+++ b/book-example/book.toml
@@ -6,6 +6,7 @@ language = "en"
 
 [output.html]
 mathjax-support = true
+offline-support = true
 
 [output.html.playpen]
 editable = true
--- a/book-example/src/format/config.md
+++ b/book-example/src/format/config.md
@@ -1,8 +1,8 @@
 # Configuration
 
-You can configure the parameters for your book in the ***book.toml*** file.
+You can configure the parameters for your book in the **_book.toml_** file.
 
-Here is an example of what a ***book.toml*** file might look like:
+Here is an example of what a **_book.toml_** file might look like:
 
 ```toml
 [book]
@@ -45,6 +45,7 @@ This is general information about your book.
 - **language:** The main language of the book, which is used as a language attribute `<html lang="en">` for example.
 
 **book.toml**
+
 ```toml
 [book]
 title = "Example book"
@@ -87,8 +88,8 @@ The following preprocessors are available and included by default:
   to say, all `README.md` would be rendered to an index file `index.html` in the
   rendered book.
 
-
 **book.toml**
+
 ```toml
 [build]
 build-dir = "build"
@@ -148,6 +149,8 @@ The following configuration options are available:
 - **theme:** mdBook comes with a default theme and all the resource files needed
   for it. But if this option is set, mdBook will selectively overwrite the theme
   files with the ones found in the specified folder.
+- **offline-support** Precache the chapters so that users can view the book
+  while offline. Available in [browsers supporting Service Worker](https://caniuse.com/#feat=serviceworkers).
 - **default-theme:** The theme color scheme to select by default in the
   'Change Theme' dropdown. Defaults to `light`.
 - **preferred-dark-theme:** The default dark theme. This theme will be used if
@@ -173,7 +176,7 @@ The following configuration options are available:
 - **playpen:** A subtable for configuring various playpen settings.
 - **search:** A subtable for configuring the in-browser search functionality.
   mdBook must be compiled with the `search` feature enabled (on by default).
-- **git-repository-url:**  A url to the git repository for the book. If provided
+- **git-repository-url:** A url to the git repository for the book. If provided
   an icon link will be output in the menu bar of the book.
 - **git-repository-icon:** The FontAwesome icon class to use for the git
   repository link. Defaults to `fa-github`.
@@ -186,7 +189,7 @@ Available configuration options for the `[output.html.playpen]` table:
   Defaults to `true`.
 - **line-numbers** Display line numbers on editable sections of code. Requires both `editable` and `copy-js` to be `true`. Defaults to `false`.
 
-[Ace]: https://ace.c9.io/
+[ace]: https://ace.c9.io/
 
 Available configuration options for the `[output.html.search]` table:
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -447,6 +447,8 @@ pub struct HtmlConfig {
     pub curly_quotes: bool,
     /// Should mathjax be enabled?
     pub mathjax_support: bool,
+    /// Cache chapters for offline viewing
+    pub offline_support: bool,
     /// An optional google analytics code.
     pub google_analytics: Option<String>,
     /// Additional CSS stylesheets to include in the rendered page's `<head>`.
@@ -609,6 +611,7 @@ mod tests {
         default-theme = "rust"
         curly-quotes = true
         google-analytics = "123456"
+        offline-support = true
         additional-css = ["./foo/bar/baz.css"]
         git-repository-url = "https://foo.com/"
         git-repository-icon = "fa-code-fork"
@@ -647,6 +650,7 @@ mod tests {
         };
         let html_should_be = HtmlConfig {
             curly_quotes: true,
+            offline_support: true,
             google_analytics: Some(String::from("123456")),
             additional_css: vec![PathBuf::from("./foo/bar/baz.css")],
             theme: Some(PathBuf::from("./themedir")),
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -9,11 +9,20 @@ use crate::utils;
 use std::borrow::Cow;
 use std::collections::BTreeMap;
 use std::collections::HashMap;
-use std::fs;
+use std::collections::hash_map::DefaultHasher;
+use std::hash::Hasher;
+use std::fs::{self, OpenOptions};
+use std::io::Write;
 use std::path::{Path, PathBuf};
 
 use handlebars::Handlebars;
 use regex::{Captures, Regex};
+
+#[derive(Default)]
+pub struct ChapterFile {
+    pub path: String,
+    pub revision: u64,
+}
 
 #[derive(Default)]
 pub struct HtmlHandlebars;
@@ -23,12 +32,46 @@ impl HtmlHandlebars {
         HtmlHandlebars
     }
 
+    fn build_service_worker(&self, build_dir: &Path, chapter_files: &Vec<ChapterFile>) -> Result<()> {
+        let path = build_dir.join("sw.js");
+        let mut file = OpenOptions::new().append(true).open(path)?;
+        let mut content = String::from("\nconst chapters = [\n");
+
+         for chapter_file in chapter_files {
+            content.push_str("  { url: ");
+
+             // Rewrite "/" to point to the current directory
+            // https://rust-lang-nursery.github.io/ => https://rust-lang-nursery.github.io/mdBook/
+            // location.href is https://rust-lang-nursery.github.io/mdBook/sw.js
+            // so we remove the sw.js from the end to get the correct path
+            if chapter_file.path == "/" {
+                content.push_str("location.href.slice(0, location.href.length - 5)");
+            } else {
+                content.push_str("'");
+                content.push_str(&chapter_file.path);
+                content.push_str("'");
+            }
+
+             content.push_str(", revision: '");
+            content.push_str(&chapter_file.revision.to_string());
+            content.push_str("' },\n");
+        }
+
+         content.push_str("];\n");
+        content.push_str("\nworkbox.precache(chapters);\n");
+
+         file.write(content.as_bytes())?;
+
+         Ok(())
+    }
+
     fn render_item(
         &self,
         item: &BookItem,
         mut ctx: RenderItemContext<'_>,
         print_content: &mut String,
-    ) -> Result<()> {
+    ) -> Result<Vec<ChapterFile>> {
+        let mut chapter_files = Vec::new();
         // FIXME: This should be made DRY-er and rely less on mutable state
         if let BookItem::Chapter(ref ch) = *item {
             let content = ch.content.clone();
@@ -47,6 +90,9 @@ impl HtmlHandlebars {
                 .to_str()
                 .chain_err(|| "Could not convert path to str")?;
             let filepath = Path::new(&ch.path).with_extension("html");
+	        let filepath_str = filepath.to_str().ok_or_else(|| {
+                Error::from(format!("Bad file name: {}", filepath.display()))
+            })?;
 
             // "print.html" is used for the print page.
             if ch.path == Path::new("print.md") {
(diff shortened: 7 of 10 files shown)


DevQps:
@jasonwilliams Sorry for the ultra late reply! I promised last Saturday but it took me longer sadly..

I checked [https://jason-williams.co.uk/book/](https://jason-williams.co.uk/book/) but when I plug off my Ethernet cable I get a "Not found page" on Firefox 68.0.2 (64-bit). Is it possible that something changed?
image

I will go check the other instructions as well.

DevQps:
@jasonwilliams Could you give me a few more pointers on the instructions?

```
    1. Checkout https://github.com/rust-lang/book
    2. Add offline-support = true to the [output.html] section
    3. Disable the check for localhost at the end of book.js in mdBook repo
    4. Build mdBook (set the binary in your path target/debug/mdBook)
    5. call mdBook serve in the book repo
```
I cloned the book repo and added the offline-support line. I guess I also need to clone your version of the mdbook repository? What do you mean exactly with disable the check for localhost? What should I verify once I did these steps?

Michael-F-Bryan:
@DevQps [this line](https://github.com/jasonwilliams/mdBook/blob/e7fbd0988359c9bae8c3ee7c111c65c4bbcbc60d/src/theme/book.js#L715-L716):

```js
  var isLocalhost =
    ["localhost", "127.0.0.1", ""].indexOf(document.location.hostname) !== -1;
```

jasonwilliams:
@DevQps could you refresh your cache and try https://jason-williams.co.uk/book/ again?
If you have the same problem, open your dev tools and go to `application`

DevQps:
> 
> 
> @DevQps could you refresh your cache and try https://jason-williams.co.uk/book/ again? (whilst online).
> Then go offline and give it another go.
> 
> If you have the same problem, open your dev tools and go to `application` then take a screenshot for us

I tried it again! I figured out that it caches the pages I have been to before, but not the pages I have been to. For example I could browse to Functions again, but not to other pages. I did a Crtl + Shift + R refresh and deleted all storage / cookies manually as well, but no change.

jasonwilliams:
@DevQps what browser are you using?
Is your dev tools showing all the pages downloaded in the service worker tab?

DevQps:
Hey Jason!,

I tried it again on Firefox and it worked. Then I tested it on Chrome again, and the first time it failed. But when I reconnected + browsed a page it worked. Could it be that either A) It takes some times (as in around 10 seconds) to fully load everything, or B) That it only works after a reload (i.e. the new service worker is run?)

So I think everything works now!

jasonwilliams:
@DevQps the dev tools tab (applications) should show you which pages have downloaded
Here is a snippet of mine:
Annotation 2019-09-15 175337

You may have clicked away before everything downloaded

DevQps:
That's probably the case! I think this MR is fine!

jasonwilliams:
@DevQps what’s the next steps?

DevQps:
I am not sure! @steveklabnik Can you review this maybe?

steveklabnik:
I am not really involved in mdbook maintenance these days; @Dylan-DPC should know.

src/theme/sw.js
```diff
@@ -0,0 +1,45 @@
...
+  staleWhileRevalidate
+);
+workbox.routing.registerRoute(
+  new RegExp("https://cdn.jsdelivr.net/clipboard.js"),
+  staleWhileRevalidate
+);
+
+// Local resources
+workbox.routing.registerRoute(
+  new RegExp(".woff2?|.ttf|.css|.js|.json|.png|.svg"),
```
weihanglo: Maybe a more robust regex?
And all registered routes above may need some modifications and escaping such as `^`.

```suggestion
  /\.(woff2?|ttf|css|js|json|png|svg)$/
```

src/theme/book.js
```diff
@@ -1,620 +1,723 @@
...
-            menu.classList.remove('bordered');
-        }
+(function serviceWorker() {
+  var isLocalhost =
+    ["localhost", "127.0.0.1", ""].indexOf(document.location.hostname) !== -1;
 
-        previousScrollTop = document.scrollingElement.scrollTop;
-    }, { passive: true });
+  if ("serviceWorker" in navigator && !isLocalhost) {
+    navigator.serviceWorker.register("sw.js").catch(function(error) {
```
weihanglo: Encountered this error when navigate to pages other than base URL. 

```
Service worker registration failed: TypeError: "ServiceWorker script 
at http://localhost:3000/format/sw.js 
for scope http://localhost:3000/format/ encountered an error during installation."
```

The problem here is that the register path is relative to current file URL. We can resolve it by adding `<base href="https://our.basedomain.com">` though this would be a huge change.

weihanglo: I feel that we could merge this PR and open another if the problem is too thorny to deal with at this time.

jasonwilliams:
@weihanglo rebased

weihanglo:
https://github.com/rust-lang-nursery/mdBook/pull/1000#discussion_r335536088

Besides this `baseURI` problem, I think we can merge this PR to make some progress. We can try to deal it later in a separate PR.

weihanglo:
@jasonwilliams 
The relative path issue has been fixed. See https://github.com/jasonwilliams/mdBook/pull/1

jasonwilliams:
@weihanglo thoughts on this?

weihanglo:
I’ve tested and all looks good to me.

ehuss:
I'm finding this difficult to test.  It looks like Firefox, Safari, and Chrome disable serviceWorker for non-https urls.  Is that correct?  Is the only way to test this involve commenting out the localhost test and only use localhost?  What is the reason for the localhost check?

I noticed you are using the Rust book as an example, do you desire to try this on rust-lang books? I'm a little wary of that, since that could be a non-trivial load increase (mainly for the user).  If so, I'm also uncertain about the motivation, since rust ships copies of all its books for offline access.

jasonwilliams:
Lets unpack this..

> I'm finding this difficult to test.

Which bit is difficult to test? Did you follow the testing instrucions [here](https://github.com/rust-lang/mdBook/pull/1000#issue-306662841)? 

> It looks like Firefox, Safari, and Chrome disable serviceWorker for non-https urls. Is that correct? 

Yes this is correct, there is an exception for localhost however (for development purposes).
I don't think this is a problem as most sites (rust-lang documentation, other books) are now https, plus it will continue to work on http it just won't cache.

> Is the only way to test this involve commenting out the localhost test and only use localhost? What is the reason for the localhost check?

That. and the rest of the instuctions [here](https://github.com/rust-lang/mdBook/pull/1000#issue-306662841) (under Testing Instructions (build)).
The localhost check was added to stop it caching when developing locally. It was added in the PR before mine.

> I noticed you are using the Rust book as an example, do you desire to try this on rust-lang books? I'm a little wary of that, since that could be a non-trivial load increase (mainly for the user). If so, I'm also uncertain about the motivation, since rust ships copies of all its books for offline access.

I don't know what load increase means.

> If so, I'm also uncertain about the motivation, since rust ships copies of all its books for offline access.

This is assuming i want to read Rust on a development environment where i had cargo/rustup installed. This is more for environments where that isn't the case (mobile phone, tablet, handheld devices). I can't load the offline documentation from rustup onto a tablet.

ehuss:
> Which bit is difficult to test? 

If I want to test from other devices (like a phone or tablet), I think I'll need an https server?  I'll try to figure out how to get one running locally.

Also, commenting things out is a little awkward, but should be fine for now.

> I don't know what load increase means.

The very first time someone lands on one of the books, it downloads the whole thing, right?  That increases page load for *everyone*, even if they are only wanting a quick look at a single page.  Testing on my slower connection, it took an extra 5-10s to load everything, which is quite significant.  

It might not be too bad, as it is all in the background, so shouldn't affect initial rendering. And hopefully it caches well?

jasonwilliams:
> If I want to test from other devices (like a phone or tablet), I think I'll need an https server? I'll try to figure out how to get one running locally.
Also, commenting things out is a little awkward, but should be fine for now.

That's the reason I set this up on a web server so you can visit and test that way.
https://jason-williams.co.uk/book/ You can still visit there and test on a mobile device.
If you wish to test it yourself follow my build instructions on a webserver.
If you want to test locally you can do, but you'll need to proxy your device to localhost on your machine.

> The very first time someone lands on one of the books, it downloads the whole thing, right? That increases page load for everyone, even if they are only wanting a quick look at a single page. Testing on my slower connection, it took an extra 5-10s to load everything, which is quite significant.

This is true and the subsequent loads will be in the background and should not affect page load performance for the page you're trying to load. The javascript kicks in and downloads after the initial page has loaded. This will also only happen the first time, after that only changed content should download and not the whole book. 

> it took an extra 5-10s to load everything
That's surprising to me, can i ask how you tested that? 
For me the speed has not changed at all, and the network tab is only monitoring requests for that page (not the service worker requests) which will have happened long before background fetching starts happening.

ehuss:
> That's surprising to me, can i ask how you tested that?

It's all in the background, but mainly just watching server-side requests and network activity on the local client.  It probably doesn't matter, the fonts and search index dominate everything.

I'm having random problems with fonts not caching.  I can't quite pin it down, because sometimes they won't load offline and sometimes they do.  I see some CORS errors, but I'm not sure what's going on.  It's still usable, but the icons at the top of the page are sometimes broken.

src/theme/index.hbs
```diff
@@ -256,6 +256,13 @@
         <script src="{{ path_to_root }}searcher.js" type="text/javascript" charset="utf-8"></script>
         {{/if}}
 
+        {{#if offline_support}}
+        <script type="text/javascript">
+            // Must insert before loading book.js
+            window.sevice_worker_script_src = "{{ path_to_root }}sw.js";
```
ehuss: BTW, this seems to be misspelled?  That is `sevice` → `service` with an `r`.  If that is not intentional, can all the instances be updated?

jasonwilliams:
@ehuss in terms of fonts what browser are you testing on and how are you testing? Are you using my endpoint? I will take a look and see if I get the same isssue

ehuss:
I'm testing your endpoint and a local one.

## Safari font caching
I can't quite figure out why the fonts don't cache all the time.  It's on iOS 12 (physical device, using airplane mode) and Safari 13 desktop.  Sometimes it works sometimes it doesn't.  I get this in the console:

```
[Error] Failed to load resource: The Internet connection appears to be offline. (crates, line 0)
[Error] Failed to load resource: The Internet connection appears to be offline. (fontawesome-webfont.woff2, line 0)
[Error] Unhandled Promise Rejection: TypeError: The Internet connection appears to be offline.
    (anonymous function)
    promiseReactionJob
[Error] Failed to load resource: The Internet connection appears to be offline. (fontawesome-webfont.woff, line 0)
[Error] Failed to load resource: The Internet connection appears to be offline. (fontawesome-webfont.ttf, line 0)
[Error] Failed to load resource: The Internet connection appears to be offline. (fontawesome-webfont.svg, line 0)
```

## Android 10 Chrome Emulator
I can't get offline to work at all on Chrome in an Android emulator. I get the following error in the console. 
```
Service worker registration failed: DOMException
```
I don't have a physical android device to test.  Do you have one? Is this something you can test?

## Any browser — search broken
Trying to load a page from the search results fails:
The `?highlight=foo` part of the URL seems to break things.

jasonwilliams:
> ## Any browser — search broken
> Trying to load a page from the search results fails:
> The ?highlight=foo part of the URL seems to break things.

This is a known issue, i didn't want to invest too much time in making search work offline if the initial work wasn't going to be accepted. So as a first MVP i think its safe to assume search does not work offline ( and can be added later on).

Having search work offline is a feature within itself and shouldn't be part of this.

ehuss:
I'm having a problem with it not invalidating the cache.  What mechanism does it use to know it needs to reload?  

I was serving one book while testing this.  I then loaded a different book (at the same URL), and only the old book would show up.  I had to go into the firefox webworkers config to delete the webworker to get it to work.  That's a little worrisome, if we turn it on for people and the only way to turn it off is to take steps like that.  And they probably won't know they are looking at an out-of-date copy.

jasonwilliams:
> What mechanism does it use to know it needs to reload?

So by default if you don't supply workbox with anything, it will use the URL as the revision key. Obviously (as urls won't change) that is not what we want.

So we pass a (revision)[https://github.com/rust-lang/mdBook/pull/1000/files#diff-22028e040c17d23dbccf8fc78c428d53R59-R61] hash. This tells workbox whether content has changed or not.

If content has changed, workbox should update the Service Worker Cache, if not nothing happens.
I've tested this previously by changing content then rebuilding, and its worked fine for me. 

> I was serving one book while testing this. I then loaded a different book (at the same URL), and only the old book would show up. 
I will need to see what steps you took to have this affect, as its not clear to me what you did.

jasonwilliams:
@ehuss the latest commit fixes search in offline mode.
The pages had query strings on the end (for highlighting) which caused search URLs to fail.

There is also a fix which adds woff2 properly to the cache (the v=4.7.0) was being missed and those resources were not cached.

Please try again at some point, its working better for me now

ehuss:
> Sometimes refreshing alone is not enough to pull down new changes (this is because browsers don't count refreshing as a new session), you may need to leave and come back by closing the tab and reopening a new tab.

Hm, this is probably what I'm running into.  This is worrisome.  On desktop Safari, the only way I've found to get it to reload is to close the tab and create a new window (quitting and restarting doesn't even work).  On mobile safari, I have yet to find a way to get it to reload (I've tried closing the tab, reloading, killing the browser, etc.).  I'm concerned that if someone were to visit a book, they would get stuck on whatever version they first loaded.

jasonwilliams:
> On mobile safari, I have yet to find a way to get it to reload (I've tried closing the tab, reloading, killing the browser, etc.). I'm concerned that if someone were to visit a book, they would get stuck on whatever version they first loaded.

iOS (v13.3) is working as expected for me.
I changed content on https://jason-williams.co.uk/book/ then ran `mdbook build`.
Either closing/reopening the tab or navigating somewhere else then back again refreshes it for me.

> Once successfully installed, the updated worker will wait until the existing worker is controlling zero clients. (Note that clients overlap during a refresh.)
https://developers.google.com/web/fundamentals/primers/service-workers/lifecycle

How are you testing? Are you properly rebuilding the book with the new branch?
For me IOS is working fine

peaceshi:
maybe just use "**Cache**" is a way. 
just cache what you read with `registerRoute`. 
then just use `strategies.StaleWhileRevalidate/CacheFirst` to fetch resources.

how to fetch? use `setInterval` with `registration.update`
https://github.com/peaceshi/GameProgrammingPatterns/blob/2f5d28d94c0885c1b03908665a015e266ce8c680/plugin/init.js#L32

because of all html pages from index.hbs, everything in index.hbs will be cached at first time. such as js/css etc.

and as PWA, if user **install** it , then asking user to download all or not.
user **not install** it. just cache what has been read.

peaceshi:
and such as debug at localhost. I think the first is **make sure** the user can read offline.

it is hard for beginner to know something, such as this:
https://github.com/peaceshi/GameProgrammingPatterns/blob/2f5d28d94c0885c1b03908665a015e266ce8c680/plugin/init.js#L26

beginner can not understand the reason why `/GameProgrammingPatterns/service-worker.js` not just `service-worker.js`

so make a "just build and run" PWA support  is necessary. **This is another topic.**

all in all, just let it run correctly at production env or just support it become PWA now.

peaceshi:
Sample book: 
https://peaceshi.github.io/GameProgrammingPatterns/index.htmls
use Microsoft Edge canary:
image
then open app. read, offline. everything runs correctly.

jasonwilliams:
@peaceshi caching only what you have already read doesn’t really solve the problem of an “offline book”, one you move to the next page it won’t work.

I think what we have now is almost there, we just need the new service worker to kick in immediately to prevent stale pages while refreshing.

There’s also 1 other question of expiration time but I don’t think that’s an issue right now (stale data may be better than no data)

peaceshi:
> once you move to the next page it won’t work.

do you test my example? install it as an `app`. then offline.

all cached exist.

in PWA app. all cached resource by service-worker should work fine.

if you want `Storage` not `Cache` at local. IndexedDB maybe better.

> There’s also 1 other question of expiration time but I don’t think that’s an issue right now (stale data may be better than no data)

This is really complex. because htmlhave been cached by sw, when sever update html, the sw will also use stale data. 
- A way to solve it is that `setInterval` with `registration.update`. to many fetch ,this is not a good way.
- another is use cache-version `unregister all sw` first , then cache all to a new version with cache API. 
See MDN. 
https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Offline_Service_workers
- also `caches.match` is better. I think.
https://developers.google.com/web/ilt/pwa/caching-files-with-service-worker#cache_then_network

About workbox... I can't connect to google to read workbox doc now. wait another day...

jasonwilliams:
@ehuss 
The latest commit sets the new service worker to skipWaiting() and takes over immediately. This means  that new changes should appear on refresh, you shouldn't need to close and reopen the browser (tab).

You may find it sometimes needs 2 refreshes after content has changed.
1. to pull down the new service worker (which will fetch changed content in the background)
2. to load the new content which the service worker changed in 1.

https://jason-williams.co.uk/book/ has been updated

ehuss:
Is it expected that the images are broken on your site? https://jason-williams.co.uk/book/ch00-00-introduction.html

I once saw broken images on my test site, but I can't reproduce it now.

I somehow got both my iOS device and simulator stuck on an old version (from a few months ago), and I can't get it to refresh or unload.  I tried deleting the cache in Settings, and a hard restart, and it made no difference.  Is there any way I can clear it?  Maybe attach with a remote console?

Is it normal to have a lot of errors in the console?  firefox, safari, and chrome all spew a bunch of stuff whether I'm online or offline.

I can't get chrome to work against my local site, and I wonder if it is because I am using an IP instead of a hostname.  When I click on something offline, I just get "ERR_INTERNET_DISCONNECTED".  Your website works, so I'm not sure what's different.  I'm using a really basic https server, so maybe it isn't serving the right headers like etags?

I'm a bit concerned about the reloading thing.  It seems that whenever someone visits a book a second time, they'll see the old version until they visit another page?  Why would the caching code display the old version if the hash has changed?  I would much prefer that this doesn't affect normal online behavior.

jasonwilliams:
> I'm a bit concerned about the reloading thing. It seems that whenever someone visits a book a second time, they'll see the old version until they visit another page? Why would the caching code display the old version if the hash has changed? I would much prefer that this doesn't affect normal online behavior.

Let me tackle this one.
When you come back to the site the new service worker downloads and takes over. Then in the background it will perform an update, we use stale while revalidate so it will continue to show you the old version while it does that. Then you can refresh to have the latest.

The only way around this is if we use JavaScript to reload the page if there’s been a update, would you prefer that?

OrangutanGaming:
Allowing either the user or the server (through `book.toml` probably) to decide if they want to cache assets might be important for some books where they may have a large amount of assets but they could either be mandatory or just minor enhancements.

Also, taking an “offline-first” approach usually solves problems, like if the device believes there’s a network connection despite it not being able to get any responses (called “Lie-Fi” in https://youtu.be/cmGr0RszHc8), so fetching from the cache then updating is probably much better than fetching then falling back to the cache, even if you choose to automatically refresh the page due to a newer version being fetched.

jasonwilliams:
> Also, taking an “offline-first” approach usually solves problems, like if the device believes there’s a network connection despite it not being able to get any responses (called “Lie-Fi” in https://youtu.be/cmGr0RszHc8), so fetching from the cache then updating is probably much better than fetching then falling back to the cache, even if you choose to automatically refresh the page due to a newer version being fetched.

Yeah I second this.
The amount of times my device tries to fetch (when there's none or little connection) and could actually just give me a cached version while I'm waiting, it can be frustrating. 

The trade off is this (both assume the user has previously visited):

A: **StaleWhileRevalidate** - User visits the documentation page, it will load straight away, user will see the same content they saw before, page updates in the background ready for the next refresh/link click. This means the user needs to refresh to get the latest update.

B: **Fetch First** - Page will attempt to pull down fresh copy, if user has poor net connection or offline it may take a while for the device to "give up" and fallback to cache. This can be annoying, but you're guaranteed to see the latest version.

Personally I like A, If "double refresh" is still an issue we can create a pop-up which tells the user their book has updated and they need to refresh to see the latest version.
I'm not convinced that needs to be solved in this issue though.

OrangutanGaming:
I think A is much better as you can automatically update if you want to by just automatically updating the page (effectively getting the positives of both) but it also gives what I think is the best update system for the user: some kind of “update is available _refresh_” pop up. This tells the user 2 things:

An update is available so they can choose to refresh if they want or if they are just in a rush for a specific bit of information they can leave it for next time

That the page is cached or at least somehow stored locally

This also can be more important when they first connect with no connection then gain a connection as the pop up will just show while browsing and they can just refresh.

OrangutanGaming:
I guess this functionality is more standard in a PWA so it probably can be delayed. In a book context, the user may find an automatic update while reading strange and unexpected, especially if they were working through a code example then suddenly the example completely changes. Since usually when reading through a book the pages are shorter and you change page fairly often, this can probably be skipped, but could be a nice PR once this is merged, especially if the update functionality is behind a config option. Assets are probably more important for books since they could make or break someone even considering using offline.

jasonwilliams:
@ehuss do you have any thoughts on the above?

ehuss:
I don't think having a reload every time the page is updated is a great user experience. For sites that are updated frequently, this means the user will be frequently greeted with a reload. Having a notice appear that the content is out-of-date with a "refresh" link might be a better option (assuming this notice only appears if the content changes). I've seen that in a few other sites, and seems reasonable (although still slightly annoying).  

Unfortunately I don't really know much about service workers, and so it is really difficult for me to understand what is going on, what possible problems may arise, or how to debug them, or how to fix them. For example, my iOS device is still not working, and I have no idea how to fix it. 😦

nihaals:
Depending on the users, you may actually want instant refreshes, for example if it's a book for a closed group that is mostly designed for offline use, where having it updated as soon as possible is fine, but I agree that for most books having a pop up saying an update is available is the best for the user. Implementing the latter first then releasing instant refreshes in a future version configurable by an option in `book.toml` would be a nice roadmap in my opinion.

nihaals:
What's currently blocking this from being merged? Having it use the cache first then fetch and give a refresh pop up?

nihaals:
It looks like fontawesome's acting weird. With "Lie-Fi", it takes 2 failed requests for the SW to return and since it for some reason isn't being cache properly it shows invalid icons every time you go to a new page. Not sure if this has been fixed since the version you have on your hosted version, but seems odd. It loads fine completely offline.

Also, not sure if it's worth adding a config table for custom regex, definitely has a use but you can just make your own `sw.js` (although would miss out of possible future updates just to add new paths). Haven't tested yet but #1253 might result in chapters not being cached too

jasonwilliams:
> What's currently blocking this from being merged?

Not much just testing.
I’m not going to keep rebasing it if the owners of this project don’t want it though so there needs to be some plan.

I can fix issues but if there’s no response there’s not much I can do.

nihaals:
Only issues I've had while testing is FA afaik. Knowing when it's done caching/refreshing the cache might be helpful on mobile (my favourite is https://stripe.com/docs/api which has a little icon at the top, although I don't know how it behaves if it's stale). Since it would be helpful with my project I'd even be willing to maintain a fork and keep rebasing/merging if the devs don't find this useful (judging from the reactions in the issue it didn't seem like a priority but since it's already implemented I'm not sure).

Are you expecting issues to be found? I feel like it would mostly be feature requests that can be added after an initial release (e.g. the popup when updated)

jasonwilliams:
@nihaals the first issue we need to sort out is StaleWhilerevalidate vs “fetch first”. I don’t think we made a decision on this. 

Personally I don’t mind SWR but @ehuss had issues with it. We need to make a decision on this to move forward, otherwise we’re at an impasse. 

I think outside of that it was fine (for me) but @ehuss hit some bugs. I couldn’t reproduce any of them so we we’re stuck.

Without any of the maintainers acknowledging or testing this PR it will never get merged.

nihaals:
Just off the top of my head
## SWR Pros/FF Cons
- "Lie-Fi" is usable
  - Your project might aim for users more likely to be affected by this or they need to prioritise it displaying content
  - The key part of one of the talks I linked is thinking "offline first", assuming good internet connection may harm the UX of users without
- Loads faster
  - Something that benefits both online and offline users
  - Possibly one of the biggest factors that can be affected by a SW that makes people more likely to see/read the book
- Can choose between fresh (refreshing when seeing the pop up) and speed (e.g. you just need to quickly check something)
  - If you find yourself always needing fresh data (can't think of an example apart from something news related?), you can change to fetch first in the SW, but switching to offline first is more involved due to a possible popup
## SWR Cons/FF Pros
- Receive stale content
  - In my opinion, not a big deal
  - Individual pages are unlikely to be huge so they would get fresh content soon after
  - A simple pop up saying they can refresh means if they need to prioritise fresh they can

I think double refresh is a hybrid approach but hurts UX and takes away the 2nd and 3rd SWR Pros

I haven't tested on iOS yet but I don't see why it wouldn't work, although workbox is fairly out of date now, there's a new major version, in fact v6 is currently in its 2nd alpha release

nihaals:
> How do they handle when content changes?

Pretty sure the standard is to show a little popup at the bottom saying there's a new version and they can update by refreshing (sometimes with a refresh button)

> This hasn't been my experience when using: [...]

I agree, this seems to be the most popular way and in my opinion the best (using the approach above). I don't think I've seen anyone do a double reload

> Having the end users also opt-in is a bit overkill and would involve some UX explaining to them what they're opting into. I think there's risk of it being confusing

I agree, offline websites is somewhat new, and for most users it definitely is. In one of the talks I linked they said that they hope having something like "this website can now be used offline" will fall out of fashion as eventually most websites will be like this, but at the moment, it's useful to simply tell the user as they probably won't *expect* offline

> Is it having problems with the FA fonts?

I also had this issue (https://github.com/rust-lang/mdBook/pull/1000#issuecomment-674434247)

> [release behind a flag]

I think this would be the best approach as creators may need to be aware of what it is doing due to them possibly having little exposure to SWs. I'm sure when people see the option added in the change log they'll be intrigued and want to test it which will help get more testers (although there shouldn't be many issues as the library used is popular and well tested)

> How do I clear the web worker cache on mobile safari?

I'm not sure as I don't use Safari, but they're dev tools seem to be lacking (which doesn't seem rare). This [support documentation](https://developers.google.com/web/ilt/pwa/tools-for-pwa-developers#interact_with_service_workers_in_the_browser) suggests that there are not Safari dev tools for SWs (although it does warn that it is possibly out of date and could be added in Safari 14). As I said above, browser unique bugs should be extremely rare and would most likely be a library issue or even a browser bug. The current library version is fairly out of date (https://github.com/rust-lang/mdBook/pull/1000#issuecomment-674454187) so updating could help if you do run into any issues

> Is it normal to have lot of errors in the console (online and offline)?

The only errors I have are relating to FA (but I believe this issue may be elsewhere, I'm not sure if it's due to a bad service worker. I'm really unsure what's causing the FA issue, I haven't used it much myself but it the setup seems fine. Possibly updating could help but I haven't tested yet

> Does chrome require a hostname or should an IP work? Or, in other words, how do I get this to work locally with chrome?

localhost definitely works, not sure if local files is part of the spec but probably safe to use a server

> For example, let's say there's a blog post, "hey, check out this new Rust feature" with a link to the docs, and a bunch of users visit that, and the docs aren't there, because they have an old cache

I didn't really think about that and it's a great point. If the page is new, the 404 shouldn't get cached so it would be fine, and if it was something like a changelog page that gets updated, I do agree that it could be confusing, however, with an update pop up I think this could be solved. The Google pop ups aren't obnoxious but are still easily visible which is probably an important balance. If someone just navigates to another page though it would update, so stale content would only be displayed on the first page they see.

It seems most of your issues are UX related. Would fixing the FA problem and having the "update is available" pop up solve those? I think the pop up could make a huge difference and could be useful to actually test on users unaware of the feature to see how well it performs (maybe once this is merged, PRing it into the Rust Book and seeing if that community notices it, finds it annoying and think it's worth the benefit of decreasing load times and having offline access. Even if a feature wasn't used in The Book I still think it could have value for some mdBook users (for example I want to create a book that would primarily be used in places where there wouldn't be great internet access or needs to be accessed no matter what, so this trade off would be trivial to me, especially when stale content would not cause major problems, another type of user who could find this useful)

jasonwilliams:
If you're referring to my site ignore the fonts errors, i think that's my server not handling that file type, I don't think its related to offline at all.

nihaals:
I was, but there was also strange FA *behaviour*, not just console errors, but I haven't built the PR and tried hosting it myself yet

ehuss:
> Would fixing the FA problem and having the "update is available" pop up solve those?

I still feel like this would not be a good user experience.  For docs that update rapidly, this would show up essentially every day they visit.  For users who read the docs sporadically (say once a month), they will essentially run into this every time they visit, which I feel is a bad experience.

Would it be possible to have it be more network-centric so that the current behavior is retained, and the "offline" behavior on engages when needed?  For example, can it:

* Always load from the network first, and display that page.
* If that load fails (after some relatively short timeout), consult the offline cache and display that.
* To accommodate the flaky network issue, when it goes into offline mode, switch to loading directly from the offline cache immediately (like this PR works), and then switch back after some period of time (say 24 hours).

Another question, I'm not sure if I asked before, but is there a way to "unload" the service worker?  Let's say someone enables this, and then decides to turn it off.  Will it be permanently be enabled for any user who visited the site before it was disabled?

jasonwilliams:
I’m ok to try “network first”. It’s not “great” when you have no network because some browsers hang for a bit before falling back to the cache but it’s probably the closest we’ll get to pleasing both scenarios.

> Another question, I'm not sure if I asked before, but is there a way to "unload" the service worker? Let's say someone enables this, and then decides to turn it off. Will it be permanently be enabled for any user who visited the site before it was disabled?

I think we can have a switch for offline on and off, if you turn it off it can unload the service worker. However I don’t think the switch is needed if we’re doing “network first” as it will make no difference to your experience as you’re fetching from the network anyway.

So we need to make some decisions here:
- If we do offline “cache first” buy default, then yes a switch would work. The user can switch it on or it can be ok by default set by the author.
- If we’re fetching from the network first anyway a switch is pointless and only adds confusion. It’s a bit like having a switch to use the HTTP cache in your browser on a website.

So maybe we can pick which direction we prefer from these?

nihaals:
If the goal is to have both online first and offline first implemented, I think having offline first as the default would be better as it's definitely where most projects seem to be going. I'm not sure how easily you could describe this switch though, you may need a short 2 sentences describing each option.

After reading a [web.dev post](https://web.dev/stale-while-revalidate/) about pretty much this problem, I thought of another idea which would also have a more user friendly configuration screen. What if we use offline first, but have 3 states:

1. The content was retrieved less than `CONFIG_OPTION_1` [minutes] ago. This means the content will be served straight from the cache. That's it. (This state could be removed but would alleviate some server time.)

2. The content was retrieved less than `CONFIG_OPTION_2` [minutes] ago (and more than `CONFIG_OPTION_1`). This means the content will be served straight from the cache, but there will also be a network request to fetch new content and refresh the cached content if needed. This would also invoke the standard "new content, refresh" pop up.

3. The content was retrieved more than `CONFIG_OPTION_2` [minutes] ago. This means the cached content is "very" stale and will be discarded. No cached content will be displayed and new content will be retrieved (as if it is their first time loading the site).

Having both of these options user configurable and book author configurable means:
- For people writing very dynamic books, they can either set (the default values of) `CONFIG_OPTION_1` and `CONFIG_OPTION_2` to the same value (maybe even 0) which means most of the time only fresh content is shown.
- For people writing books that want long cache times and aren't very dynamic, they can set `CONFIG_OPTION_1` and `CONFIG_OPTION_2` to larger numbers (or even -1, although that shouldn't really be allowed for `CONFIG_OPTION_1`) so users will get their cached content and get the pop up when there is an update.


- `CONFIG_OPTION_1` <= `CONFIG_OPTION_2`
- `CONFIG_OPTION_1` != `-1` (to avoid books *never* being updated)
- `CONFIG_OPTION_2` *can* equal `-1` (standard offline first behaviour)

ehuss:
> I’m ok to try “network first”. It’s not “great” when you have no network because some browsers hang for a bit before falling back to the cache but it’s probably the closest we’ll get to pleasing both scenarios.

If it can do something like I described above, I think the delay should only be visible once over some time period (say 24 hours).  I'm uncertain how often that would happen in practice.  What kind of scenarios would cause it to go offline?  

> If we’re fetching from the network first anyway a switch is pointless and only adds confusion

Yes, my intent was that if it can automatically switch to offline mode, there would not be a need for a user-visible switch.

My question was more for the publisher side. If I publish a book with this turned on, and it causes problems for my readers, I want to turn it off for everyone.  AIUI, if you just disable the config option, it doesn't disable the service worker, and there isn't a practical way for users to unload the service worker, leaving it permanently running on their system.  Would it be possible, if the option is turned off, if there was some javascript that removed the service worker if it exists?

nihaals:
@ehuss What do you think about https://github.com/rust-lang/mdBook/pull/1000#issuecomment-687874288?

ehuss:
> What do you think about [#1000 (comment)](https://github.com/rust-lang/mdBook/pull/1000#issuecomment-687874288)?

I guess I don't understand the difference between option1 and option2.  Does workbox.js always issue a request in the background?  I thought it would always check.  What would be the purpose of completely disabling that background request? And for option 3, what would be the purpose of never showing the cached entry? Isn't that really important, like the whole point of offline viewing? I would expect it to show the cached content for as long as possible if it is unable to update.

Also, I may be a bit unclear by what you mean by "user", as there are two different users, the reader and the publisher. Perhaps it would be best to use that terminology when talking about the interaction.  I would personally prefer to avoid complicating things for the reader as much as possible.

nihaals:
Some pseudo-code that should hopefully explain it better:
```python
def fetch_and_cache(path: str) -> Content:
    # Probably wouldn't actually use a path when refreshing as the entire book should be cached (config option to only cache visited pages in the future?)
    content = get(path)
    set_cache(path, content)  # store content with path and timestamp
    if content_already_displayed():
        # Specifically for the background refresh, true if the user can already see content but this is still running
        # Would also check if the fresh version is actually different to the previous version
        show_popup('Update available, refresh to see')
    return content

def get_content(path: str, config_option_1: int, config_option_2: int) -> Content:
    """Gets content at path respecting cache options"""
    if config_option_1 > config_option_2:
        raise ValueError()
    if config_option_1 <= -1:
        raise ValueError()
    if config_option_2 < -1:
        raise ValueError()
    if config_option_2 == -1:
        # Simplify this snippet's logic
        # Would actually behave like infinity
        config_option_2 = 2^128

    cached_content, timestamp = get_cached_content(path)  # timestamp is a UNIX timestamp for simplicity
    if cached_content is None:
        return fetch_and_cache(path)
    age = now() - timestamp
    if age < config_option_1:
        # Only show cached content, no cache refresh
        return cached_content
    if config_option_1 <= age < config_option_2:
        # Concurrently refresh cache and show cached content (background fetch)
        run_in_background(fetch_and_cache(path))
        return cached_content
    if age >= config_option_2:
        # Fresh content *only*
        set_cache(path, None)
        return fetch_and_cache(path)
```

Publishers would be able to set default values for `config_option_1` and `config_option_2` which the reader can change if they want

> I guess I don't understand the difference between option1 and option2

State 1 will only show cached content, state 2 will show cached content while also refreshing the cache (useful if they are quickly clicking through pages to remove some origin server load, expected value (`config_option_1`) around a minute)

>  Does workbox.js always issue a request in the background? I thought it would always check.

I believe there is a lower level API where you have more control, otherwise you can simply not use workbox and write the serviceworker directly

> What would be the purpose of completely disabling that background request?

To remove unneccesary server load and egress (anyone could benefit from having this at a few seconds as it's unlikely the server updating within those 5 seconds will be important since they are already browsing). It would also help the client if they have a bandwidth cap like on cellular data (refreshing the cache every 5 seconds is already probably more often than they would want so if the publisher sets a low value the reader can increase it)

> And for option 3, what would be the purpose of never showing the cached entry? Isn't that really important, like the whole point of offline viewing? I would expect it to show the cached content for as long as possible if it is unable to update.

State 3's user experience doesn't feel that much different from online first, many users on mobile stop loading a page after a few seconds and the timeout for online first would need to be high enough to be useful. A possible extra boolean config option could be added to change state 3 (or just replace the state 3 in the above snippet with this one if it doesn't seem useful):
```python
    if age >= config_option_2:
        # Force fetch new content with timeout, fall back to cache
        with timeout(seconds=5):
            return fetch_and_cache(path)
        return cached_content
```
I think this is probably the best way to implement the advantage of online only since the only benefit is showing fresher content after a period of inactivity (which works well with the final state)

ehuss:
> To remove unneccesary server load and egress

Does this PR not use the same caching logic as the browser normally does?  I would expect it to have the same behavior as normal online behavior (doing whatever a browser normally does to determine if it needs to access the network and checking if a page is out-of-date).

nihaals:
You can't cache the initial request without SWs AFAIK, so normal online behaviour doesn't exist (for the actual pages)

In this implementation, any normal browser caching stuff should be bypassed as that would cause conflicts and you wouldn't have control over it. By having it under a config option, the publisher and reader can actually control the no fetch caching of pages (which I think is objectively better?)

sanmai-NL:
@ehuss Please note [there is a significant use case for this](https://github.com/rust-lang/mdBook/issues/546#issuecomment-724000747).

jasonwilliams:
- if someone can rebase this PR to my branch that would be great, If not I don’t know when I’ll get around to doing it.
- I think we should allow the publisher to choose swr or network first in their settings and support both. If they choose swr then we can offer a pop up when the data has updated. This means we don’t need to set arbitrary times. I don’t mind defaulting to network first.
I’ve taken a look at many rust books and most of them don’t even update weekly so realistically people won’t get this pop up very often, barely at all. Remember that it will also update in the background.

sanmai-NL:
@jasonwilliams I don't have write access..

jasonwilliams:
@sanmai-NL you should be able to clone https://github.com/jasonwilliams/mdBook/tree/offline_support and rebase it against upstream then do a pull request back up

nihaals:
Just saw this

image

It reminded me that client side apps already have something like the update pop up we were talking about and I don't think that causes many UX problems (although this is only if you leave the tab open during an update, if you go to it for the first time in a while you still get the new assets unless they also have a service worker)

jasonwilliams:
@nihaals if the publisher opts in to offline first we can implement that option.

jasonwilliams:
I still believe there is a significant use case for mdbooks to support offline usability. But won't have time to catch up on this now, im happy if someone else wanted to pick up on this from where I left off

GuillaumeGomez:
I opened https://github.com/rust-lang/mdBook/issues/3201 so we can discuss about it. Thanks for the preliminary implementation in any case. Closing since it's now very outdated.

PR #1003, opened 2019-08-14
Implement a `rustdoc_include` preprocessor

### LINKED ISSUE #618: Partial include should show full file instead of stub when showing hidden lines.

carols10cents:
Fixes #618 

This would make work on The Book to extract code listings into external, complete, testable examples easier.

I'm going to mostly let the documentation speak for itself; if it doesn't completely explain this feature, I'd like to fix it!

I'm also totally open to a different name for this feature 🤷‍♀

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ matrix:
       env: TARGET=x86_64-unknown-linux-gnu
     - rust: nightly
       env: TARGET=x86_64-unknown-linux-gnu
-    - rust: 1.34.0 # Minimum required Rust version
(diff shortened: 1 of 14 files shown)


carols10cents:
Range `.contains` was stabilized in 1.35, so I bumped the minimum supported Rust version from 1.34 to 1.35. Let me know if that's not OK and I'll rewrite where I use `.contains`.

maisiliymBridj:
Is there a tool to read those markdown files from an editor? This makes reading documentation much more difficult for people who cant afford to run an addidional operating system (web engine) just to read docs.

PR #1021, opened 2019-09-06
Upgrade pulldown_cmark to 0.6

marcusklaas:
This updates pulldown_cmark to the latest release version, which has some breaking changes.


ehuss:
Thanks so much for updating this!  

I ran into an issue taking it for spin: https://github.com/raphlinus/pulldown-cmark/issues/398  I'm a little concerned about merging this when it panics on a couple rust-lang books.  Can we maybe wait for a fix for that?

marcusklaas:
Fixing the panics before upgrading seems wise indeed. Thanks for reporting these. I'll get right on them.

marcusklaas:
I'll update this PR to resolve the conflict.

marcusklaas:
Done!

PR #1025, opened 2019-09-13
Update to notify-5.0.0-pre.1

XAMPPRocky:
this PR updates the `notify` dependency to `5.0.0-pre.1`. Normally I think we should wait for a stable version of dependencies, however `notify-4.0` which depends on `fsevent-0.4` can no longer be built on macOS which means that `cargo install mdbook` now always fails. The watch and serve feature is very important for the CLI tool so I think we include this dependency until a more suitable solution is found. At the very least people can clone this fork to build their mdbook on macOS until then.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,6 +30,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "anymap"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "atty"
 version = "0.2.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


ehuss:
Thanks for the PR!  I'm trying to understand what's going on. It looks like `notify` is maybe transitioning to a new maintainers?  And it looks like the offending `fsevent` release has been yanked, so maybe this isn't necessary right now?

XAMPPRocky:
@ehuss Yeah, with the `fsevent` release being yanked this is no longer needed.

Dylan-DPC:
@ehuss. Yes notify is currently under transition. A new organisation has been created but the repo and other related shenanigans are yet to be moved or updated. But... It will survive, it will survive! 😆

Issue #1024, opened 2019-09-13
How to document a project with files in source hierarchy with mdbook


andrewdavidmackenzie:
I have a project with a number of markdown files throughout the source code hierarchy, and I also have a "book" in a docs directory.

```
+---subproject1
|    +---README.md
+---subproject2
|    +---README.md
+---docs
     +--book.toml
     +--src/
        +--SUMMARY.md
```

I would like to document this project using mdbook, with the ability to refer to markdown files in the source directories from markdown files under /docs.

I understand (and tests seem to confirm it) that I cannot refer to markdown files via links that are NOT under the "src" directory specified for the book. This doesn't produce a build failure, but the links fail in the deployed book on gh-pages.

My Hack
=
So, I have a Makefile to build it that copies the markdown files *outside* the docs directory into the docs/src directory, preserving directory structure, and then I can refer to them using links in the other markdown files I write in my book.

Resulting in this:
```
+---subproject1
|    +---README.md
+---subproject2
|    +---README.md
+---docs
     +--book.toml
     +--src/
        +--SUMMARY.md
        +---subproject1
         |    +---README.md
        +---subproject2
        |    +---README.md
```

Alternative
=
So, I tried to organize my project where the "src" file is the root, and I should be able to refer to any file in the entire hierarchy.

```
+---subproject1
|    +---README.md
+---subproject2
|    +---README.md
+--book.toml
      (src = ".")
+--SUMMARY.md
```

In the docs on [mdbook](https://rust-lang-nursery.github.io/mdBook/cli/build.html) it states: 
`Note: Make sure to run the build command in the root directory and not in the source directory.`

When I try to build this, it fills the disk on my machine (:-) ), which is probably why it's recommended against.

Is there a way to do this without my original hack?

ehuss:
You can refer to files outside of the book, it just depends on how it is deployed.  For example, the rust-lang books all deploy to https://doc.rust-lang.org/.  Each book can do relative links to other things outside of the book with relative links. For example, [the reference](https://doc.rust-lang.org/reference/) links to the [standard library docs](https://doc.rust-lang.org/std/) with links like `../std/primitive.char.html`.  The exact structure will depend on how you deploy your documentation.

So, I think for your example, you would keep your original structure, and have a build script which builds the book, and then copies all the relevant pieces to a directory for final packaging and deployment.

Alternatively you can `{{#include}}` files from outside the book as well.

andrewdavidmackenzie:
[...]

PR #1026, opened 2019-09-14
Avoid a possible recursive copy when destination (build_dir) is underneath source directory

andrewdavidmackenzie:
If the user makes a mistake in config `mdbook` could have started a never-ending copy that only ends when your disk is full. This change avoids that.

It also makes it possible to create a valid (if not recommended) configuration where your `build-dir` is underneath the `src` directory. One reason for possibly wishing to do that is when you desire `src` to be the root directory of a project.

Relates to #1024


andrewdavidmackenzie:
Any feedback on this PR?

ehuss:
I'm having trouble figuring out the configuration where this is a problem.  If I set `build.build-dir` to `src/foo`, I don't see any problems happening.  What is a configuration where it causes recursion?

andrewdavidmackenzie:
When you have the book.toml file in the root of your project.

in book.toml you have `src = "."`

and the build dir is also under the root somewhere, something like

```
[build]
build-dir = "target"
```

So, during the build all the contents of "src" (so all files - except filtered .md files) are copied into "target", including "target" itself - recursing.

in short: when "build-dir" is a subdir of "src".

src/utils/fs.rs
```diff
@@ -113,12 +115,18 @@ pub fn copy_files_except_ext(
         let entry = entry?;
         let metadata = entry.metadata()?;
 
-        // If the entry is a dir and the recursive option is enabled, call itself
+        // If the entry is a dir and the recursive option is enabled , call itself
```
ehuss: Can this extra space be removed?

src/utils/fs.rs
```diff
@@ -95,13 +95,15 @@ pub fn copy_files_except_ext(
     from: &Path,
     to: &Path,
     recursive: bool,
+    avoid_dir: Option<&PathBuf>,
```
ehuss: This is a public API, so this is a breaking change.  However, I don't see anyone externally using it.  

@Dylan-DPC, how do you feel about being stringent about API compatibility like this?

book-example/src/cli/build.md
```diff
@@ -35,5 +35,6 @@ not specified it will default to the value of the `build.build-dir` key in
 
 -------------------
 
-***Note:*** *Make sure to run the build command in the root directory and not in
-the source directory*
+***Note:*** *The build command copies all files (excluding files with `.md` extension) from the source directory
+into the build directory. If the build directory is under the source directory it is not copied as this would
```
ehuss: I don't think this second sentence is necessary.  I don't think there is a need to call out all the incorrect behaviors it tries to avoid.

src/utils/fs.rs
andrewdavidmackenzie: Let me know what you want to do here - don't see an easy way to make this PR work without this or a similar change.... so when OK-ed by you guys I'll rectify these three comments.

andrewdavidmackenzie: @ehuss I should have some time this week and would like to get back to this PR and close it if possible. Can you let me know what you want to do about the API and breaking change in order to resolve it? Thanks

ehuss: Hm, I think it would be ok to just change it. We can change it back if it causes problems for anyone.

andrewdavidmackenzie:
OK, I hope that addresses the two comments on texts in the PR.

andrewdavidmackenzie:
Thanks. I'll look out for the next release to switch back to this from my fork.

PR #1033, opened 2019-09-22
Log deserialization errors for [html.config]

### LINKED ISSUE #1032: Do not ignore `HtmlConfig` syntax errors silently
`HtmlConfig` parsing error are silently ignored and the default values are used. This makes it hard for users to troubleshoot typos in their `book.toml`.

I tried this `book.toml`:

```toml
[output.html]
default-theme = "navy"
curly-quotes = "true"
# should have been: curly-quotes = true
```

*I expected to see this happen*: mdBook displays a warning or error message when building if the configuration contains a syntax error, like
`[WARNING] (mdbook::book): Configuration value for 'output.html.curly-quotes' is invalid, expected a boolean but got string "true".`
It applies the valid config nodes and sets the `default-theme` to `navy`, but uses the default value for `curly-quotes`.

*Instead, this happened*: The build succeeds without any warning, and ignores the valid config nodes under `output.html`. So the `default-theme` falls back to default.

## Meta

mdBook version b4bb44292d707c720dc75851500b3249d3a93ad3

TjeuKayim:
Fixes #1032 

Log deserialization errors in `Config::html_config()`, instead of ignoring them.
After implementing this, the log appeared three times, caused by multiple calls to `html_config()`. Therefore, I de-duplicated that code.

## Example
Using this `book.toml`:
```toml
[output.html]
default-theme = "navy"
curly-quotes = "true"
# should have been: curly-quotes = true
```
The build log displays the error:
```
[~/src/mdBook]% cargo run -- build ./book-example
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/mdbook build ./book-example`
2019-09-22 21:19:14 [INFO] (mdbook::book): Book building has started
2019-09-22 21:19:14 [INFO] (mdbook::book): Running the html backend
2019-09-22 21:19:14 [ERROR] (mdbook::utils): Error: Parsing configuration [output.html]
2019-09-22 21:19:14 [ERROR] (mdbook::utils):    Caused By: Couldn't deserialize the value
2019-09-22 21:19:14 [ERROR] (mdbook::utils):    Caused By: invalid type: string "true", expected a boolean for key `curly-quotes`
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -62,6 +62,7 @@ use toml_query::insert::TomlValueInsertExt;
 use toml_query::read::TomlValueReadExt;
 
 use crate::errors::*;
+use crate::utils;
 
 /// The overall configuration object for MDBook, essentially an in-memory
 /// representation of `book.toml`.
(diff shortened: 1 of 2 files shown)


ehuss:
Thanks!

PR #1035, opened 2019-09-24
Added line numbers to editable sections of code.

### LINKED ISSUE #1028: How to add line numbers to default playpen?
I notice that the Ace editor can be configured to show/hide line numbers by setting `showLineNumbers`, but `mdbook` explicitly [sets this to false](https://github.com/rust-lang-nursery/mdBook/blob/master/src/theme/playpen_editor/editor.js#L13).

I couldn't see a way to pass configuration from my `mdbook` project's `.toml` to override this. IMO it would be a big improvement if book authors could configure their editor to show or hide line numbers.

andymac-2:
- Added line numbers to config struct
- Added playpen_line_numbers field to hbs renderer.
- Added section to set `window.playpen_line_numbers = true` in page template
- Use line number global variable to show line numbers when required.
- Updated docs

Fixes #1028
Fixes [Line Numbers for Code Editor #1193](https://github.com/rust-lang/rust-by-example/issues/1193)

### DIFF
--- a/book-example/book.toml
+++ b/book-example/book.toml
@@ -9,6 +9,7 @@ mathjax-support = true
 
 [output.html.playpen]
 editable = true
+line-numbers = true
 
 [output.html.search]
 limit-results = 20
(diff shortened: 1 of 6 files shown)


ehuss:
Thanks!

PR #1037, opened 2019-09-25
Automatically use a dark theme according to 'prefers-color-scheme'

Flying-Toast:
This PR makes mdBook automatically use a dark theme if [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) is set to 'dark'.

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -43,7 +43,7 @@
         <!-- Provide site root to javascript -->
         <script type="text/javascript">
             var path_to_root = "{{ path_to_root }}";
-            var default_theme = "{{ default_theme }}";
+            var default_theme = window.matchMedia("(prefers-color-scheme: dark)").matches ? "{{ preferred_dark_theme }}" : "{{ default_theme }}";
         </script>
 
         <!-- Work around some values being stored in localStorage wrapped in quotes -->
--- a/book-example/src/format/config.md
+++ b/book-example/src/format/config.md
@@ -150,6 +150,8 @@ The following configuration options are available:
   files with the ones found in the specified folder.
 - **default-theme:** The theme color scheme to select by default in the
   'Change Theme' dropdown. Defaults to `light`.
+- **preferred-dark-theme:** The default theme to use if the browser
+  requests the dark version of the site. Defaults to `navy`.
 - **curly-quotes:** Convert straight quotes to curly quotes, except for those
   that occur in code blocks and code spans. Defaults to `false`.
 - **mathjax-support:** Adds support for [MathJax](mathjax.md). Defaults to
--- a/book-example/src/format/config.md
+++ b/book-example/src/format/config.md
@@ -218,6 +218,7 @@ description = "The example book covers examples."
 [output.html]
 theme = "my-theme"
 default-theme = "light"
+preferred-dark-theme = "navy"
 curly-quotes = true
 mathjax-support = false
 google-analytics = "123456"
--- a/book-example/src/format/config.md
+++ b/book-example/src/format/config.md
@@ -150,8 +150,10 @@ The following configuration options are available:
   files with the ones found in the specified folder.
 - **default-theme:** The theme color scheme to select by default in the
   'Change Theme' dropdown. Defaults to `light`.
-- **preferred-dark-theme:** The default theme to use if the browser
-  requests the dark version of the site. Defaults to `navy`.
+- **preferred-dark-theme:** The default dark theme. This theme will be used if
+  the browser requests the dark version of the site via the
+  ['prefers-color-scheme'](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme)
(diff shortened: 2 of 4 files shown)


ehuss:
Thanks!  Can you also update the documentation for the new option at https://github.com/rust-lang-nursery/mdBook/blob/master/book-example/src/format/config.md#html-renderer-options?

I'm also thinking, if a book defines its own themes, this will break them if they don't have a theme named "navy".  Perhaps it should somehow detect if the theme doesn't exist?

Flying-Toast:
@ehuss I've updated the documentation.
What should the behavior be if the navy theme doesn't exist? Do you think it would make sense to include a dark theme by default with the book, like what is done with 'light'?

Flying-Toast:
Also, why is the CI build failing from a file not touched by this PR? Should I fix that to make the build pass, or leave it as-is?

Flying-Toast:
@ehuss I've added the new option to the documentation.

Isn't the current behavior for a missing 'navy' theme the same as for missing the default 'light' theme (i.e. removing the `.light {}` rule from variables.css)?
What if we created a new theme called 'dark' (or renamed an existing one) to go with the default 'light' theme? That would make it more obvious that the theme is a generic/default one, rather than arbitrarily choosing one of the dark themes as the default.

Also, it seems the CI build is failing from a file not touched by this PR. Should I fix that to make the build pass or leave it as-is?

ehuss:
> Isn't the current behavior for a missing 'navy' theme the same as for missing the default 'light' theme (i.e. removing the `.light {}` rule from variables.css)?

Presumably if someone customizes the themes, they either have a "light" theme, or change the default in the config.  

> What if we created a new theme called 'dark' (or renamed an existing one) to go with the default 'light' theme? That would make it more obvious that the theme is a generic/default one, rather than arbitrarily choosing one of the dark themes as the default.

That might be a little clearer, but I would prefer not to make intrusive changes like that.  Can the javascript just query if the default dark class is defined, and if not just fall back to the default theme (as-if the `prefers-color-scheme` query is false)?

> Also, it seems the CI build is failing from a file not touched by this PR.

Looks like the new rust release today had a minor rustfmt change.  I went ahead and fixed it, so no need to do anything about it.

Flying-Toast:
I can't figure out how to detect if a theme is defined or not in javascript. How would you recommend doing that?

Flying-Toast:
I can't figure out how to detect if the theme is implemented from within javascript. How would you recommend doing that?

src/theme/index.hbs
```diff
@@ -43,7 +43,11 @@
         <!-- Provide site root to javascript -->
         <script type="text/javascript">
             var path_to_root = "{{ path_to_root }}";
-            var default_theme = window.matchMedia("(prefers-color-scheme: dark)").matches ? "{{ preferred_dark_theme }}" : "{{ default_theme }}";
+            var default_theme = window.matchMedia("(prefers-color-scheme: dark)").matches &&
+                // Fallback to the default theme if preferred_dark_theme isn't available
+                Array.from(document.querySelector('link[href="{{ path_to_root }}css/variables.css"]').sheet.rules)
```
Flying-Toast: @ehuss This is the only way I can find to detect if the theme is available or not. It's *really* janky though (it searches through the stylesheets for a rule affecting the theme). Is there a different way you had in mind/know of?

Flying-Toast: Or, instead of falling back to the light theme from within JS, we could just have a note in the docs to set `preferred-dark-theme` to the same value as `default-theme` to disable it.

> Presumably if someone customizes the themes, they either have a "light" theme, or change the default in the config.

Could that same logic also apply to `preferred-dark-theme`?

Flying-Toast:
What if we have a note in the docs to set preferred-dark-theme to the same value as default-theme to disable it?

> Presumably if someone customizes the themes, they either have a "light" theme, or change the default in the config.

Could that same logic also apply to preferred-dark-theme?

Flying-Toast:
@ehuss What if we have a note in the docs to set preferred-dark-theme to the same value as default-theme to disable it?

> Presumably if someone customizes the themes, they either have a "light" theme, or change the default in the config.

Could that same logic also apply to preferred-dark-theme?

ehuss:
> @ehuss What if we have a note in the docs to set preferred-dark-theme to the same value as default-theme to disable it?

That would be a good thing to add to the docs.

But I'd still prefer not to make a breaking change for anyone that has custom themes.  It does seem annoyingly difficult to query from javascript.

How about a compromise, for now set the default to the same as `default-theme`, and then in 0.4 we'll switch it to `navy`?  Anyone who wants to try it out now can just manually set their config.  How does that sound?

Flying-Toast:
@ehuss Okay, sounds good. I've changed it as you suggested.

ehuss:
Thanks!  I'll open an issue to track changing the default.

PR #1045, opened 2019-10-02
Make serve and watch respect .gitignore

fmanco:
When editing files with some editors (e.g. vim) temporary files are
constantly being modified causing mdbook serve and watch to constantly
rebuild the project.

To avoid this leverage the ignore crate to load the project's root
.gitignore file and skip changes on files ignored.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,6 +2,14 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
+version = "0.6.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "memchr 2.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "aho-corasick"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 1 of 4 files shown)


ehuss:
Sorry, I'm a bit confused. Was it just coincidence that this and #1044 were opened within a few hours of each other?

segfaultsourcery:
@ehuss It's likely a hilarious coincidence. I did open a thread on /r/rust asking for input (I'll be making good use of that advice in a bit), so maybe @fmanco wanted to provide me with a better example? :)

fmanco:
What's better than 1 PR... 2 PRs. Kidding, this was a coincidence. @segfaultsourcery this is my first code in Rust which I'm learning just now, better examples is likely not something you'll get from me any time soon 😄 .

I'm fine with dropping this and just going with #1044 if you prefer. Only did this for learning and because the constant refreshing was annoying me. As long as we get something I'm happy.

segfaultsourcery:
I've only been writing it for a couple of months, myself. I'm also happy either way, I just want the itch scratched.

fmanco:
Let's go with yours @segfaultsourcery I'm not going to be able to put much time into this and you already got some feedback from r/rust which is great. Thanks for working on it.

segfaultsourcery:
Oh I was just going to wait for input from the others on how to proceed.
Anyway, you have one very neat thing that I didn't think of, and that's
adding the command line param for enabling gitignore. In my case, it's just
always enabled. If you want, you can make a pull request to me so we both
get our names on this. I can also loop you in if I get any requests.

fmanco:
@segfaultsourcery I had no time to follow up on this, I'm glad your PR got merged. Once I have a bit of time I will make a new PR for the command line flag.

segfaultsourcery:
Hey! No worries.

It turns out that they didn't want a flag (check the discussion in the
pull), so instead were searching backwards for a .gitignore through the
ancestors of the current folder.

I don't think it's released yet, but it does work if you `cargo install
--git` the master branch.
Just add a line with "*.swp" (I think it was for you?) and it should work.

fmanco:
Awesome. Thanks for the work. This is great.

PR #1047, opened 2019-10-03
Fix Search::use_boolean_and documents

rnitta:
The comment and document say that `Search::user_boolean_and` defaults to `true`, but in fact, it is `false` by default.

https://github.com/rust-lang-nursery/mdBook/blob/a6f317e352855fe058af82d18cae5541c6b356d3/src/config.rs#L536

It seems that changing default behavior is not good, so I changed the related comment and the related doc.


ehuss:
Thanks!

PR #1050, opened 2019-10-04
Option to display copy buttons.

### LINKED ISSUE #1043: [Feature Request] Copy code button should be optional
The copy code button functionality should configurable in `book.toml`.

Particularly on mobile, the copy button is an eye-sore and is mostly useless, since no one is going to be able to make use of copied code on non-computers. On mobile the button is also relatively large compared to the rest of the content and seems to take away from what is otherwise a very clean and minimal design.

Also, for some books, the code blocks then to contain code that is incomplete or just are meant to show a skeleton of how actual code might look like and there isn't any value in copying that code.

It seems the way the code is written in `index.js` it is non-trival to get rid of this functionality; getting rid of the code that creates the copy button ends up breaking the functionality of other buttons, such as the sidebar. It could just be that I do not understand the way `index.js` is structured enough to be able to remove this functionality.

Setting the css for `.fa-copy.fa` to `display: none` works, but feels like a dirty hack.

andymac-2:
- Added field to playpen data structure, book.toml
- Communicate through window.playpen_copyable
- Javascript updated to check before displaying copy buttons.

Fixes #1043


src/config.rs
```diff
@@ -476,6 +476,8 @@ impl HtmlConfig {
 pub struct Playpen {
     /// Should playpen snippets be editable? Default: `false`.
     pub editable: bool,
+    /// display the copy button
```
weihanglo: Uppercase the first letter might be better and more consistent across comments. Also, we love to document  default values especially when it is not the default of that type.
```suggestion
    /// Display the copy button. Default: `true`.
```

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -445,6 +445,9 @@ fn make_data(
             data.insert("playpen_line_numbers".to_owned(), json!(true));
         }
     }
+    if html.playpen.copyable {
```
weihanglo: It should be `html_config`.

```suggestion
    if html_config.playpen.copyable {
```

andymac-2:
Thank you!

PR #1051, opened 2019-10-04
I fixed a small gitignore bug

segfaultsourcery:
The way it would join the paths could in some cases throw away the last part of the folder hierarchy before attaching the filename. Using `.join()` instead of `.with_file_name()` solves the issue.

### DIFF
--- a/src/cmd/watch.rs
+++ b/src/cmd/watch.rs
@@ -53,31 +53,46 @@ fn remove_ignored_files(book_root: &PathBuf, paths: &[PathBuf]) -> Vec<PathBuf> 
         return vec![];
     }
 
-    let gitignore_path = book_root.join(".gitignore");
-
-    match gitignore::File::new(gitignore_path.as_path()) {
-        Ok(exclusion_checker) => paths
-            .iter()
-            .filter(|path| match exclusion_checker.is_excluded(path) {
-                Ok(exclude) => !exclude,
-                Err(error) => {
-                    warn!(
-                        "Unable to determine if {:?} is excluded: {:?}. Including it.",
-                        &path, error
-                    );
-                    true
+    match find_gitignore(book_root) {
+        Some(gitignore_path) => {
+            match gitignore::File::new(gitignore_path.as_path()) {
+                Ok(exclusion_checker) => filter_ignored_files(exclusion_checker, paths),
+                Err(_) => {
+                    // We're unable to read the .gitignore file, so we'll silently allow everything.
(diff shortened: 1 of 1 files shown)


src/cmd/watch.rs
```diff
@@ -53,7 +53,7 @@ fn remove_ignored_files(book_root: &PathBuf, paths: &[PathBuf]) -> Vec<PathBuf>
         return vec![];
     }
 
-    let gitignore_path = book_root.with_file_name(".gitignore");
+    let gitignore_path = book_root.join(".gitignore");
```
ehuss: This as a few minor issues:

- Not all books are in the root of a git repository.  For example, in mdbook itself, it is `mdbook/book-example`.  I would like to, at a minimum, remove the `Unable to read gitignore file` message, at least for file-not-found.
- Not all books use git.
- This doesn't really support global git configuration.  Short of pulling in something like `git2`, there's probably not much that can be done, and I think that might be too heavy of a dependency.

Can this be changed to recursively look up the directory tree for `.gitignore`?  And remove the error if no file is found.

segfaultsourcery: Those are good points. I was thinking that maybe the simplest approach would be to specify where the file is? The now closed pull request 1045 did it that way and I think it's a pretty good idea. Would it be okay to specify a flag (like 1045's -i) followed by an optional path to a .gitignore file? If it's not given, it could either look in the root where it looks now, or try to find it recursively.

If we go the recursive path, can you give me a pointer for how to do it efficiently?

ehuss: I'd be fine with just recursively searching for it.  I'm not a big fan of adding lots of rarely used flags, and I think in most cases it should be sufficient, at least as a start.

As for how to do it, `Path` has an [`ancestors`](https://doc.rust-lang.org/std/path/struct.Path.html#method.ancestors) method which you can quickly iterate (joining with `.gitignore` along the way).  Maybe something like:

```rust
    if let Some(path) = book_root
        .ancestors()
        .map(|p| p.join(".gitignore"))
        .find(|p| p.exists())
    {
        println!("{:?}", path);
    }
```

segfaultsourcery: Oh wonderful. Thanks. The .gitignore is now searched for using your snippet, and I silenced the error.

Thanks for your input!

ehuss: Thanks!

segfaultsourcery: Hooray! I'm so happy to have contributed! :)

PR #1055, opened 2019-10-05
Fix (command:clean): removes error message 'dir not found' if 'clean' is run multiple times

### LINKED ISSUE #1046: Better message when `mdbook clean` is run on clean workspace
Currently, if you run `mdbook clean` twice, you get the following error: 

```
$ mdbook clean
2019-10-02 20:18:03 [ERROR] (mdbook::utils): Error: Unable to remove the build directory
2019-10-02 20:18:03 [ERROR] (mdbook::utils):    Caused By: No such file or directory (os error 2)
```

Would it make sense to show a better message to the user when running `mdbook clean` on an already clean workspace? 

If you think this proposal makes sense, I could perhaps do this but I would need suggestions from you. If you think I am missing something and this is not a good idea, please close this issue.

Thank you.

amanjeev:
Fixes #1046 

Please let me know if I am doing something wrong here.

### DIFF
--- a/src/cmd/clean.rs
+++ b/src/cmd/clean.rs
@@ -29,7 +29,10 @@ pub fn execute(args: &ArgMatches) -> mdbook::errors::Result<()> {
         Some(dest_dir) => dest_dir.into(),
         None => book.root.join(&book.config.build.build_dir),
     };
-    fs::remove_dir_all(&dir_to_remove).chain_err(|| "Unable to remove the build directory")?;
+
+    if dir_to_remove.exists() {
+        fs::remove_dir_all(&dir_to_remove).chain_err(|| "Unable to remove the build directory")?;
+    }
 
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


src/cmd/clean.rs
```diff
@@ -29,7 +30,11 @@ pub fn execute(args: &ArgMatches) -> mdbook::errors::Result<()> {
         Some(dest_dir) => dest_dir.into(),
         None => book.root.join(&book.config.build.build_dir),
     };
-    fs::remove_dir_all(&dir_to_remove).chain_err(|| "Unable to remove the build directory")?;
+
+    let path_to_dir_to_remove = Path::new(&dir_to_remove);
```
ehuss: This line isn't necessary, dir_to_remove is already a Path.

amanjeev: I think I was relying on my IDE too much. Checked for `.exists()` and the IDE could not find it so moved on but I was mistaken. Let me push the fix.

ehuss:
Thanks!

PR #1065, opened 2019-10-10
Reapply: Move hiding of boring lines into static content (#846)

benediktwerner:
Was reverted because of formatting in #1064

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -6,6 +6,7 @@ use crate::renderer::{RenderContext, Renderer};
 use crate::theme::{self, playpen_editor, Theme};
 use crate::utils;
 
+use std::borrow::Cow;
 use std::collections::BTreeMap;
 use std::collections::HashMap;
 use std::fs;
@@ -584,6 +585,7 @@ fn fix_code_blocks(html: &str) -> String {
 }
 
 fn add_playpen_pre(html: &str, playpen_config: &Playpen) -> String {
+    let boring_line_regex = Regex::new(r"^(\s*)#(#|.)(.*)$").unwrap();
     let regex = Regex::new(r##"((?s)<code[^>]?class="([^"]+)".*?>(.*?)</code>)"##).unwrap();
     regex
         .replace_all(html, |caps: &Captures<'_>| {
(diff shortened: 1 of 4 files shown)


ehuss:
@benediktwerner This appears to have broken hidden line removal in some cases.  Do you think you can take a look for a fix?

Here are some examples:
https://doc.rust-lang.org/nightly/reference/patterns.html#range-patterns
https://doc.rust-lang.org/nightly/reference/runtime.html#the-panic_handler-attribute

benediktwerner:
@ehuss Sure, although I didn't create the original PR I just took a quick look at it and created #1088 which I think fixes the issue in the first link.

The issue in the second link is that the code block is marked with `ignore` and is therefore not converted to a playpen: https://github.com/benediktwerner/mdBook/blob/master/src/renderer/html_handlebars/hbs_renderer.rs#L597

I'm not sure what the best fix for this would be. Is it even correct that code blocks like that aren't converted to a playpen? According to the documentation `ignore` only prevents `rustdoc` from running the code.

ehuss:
I think it is correct that `ignore`d blocks should not use the playpen, since they are unlikely to work.

Would it work to process all `#` lines for all `rust` code blocks to match the old behavior?

benediktwerner:
The thing is `ignore`d blocks don't even have a button to show hidden lines, so the only way for these lines to show up is when using the copy button. Also, the example in the link runs just fine on rust playground and otherwise there is also the `no_run` tag that seems to only hide the run button.

So I think the example in the link should be displayed in a playpen and if some code is not runnable it should use `ignore,no_run` but still be displayed in a playpen with the "Show hidden lines" button.

benediktwerner:
Also, the `noplaypen` class should probably be removed as well. The `no_run` class is enough to hide the `run` button but doesn't prevent hiding lines. For compatibility, it could be changed to do the same as `no_run`. Although, if there is another reason for `noplaypen` to exist then that would be a good reason to hide lines even without a playpen but as far as I can tell from the PR it was added only to hide the play button.

ehuss:
I see what you're saying that there is not a strong reason for `ignore` blocks to have hidden lines.  However, it might be useful in some cases, such as temporarily disabling an example, or using the hidden lines to convey intent to people working on the documentation, but not important to readers.  I agree the `runtime` example shouldn't be `ignore`, but it needed to be when it was written.  Now we can remove the `ignore` field, and it will work without needing to rewrite it.

I think `noplaypen` was intentionally added to have something to be run by `mdbook test`, but is otherwise not desired for the user to try to run.  `no_run` would disable the testing the code.

benediktwerner:
I actually do think there might be reasons to have hidden lines in `ignore` blocks. What I want to say is that `ignore` blocks should simply become playpens because the `ignore` should only target `rustdoc` and people should use `no_run` or `noplaypen` if they don't want a run button or a playpen. If `ignore` blocks become playpens lines would be hidden properly without any other change and there would also be a button to show the hidden lines. So basically I want to remove the check for `ignore` here: https://github.com/benediktwerner/mdBook/blob/master/src/renderer/html_handlebars/hbs_renderer.rs#L612

In any case, I now changed my PR to always hide lines as long as the language is `rust` so it doesn't matter that much anymore.

Issue #1067, opened 2019-10-11
Sidebar can't scroll to current active section


ngolin:
Sidebar on the left can't auto scroll to the current active section. This is something to do with the `book.js`. I'm ready to propose a PR.

mrec:
Hi - I'm not an mdBook maintainer and have no authoritah over PRs, but wanted to mention that `scrollIntoViewIfNeeded` is [nonstandard and very poorly supported](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoViewIfNeeded). Does the standard [`scrollIntoView`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) not work here?

ngolin:
> Hi - I'm not an mdBook maintainer and have no authoritah over PRs, but wanted to mention that `scrollIntoViewIfNeeded` is [nonstandard and very poorly supported](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoViewIfNeeded). Does the standard [`scrollIntoView`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) not work here?

Now it's `.scrollIntoView({ block: 'center' });`.

ehuss:
I think this should be more-or-less fixed with #1052 and #1161.

PR #1052, opened 2019-10-05
Fix #1029 sidebar not auto-scrolling

morphologue:
This fixes https://github.com/rust-lang-nursery/mdBook/issues/1029

PR #1068, opened 2019-10-11
Fix: Scroll sidebar to current active section (#1067)

ngolin:
Sidebar on the left can't auto scroll to the current active section. This is something to do with the `book.js`.

**Edit:** I just aware that there already has been a [PR #1052](https://github.com/rust-lang-nursery/mdBook/pull/1052) fixing this problem. But I think my PR is a bit cleaner and using the nice Web API. My solution places the active item [in the middle of the sidebar](https://ngolin.github.io/rust-by-example/fn.html) whenever possible, which I think it's a bit nicer. 😉 

> Note: This [on the top of the sidebar](https://rustbook-sidebar-autoscroll.s3-ap-southeast-2.amazonaws.com/ch09-00-error-handling.html) by @morphologue.

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -521,10 +521,12 @@ function playpen_text(playpen) {
     }, { passive: true });
 
     // Scroll sidebar to current active section
-    var activeSection = sidebar.querySelector(".active");
-    if (activeSection) {
-        sidebarScrollBox.scrollTop = activeSection.offsetTop;
-    }
+    window.addEventListener('load', function() {
+        var activeSection = sidebar.querySelector(".active");
+        if (activeSection) {
+            activeSection.scrollIntoViewIfNeeded();
+        }
+    });
 })();
 
 (function chapterNavigation() {
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -523,7 +523,8 @@ function playpen_text(playpen) {
     window.addEventListener('load', function() {
         var activeSection = sidebar.querySelector(".active");
         if (activeSection) {
-            activeSection.scrollIntoViewIfNeeded();
+            // https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView
+            activeSection.scrollIntoView({ block: 'center' });
         }
     });
 })();
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -435,7 +435,6 @@ function playpen_text(playpen) {
 (function sidebar() {
     var html = document.querySelector("html");
     var sidebar = document.getElementById("sidebar");
-    var sidebarScrollBox = document.getElementById("sidebar-scrollbox");
     var sidebarLinks = document.querySelectorAll('#sidebar a');
     var sidebarToggleButton = document.getElementById("sidebar-toggle");
     var sidebarResizeHandle = document.getElementById("sidebar-resize-handle");
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -84,7 +84,7 @@
         </script>
 
         <nav id="sidebar" class="sidebar" aria-label="Table of contents">
-            <div id="sidebar-scrollbox" class="sidebar-scrollbox">
+            <div class="sidebar-scrollbox">
                 {{#toc}}{{/toc}}
             </div>
             <div id="sidebar-resize-handle" class="sidebar-resize-handle"></div>


ngolin:
@morphologue 👍 👍

ehuss:
I like the idea in theory, but [`scrollIntoViewIfNeeded`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoViewIfNeeded) is not widely supported. In particular it doesn't work on Firefox at all.

Is there some other way to accomplish it without `scrollIntoViewIfNeeded`?

ngolin:
> I like the idea in theory, but [`scrollIntoViewIfNeeded`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoViewIfNeeded) is not widely supported. In particular it doesn't work on Firefox at all.
> 
> Is there some other way to accomplish it without `scrollIntoViewIfNeeded`?

Yes, I love Firefox. [`scrollIntoView({ block: 'center' })`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) got much widely supported.

src/theme/book.js
```diff
@@ -521,10 +520,13 @@ function playpen_text(playpen) {
     }, { passive: true });
 
     // Scroll sidebar to current active section
-    var activeSection = sidebar.querySelector(".active");
-    if (activeSection) {
-        sidebarScrollBox.scrollTop = activeSection.offsetTop;
-    }
+    window.addEventListener('load', function() {
```
morphologue: Is it necessary to add an event listener here? This script is loaded at the end of `index.hbs` after all the HTML for the sidebar, so that's why the existing code was able to directly query for the `.active` element.

ngolin: I'm just too careful. Just querying the `.active` element is not enough, maybe at the end there are some CSS or JS change the width or height of elements dramatically, which would break the scrolling. So I wait until the page totally rendered, that is window loaded.

morphologue: It's a good instinct to be careful, however the `load` event doesn't guarantee that JavaScript has finished manipulating the DOM. It just means that HTML and dependent resources (such as images and stylesheets) have been loaded: https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event. Given that there is other code in this file which assumes that the DOM has stabilised, I think this event listener is a bit unnecessary. Of course it's not harmful though either (unless some images take a long time to load).

trstruth:
I'm going through the rust book at the moment and this is driving me nuts!  Glad to see a PR has already been submitted and reviewed.  Any idea when it'll get merged?

morphologue:
> I'm going through the rust book at the moment and this is driving me nuts! Glad to see a PR has already been submitted and reviewed. Any idea when it'll get merged?

Hi @trstruth, my PR https://github.com/rust-lang/mdBook/pull/1052 was merged on the 6th of October and included in release [v0.3.2](https://github.com/rust-lang/mdBook/releases/tag/v0.3.2.) The problem is that the [public Rust book](https://doc.rust-lang.org/book/) seems to have been built using an older version. I am temporarily hosting a version of the [Rust book with my fix](https://rustbook-sidebar-autoscroll.s3-ap-southeast-2.amazonaws.com/index.html) which you might find more convenient for now.

In case you're wondering about the difference between this PR and mine, mine just fixed the existing code (which was trying to make the active selection show at the top of the screen), whereas this PR makes more substantial changes, and the active selection show in the middle of the screen.

trstruth:
@morphologue thanks for the info, I've opened an issue on the rust book repo describing the situation.  Will use your temporary host in the meantime :)

ngolin:
> @morphologue thanks for the info, I've opened an issue on the rust book repo describing the situation. Will use your temporary host in the meantime :)

I also host all the books [here](https://ngolin.github.io/book/).

src/theme/book.js
ehuss: The addition of this event listener causes significant jumping and flickering on chrome and safari browsers.  Would it be possible to remove it?

avinassh:
any update on this PR, when this will be merged?

ehuss:
@avinassh I'm waiting for the fixes for the addition of the event listener (https://github.com/rust-lang/mdBook/pull/1068#discussion_r348681908).  I've been intending to make the changes myself when I have the time, but I haven't been able to, yet.  If someone wants to resubmit the PR with the necessary fixes, that would be great.  It would also be useful to fully test this on a wide array of browsers and platforms, and report how well it works.

PR #1069, opened 2019-10-13
Better automatic dark mode

benediktwerner:
This primarily changes two things:
1. The default theme isn't stored as the selected theme to `localStorage` anymore. This is mainly useful for people with dynamic dark modes (i.e. only at night) because the dark mode preference is now recomputed properly on each reload as long as they didn't select a theme.
2. I added some CSS to the default themes that enable a dark mode in `no-js` mode on `prefers-color-scheme: dark`.

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -321,7 +321,7 @@ function playpen_text(playpen) {
         themeToggleButton.focus();
     }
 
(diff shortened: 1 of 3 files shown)


Dylan-DPC:
Thanks :)

PR #1070, opened 2019-10-18
Fix: Sticky header disappears by momentum scroll

rnitta:
In the default theme, there is sticky header.

sticky_header

The header will be hidden when the page is scrolled down. like:

スクリーンショット 2019-10-18 15 28 58

This behavior make a problem with momentum scroll.
With momentum scroll, `document.scrollingElement.scrollTop` can be negative number, so scrolling up above the upper limit of a page makes such logs below:

```
[Log] document.scrollingElement.scrollTop:0 (book.js, line 593)
[Log] document.scrollingElement.scrollTop:-1 (book.js, line 593)
[Log] document.scrollingElement.scrollTop:-2 (book.js, line 593)
[Log] document.scrollingElement.scrollTop:-4 (book.js, line 593)
[Log] document.scrollingElement.scrollTop:-10 (book.js, line 593)
[Log] document.scrollingElement.scrollTop:-12 (book.js, line 593)
[Log] document.scrollingElement.scrollTop:-11 (book.js, line 593)
[Log] document.scrollingElement.scrollTop:-10 (book.js, line 593)
[Log] document.scrollingElement.scrollTop:-7 (book.js, line 593)
[Log] document.scrollingElement.scrollTop:-4 (book.js, line 593)
[Log] document.scrollingElement.scrollTop:-2 (book.js, line 593)
[Log] document.scrollingElement.scrollTop:-1 (book.js, line 593)
[Log] document.scrollingElement.scrollTop:0 (book.js, line 593)
```

The last part of logs is detected as "scrolling down", so

Oct-18-2019 15-38-59

the header disappear.
I Fixed it by coercing the negative value of `document.scrollingElement.scrollTop` into 0.

this problem can be reproduce with macOS safari and iOS safari and maybe other momentum scrollable browsers.

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -591,6 +591,6 @@ function playpen_text(playpen) {
             menu.classList.remove('bordered');
         }
 
-        previousScrollTop = document.scrollingElement.scrollTop;
+        previousScrollTop = Math.max(document.scrollingElement.scrollTop, 0);
     }, { passive: true });
 })();


jdxcode:
if this drives you nuts because the header is hiding the top part you're trying to read and the book you're reading does not yet have this patch, add this custom css to make the header always visible:

```css
html:not(.sidebar-visible) #menu-bar:not(:hover).folded > #menu-bar-sticky-container {
    transform: none;
}
```

Issue #1017, opened 2019-08-25
Reduce dependency tree


bheisler:
Hey, thanks for the work on mdbook!

Would it be possible to try to prune mdBook's dependency tree?

My documentation build in Travis needs to download and compile ~300 crates to install mdbook on every build. This takes a long time (~10 minutes for mdbook). I then install `mdbook-linkcheck`, which depends on mdbook and so it has to download and recompile the whole tree again. The long compile times also make it more painful to install and use mdbook normally, or to update from older versions.

ehuss:
Binary releases are made available at https://github.com/rust-lang-nursery/mdBook/releases, which is what I use to reduce CI time.  That may not help with `mdbook-linkcheck`, though.  There is also a number of Cargo features which you can turn off to reduce the compile time (almost twice as fast with all of them off).

There are a few deps that could maybe be trimmed, but they will unlikely make a noticeable impact. The slower dependencies (syn, serde, regex, clap) can't really be removed.

You can also try to take advantage of CI caching.  sccache is one option.  You can also just cache `~/.cargo/bin`.  I've also been experimenting with the unstable `-Z install-upgrade` cargo option to make this easier (such as [here](https://github.com/rust-lang-nursery/edition-guide/blob/e58bc4ca104e890ac56af846877c874c432a64b5/.travis.yml#L5)).

bheisler:
Ah, I hadn't known about the cargo features; that will help.

I currently force-install on every build to make sure that it has the latest version, so caching doesn't really help there. I'll keep an eye on the `install-upgrade` option; hopefully it's stabilized soon.

Lokathor:
@bheisler until then try something like this

```yml
cache:
  cargo

before_script:
  - test -x $HOME/.cargo/bin/cargo-install-update || cargo install cargo-update
  - test -x $HOME/.cargo/bin/mdbook || cargo install mdbook
  - cargo install-update -a
```

Michael-F-Bryan:
I use `mdbook` for several documenting several closed-source projects and use a similar technique to @Lokathor. The only difference being I'll check the `--version` and do a `--force` install otherwise.

Another optimisation you can use is `cargo install --debug mdbook` to skip optimisations.

> That may not help with mdbook-linkcheck, though.

Thanks for pointing that out @ehuss. There's some work I'd like to do soon, so I'll see if I can upload some release binaries for `mdbook-linkcheck` with the next release.

[...]

PR #1076, opened 2019-10-25
WIP: Improve crate compilation times

Michael-F-Bryan:
Hopefully this will make improvements towards #1017. As I'm going I'll try to post numbers so we can see how changes affect compilation times.

Initial breakdown of compilation times using `cargo build -Ztimings`: 
[timings.zip](https://github.com/rust-lang-nursery/mdBook/files/3772170/timings.zip)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,7 +31,7 @@ jobs:
             rust: stable
           - build: msrv
             os: ubuntu-latest
-            rust: 1.35.0
+            rust: 1.37.0
     steps:
     - uses: actions/checkout@master
     - name: Install Rustup
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -189,37 +189,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "darling"
-version = "0.8.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "darling_core 0.8.6 (registry+https://github.com/rust-lang/crates.io-index)",
- "darling_macro 0.8.6 (registry+https://github.com/rust-lang/crates.io-index)",
-]
-
-[[package]]
-name = "darling_core"
-version = "0.8.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "fnv 1.0.6 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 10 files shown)


Michael-F-Bryan:
After updating `Cargo.lock` with `cargo update` and manually upgrading `env_logger` from 0.6 to 0.7, compilation time dropped from 97.7s to 85.9s.

I'm guessing the compilation improvements are because we aren't compiling multiple versions of the same crate.

[bump-dependency-versions.zip](https://github.com/rust-lang-nursery/mdBook/files/3772192/bump-dependency-versions.zip)

Michael-F-Bryan:
For some reason, removing the itertools crate (it was only used two or three times to skip an allocation when concatenating lines) and running `cargo clean && cargo build -Ztimings` again dropped compilation times on my machine to 60s.

[remove-itertools.zip](https://github.com/rust-lang-nursery/mdBook/files/3772391/remove-itertools.zip)

I don't think `itertools` was to blame... maybe it had something to do with build order or a warm page cache?

Dylan-DPC:
@ehuss yeh most likely this will be a breaking change so might as well release a new major version when finished which would also justify the msvr bump

Michael-F-Bryan:
I'm not sure why, but removing `toml_query` from the build actually **increased** compile times from 60s to 77s...

New dependency graph:

dependency-graph

Timings: [remove-toml-query.zip](https://github.com/rust-lang-nursery/mdBook/files/3772841/remove-toml-query.zip)

Michael-F-Bryan:
I swapped out iron with [simple-server](https://github.com/steveklabnik/simple-server) and the "build" stage in CI went from 4 mins 1 second to 3 mins 52 seconds.

dependency-graph


Timings: 
[remove-iron.zip](https://github.com/rust-lang-nursery/mdBook/files/3772963/remove-iron.zip)

ehuss:
@Michael-F-Bryan Is this ready to go? You still have WIP in the title.   I really like the changes, and it looks good to go to me.

Michael-F-Bryan:
I'm not certain it makes much of a difference. When running on my laptop I found the results super noisy and inconsistent, and I wasn't confident the compile time improvements weren't just due to something like the page cache or CPU governor changing clock frequency...

ehuss:
Thanks for getting this going!  Closed via #1211.

PR #1079, opened 2019-10-28
Lowercase matching the theme name

### LINKED ISSUE #1078: Non `light` Default Theme Isn't Served On Page Load
When using the `default-theme` configuration option on `mdbook v0.3.3` the "default theme" is not the theme rendered when opening `book/index.html` or connecting to `localhost:3000` with `mdbook serve` running, it appears to always render `light`. 

here's my book.toml:
```toml
[book]
authors = ["Ricky"]
multilingual = false
src = "src"
title = "Intro To Rust"

[output.html]
default-theme = "Ayu"
preferred-dark-theme = "Ayu"
```

The page shows I have `Ayu` selected as the default them but renders as `light`, tried clearing all caches and removing `book/` and rebuilding. Happening to me on all the following browsers (Safari `13.0.2 (14608.2.40.1.3)`, Firefox Nightly `72.0a1 (2019-10-28) (64-bit)`, Firefox `70.0 (64-bit)`). Image shows `Ayu (default)` but obviously not `Ayu` theme, if I then select it will then display.

deg4uss3r:
Using `.to_lowercase()` on the theme matching to avoid needing exact capitalization in the `book.toml`. Fixes #1078

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -407,13 +407,13 @@ fn make_data(
     }
 
     let default_theme = match html_config.default_theme {
-        Some(ref theme) => theme,
-        None => "light",
+        Some(ref theme) => theme.to_lowercase(),
+        None => "light".to_string(),
     };
     data.insert("default_theme".to_owned(), json!(default_theme));
 
     let preferred_dark_theme = match html_config.preferred_dark_theme {
-        Some(ref theme) => theme,
+        Some(ref theme) => theme.to_string(),
         None => default_theme,
     };
     data.insert(
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -413,7 +413,7 @@ fn make_data(
     data.insert("default_theme".to_owned(), json!(default_theme));
 
     let preferred_dark_theme = match html_config.preferred_dark_theme {
-        Some(ref theme) => theme.to_string(),
+        Some(ref theme) => theme.to_lowercase(),
         None => default_theme,
     };
     data.insert(


src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -407,13 +407,13 @@ fn make_data(
...
-        Some(ref theme) => theme,
-        None => "light",
+        Some(ref theme) => theme.to_lowercase(),
+        None => "light".to_string(),
     };
     data.insert("default_theme".to_owned(), json!(default_theme));
 
     let preferred_dark_theme = match html_config.preferred_dark_theme {
-        Some(ref theme) => theme,
+        Some(ref theme) => theme.to_string(),
```
ehuss: Is there some reason this doesn't use `.to_lowercase()` to match the `default_theme` behavior?

deg4uss3r: Resolved, good catch

deg4uss3r:
@ehuss Doesn't the PR #804 get the themes and then lower case them which is why this is necessary? 

https://github.com/rust-lang/mdBook/blob/master/src/renderer/html_handlebars/helpers/theme.rs#L23

ehuss:
@deg4uss3r I think that lowercase only applies to showing the (default) indicator.  The CSS class name is used directly as-is, and I believe those are case sensitive.

deg4uss3r:
@ehuss Ah, I think you're right, [here](https://github.com/rust-lang/mdBook/blob/c9dae170f38bae0907a86eff36406f1231b580ec/src/theme/index.hbs#L110-L114) seems to be the real cause to my confusion making the list of themes capitalized which is what I put into the config, rather than the true matching lowercase value.

Do you think this solution interferes with the actual style sheet selection? If so, I'm happy to rework it to get the config option to be case-insensitive if I can be pointed in the right direction.

ehuss:
> Do you think this solution interferes with the actual style sheet selection? If so, I'm happy to rework it to get the config option to be case-insensitive if I can be pointed in the right direction.

No, I think it's fine.  Adding a theme is a nontrivial task, and I don't suspect people use capitalized css names much anyways.

PR #1088, opened 2019-11-04
Fix hiding of empty boring lines

benediktwerner:
#1065 introduced a regression where lines in a playpen starting with `#` but nothing after that are now not hidden anymore. This PR fixes that regression and also adds a test case for this behavior.

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -652,26 +652,28 @@ lazy_static! {
 }
 
 fn hide_lines(content: &str) -> String {
-    let mut prev_line_hidden = false;
     let mut result = String::with_capacity(content.len());
     for line in content.lines() {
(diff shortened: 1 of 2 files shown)


ehuss:
Attributes also seem to be broken.  For example:

```rust
#![no_std]
#![warn(deprecated_in_future)]
#![warn(missing_docs)]
```

This is hiding the attributes, and when you expand they are corrupted by removing the `#`.

benediktwerner:
Well, as far as I can tell this also wasn't handled properly in the old code and only happened by "accident" because macros are but into a `<span>` by syntax highlighting and the line then doesn't start with a `#` anymore but now the lines get hidden before syntax highlighting. I think the correct way has always been to use `##![no_std]` which then gets converted to a `#![no_std]` and stays visible.

But I now changed it so that it doesn't hide lines starting with `#!` or `#[`.

Issue #812, opened 2018-11-03
editions support for mdbook test


vorner:
Hello

I can't find a way to turn on the 2018 edition in `mdbook test`. I guess this is more of a documentation issue, because after all, there's the 2018 edition of The Book, so it must work somehow. But some kind of option somewhere or explanation how to do that in docs would be nice.

steveklabnik:
Mdbook calls out to rustdoc, which supports “edition2018” and “edition2015” as a modifier to the triple backtick code blocks.

tkaitchuck:
It would be useful to have a config option to set a default. That would avoid having to type it every time, and prevent the loss of syntax highlighting within the IDE.

kpp:
@steveklabnik Looks like `edition2018` is ignored by rustdoc.

src/chapter_1.md:

    # Chapter 1
    
    ```edition2018
    async fn hello() {}
    fn main() {
        assert_eq!(1, 2);
    }
    ```

```
$ rustdoc --test src/chapter_1.md 

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

However with:


    ```rust
    async fn hello() {}
    fn main() {
        assert_eq!(1, 2);
    }
    ```

Works as expected:

```
$ rustdoc --edition 2018 --test src/chapter_1.md 

running 1 test
test src/chapter_1.md - Chapter_1 (line 4) ... FAILED

failures:

---- src/chapter_1.md - Chapter_1 (line 4) stdout ----
Test executable failed (exit code 101).

stderr:
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `2`', src/chapter_1.md:4:5
```

ehuss:
Yea, this was just recently discovered.  There is a PR (https://github.com/rust-lang/rust/pull/66238) to fix that.

kpp:
How about adding `edition=2018` option to mdbook config?

ehuss:
> How about adding `edition=2018` option to mdbook config?

There was a recent PR to add that (#1086), but it had a problem so had to be reverted. I'm looking at fixing it, but I wanted to change how the config was defined.  It's near the top of my list, so hopefully soon!

PR #1100, opened 2019-11-17
Add support for Rust edition

kpp:
Books will have to replace ` ```rust` with ` ```edition2018` everywhere to make it work, but it would be useful to change it in config only.

According to https://github.com/rust-lang/mdBook/issues/812#issuecomment-554758875

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -1,5 +1,5 @@
 use crate::book::{Book, BookItem};
-use crate::config::{Config, HtmlConfig, Playpen};
+use crate::config::{Config, HtmlConfig, Playpen, RustEdition};
 use crate::errors::*;
 use crate::renderer::html_handlebars::helpers;
 use crate::renderer::{RenderContext, Renderer};
@@ -81,7 +81,7 @@ impl HtmlHandlebars {
             debug!("Render template");
             let rendered = ctx.handlebars.render("index", &ctx.data)?;
 
-            let rendered = self.post_process(rendered, &ctx.html_config.playpen);
+            let rendered = self.post_process(rendered, &ctx.html_config.playpen, ctx.edition);
 
             // Write to file
             debug!("Creating {}", filepath.display());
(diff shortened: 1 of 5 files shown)


ehuss:
Oh, thanks for pushing this forward!

My preference would be to have a new top-level `[rust]` table where rust-specific configuration can go.  We can add more things to it in the future (maybe even things like dependencies).  

As for forwarding the configuration into the playground, it shouldn't be too hard.  I think it would involve adding another field to `RenderItemContext` in hbs_renderer so the relevant information can make it into `add_playpen_pre`.

kpp:
I feel bugs in the future. ` ```rust` blocks are replaced with `class="language-rust"`. How about pure ` ```edition2018` code blocks?

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -612,10 +619,24 @@ fn add_playpen_pre(html: &str, playpen_config: &Playpen) -> String {
                 if (!classes.contains("ignore") && !classes.contains("noplaypen"))
                     || classes.contains("mdbook-runnable")
                 {
+                    let contains_e2015 = classes.contains("edition2015");
+                    let contains_e2018 = classes.contains("edition2018");
+                    let edition_class = if contains_e2015 || contains_e2018 {
+                        // the user forced edition, we should not overwrite it
```
kpp: This allow users to write ` ```rust,edition2015` even with book.edition=2018 and their code will be sent to playpen with edition2015. I don't know how to implement the same behavior in `mdbook test`.

ehuss: `mdbook test` with a nightly toolchain honors the `edition2015` annotation.  1.41 will honor it as of https://github.com/rust-lang/rust/pull/66238.

ehuss:
> I feel bugs in the future. `` ```rust `` blocks are replaced with `class="language-rust"`. How about pure `` ```edition2018 `` code blocks?

Hm, if I understand the question correctly:  rustdoc treats code blocks with annotations like `edition2018` or `no_run` as rust blocks, but mdbook does not render them as rust blocks. Is that what you are referring to?

That sounds like a separate issue, and something we can maybe deal with separately.

src/renderer/html_handlebars/hbs_renderer.rs
kpp: Will it resolve in an error if there will be `edition=2015` in bool.toml and ` ```edition2018` in the block of code?

ehuss: No, the inline annotation overrides the global default.

kpp:
> That sounds like a separate issue

I agree.

ehuss:
Can you also move it under a separate `[rust]` table?

kpp:
It will take some time. I am not sure I will do that in a few days.

ehuss:
Completed via #1163.

Issue #1020, opened 2019-08-29
Disable / Modify search hot key


jparris:
Hi there,

Using the {{#include file}} operator I've included a html file that contains a interactive javascript widget. Everything works perfectly expect for the fact that the "s" key press triggers search which then begins consuming the key presses.

I'd like to be able to re-bind or disable the search hot key.

-Jon

brycefisher:
Ditto! I want to keep search but enable my users to press S without triggering search, but right now, there's no clean way to disable that hot key on the page I want via configuration.

omega-tree:
This would be very useful!

nihalxkumar:
Has there been any solution for this? I would like to have an option to modify the `s` key binding

PR #1101, opened 2019-11-20
Use ALT+S to trigger search

brycefisher:
As currently implemented, any time the S key is pressed outside of the search interface, that keystroke is intercepted and the search interface is displayed. This causes problems when interactive widgets are embedded on the page (as raised in https://github.com/rust-lang/mdBook/issues/1020).

This PR instead switches to ALT+S so that interactive widgets can be displayed on the page (and more generally, makes the S key available to JavaScript running on the page). Rationale for ALT+S:

 * Current implementation uses 'S' -- so we want to stay close to that
 * 'S' is a great mnemonic device for _search_
 * Windows desktop applications uses CTRL+S for save, which would likely cause the page to be downloaded
 * MacOS desktop applications use CMD+S (or `event.keyMeta` in javascript) to save 
 * Ubuntu follows one of these conventions (can't remember which offhand...maybe both?)
 * ALT+S appears not to be in widespread usage on either of those platforms
 * ALT+S is only key-stroke different than current implementation

### DIFF
--- a/src/theme/searcher/searcher.js
+++ b/src/theme/searcher/searcher.js
@@ -323,7 +323,7 @@ window.search = window.search || {};
             }
             showSearch(false);
             marker.unmark();
-        } else if (!hasFocus() && e.keyCode === SEARCH_HOTKEY_KEYCODE) {
+        } else if (!hasFocus() && e.keyCode === SEARCH_HOTKEY_KEYCODE && e.altKey) {
             e.preventDefault();
             showSearch(true);
             window.scrollTo(0, 0);


ehuss:
Thanks for the PR!  I don't think we can make this change, because Alt-S is used on Windows to access menus.

I imagine we could make this a configuration option.  Even better: is it possible to disable/override the `S` hotkey when your widget is active or has focus?

brycefisher:
My pleasure!! Yes configuration option would be much better.

Perhaps a new key in book.toml under the "search" options for "hot_key"? Implementation wise, the simplest mechanism would be a token inside search.js that gets replaced inside mod.rs. Too hacky?

ehuss:
Yea, it would go in the search option.  I'd avoid embedding javascript in the config. 
Is it not possible to just disable the hotkey in more situations?  Like if any input field has focus?

brycefisher:
That's a promising idea to explore...if the `event.target` is a form element, then bail out...That might work!

ehuss:
I'm going to close since there hasn't been much activity for a while.  If you want to pick this back up, feel free to open a new PR!

PR #1103, opened 2019-11-26
Add --features to CI recipe

matklad:
Hi!

I've noticed that salsa's CI compiles a bunch of seemingly unnecessary crates, like mio. Looks like they are coming from `cargo install mdbook` step. I guess for CI it makes sense to just recommend to disable watching and serving?


Dylan-DPC:
Thanks

Issue #1112, opened 2019-12-10
Apply selected theme to all pages when viewing mdBook locally (file:///) in Firefox


20k-ultra:
When I select a theme and navigate to another page the theme is not persisted. This is because mdBooks stores the selected theme in localStorage but since these files are being served locally via file:/// each file is treated as its own domain and so the theme selected is only available to the HTML page that set the value. 

I've successfully created a way to persist the theme across all pages using sessionStorage. I'd like to create a PR with this mechanism implemented but would like to hear feedback first. My plan is to create standard functions for handling the themes such as setting, remembering, getting, and applying. This way the getting theme code can handle the complexity of checking sessionStorage then localStorage then defaulting to the default theme. 

The potential risks to consider is that sessionStorage is available to all javascript to access which could make another script/website override the mdbook-theme of your site. For example, imagine viewing a mdBook on rust-lang.org domain and another on crates.io. If you set the theme on crates.io it would then apply to rust-lang.org since the priority is to look at sessionStorage for remembered themes selected. Meaning when you select another page on rust-lang.org the mdBook will apply the theme you picked on crates.io.

A solution for this problem is only use sessionStorage when mdBook is being served via file:///. This means mdBook will perform correctly for each domain and run the same as it is today but for people using the offline docs from rustup they will be able to have the theme they picked consistently be applied.

I made a [gist](https://gist.github.com/20k-ultra/41edc0c131ae206c9fcf9a222b19c709) of my current code just to give a better idea if what I'm explaining isn't clear. I'll have to add the logic to only check sessionStorage if mdBook is being served locally. 

One last gotcha is sessionStorage is only used for the life of the browser session. Once it is closed then the values are lost. That's ok though since if we set the theme and navigate around a mdBook it will set the localStorage value for each of these files as the user browses. When they return it will again set the theme into sessionStorage since the theme was persisted in localStorage too! In the off chance they close their browser and open a never before seen page in mdBook they would just have to set the theme again which is much better then the current situation.

20k-ultra:
I think since this problem only exists for people that view books locally then it should only work when viewing locally. I saw in the code that google analytics code is only added if the user is not locally viewing the book so this logic exists already.

I'm working on this now and will just make PR when it's done.

20k-ultra:
Ok, discovered a few things. One, Chrome shares localStorage across the file:/// "domain" so this problem doesn't exist for Chrome users however I only use Firefox and just have Chrome to test behavior between browsers.

[...]

PR #1115, opened 2019-12-13
refactor(js): persist selected theme in cookie when viewing book locally

20k-ultra:
I created issue #1112 because I am learning rust and use the offline docs provided from the rustup toolchain. Since the files are being served locally the theme I selected would not persist between page changes on Firefox (Developer Edition 72.0b5).

This PR provides a cookie mechanism to persist the selected theme across pages when viewing mdBooks locally.

For an easily reproducible instance of this bug simply us mdBook on the book-example provided in the repository and view that book locally in Firefox. Select a theme and notice how the theme defaults once you change pages. 

Now build mdBook with the code from this PR and see that themes a persisted across page changes.

The code is meant to only perform this check if a book is being viewed locally so this will not change behavior for books served from a domain.

I removed line 287 also because it was throwing an error for invalid selector and I couldn't figure out why that code existed. I understand it was trying to focus a button but even if it got the CSS selector correct the focus did nothing.

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -284,7 +284,6 @@ function playpen_text(playpen) {
     function showThemes() {
         themePopup.style.display = 'block';
         themeToggleButton.setAttribute('aria-expanded', true);
-        themePopup.querySelector("button#" + document.body.className).focus();
     }
 
     function hideThemes() {
@@ -293,7 +292,7 @@ function playpen_text(playpen) {
         themeToggleButton.focus();
     }
 
-    function set_theme(theme, store = true) {
+    function setTheme(theme, store = true) {
         let ace_theme;
 
         if (theme == 'coal' || theme == 'navy') {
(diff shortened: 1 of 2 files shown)


ehuss:
I don't see any problems with setting themes in the major browsers.  I have Firefox 71, is this maybe new behavior in 72?

20k-ultra:
I disabled all my Firefox add-ons and the behaviour still exists but I can confirm that when I view an mdBook locally on someone else's machine in Firefox (didn't get the version) the theme was persisted across page changes. I'll try to figure out what it is about my environment that is causing this.

20k-ultra:
Firefox ESR version 68 works correctly on my Debian 9 machine.
Firefox Developer 72.0b6 (beta build) localStorage is not working correctly.

I'll continue this conversation over at the appropriate Firefox bug reporting. This PR does fix the problem but really the issue is not mdBook so I guess it's not worth introducing new code when it only affects a small group of users with this browser and in a unstable beta release channel.

ntc2:
@20k-ultra did you ever figure out why this was happening on your machine? I'm having the same problem with Firefox 114.0 right now, where the themes don't persist when I view the Rust book locally with file:///...

Issue #893, opened 2019-03-25
Zero exit code when plugin not found


chinedufn:
## Problem

I started adding the `mdbook-linkcheck` plugin to my book build process and was confused when CI was still passing even though I had forgotten to install the plugin in my ci config.

```
#!/bin/bash -eo pipefail
(cd book && mdbook build) && cargo doc --no-deps --document-private-items -p foo && cp -R target/doc book/book/api
2019-03-25 16:25:00 [INFO] (mdbook::book): Book building has started
2019-03-25 16:25:00 [INFO] (mdbook::book): Running the html backend
2019-03-25 16:25:01 [INFO] (mdbook::book): Running the linkcheck backend
2019-03-25 16:25:01 [INFO] (mdbook::renderer): Invoking the "linkcheck" renderer
2019-03-25 16:25:01 [WARN] (mdbook::renderer): The command wasn't found, is the "linkcheck" backend installed?
2019-03-25 16:25:01 [WARN] (mdbook::renderer): 	Command: mdbook-linkcheck
 Documenting tw v0.1.11 (/root/foo-barg/app/crates/foo-bar-cli)
    Finished dev [unoptimized + debuginfo] target(s) in 2.16s

```

## Expected Behavior

Would expect that `mdbook build` would fail with a non zero exit code if we tried to use a plugin that it couldn't find.

Dylan-DPC:
hi @chinedufn can you send a PR that makes the required change?

segfaultsourcery:
This has been sitting idle for a while, and it's fairly low hanging fruit, so I fixed it.

ehuss:
I think this is closed by #1122.

PR #1122, opened 2019-12-31
Fixed zero exit code when plugin not found

segfaultsourcery:
In response to issue #893.

### DIFF
--- a/book-example/src/for_developers/backends.md
+++ b/book-example/src/for_developers/backends.md
@@ -331,7 +331,7 @@ the usual `RUST_LOG` to control logging verbosity.
 
 ## Handling missing backends
 
-If you should enable a backend that isn't installed, the default behavior is to throw an error:
+If you enable a backend that isn't installed, the default behavior is to throw an error:
 
 ```text
 The command wasn't found, is the "wordcount" backend installed?
--- a/book-example/src/for_developers/backends.md
+++ b/book-example/src/for_developers/backends.md
@@ -329,6 +329,36 @@ generation or a warning).
 All environment variables are passed through to the backend, allowing you to use
 the usual `RUST_LOG` to control logging verbosity.
 
+## Handling missing backends
+
+If you should enable a backend that isn't installed, the default behavior is to throw an error:
+
+```text
+The command wasn't found, is the "wordcount" backend installed?
+```
+
+This behavior can be changed by marking the backend as optional.
+
+```diff
+  [book]
+  title = "mdBook Documentation"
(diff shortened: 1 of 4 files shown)


segfaultsourcery:
I noticed that too, but I'm honestly not sure who I should ask.

azerupi:
I am not sure either. @Michael-F-Bryan do you have an idea?

segfaultsourcery:
I guess an alternative could be to hide it behind a flag, like `--abort-on-missing-plugin`, or so, but I also know that excessive flags are (rightfully) frowned upon.

Michael-F-Bryan:
I'm surprised that missing backends are allowed. I guess one reason you may want that is if the `book.toml` enables link checking but you don't have `mdbook-linkcheck` installed. I could see how that might be annoying.

Instead of a command line flag, you could add an extra item to the `[build]` section of `book.toml`. For example `missing-plugin = "ignore"` (or `"warn"` or `"fatal"`, where the default is `"warn"`). That'd involve creating a `MissingPluginBehaviour` enum and using that to dictate what happens if a backend/preprocessor doesn't exist.

segfaultsourcery:
It surprised me as well. It seems like a design decision that's definitely going to bite someone at some point (and it seems like it did). I like the idea with the `[build]` item. I'm tempted to write it now, but I think I'll wait for someone to decide for me.

Maybe @ehuss wants to decide? :)

segfaultsourcery:
Or maybe @Dylan-DPC ?

Dylan-DPC:
looks good to me. .will wait for a review from @ehuss

ehuss:
What about adding an `optional` key to the `output` table?  Similar to `preprocessor.*.renderers` as a way to control whether or not the renderer is used.

```
[output.linkcheck]
optional = true
```

Or maybe `ignore-if-missing` if `optional` is too vague?

`build.missing-plugin` is also fine, but doesn't give much fine-grain control.

Michael-F-Bryan:
`optional = true` seems like an intuitive solution. 

We'd need to update the docs so 3rd party backends (and preprocessors?) know about the convention and don't try to use the key as part of their config, although I doubt that happens in practice at the moment anyway.

segfaultsourcery:
Thank you for your input. How does this look?

ehuss:
Seems ok.  Can you add documentation?  And rebase on master.

segfaultsourcery:
I wrote a little blurb in the docs, but I'm not sure how to go about rebasing to master.

Issue #167, opened 2016-08-28
Allow linking to sections from summary


dtolnay:
In SUMMARY.md I have a link to a section like `[Section header](page-title.md#section-header)`. This works in GitBook but mdBook creates empty files `page-title.md#section-header` and `book/page-title.md#section-header` (file names containing `#`).

azerupi:
To have a little more context here, what would be the use case for this feature?

dtolnay:
Sections A and B for which both of the following are true:

- I want A and B on the same page because users will want to have both visible at once, cross-referencing between them.
- A large percentage of visitors to my site are looking specifically for sections A and B so I want to provide links directly to them in a predictable place.

azerupi:
I assume this requirement is for the Serde book? Particularly, [here is where it is used currently](https://serde.rs/attributes.html#container-attributes).

I do see the appeal for such a feature, but there are some drawbacks too. 

- I wouldn't want to allow arbitrary links in the summary, even the proposed `page-title.md#section-header` could easily break the hierarchy. If added I think we should go with `#section-header` alone and it would automatically refer to the parent chapter.

- There are some edge cases the parser would have to worry about. For example: Section links can't contain sub-chapters, they also can't be at the root level.

- The current implementation assumes all links from the summary are files. We could add some dirty checks to make it work, but if we want to do it properly, the concerned parts should be rewritten to support it.

Getting this right seems to be more complicated than it might appear. When you scroll to the sub-section the sidebar has to update the *current* chapter (= blue highlight) and the url should be updated accordingly to allow for easy linking. Gitbook handles this well. But this feature seems to break the keyboard navigation in Gitbook.

Michael-F-Bryan:
@dtolnay, I believe you're using GitBook for [serde.rs] at the moment, would you still say this is something you'd like to see in `mdbook`? 

[serde.rs]: https://serde.rs/

dtolnay:
My summary no longer includes any links to sections so I would not use this. Thanks for checking.

buschtoens:
I'm still very interested in this feature! At the very least, I would recommend to throw an error with a link to this issue when trying to `build` a `SUMMARY.md` that has `#anchors`.

I have very little Rust experience so far (front end dev primarily), but can I potentially help with this? Would this feature be appreciated / accepted to begin with?

tglman:
Hi All,

[...]

PR #1128, opened 2020-01-22
add support for links with anchors from summary, issue #167

tglman:
As the title discribe

### DIFF
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -1,4 +1,5 @@
 use std::collections::VecDeque;
+use std::ffi::OsStr;
 use std::fmt::{self, Display, Formatter};
 use std::fs::{self, File};
 use std::io::{Read, Write};
@@ -159,6 +160,8 @@ pub struct Chapter {
     pub path: Option<PathBuf>,
     /// An ordered list of the names of each chapter above this one, in the hierarchy.
     pub parent_names: Vec<String>,
+    /// The link anchor, if this character is reached from a anchor.
+    pub anchor: Option<String>,
 }
 
 impl Chapter {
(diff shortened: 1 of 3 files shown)


tglman:
Hi all,

Any chance to have a review of this ?

Regards

ehuss:
I can't get this to work.  Adding `foo.md#bar` just creates a file `src/foo.md#bar`, and the sidebar doesn't work at all.  How is it supposed to work? 

Also, how are you addressing all of the concerns listed at https://github.com/rust-lang/mdBook/issues/167#issuecomment-243005676?  It seems like it's a pretty major shift to mdBook's understanding that every chapter is a single file.

tglman:
Hi @ehuss 

Thanks for your feedback, so the check for the file and the anchor are done here: https://github.com/rust-lang/mdBook/pull/1128/files#diff-3764eab707e02a92f9da67865a327055R237

So if the file "src/foo.md" do not exists for now it consider the "foo.md#bar" as the file name, i think this is happening in your case, let me know if you prefer a different behaviour, I've done this in this way to keep the behaviour compatible do previous the change.

for the concern you linked:

>  I wouldn't want to allow arbitrary links in the summary, even the proposed page-title.md#section-header could easily break the hierarchy. If added I think we should go with #section-header alone and it would automatically refer to the parent chapter.

I didn't handle this.

>    There are some edge cases the parser would have to worry about. For example: Section links can't contain sub-chapters, they also can't be at the root level.

I didn't handle this.

>   The current implementation assumes all links from the summary are files. We could add some dirty checks to make it work, but if we want to do it properly, the concerned parts should be rewritten to support it.

I did handle this, if the link contains an anchor and  the file name is rewritten without the anchor and the anchor is stored in the Chapter structure to generate afterward the TOC.

I will do a double check of it, I'm currently testing this on a existing doc here: https://github.com/orientechnologies/orientdb-docs/commits/mdbook

Let me know as well if you prefer that I handle this in a different way.

Regards

tglman:
Hi @ehuss, 

Just to make you aware, I noticed that because I modified the Chapter structure, that is serialized and deserialized for be passed to the pre-processors, if you have a pre-processor compiled with the old version of mdbook will remove the "anchor" because in the old version of the json format used to pass informations they are not present.

Not sure how this could actually be fixed in a generic way, suggestion are welcome here.

Regards

PR #1130, opened 2020-01-24
Update handlebars to 3.0

sunng87:
3.0 provides better render performance and I recommend the community to upgrade.


sunng87:
The failing CI seems to be an issue with rustup setup.

sunng87:
Is it possible to rerun these checks?

sunng87:
@ehuss thank you! Let me know if there is any issue with the new version of handlebars

PR #1133, opened 2020-01-30
fix: ayu theme meta highlighting

EverlastingBugstopper:
Was previously highlighting `#[derive(Clone)]` like it was a comment. This updates the css to match the grouping of other themes.

For an example see https://rust-lang.github.io/async-book/02_execution/04_executor.html. Note the `#[derive(Clone)]` highlighting color in line 7 of the following screenshots

For reference, here is the regular theme:

image

And here is the ayu theme today:

image

And here is the ayu theme after the change proposed in this PR:

image

### DIFF
--- a/src/theme/ayu-highlight.css
+++ b/src/theme/ayu-highlight.css
@@ -12,8 +12,7 @@ Original by Dempfi (https://github.com/dempfi/ayu)
 }
 
 .hljs-comment,
-.hljs-quote,
-.hljs-meta {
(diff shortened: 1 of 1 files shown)


Byron:
Failure seems related to the issue fixed via #1131 .

EverlastingBugstopper:
@Byron - I don't think so? Willing to be corrected but this seems like an issue with the css. I've updated my submission to include reference screenshots.

PR #1148, opened 2020-02-17
Add "Suggest an edit" link next to "Git repository"

xkr47:
Includes new configuration option `git-repository-edit-baseurl` for
supporting non-GitHub repository layouts.

### DIFF
--- a/book-example/src/format/config.md
+++ b/book-example/src/format/config.md
@@ -178,6 +178,12 @@ The following configuration options are available:
   an icon link will be output in the menu bar of the book.
 - **git-repository-icon:** The FontAwesome icon class to use for the git
   repository link. Defaults to `fa-github`.
+- **git-repository-edit-baseurl:** The base url for suggesting an edit
+  to individual pages/chapters of the book. If **git-repository-url** is defined,
+  defaults to **git-repository-url**/blob/master which works for e.g. GitHub.
+  The page source path is appended to this url, e.g. `/src/SUMMARY.md`. So when
+  this or **git-repository-url** is configured an icon link will be output in
+  the menu bar of the book.
   
 Available configuration options for the `[output.html.fold]` table:
 
--- a/book-example/src/format/config.md
+++ b/book-example/src/format/config.md
@@ -178,12 +178,15 @@ The following configuration options are available:
   an icon link will be output in the menu bar of the book.
 - **git-repository-icon:** The FontAwesome icon class to use for the git
   repository link. Defaults to `fa-github`.
-- **git-repository-edit-baseurl:** The base url for suggesting an edit
-  to individual pages/chapters of the book. If **git-repository-url** is defined,
-  defaults to **git-repository-url**/blob/master which works for e.g. GitHub.
-  The page source path is appended to this url, e.g. `/src/SUMMARY.md`. So when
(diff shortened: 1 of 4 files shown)


ehuss:
I would prefer if this wasn't enabled by default.

Also, the default of `/blob/master` results in broken links if the git-repository-url isn't the root of a github project.  I don't think we should assume people are using github, or have the book in the root of the repo, or are pointing links to the project root.

I also don't quite see the point of a "suggest an edit" link that just takes you to the source browser.  It seems to be very similar to the existing git repo link.  

I wonder if using a github `/edit/` link would make more sense?  Or maybe just make some general mechanism for easily adding arbitrary icons of any kind?

xkr47:
Well the repository link already defaulted to the github icon, so I thought assuming github was ok. But you have valid point.

I think it is good to have a direct link to the chapter being viewed, so the random reader won't have to figure out the repo directory structure and file naming conventions to contribute. I'm perfectly fine with changing the link to point to the edit page directly. Otoh if we won't assume github, which means the user always has to enter the edit baseurl separately, we won't be assuming anything about the url anyway so it's just up to documentation to give good examples?

xkr47:
Bitbucket format for edit urls is `https://bitbucket.org/<user>/<repo>/src/<branch>/<path>?mode=edit` so maybe we need to use a template string instead of base url?

ehuss:
Making `git-repository-url` a template with something like `{path}` substituting the relative path to the source file sounds good.  That would allow the book author to decide exactly what they want to do (always link to the root, link to the source page, link to an edit page, etc.).

xkr47:
And having a separate setting for it that solely controls the appearance of the "edit" icon also lets the owners decide whether they want  repo link, a direct edit link, or both.. Ok will try to get this done soon, this is my first day coding rust after all :)

ehuss:
The index pages seem to be broken.  For example, in the book-example adding:

```toml
git-repository-edit-url-template = "https://github.com/rust-lang/mdBook/edit/master/book-example/{path}"
```

the front page tries to load index.md instead of README.md.   Same with if you go to `cli/index.html` it tries to edit `cli/index.md` instead of `cli/README.md`.

xkr47:
Good catch, I will investigate when I can!

netlander:
Any idea when this is going to get merged?

rubenmoor:
> Bitbucket format for edit urls is `https://bitbucket.org/<user>/<repo>/src/<branch>/<path>?mode=edit` so maybe we need to use a template string instead of base url?

The exact url you can put in the `index.hbs` to keep the config file clean. All that I am missing really is a `{{ md-file-path }}` in the `index.hbs` that puts in the markdown-file, the `<path>`. Shouldn't that also be easier all in all?

flavio:
> Any idea when this is going to get merged?

I'm interested in this PR too. I see the PR got 1 approval, is that just a matter of doing a rebase and solving the conflicts? I would be happy to help with that if that's the case

ehuss:
@flavio it looks like there were still some issues with index handling. If you want to take it and try to fix that, I think that would be fine since it has been over a year since we last heard from the author.

ehuss:
This has been merged as #1506.

PR #1153, opened 2020-02-29
Bring back draft chapters

azerupi:
According to discussion in PR #1109 I have implemented the "draft chapters" by making the path an option. While it works, it feels a lot dirtier with all the `if let`.

Maybe this version can be improved to remove some of this 'hack' feeling?
I will leave both PRs open so that they can be compared and I leave it to your appreciation which version you want to go with :slightly_smiling_face:

### DIFF
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -39,17 +39,19 @@ fn create_missing(src_dir: &Path, summary: &Summary) -> Result<()> {
         let next = items.pop().expect("already checked");
 
         if let SummaryItem::Link(ref link) = *next {
-            let filename = src_dir.join(&link.location);
-            if !filename.exists() {
-                if let Some(parent) = filename.parent() {
-                    if !parent.exists() {
-                        fs::create_dir_all(parent)?;
+            if let Some(ref location) = link.location {
+                let filename = src_dir.join(location);
+                if !filename.exists() {
+                    if let Some(parent) = filename.parent() {
+                        if !parent.exists() {
+                            fs::create_dir_all(parent)?;
+                        }
                     }
-                }
-                debug!("Creating missing file {}", filename.display());
+                    debug!("Creating missing file {}", filename.display());
 
-                let mut f = File::create(&filename)?;
-                writeln!(f, "# {}", link.name)?;
+                    let mut f = File::create(&filename)?;
+                    writeln!(f, "# {}", link.name)?;
+                }
             }
 
             items.extend(&link.nested_items);
@@ -152,7 +154,7 @@ pub struct Chapter {
     /// Nested items.
     pub sub_items: Vec<BookItem>,
     /// The chapter's location, relative to the `SUMMARY.md` file.
-    pub path: PathBuf,
+    pub path: Option<PathBuf>,
     /// An ordered list of the names of each chapter above this one, in the hierarchy.
     pub parent_names: Vec<String>,
 }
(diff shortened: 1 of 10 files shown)


src/book/book.rs
```diff
@@ -215,40 +237,46 @@ fn load_summary_item<P: AsRef<Path>>(
...
+fn load_chapter<P: AsRef<Path> + Clone>(
     link: &Link,
     src_dir: P,
     parent_names: Vec<String>,
 ) -> Result<Chapter> {
-    debug!("Loading {} ({})", link.name, link.location.display());
-    let src_dir = src_dir.as_ref();
+    let mut ch = if let Some(ref link_location) = link.location {
+        debug!("Loading {} ({})", link.name, link_location.display());
+        let src_dir = src_dir.as_ref();
```
ehuss: I think the `+ Clone` requirements can be removed if this line is moved up so that it has scope in the whole function (and remove the call to `.clone()`).

src/book/mod.rs
```diff
@@ -253,26 +253,28 @@ impl MDBook {
...
-                        .args(&library_args)
-                        .output()?;
-
-                    if !output.status.success() {
-                        bail!(ErrorKind::Subprocess(
-                            "Rustdoc returned an error".to_string(),
-                            output
-                        ));
+                if let Some(ref chapter_path) = ch.path {
+                    if !chapter_path.as_os_str().is_empty() {
```
ehuss: Can the rightwards drift be reduced?  Maybe something like:

```rust
let chapter_path = match ch.path {
    Some(path) => { if path.as_os_str().is_empty() { continue; } else { path } }
    None => continue,
};
```

src/renderer/html_handlebars/search.rs
```diff
@@ -71,11 +71,21 @@ fn render_item(
...
 ) -> Result<()> {
     let chapter = match *item {
-        BookItem::Chapter(ref ch) => ch,
+        BookItem::Chapter(ref ch) => {
+            if let Some(_) = ch.path {
+                ch
+            } else {
+                return Ok(());
+            }
+        }
```
ehuss: Can probably be simplified.

```suggestion
        BookItem::Chapter(ref ch) if !ch.is_draft_chapter() => ch,
```

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -31,74 +31,80 @@ impl HtmlHandlebars {
...
-                    .get("book_title")
-                    .and_then(serde_json::Value::as_str)
-                    .unwrap_or("");
-
-                title = match book_title {
-                    "" => ch.name.clone(),
-                    _ => ch.name.clone() + " - " + book_title,
-                }
-            }
+            if let Some(ref path) = ch.path {
```
ehuss: Reduce some of the rightwards drift here, too. Something like:

```rust
        let (ch, path) = match item {
            BookItem::Chapter(ch) if !ch.is_draft_chapter() => (ch, ch.path.as_ref().unwrap()),
            _ => return Ok(())
        };
```

src/book/book.rs
azerupi: Good catch! :)

azerupi:
Thanks, those were some really good suggestions :+1:

azerupi:
Hi @ehuss,
I have made the requested changes, is there anything else? 
I see that you have planned this PR for v0.4.0 does this mean that we need to wait before merging?

I will already close the other PR 🙂

kalindkaria:
This is a really good feature to have, in fact I used it in most of my docs.
But it doesn't seem to work with the preprocessor `mdbook-toc` built by @badboy.
It would be great if someone can help me with this.
Thanks.

PR #1161, opened 2020-03-09
Scroll sidebar to middle instead of top.

ehuss:
This is a re-submission of #1068. It changes the side bar so that the current page is in the vertical center instead of the top of the window.  From the previous PR, I removed the onload event handler since it caused some jumping, and it doesn't seem to be necessary.

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -505,10 +505,12 @@ function playpen_text(playpen) {
     }, { passive: true });
 
     // Scroll sidebar to current active section
-    var activeSection = document.getElementById("sidebar").querySelector(".active");
-    if (activeSection) {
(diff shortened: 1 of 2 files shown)


ehuss:
@Dylan-DPC what do you think?  I kinda like having it scroll to the middle of the page, where you can see the pages just before and after where you currently are.  But the preference is somewhat arbitrary. 

I tested on a variety of browsers, and it seems to work well.  Some older browsers, like Edge, don't support the "center" property, but do support scrolling, and it scrolls to the top of the page like it does today.

Dylan-DPC:
I like the idea. Thanks. Merging! (sorry didn't realise I was pinged :D )

PR #1162, opened 2020-03-09
Don't highlight code spans in headers.

ehuss:
This changes it so that code spans in headers are not highlighted as code, and don't change the font size.


ehuss:
I've always felt like the formatting for these didn't look very good.  Any opinion?

Before:
image

image

image


After:
image

image

image

I'm not 100% happy with the new styling, but seems less jarring to me.

Dylan-DPC:
Yeh looks fine to me. Merging :)

PR #1163, opened 2020-03-09
 Make new [rust] section to config and  place edition under it 

kngwyu:
On #1100.
Made new top-level configuration `[rust]` and moved edition config under it.
cc: @kpp @ehuss


kngwyu:
Cool, thanks!
I'm looking forward to see `edition = 2018` simplifies real playpens :)

Issue #1160, opened 2020-03-06
server: reload on vim swap file update


zdenek-crha:
I am using `vim` to edit files while `mdbook server is running`. Every time `vim` updates it swap file I get page reload.

Swap file updates are happening even without file save, about every 30 seconds if there was some activity in the file. Constant reloads in `mdbook serve` is quite annoying, especially when editing long page and each reload jumps to the top.

I think `mdbook watch` will have same issue, but since files are opened from filesystem and there is no websocket connection to trigger page reload on change, it is not as noticeable.

I've looked for ways to pass list of exclude patterns either on command line or through `book.toml`, but there seems to be no option for that.

**Version:** `mdbook v0.3.5`

**`mdbook serve` log**

The log is pruned to show only lines with detected file changes for better illustration of rebuild frequency. In reality, every `Files changed` line is followed by book build and websocket accepting new connection on tcp (page reload).

From list of changed files it can be seen that `mdbook` updates built data due to swap change about 5x more frequently than actual number of file changes.

```
12:15:31 (mdbook::cmd::serve): Serving on: http://localhost:3333
12:15:31 (ws): Listening for new connections on [::1]:3001.
12:15:31 (mdbook::cmd::watch): Listening for changes...
12:15:37 (mdbook::cmd::serve): Files changed: ["/home/user/mdbook/src/chapter.md", "/home/user/mdbook/src/.chapter.md.swp"]
12:15:44 (mdbook::cmd::serve): Files changed: ["/home/user/mdbook/src/.chapter.md.swp"]
12:16:11 (mdbook::cmd::serve): Files changed: ["/home/user/mdbook/src/.chapter.md.swp"]
12:16:24 (mdbook::cmd::serve): Files changed: ["/home/user/mdbook/src/chapter.md", "/home/user/mdbook/src/.chapter.md.swp"]
12:16:32 (mdbook::cmd::serve): Files changed: ["/home/user/mdbook/src/.chapter.md.swp"]
12:17:03 (mdbook::cmd::serve): Files changed: ["/home/user/mdbook/src/.chapter.md.swp"]
12:17:10 (mdbook::cmd::serve): Files changed: ["/home/user/mdbook/src/chapter.md", "/home/user/mdbook/src/.chapter.md.swp"]
12:17:18 (mdbook::cmd::serve): Files changed: ["/home/user/mdbook/src/.chapter.md.swp"]
[...]

PR #1164, opened 2020-03-10
Mention that .gitignore is used for watch excludes (#1160)

zdenek-crha:
This PR is adding documentation of `.gitignore` file use for `watch` and `serve` command exclude patterns when checking for modified files.

It is a follow-up to #1160 and incorporates answer from @ehuss and my observation that global `.gitignore` files are not used.

I confirmed that by checking the code, but feel free to let me know if I've overlooked something.

### DIFF
--- a/book-example/src/cli/serve.md
+++ b/book-example/src/cli/serve.md
@@ -46,3 +46,14 @@ The `--dest-dir` (`-d`) option allows you to change the output directory for the
 book. Relative paths are interpreted relative to the book's root directory. If
 not specified it will default to the value of the `build.build-dir` key in
 `book.toml`, or to `./book`.
+
+#### Specify exclude patterns
+
+The `serve` command will not automatically trigger a build for files listed in
+the `.gitignore` file in the book root directory. The `.gitignore` file may
+contain file patterns described in the [gitignore
+documentation](https://git-scm.com/docs/gitignore). This can be useful for
+ignoring temporary files created by some editors.
+
+_Note: Only `.gitignore` from book root directory is used. Global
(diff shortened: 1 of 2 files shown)


book-example/src/cli/serve.md
```diff
@@ -46,3 +46,12 @@ The `--dest-dir` (`-d`) option allows you to change the output directory for the
 book. Relative paths are interpreted relative to the book's root directory. If
 not specified it will default to the value of the `build.build-dir` key in
 `book.toml`, or to `./book`.
+
+#### Specify exclude patterns
+
+The `serve` command can that list of exclude patterns to use when deciding if
+file change should trigger build or not. Patterns are read from `.gitignore`
+file in the book root directory.
```
ehuss: The first sentence seems a little awkward, I'm not sure if there's a typo somewhere? 

Maybe something like this?

```
The `serve` command will not automatically trigger a build for files listed in
the `.gitignore` file in the book root directory. The `.gitignore` file may
contain file patterns described in the [gitignore
documentation](https://git-scm.com/docs/gitignore). This can be useful for
ignoring temporary files created by some editors.
```

zdenek-crha: I've been editing the text multiple times to get as concise and understandable version as possible. Guess it was not as understandable as I thought :-)

I've used the text you have suggested as it is way clearer. I've left the note about global `.gitignore` not being used (since that point bite my). And sorry for taking that long on such simple change.

PS: If amending commits in pull-request is not the way to go in `mdBook` project, let me know.  Unless I need to do significant changes I default to amends to not spam git history with tons of 'fixed comments' commits.

Issue #1096, opened 2019-11-13
SUMMARY.md can no longer contain HTML comments


Michael-F-Bryan:
In one of my books I comment out not-yet-written chapters using HTML comments. After updating from `v0.3.4` to `v0.3.5`  running `mdbook build` now fails while parsing `SUMMARY.md`.

```console
$ mdbook build
2019-11-13 10:00:58 [ERROR] (mdbook::utils): Error: Summary parsing failed
2019-11-13 10:00:58 [ERROR] (mdbook::utils):    Caused By: There was an error parsing the numbered chapters
2019-11-13 10:00:58 [ERROR] (mdbook::utils):    Caused By: Unable to get last link because the list of SummaryItems doesn't contain any Links
```

I believe the regression was introduced in #1021 when `pulldown-cmark` was updated.

Michael-F-Bryan:
Minimal, reproducible example:

```markdown
# Summary

- [First](./first.md)
<!-- this is a comment -->
- [Second](./second.md)
```

EasiestSoft:
I also hope this will be resolved soon

PR #1167, opened 2020-03-17
Fix SUMMARY's parse_numbered with comment

dalance:
This PR fixes https://github.com/rust-lang/mdBook/issues/1096.

If `SUMMARY` contains comment like below:

```markdown
- [First](./first.md)
<!-- this is a comment -->
- [Second](./second.md)
```

`- [Second]` generate `Event::Start(Tag::List(..))`, but `items` is empty because it is not nested.
The empty can be ignored.

### DIFF
--- a/src/book/summary.rs
+++ b/src/book/summary.rs
@@ -377,6 +377,10 @@ impl<'a> SummaryParser<'a> {
                     items.push(item);
                 }
                 Some(Event::Start(Tag::List(..))) => {
+                    // Skip this tag after comment bacause it is not nested.
+                    if items.is_empty() {
+                        continue;
+                    }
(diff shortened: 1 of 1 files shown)


src/book/summary.rs
```diff
@@ -377,6 +377,9 @@ impl<'a> SummaryParser<'a> {
                     items.push(item);
                 }
                 Some(Event::Start(Tag::List(..))) => {
+                    if items.is_empty() {
+                        continue;
```
ehuss: Thanks!  Can you add a comment here explaining why it is doing this?

dalance: Thanks!
I added a comment.

PR #1169, opened 2020-03-19
Add config option to disable print html, css, and icon

### LINKED ISSUE #1177: Add ability to disable the Print button and print.html
In some cases, I would like to avoid providing an easy way for the reader to simply snatch the entire content as one comfortable PDF. I would love the option to disable the button and of course, the generation of the `print.html` page.

Since [the config documentation page](https://github.com/rust-lang/mdBook/blob/master/book-example/src/format/config.md) does not contain the word "print", I am guessing this feature does not exist.

rossmacarthur:
This PR adds a new config option which allows you to disable all "print" support.

```toml
[output.html.print]
enable = false
```

Resolves #1177

### DIFF
--- a/book-example/src/format/config.md
+++ b/book-example/src/format/config.md
@@ -187,6 +187,9 @@ The following configuration options are available:
 - **additional-js:** If you need to add some behaviour to your book without
   removing the current behaviour, you can specify a set of JavaScript files that
   will be loaded alongside the default one.
+- **print:** A subtable for configuration print settings. mdBook by default adds
+  support for printing out the book as a single page. This is accessed using the
+  print icon on the top right of the book.
 - **no-section-label:** mdBook by defaults adds section label in table of
   contents column. For example, "1.", "2.1". Set this option to true to disable
   those labels. Defaults to `false`.
@@ -216,6 +219,11 @@ The following configuration options are available:
   site*][custom domain]).
 
 [custom domain]: https://docs.github.com/en/github/working-with-github-pages/managing-a-custom-domain-for-your-github-pages-site
+
+Available configuration options for the `[output.html.print]` table:
+
+- **enable:** Enable print support. When `false`, all print support will not be
+  rendered. Defaults to `true`.
 
 Available configuration options for the `[output.html.fold]` table:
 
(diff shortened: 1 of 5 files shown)


Dylan-DPC:
No issues. We are open to config changes

ehuss:
Yea, seems fine.

Just one minor thing, I kinda prefer to avoid double negatives (no-foo=false).  Maybe make the option name "print-enabled" with a default of `true`?  Or maybe just `print`?  

Also, I'm just curious, what's your motivation for this?

DannyBen:
To answer the motivation question, at least from my standpoint:

First, I must say It is super great that this feature exists in the first place - I usually want it enabled.

The general use case for disabling it, is to prevent the ease of copyright violation. Sometimes we use mdbook to build a static site, and like most other websites, we do not want to provide a "download the entire site as PDF" option. If you want to read it, you come here.

A more particular use case I have (but with the same principle), is that I want to share a book that is normally only used internally, with someone external for a short period for review. The material is confidential, and should not be downloaded, only read.

rossmacarthur:
@ehuss I changed the config setting to be a subtable. I think this is better because we can add extra settings later if we want and it mirrors the way `fold` settings are done.

ehuss:
I would kinda prefer to avoid adding a new subtable, unless there is some expectation that there will actually be additional keys. If in the future we decide we need additional keys, we can always upgrade it to a subtable.  I think a simple `print = false` would be sufficient.

rossmacarthur:
I don't need this anymore and I don't think I have time to continue working on this PR. Someone else is welcome to fork my branch and open a new PR.

kalindkaria:
Hi, can I know why this PR is closed without merging the feature of **disable printing** to master branch?
This is a good feature to have, even I would like to have it in mdBook.

ehuss:
I'll go ahead and merge. The implementation looks good, and I don't feel too strongly about whether or not the config is a separate table. I can image extra print options being added in the future.

ehuss (approved):

kalindkaria:
Thanks @ehuss for the quick response. I can now use the following to disable the print option.
```toml
[output.html.print]
enable = false
```
But, the **print** button at top-right is still visible, yeah I agree on clicking it returns 404, as `print.html` file doesn't exist now.
Is there a way to even remove that button or this feature is not yet requested for or added?

ehuss:
@kalindkaria Oops, thanks for checking!  I should have reviewed more closely, there is a `!` where it shouldn't have been.  Posted #1335 to fix.

kalindkaria:
Yes! @ehuss Thanks alot for adding this feature.

PR #1171, opened 2020-03-21
implement support for book parts

mark-i-m:
Implements support for breaking the list of chapters in `SUMMARY.md` into parts. This is useful for really long books that (e.g. rustc-dev-guide).

Some notable things:
- I think this makes breaking changes to some public APIs of the `mdbook` crate.
- I did some refactoring on the `SUMMARY.md` parser that I think cleans things up a bit. In particular, I gave the parser the ability to lookahead 1 token, which got rid of some edge cases and prevents every combinator from having to figure out if they are missing their first token. The new convention is that if your combinator discovers it has gone too far, it puts back the last token it read.

Here is a sample of what it looks like:

image

### DIFF
--- a/book-example/src/format/summary.md
+++ b/book-example/src/format/summary.md
@@ -22,9 +22,11 @@ allow for easy parsing. Let's see how you should format your `SUMMARY.md` file.
    [Title of prefix element](relative/path/to/markdown.md)
    ```
 
-3. ***Part Title:*** An optional title for the next collect of numbered
-   chapters. The numbered chapters can be broken into as many parts as
-   desired.
+3. ***Part Title:*** Headers can be used as a title for the following numbered
+   chapters. This can be used to logically separate different sections
+   of book. The title is rendered as unclickable text.
+   Titles are optional, and the numbered chapters can be broken into as many
+   parts as desired.
 
 4. ***Numbered Chapter*** Numbered chapters are the main content of the book,
    they will be numbered and can be nested, resulting in a nice hierarchy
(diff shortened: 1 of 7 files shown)


mark-i-m:
All tests are passing locally for me.

terretta:
**+1** -- looking for this.

Currently, mdbook knows about this SUMMARY.md syntax but doesn't leverage:
https://github.com/rust-lang/book/blob/2c942dc094f4ddcdc7aba7564f80782801197c99/second-edition/src/SUMMARY.md#basic-rust-literacy

Using ## to indicate parts in SUMMARY.md is used at GitBook as well. This PR allows rendering of a matching Nav, while the current code drops the part names entirely.

mark-i-m:
It's been a while -- I don't exactly remember, but I think I only implemented 1 level of headers in this PR...

ehuss:
Thanks for your patience!

I can't quite see why summary chapters are split into separate `Part`s. Like, why couldn't headers be just another variant of `SummaryItem` and treated similarly to `SummaryItem::Separator`?  Is there somewhere that structure is necessary?

Also, this will need some documentation (probably in https://github.com/rust-lang/mdBook/blob/master/book-example/src/format/summary.md).

mark-i-m:
@ehuss I don't really remember why I did that; I don't think it occurred to me when I was modifying the summary parser. It doesn't look that hard to change, if you would like me to.

I can also update the documentation you pointed to. Does the comment on `parse_summary` look roughly like what you want in the documentation?

ehuss:
I don't know if there will be any problems, but I would probably prefer using `SummaryItem`.  The `BookItem` is designed to be recursive, but AFAICT the part titles are not.  And, it doesn't really change the "structure" of the book, so I think it would be good to avoid changing the API in that way.

mark-i-m:
@ehuss Done. I think it was a good idea, it reduced the diff somewhat.

PR #1173, opened 2020-03-22
Support anchors in SUMMARY

mcarton:
While porting a [book of mine](https://github.com/mcarton/rust-derivative/tree/master/doc) from GitBook to mdBook, I noticed that anchors are not supported in the SUMMARY file:

For example, given the following summary:

```markdown
- [Foo](Foo.md)
    - [Subfoo 1](Foo.md#anchor-1)
    - [Subfoo 2](Foo.md#anchor-2)
- [Bar](Bar.md)
```

The following wrong things are going to happen when running `mdbook build`:

- Files `doc/Foo#anchor-1.md` and `doc/Foo#anchor-2.md` are going to be created.
- The navigation panel will only ever highlight "Subfoo 2" whether you click on "Foo", "Subfoo 1" or "Subfoo 2".
- The navigation arrows don't work: the left arrow can't go back, the right arrow points to the last sub-section.
- All 3 links in the `Foo` section point to `Foo.html`, which is empty.

### DIFF
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -320,7 +320,7 @@ And here is some \
             .write_all(DUMMY_SRC.as_bytes())
             .unwrap();
 
-        let link = Link::new("Chapter 1", chapter_path);
+        let link = Link::new("Chapter 1", chapter_path, None);
 
         (link, temp)
     }
@@ -336,7 +336,7 @@ And here is some \
             .write_all(b"Hello World!")
             .unwrap();
 
-        let mut second = Link::new("Nested Chapter 1", &second_path);
+        let mut second = Link::new("Nested Chapter 1", &second_path, None);
         second.number = Some(SectionNumber(vec![1, 2]));
 
         root.nested_items.push(second.clone().into());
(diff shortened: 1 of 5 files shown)


ehuss:
@Dylan-DPC I think it would be good to have some discussion before merging big changes like this.  Some issues:

- There is an existing (albeit broken) PR for this at #1128.
- There were a number of concerns raised in https://github.com/rust-lang/mdBook/issues/167#issuecomment-243005676.  I think they should be addressed first.
- This is a pretty fundamental change to how mdbook previously assumed every chapter is a separate file.  I suspect it will have large ramifications, and will break external tools.
- This seems to be broken for focus on the side-bar.  It seems to always focus the last element.
- This breaks backwards compatibility due to the API changes.

mcarton:
Don't hesitate to revert this if it causes any issue, however:

> This is a pretty fundamental change to how mdbook previously assumed every chapter is a separate file. I suspect it will have large ramifications, and will break external tools.

It shouldn't break any existing book though, because files with anchors are completely broken now.

> This seems to be broken for focus on the side-bar. It seems to always focus the last element.

This is incorrect: it always focuses the main section, eg. for my book it will display as

image

even if you click on the `Debug.md#format-with` link. This is IMO a good thing. To focus the right sub-section would need CSS or JS magic AFAIK.

Dylan-DPC:
@ehuss ah my bad didnt see the issue.

Dylan-DPC:
reverted in #1175

PR #1188, opened 2020-04-14
Remove google fonts by serving them locally

### LINKED ISSUE #847: Remove Google surveillance

pheki:
This is #848 +
- Apache2 license added to the fonts directory with link in the .css file (edit: I changed the link to the license in the apache site)
- All scripts (charsets) added to the fonts and they're also updated (used https://google-webfonts-helper.herokuapp.com/fonts/)
- Other minor improvements

Fixes #847 

I may also remove the .woff variants as the only current browser that doesn't support it is IE, which would anyway fallback to sans-serif.

I propose that we first fix this long standing issue and then add an option to make these fonts optional for those that use custom themes, for example.

@ehuss can you give me your thoughts?

Thanks!


ehuss:
Thanks for picking this up!

Just a few small things, and I think I'd be happy merging this:

* I think Source Code Pro needs a copy of the OFL license (at least the terms seem to indicate it needs to be included when bundled).
* I'd be happy to drop the `woff` fonts. AFAIK, the older browsers which just use default fonts instead, right? 
* I think there should be a mechanism to skip copying the fonts to the output directory if you are not using them.  I think I personally would prefer, if `fonts.css` is overridden, to skip copying the fonts. I think that would be nice, as it would avoid extra configuration needed, and if you are overriding `fonts.css`, you're probably specifying your own fonts anyways. I'm not sure how difficult that will be to detect, though.
* The documentation needs to be updated. I think [`theme/README.md`](https://github.com/rust-lang/mdBook/blob/master/book-example/src/format/theme/README.md) would be a good place to indicate that fonts can be overridden, and what the implications are.

pheki:
Hey!

First of all, I didn't want the .woff files to end up in the git history (as they sum ~ 500 kB) and the commits were very confusing, so I just force pushed a single simpler commit with the authors of both #848 and #1189 as co-authors.

Addressing your points:

- OFL license added 👍 
- Yes, older browsers will fallback to system fonts, up to the last font specified in general.css). In the case of Open Sans any sans-serif font will be used and in the case of Source Code Pro any monospace font will be used (with [a lot of fallbacks](https://github.com/rust-lang/mdBook/blob/e8d4bc52e176bfe32c3e52bc6c3db21afa6b2b31/src/theme/css/general.css#L24) apparently)
- I've actually implemented skipping copying fonts when fonts.css is overridden, but found it pretty clunky. For example, creating a theme with `mdbook init --theme` (which generates the default theme) would stop fonts from being copied. I've kept that implementation at [another branch](https://github.com/pheki/mdBook/tree/remove-google-surveillance-theme). The implementation in this PR actually adds fonts.css to `/fonts` and not to the theme, preventing it and all fonts from being copied with a new option: `no-copy-fonts`. I'd like it better for the option to be `copy-fonts` but then I'd have we'd have to provide `impl Default for HtmlConfig` instead of [deriving](https://github.com/pheki/mdBook/blob/81a7de4cdce20b08a5d87a827f4ddb2642210359/src/config.rs#L436) and I'm not sure about your opinion on this.
- Docs for the configuration file updated.

What do you think?

lzutao:
> I didn't want the .woff files to end up in the git history

How about using `build.rs` to download fonts at build time?

ehuss:
Ah, good catch!  I think I would prefer to avoid the `no-` prefix. I think it would be fine to have a manual Default impl.

ehuss:
> How about using `build.rs` to download fonts at build time?

I would prefer not to have a network dependency like that.

pheki:
> Ah, good catch! I think I would prefer to avoid the `no-` prefix. I think it would be fine to have a manual Default impl.

Great! In the next few days I'll fix this.

pheki:
Done! I also forgot to copy the fonts' licenses themselves to the output directory, which is now fixed 😅 .

PR #1189, opened 2020-04-14
Self-host fonts, add option to not copy to directory

okaneco:
Vendors Open Sans and Source Code Pro as woff/woff2, adds respective licenses for each.
Supports character sets for:
    - `open-sans-v17-latin_vietnamese_latin-ext_greek-ext_greek_cyrillic-ext_cyrillic`
    - `source-code-pro-v11-latin_vietnamese_latin-ext_greek_cyrillic-ext_cyrillic`
Add option to disable inclusion of woff/woff2

Open Sans is licensed under Apache 2.0, Source Code Pro is licensed under SIL Open Font License. The fonts were downloaded with google-webfonts-helper. 

I added controls to the HTML config struct to make the fonts optional. By default, the fonts are enabled with both woff2/woff files. Users can disable the self-hosted fonts, or disable only the woff format files which are supported by older browsers and larger in size.

```toml
[output.html.font]
enable = true
woff = true
```


ehuss:
Fixed via #1188.

PR #1205, opened 2020-04-29
Add flag --websocket-public-path to serve cmd

cwwarren:
This only changes the `livereload_url` as rendered into the HTML.

This is to support running mdbook behind an existing reverse proxy
without exposing an additional port for mdbook.

For example:

- https://example.com/   -> 10.0.0.1:3000 mdBook HTTP server
- https://example.com/ws -> 10.0.0.1:3001 mdBook WS server

There are no backwards compatibility concerns. Unless the new flag is given behavior stays the same.

I'm fully aware this is a unique use case and will probably will make some people scratch their heads. This is for an internal collaborative documentation tool using mdbook and web based IDEs and we want to expose live mdbook previews to other members of the team through our regular reverse proxy infrastructure.

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -53,6 +53,15 @@ pub fn make_subcommand<'a, 'b>() -> App<'a, 'b> {
                 ),
         )
         .arg(
+            Arg::with_name("websocket-public-path")
+                .long("websocket-public-path")
+                .takes_value(true)
+                .empty_values(false)
+                .help(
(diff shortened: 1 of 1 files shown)


ehuss:
I have a local branch that changes it so that the websocket is not on a separate port (it serves on http://host/__livereload). Do you think your reverse proxy would have problems with that?

cwwarren:
@ehuss

Nope, that would be a much better solution. This was the quick and easy solution.

ctlajoie:
@ehuss I would definitely love to have that. It solves all the reverse proxy issues

Issue #353, opened 2017-06-23
Rethink index.hbs organization


budziq:
As discussed in https://github.com/azerupi/mdBook/pull/338#discussion_r123618836

Discussion points:

index.hbs is currently coupled to hbs_renderer and book.js. Book substituting the index with their own, takes a large maintenance burden and looses bugfixes and new features. 
- Consider splitting index.hbs into core.hbs (all the js includes and core styles) and index.hbs that would be almost empty and easily replaceable.
- possibly make the other parts replaceable (sidebar, top bar , theme chooser)
- consider adding optional edit button like in [rust-cookbook](https://github.com/brson/rust-cookbook/issues/82)

PR #1206, opened 2020-05-01
HTML head customisation with `head.hbs`

GillesRasigade:
Allow to define custom attributes within the HTML <head> tag thanks to a `head.hbs` template file.

A typical use case is the addition of mobile capabilities for PWA deployment for example to render the doc in fullscreen once added to the home screen (Android or iOS):

```html
<meta name="apple-mobile-web-app-capable" content="yes" />
<meta name="viewport" content="width=device-width, initial-scale=1.0">
<meta name="mobile-web-app-capable" content="yes">
```

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -1,289 +1,315 @@
 <!DOCTYPE HTML>
 <html lang="{{ language }}" class="sidebar-visible no-js {{ default_theme }}">
-    <head>
-        <!-- Book generated using mdBook -->
-        <meta charset="UTF-8">
-        <title>{{ title }}</title>
-        {{#if is_print }}
-        <meta name="robots" content="noindex" />
-        {{/if}}
 
-        <meta content="text/html; charset=utf-8" http-equiv="Content-Type">
-        <meta name="description" content="{{ description }}">
-        <meta name="viewport" content="width=device-width, initial-scale=1">
-        <meta name="theme-color" content="#ffffff" />
+<head>
+    <!-- Book generated using mdBook -->
+    <meta charset="UTF-8">
+    <title>{{ title }}</title>
+    {{#if is_print }}
+    <meta name="robots" content="noindex" />
+    {{/if}}
 
-        <link rel="shortcut icon" href="{{ path_to_root }}{{ favicon }}">
-        <link rel="stylesheet" href="{{ path_to_root }}css/variables.css">
-        <link rel="stylesheet" href="{{ path_to_root }}css/general.css">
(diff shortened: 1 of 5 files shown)


src/renderer/html_handlebars/helpers/toc.rs
```diff
@@ -75,6 +75,7 @@ impl HelperDef for RenderToc {
                 ("", 1)
             };
 
+            #[warn(clippy::if_same_then_else)]
```
ehuss, line 78: I'm not sure why this is added.

src/theme/index.hbs
```diff
@@ -1,289 +1,315 @@
...
-        <script type="text/javascript">
-            var html = document.querySelector('html');
-            var sidebar = 'hidden';
-            if (document.body.clientWidth >= 1080) {
-                try { sidebar = localStorage.getItem('mdbook-sidebar'); } catch(e) { }
-                sidebar = sidebar || 'visible';
-            }
-            html.classList.remove('sidebar-visible');
-            html.classList.add("sidebar-" + sidebar);
-        </script>
```
ehuss, line 3: I'd prefer to avoid unrelated formatting changes. Can you remove all the whitespace changes and formatting changes?

ehuss:
@Dylan-DPC does this seem like a reasonable idea to you? I suspect there might be some longer term issues with making it easier to extend the templates (#353 ?), but this seems like a simple, incremental improvement.

Dylan-DPC:
i don't see any issue with this

src/renderer/html_handlebars/helpers/toc.rs
GillesRasigade: I added this to avoid having a error after executing `clippy`:
Screenshot from 2020-05-10 10-17-35

src/theme/index.hbs
GillesRasigade: Code auto-formatting has been reverted.

GillesRasigade:
Documentation added. Comments answered. Thanks for your reviews.

src/renderer/html_handlebars/helpers/toc.rs
ehuss: Ah, I see.  I just pushed a change to rewrite it: #1214.  I usually prefer unrelated changes like this to be separate.

code-ink-space:
The documentation might be confusing.
https://rust-lang.github.io/mdBook/format/theme/index.html

It says:

I think it should say "under" or "inside" the src folder because that's what worked for me. Not having `theme` next to `src` but inside it.
Doing `mdbook init --theme` made me see the correct directory structure.

PR #1207, opened 2020-05-03
Enable `MDBOOK_BOOK`  to overwrite `book.toml`

toyboot4e:
[Config::update_from_env](https://github.com/rust-lang/mdBook/blob/master/src/config.rs#L134) says:

```rust
/// > This means, if you so desired, you could override all book metadata
/// > when building the book with something like
/// >
/// > ```text
/// > $ export MDBOOK_BOOK="{'title': 'My Awesome Book', authors: ['Michael-F-Bryan']}"
/// > $ mdbook build
/// > ```
```

This is wrong in two ways:

1. We can't overwrite `book.toml` with `MDBOOK_BOOK`
2. The example value of `MDBOOK_BOOK` is a wrong JSON string (missing/wrong quotes)

So I fixed them so that the example works. But I didn't add any automated test.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -142,6 +142,22 @@ impl Config {
             let parsed_value = serde_json::from_str(&value)
                 .unwrap_or_else(|_| serde_json::Value::String(value.to_string()));
 
+            if matches!(&*key, "book" | "build") {
+                if let serde_json::Value::Object(ref map) = parsed_value {
+                    // To `set` each `key`, we wrap them as `prefix.key`
+                    let prefix = &key; // "book" or "build"
+                    let mut s = String::new();
+                    for (k, v) in map {
+                        s.clear();
+                        s.push_str(prefix);
+                        s.push('.');
+                        s.push_str(k);
+                        self.set(&s, v).expect("unreachable");
+                    }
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -142,6 +142,22 @@ impl Config {
...
+                    // To `set` each `key`, we wrap them as `prefix.key`
+                    let prefix = &key; // "book" or "build"
+                    let mut s = String::new();
+                    for (k, v) in map {
+                        s.clear();
+                        s.push_str(prefix);
+                        s.push('.');
+                        s.push_str(k);
+                        self.set(&s, v).expect("unreachable");
+                    }
```
ehuss, line 156: I think it would be better to simplify, maybe something this like:

```suggestion
                    for (k, v) in map {
                        let full_key = format!("{}.{}", key, k);
                        self.set(&full_key, v).expect("unreachable");
                    }
```

toyboot4e: Thank you very much. I tried to reduce the string allocation, but simplicity may be more important in this case. I'll try commiting your suggestion.

toyboot4e:
Thank you very much for your review!

PR #1210, opened 2020-05-05
Update toml-query to 0.10

Xanewok:
...in hopes of eventually bumping transitive dependencies to remove syn <1.0 from the dependency tree.

The only dependency holding us back seems to be elasticlunr > strum_macros, which would need to be fixed upstream.

The main goal of this PR is to remove syn <1.0 from transitive dependencies in the Rust tree.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -185,38 +185,39 @@ version = "0.1.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "quote 1.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
- "syn 1.0.8 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


ehuss:
Thanks for the PR!  I'm planning on removing toml-query completely, though.

Unfortunately removing the old syn will be difficult because mdbook uses elasticlunr and select which aren't maintained much.  I think select will be particularly hard, because the newer html5ever removes some fundamental things it uses.

Xanewok:
I know there are some issues wrt feature unification for build- and regular deps but is the same for dev-dependencies (which select is) as well? If that would not be an issue, then assuming the elasticlunr-rs PR is merged and a new version is released, we just might make it.

Xanewok:
Since #1211 will supersede this PR, I'll close this.

Thanks for working on making the codebase more maintainable!

ehuss:
Oh, yea, for non-dev builds, I guess select isn't as important.

PR #1211, opened 2020-05-05
Update dependencies.

### LINKED ISSUE #1107: Is the feature `output` or `debug` un-used?
https://github.com/rust-lang/mdBook/blob/e3717ad47bf5a659c7b7084b8b006fa05e9d9ff8/Cargo.toml#L60

Grepping all `feature` doesn't show anything about `output` feature:
```console
% git grep -F -w 'feature' -- '*.rs'
src/cmd/mod.rs:6:#[cfg(feature = "serve")]
src/cmd/mod.rs:9:#[cfg(feature = "watch")]
src/cmd/serve.rs:1:#[cfg(feature = "watch")]
src/cmd/serve.rs:114:    #[cfg(feature = "watch")]
src/config.rs:533:    /// Enable the search feature. Default: `true`.
src/main.rs:40:    #[cfg(feature = "watch")]
src/main.rs:42:    #[cfg(feature = "serve")]
src/main.rs:50:        #[cfg(feature = "watch")]
src/main.rs:52:        #[cfg(feature = "serve")]
src/renderer/html_handlebars/hbs_renderer.rs:104:    #[cfg_attr(feature = "cargo-clippy", allow(clippy::let_and_return))]
src/renderer/html_handlebars/hbs_renderer.rs:368:        #[cfg(feature = "search")]
src/renderer/html_handlebars/hbs_renderer.rs:471:    if cfg!(feature = "search") {
src/renderer/html_handlebars/hbs_renderer.rs:482:             search feature"
src/renderer/html_handlebars/mod.rs:8:#[cfg(feature = "search")]
src/theme/mod.rs:5:#[cfg(feature = "search")]
src/theme/searcher/mod.rs:2://! the "search" cargo feature is disabled.
tests/dummy_book/src/first/nested-test-with-anchors.rs:1:// The next line will cause a `testing` test to fail if the anchor feature is broken in such a way
tests/dummy_book/src/first/nested-test-with-anchors.rs:7:// The next line will cause a `rendered_output` test to fail if the anchor feature is broken in
tests/rendered_output.rs:510:#[cfg(feature = "search")]
```

ehuss:
Some updates to drop old dependencies.  The intent here is to keep things updated and more maintainable.

- Switch from error_chain to anyhow.
- Update to pulldown-cmark 0.7.
- Use warp instead of iron.
- Remove toml-query.
- Remove itertools.
- Update everything else.

Closes #1107


ehuss:
@Dylan-DPC Don't merge this, yet. I wanted to get feedback first. Also, I'd like to get some more PRs merged first and avoid conflicts.

This is based on #1076, with some different changes.  This is actually a little slower to build due to warp/hyper, but for me just by a few seconds.

I decided to go with warp for the web server. It has pretty good websocket support, among other things. This allows the websocket to serve from the same port which simplifies a lot of things, and makes it easier to run multiple servers.  I've also noticed it seems a little faster, although I haven't measured.

Dylan-DPC:
converted to draft xD

Xanewok:
@ehuss there's also a new [release of elasticlunr-rs](https://github.com/mattico/elasticlunr-rs/commit/e41e5c66d2ad41f01fb9d1aee2a7935e73ab2ad7) with `strum` crates bumped to 0.16, which is another step towards syn-0.15-less dep tree, so it'd be great to also include that here as well :smiley:

ehuss:
Thanks @Xanewok.  Unfortunately the new version brings in a large number of new dependencies, so there is a pretty significant net increase in build times.  🤷

Xanewok:
Ah, that's unfortunate. Let's see if we can fix it upstream to not unconditionally include Chinese/Japanese support.

Xanewok:
Pushed https://github.com/mattico/elasticlunr-rs/pull/28 that could help us cut down the newly introduced dependencies of elasticlunr if we don't enable support for Chinese/Japanese indexing.

Xanewok:
A 2.3.8 version of elasticlunr-rs is released which allows to opt out of the extra crates needed to support Chinese/Japanese, so we might cut down the number new of dependencies.

src/cmd/serve.rs
```diff
@@ -117,7 +98,6 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
         info!("Building book...");
 
         // FIXME: This area is really ugly because we need to re-set livereload :(
```
Xanewok, line 100: Is this still applicable? Do we need to re-set this again?

ehuss: Yes. The book is loaded from scratch (and thus the config is reloaded from scratch), and it needs to inject the URL into the output.

ehuss:
Thanks for the help with elasticlunr, that makes a big difference!

ehuss:
@Dylan-DPC I think this should be ready to go.

PR #1225, opened 2020-05-15
fontawesome upgrade to 5.13; incl free brand solid

rubenmoor:
Straightforward and tested.

However, for backwards compatibility, there are at least two options. At the moment, the upgrade breaks old `index.hbs` files - which is somewhat unnecessary. These old `index.hbs` are encouraged to be kept as theme overrides and - only if that is the case - would have to be updated manually.

The reason is simply that FontAwesome changed their filename from `font-awesome.css` to `fontawesome.css` without the hyphen. The files are rewritten anyways, thus

* option A: just rewrite it to the old `font-awesome.css`
* option B: write two files: `font-awesome.css` and `fontawesome.css`

The new way to write would cause less confusion in the long run, being consistent with FontAwesome's naming.


Nukesor:
Hey :)

Just stumbled upon the issue that the bundled font awesome lib is still version `4.7.0`, which was released in 2016.
Is there any chance we can get this updated and merged at some point?

Or rather, is there an official interest in bumping the font awesome version?

ehuss:
I'm going to close as handling of font-awesome has switch to a different approach in https://github.com/rust-lang/mdBook/pull/1330. Thanks for the PR, though!

PR #1230, opened 2020-05-18
Add SVG favicon.

### LINKED ISSUE #1219: Add a white glow or background on the favicon.
Hi,

The mdBook favicon is black and transparent. On dark browser themes, it is not easy to see. Since rust book uses it and the rust documentation and github also have this particularity, it's hard to find the tab among the other ones.

Not really important, but if it is cheap to fix that would help :)

My tabs right now:
image

### LINKED ISSUE #1223: How to use favicon.svg?
Is there any way to replace `favicon.png` with `favicon.svg`  in `book.toml`?

ehuss:
This adds an SVG favicon which inverts on browsers/operating systems that support "dark" media query.

Dark:
image

Light:
image

This doesn't really work on Safari, which has it's own collection of images I'm not in the mood to add. Mostly tested with Firefox and Chrome, Windows and macOS. My Linux system doesn't seem to have a dark/light switch I can easily find.

Closes #1223
Closes #1219


src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -415,7 +416,6 @@ fn make_data(
         "description".to_owned(),
         json!(config.book.description.clone().unwrap_or_default()),
     );
-    data.insert("favicon".to_owned(), json!("favicon.png"));
```
Elinvynia, line 418: Shouldn't this also somehow be replaced?

ehuss: In `index.hbs` I replaced `{{ favicon }}` with just the literal text `favicon.png` and `favicon.svg`.  There didn't seem to be a reason to have this because it is a fixed value and is not configurable.

Issue #1235, opened 2020-05-23
Clean build directory before preprocessors run instead of after.


XAMPPRocky:
The change shown below introduced a regression in the https://forge.rust-lang.org where the our precessor can no longer generate HTML redirects that we need to maintain compatibility. (<https://github.com/rust-lang/rust-forge/issues/336>). I would propose to perform the clean step *before* preprocessors are run.

https://github.com/rust-lang/mdBook/blob/6c4c3448e35b527253e0d0cd138a70c078514eb0/src/renderer/html_handlebars/hbs_renderer.rs#L315-L318

ehuss:
I think this is a duplicate of #1087.  It was intentionally changed, and there is some discussion there of possible solutions. It would be nice to have someone would make a proposal. I've only skimmed it, but it looks like there's an interesting discussion of allowing preprocessors to add "resources", which I think would also solve #1222.

XAMPPRocky:
Well I think it should be reverted until there is a solution, otherwise I'll have to pin or maintain a fork of mdbook.

ehuss:
I appreciate that it is frustrating when something breaks, but it has been almost 10 months since the change, it would also cause breakage to revert it. I think a more constructive discussion on #1087 would be more fruitful, and provide solutions for everyone. I suspect it is not too difficult to come up with a design for preprocessors to indicate additional resources they want to include.  I encourage you (or anyone) to work with @Michael-F-Bryan to come up with some solution, I'd be happy to review.

Alternatively, if it is just redirect support, I would be happy to have a first-class solution for that (#430).  I know it has come up many times, and would be very valuable.  I'd by happy to discuss that, as well.

Closing as a duplicate of #1087.

XAMPPRocky:
> I appreciate that it is frustrating when something breaks, but it has been almost 10 months since the change, it would also cause breakage to revert it.

My problem is not the time scale, it's the fact that this project has no respect for semver, this isn't the first time patch releases have broken forge.

Dylan-DPC:
Breaking compatibility accidentally doesn't mean we are not respecting semver

XAMPPRocky:
> I think this is a duplicate of #1087. It was intentionally changed

Michael-F-Bryan:
@XAMPPRocky, how are you generating the redirects? You mention before/after preprocessors, so it sounds like you've written your own `Preprocessor` to insert redirects?

[...]

PR #1237, opened 2020-05-26
Implement URL redirecting

### LINKED ISSUE #430: Facilitate maintaining URLs with redirect mapping
Most of the books being authored by *mdBook* are live documentations of projects under development, meaning that they require constant updates to the content and addition and removal of sections and chapters. Besides updating content to some URL or adding new URLs, all other kinds of maintenance/update changes result in broken URLs, which is not optimal for these projects, from a *being reference documentation* point of view. Because of that, IMHO it's needed to facilitate keeping URLs alive by having a simple way to redirect old URLs to new ones.

A simple solution that should cover majority of these cases is to have a redirect-map file, having the old path as the first field, and new path as the second. During build, HTML files with redirect head meta tags are created for the old paths, redirecting user to the new path. In practice, this file would only grow and there would never be deletions from the file. Also, a warning or error will be raised if a file path matches a redirect path, preventing conflicts between content and redirect URLs.

What do you think?

Michael-F-Bryan:
This is a fairly straightforward implementation based on the comments from #430.

Fixes #430, CC #1235.

---

Some potential future work would be allowing the user to configure redirects inline with a `[output.html.redirect]` table in `book.toml` or reading from a `key = value` file discovered by having some sort of `redirect = "redirects.txt"` under the `[output.html]` table. 

This is easy enough to implement in a backwards-compatible way (backwards-compatible for end users, it would require a library change) by swapping the `HashMap<String, String>` for some `RedirectsMap` enum that is either a `HashMap` or ` PathBuf`, and giving the `RedirectsMap` a `load_redirects() -> Result<impl Iterator<Item = (String, String)>, Error>` method.

I've decided to skip that for now, seeing as it's not really needed for @XAMPPRocky's use case and seems like overkill.

### DIFF
--- a/book-example/src/format/config.md
+++ b/book-example/src/format/config.md
@@ -198,6 +198,10 @@ The following configuration options are available:
   an icon link will be output in the menu bar of the book.
 - **git-repository-icon:** The FontAwesome icon class to use for the git
   repository link. Defaults to `fa-github`.
+- **redirect:** A subtable used for generating redirects when a page is moved.
+  The table contains key-value pairs where the key is the path to the moved
+  page, relative to the build directory and the value can be an ordinary URL
+  or the *absolute* path to another page in the book
 
 Available configuration options for the `[output.html.fold]` table:
 
@@ -281,6 +285,10 @@ boost-paragraph = 1
 expand = true
 heading-split-level = 3
 copy-js = true
+
+[output.html.redirect]
+"other-installation-methods.html" = "/infra/other-installation-methods.html"
+"bibliography.html" = "https://rustc-dev-guide.rust-lang.org/appendix/bibliography.html"
 ```
 
 ### Markdown Renderer
@@ -291,7 +299,7 @@ conjunction with `mdbook test` to see the Markdown that `mdbook` is passing
 to `rustdoc`.
 
 The Markdown renderer is included with `mdbook` but disabled by default.
-Enable it by adding an emtpy table to your `book.toml` as follows:
+Enable it by adding an empty table to your `book.toml` as follows:
 
 ```toml
 [output.markdown]
--- a/book-example/src/format/config.md
+++ b/book-example/src/format/config.md
@@ -289,7 +289,7 @@ heading-split-level = 3
 copy-js = true
 
 [output.html.redirect]
-/appendices/"bibliography.html" = "https://rustc-dev-guide.rust-lang.org/appendix/bibliography.html"
+"/appendices/bibliography.html" = "https://rustc-dev-guide.rust-lang.org/appendix/bibliography.html"
 "/other-installation-methods.html" = "../infra/other-installation-methods.html"
 ```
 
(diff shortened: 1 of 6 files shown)


Michael-F-Bryan:
@ehuss, would you be able to review this when you've got the chance?

book-example/src/format/config.md
```diff
@@ -198,6 +198,10 @@ The following configuration options are available:
   an icon link will be output in the menu bar of the book.
 - **git-repository-icon:** The FontAwesome icon class to use for the git
   repository link. Defaults to `fa-github`.
+- **redirect:** A subtable used for generating redirects when a page is moved.
+  The table contains key-value pairs where the key is the path to the moved
+  page, relative to the build directory and the value can be an ordinary URL
+  or the *absolute* path to another page in the book
```
ehuss, line 204: I'm a bit confused why this emphasizes an *absolute* path.  If there is a redirect `"foo/bar.html" = "relative.html"`, I believe that will redirect to `foo/relative.html` correctly, will it not?  Can you say more why it says absolute here?

ehuss:
Awesome, thanks!  

I imagine another enhancement would be to support `#` anchor redirects.  I've used these heavily when splitting chapters up (like https://github.com/rust-lang/reference/blob/master/src/types-redirect.html).

book-example/src/format/config.md
Michael-F-Bryan: Reading through that again, I think I got my logic back-to-front when writing the docs.

We want the pages which were moved (the bit before `=`) to be specified using absolute paths so there's no ambiguity over which file inside the build dir we're talking about. (i.e. it starts with `/`, therefore it's relative to the renderer's build directory)

You're correct about the redirect target (the bit after `=`) needing no constraints. We're just substituting the value into `<meta http-equiv="refresh" content="0;URL='{{url}}'">` and the browser would take care of everything for us, so pretty much anything you'd use in a normal markdown link is fair game.

Michael-F-Bryan:
> I imagine another enhancement would be to support # anchor redirects. I've used these heavily when splitting chapters up (like https://github.com/rust-lang/reference/blob/master/src/types-redirect.html).

Can you explain the logic behind this, and how it's intended to work? It looks like you're handling the case where a user follows a link to a particular section on a particular page (e.g. `/index.html#overview`) and the section has been moved to its own section.

I like the idea, but feel the implementation would be a lot more involved and out of scope for this PR. The way we prepare data for generating normal chapters is a bit of a code monster, so updating existing code to thread a chapter's specific redirect table through to Handlebars would be a lot more complex than tacking a new stage on the end, like we did here.

ehuss:
Yea, it can definitely be done later. I might do it if I feel motivated. 

This works in the same way, just that it is inspecting the anchor instead of being an unconditional redirect. For example, https://doc.rust-lang.org/reference/types.html#boolean-type redirects to the new page (https://doc.rust-lang.org/reference/types/boolean.html).  This was once one large page that got split up into individual pages.  There were a lot of links in the ecosystem to these headers, and I didn't want to have them go to the wrong place. The Cargo book might be a better example, where sections got moved to all sorts of different locations.

PR #1244, opened 2020-06-04
fix: don't jump away from form input to search bar

brycefisher:
Addresses #1020 with the suggestion from prior PR #1101


brycefisher:
Wahoo!! thanks @ehuss

Issue #1202, opened 2020-04-26
Github Actions CI example for the doc


rgarrigue:
Hi there

I played with mdbook and came up with https://github.com/rgarrigue/songbook/. It includes a Github Action pipeline "tag => build => publish epub as a release". Would you be interested in a PR to add this to the doc ? 

Best regards,

fosskers:
Yes please.

PR #1248, opened 2020-06-10
GitHub Actions example

rgarrigue:
Following #1202

### DIFF
--- a/book-example/src/continuous-integration.md
+++ b/book-example/src/continuous-integration.md
@@ -87,3 +87,92 @@ deploy: book
 		git commit -m "deployed on $(shell date) by ${USER}" && \
 		git push origin gh-pages
 ```
+
+## Alternative example running GitHub Actions
+
+Here's a GitHub Actions workflow definition that will build your book, upload an epub version to Github Release and publish the html as a GitHub Pages, for any push. Just copy / paste this in your repository's `.github/workflows/mdbook.yml` ! 
+
+```yaml
+name: Generate ebook
+
+on: 
+  push:
+
+jobs:
+  ebook:
+    name: Build and upload
+    runs-on: ubuntu-latest
+    steps:
+      # Install Rust & its package manager Cargo
+      - name: Install cargo
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+
+      # Cache installation assets
+      - name: Cache cargo registry
+        uses: actions/cache@v1
+        with:
+          path: ~/.cargo/registry
+          key: ${{ runner.os }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
+      - name: Cache cargo index
+        uses: actions/cache@v1
+        with:
+          path: ~/.cargo/git
+          key: ${{ runner.os }}-cargo-index-${{ hashFiles('**/Cargo.lock') }}
+      - name: Cache cargo build
+        uses: actions/cache@v1
+        with:
+          path: target
+          key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
+
+      # Install mdbook and requirements
+      - name: Install mdbook and epub plugin
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: mdbook mdbook-epub
+
+      # Ebook generation
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      # Here you can add one or more steps, like generating a summary or whatever
+      # - name: Generate summary
+      #   run: ./generate_summary.sh
+
+      - name: Generate ebook from markdown
+        run: mdbook build
+
+      # ePub upload as a Github release asset
+      - name: Create Release
+        id: create-release
+        uses: actions/create-release@v1.0.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
+          draft: false
+          prerelease: false
+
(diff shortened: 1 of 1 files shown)


fosskers:
Oh heck yes! And I just finished a Rust book today too :) I'll test this out.

book-example/src/continuous-integration.md
```diff
@@ -87,3 +87,92 @@ deploy: book
...
+          asset_name: songbook.epub
+          asset_content_type: application/epub+zip
+
+      # HTML publication as Github Page
+      - name: Publish HTML
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./book/html
+          publich_branch: ${{ github.ref }}
```
fosskers, line 177: Oh, slight typo here?

fosskers:
I tested it and it got it working: https://fosskers.github.io/aura/

Here's the config I used:
```yaml
name: Book

on:
  push:
    branches:
      - master

jobs:
  book:
    name: Build and Deploy
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v2

      - name: Install mdbook
        uses: peaceiris/actions-mdbook@v1
        with:
          mdbook-version: "0.3.7"

      - name: Generate book from markdown
        run: |
            cd book
            mdbook build

      - name: Publish HTML
        uses: peaceiris/actions-gh-pages@v3
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
          publish_dir: ./book/book/
```
**Note:** It may be more appropriate to use https://github.com/marketplace/actions/mdbook-action than installing `mdbook` through `cargo`. Thoughts?

rgarrigue:
The fastest solution if you can compare, since time is money on CI. 

Coming up with this was 50% getting an epub on my ereader 50% having a look at how to build rust stuff on CIs. So I went for vanilla Rust.

book-example/src/continuous-integration.md
rgarrigue: Good catch, I'll fix it as soon as my laptop end up booting

fosskers:
I did try both (with cargo and without) and the latter was *much* faster since you don't need to build dependencies nor cache anything. 

Although I didn't build an epub - not sure if that's available natively.

rgarrigue:
I'ld say, add both. Simple case, use dedicated action. More complex case with plugins, here's a rust based example. Could serve as an inspiration for other CIs too.

dabreegster:
@fosskers: Sorry to be a bit off-topic, but I've been trying to adapt https://github.com/fosskers/aura/blob/master/.github/workflows/book.yaml to my repo. The deployment succeeds and I've mimicked your directory structure, but I can't get the github pages to find the book directory. Could you please share your github pages settings?
Screenshot from 2020-08-09 15-07-35

fosskers:
```
      - name: Publish HTML
        uses: peaceiris/actions-gh-pages@v3
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
          publish_dir: ./book/book/
```
^ Be careful of the `publish_dir` here. A run of `mdbook` *creates* a `./book` directory with all the release assets. Since Aura's book *source* is already in a subdir called `book`, the full path then becomes `/book/book/`.

dabreegster:
> Be careful of the publish_dir here

Thanks! I got it working. For anyone else struggling:
- https://github.com/dabreegster/abstreet/blob/master/.github/workflows/book.yml
- In project settings, set the branch to `gh-pages` and the directory as `/`. The `gh-pages` branch is automatically created by the workflow, which is a little confusing. I at first interpreted this to mean the repo should have a separate branch just for docs.
- Make sure there's not a jekyll `_config.yml` around, if you were previously using that for github pages. I think it caused some strange conflicts.
- Put `book` in the repo's root dir, aka, `book/book.toml` should exist
- Even after the github workflow succeeds, there's still a slight replication delay of about 30-60s on the hosted https://dabreegster.github.io/abstreet/ site. Different browser / clearing cache doesn't affect this.

rgarrigue:
@fosskers can we merge ?

fosskers:
Should be good? Although it looks like there's a merge conflict now.

nikomatsakis:
Having this readily documented is great! 

That said, I see there are a bunch of actions in the marketplace:

* https://github.com/marketplace/actions/deploy-mdbook
* https://github.com/marketplace/actions/mdbook-action

I myself have been using deploy-mdbook, maybe we want to point at those as well?

fosskers:
Right, it has been some time since I looked at this last and had forgotten some details. I myself use `actions-mdbook` but the example here is installing `mdbook` manually. Advertising some off-the-self action is better, I'm thinking.

nikomatsakis:
I know that I was confused by the plethora of choices.

nikomatsakis:
I'm a bit wary because I know that it's hard to keep up with the maintenance burden here as is, so maybe that's not a good suggestion.

fosskers:
Likely whatever approach is advertised by this book will de facto be the official recommendation (this is a `rust-lang` project after all), so we should be careful about it.

rgarrigue:
For the record, I went this rust + cargo way instead of off-the-shelf since I was more interested in a generic rust approach where I know what I'm in control hence future proof, VS a convenient blackbox.

Anyway, the choice is up to you, I'll let you fix it / close it. I'm just cleaning up my dangling PRs list nowadays, hence the ping.

ehuss:
Thanks for the PR!  I am moving the CI examples to the wiki: https://github.com/rust-lang/mdBook/wiki/Automated-Deployment.  I think most of these are now covered there, but feel free to update pages there.

PR #1256, opened 2020-06-13
Remove CloudFlare CDN request

shreyasminocha:
Similar in motivation to #847.


src/theme/index.hbs
```diff
@@ -40,7 +40,7 @@
 
         {{#if mathjax_support}}
         <!-- MathJax -->
-        <script async type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS-MML_HTMLorMML"></script>
+        <script async type="text/javascript" src="{{ path_to_root }}mathjax.js"></script>
```
stappersg, line 43: I do mis the _?config=TeX-AMS-MML_HTMLorMML_ suffix.

So I'm asking  should it be
```
<script async type="text/javascript" src="{{ path_to_root }}mathjax.js?config=TeX-AMS-MML_HTMLorMML">
</script>
```
?

Dylan-DPC:
@shreyasminocha can you resolve the conflicts? thanks

shreyasminocha:
@Dylan-DPC done!

ehuss:
This doesn't seem to work.  The file is not copied to the output. All theme files require special handling in code. Also, I would prefer if it wasn't copied if mathjax is disabled.

shreyasminocha:
I don't have much rust in me, but would duplicating and editing lines responsible for the other static files in `src/theme/mod.rs` and `src/renderer/html_handlebars/hbs_renderer.rs` work? @ehuss

ehuss:
Yea, I think that's roughly correct.  In `copy_static_files` it should check if mathjax is enabled (I think there is a field in `html_config`).

shreyasminocha:
Thanks.

I ran it a couple of times locally and it seems to be working as expected now.

ehuss:
This still doesn't appear to work.  The MathJax samples aren't rendering, and I see these errors on the console:

```
Loading failed for the <script> with source “http://localhost:3000/extensions/MathMenu.js?V=2.7.1”. mathjax.html:1:1
Loading failed for the <script> with source “http://localhost:3000/extensions/MathZoom.js?V=2.7.1”. mathjax.html:1:1
```

unpavedmop:
i cant use mdbook until this is fixed. how did cf slip in after the google debacle?

those interested in this may also like:

https://github.com/rust-lang/rust/issues/40552
https://github.com/rust-lang/rust/issues/75263

ssokolow:
@unpavedmop Every fix takes them *some* time. (And, while I agree that I wouldn't want to be using any external resources, I don't use mdBook currently and, if I did, I have no need for MathJax, so I'm not going to be the one fixing it.)

You should be able to resolve the problem for yourself just by either turning off MathJax support for using the [custom theme](https://github.com/rust-lang/mdBook/blob/3d55375f618594d62707eab18b98d32c52b4c9b7/guide/src/format/theme/README.md) support to replace the built-in Handlebars template with one that uses a local copy of MathJax.

Heck, given how easy it is to build a Rust project from source, it should be trivial for you make a PR to fix the problem for all of us.

You got me curious, so I did a quick repository search for "cloudflare" using GitHub's search box. The offending line is inside an `{{#if mathjax_support}}` block in `theme/index.hbs`. Change that to use `{{ path_to_root }}` instead.

I then searched for something that already uses `{{ path_to_root }}`... in this case `highlight.css`, and it looks like you need to do the following to add another file to the bundled local resources:

1. Add `include_bytes!` lines for the MathJax files to `theme/mod.rs`
2. Add lines to `copy_across_theme` in `book/init.rs` to copy the data to output files.
3. Add lines to `copy_static_files` in `renderer/html_handlebars/hbs_renderer.rs` to copy the data to output files. (I didn't dig deep enough to figure out why this has to be done in two places.)
4. Add lines to the `copy_theme` test in `tests/init.rs` to test the functionality you added.
5. Update `format/theme/README.md` to add the MathJax files to the list of bundled files that can be overridden.

ssokolow:
Oops. I didn't sleep well last night and thought this was an issue with someone *asking* for external MathJax support to be removed, not a PR *providing* a way to do it.

Given that I really don't like sites randomly relying on external domains (especially now that cache isolation makes external CDNs purely a liability) and it seems like a very ugly hack to make a custom theme that's identical to the default except for switching to local MathJax, could this (or a corresponding feature request) be opened back up?

shreyasminocha:
Yeah I don't think I'll personally get around to finishing this, but people are welcome to continue from here.

stappersg:
On Fri, Feb 12, 2021 at 05:35:37PM -0800, Shreyas Minocha wrote:
> but people are welcome to continue from here.

Express that by re-opening this issue

KarlWithK:
I just want to say thanks for working on this! I care a lot about my privacy, and I think it's important that we remove something like this.

ehuss:
I'm going to close since this has gone stale.  If someone is willing to pick it back up, feel free to create a new PR.

PR #1257, opened 2020-06-19
Fix highlighting theme order of precedence

shreyasminocha:
> the files used for syntax highlighting can be overridden with your own

> Now your theme will be used instead of the default theme.

With the old order of stylesheets, the default theme would override the
custom one.

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -29,9 +29,9 @@
         {{/if}}
 
         <!-- Highlight.js Stylesheets -->
-        <link rel="stylesheet" href="{{ path_to_root }}highlight.css">
         <link rel="stylesheet" href="{{ path_to_root }}tomorrow-night.css">
         <link rel="stylesheet" href="{{ path_to_root }}ayu-highlight.css">
+        <link rel="stylesheet" href="{{ path_to_root }}highlight.css">
 
         <!-- Custom theme stylesheets -->
         {{#each additional_css}}


ehuss:
Hi, thanks for the PR!

Can you say more about what motivates this change?  The order shouldn't matter, because the CSS is enabled/disabled so that only one highlight file is enabled at once.

After looking at it closer, I noticed that the selector is broken.  `document.querySelector("[href$='highlight.css']")` is ambiguous because it does an "ends with" which matches both `highlight.css` and `ayu-highlight.css`.  I'm not sure how it ever worked properly, though for me I never noticed any problems.  Do you have a specific browser that is not highlighting properly?

What do you think about a change like the following?

```diff
diff --git a/src/theme/book.js b/src/theme/book.js
index 849f8cefe..8183dd4ad 100644
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -277,9 +277,9 @@ function playpen_text(playpen) {
     var themePopup = document.getElementById('theme-list');
     var themeColorMetaTag = document.querySelector('meta[name="theme-color"]');
     var stylesheets = {
-        ayuHighlight: document.querySelector("[href$='ayu-highlight.css']"),
-        tomorrowNight: document.querySelector("[href$='tomorrow-night.css']"),
-        highlight: document.querySelector("[href$='highlight.css']"),
+        ayuHighlight: document.querySelector("[href$='/ayu-highlight.css']"),
+        tomorrowNight: document.querySelector("[href$='/tomorrow-night.css']"),
+        highlight: document.querySelector("[href$='/highlight.css']"),
     };
 
     function showThemes() {
```

shreyasminocha:
> Do you have a specific browser that is not highlighting properly?

Yeah, It didn't seem to work with Firefox 68.6.0esr.

> The order shouldn't matter, because the CSS is enabled/disabled so that only one highlight file is enabled at once.

Oops, you're right 😅

> What do you think about a change like the following?

Yeah, I think that would do it. Should I push a new commit?

shreyasminocha:
> Do you have a specific browser that is not highlighting properly?
> >
> > Yeah, It didn't seem to work with Firefox 68.6.0esr.

Hmm, I can reproduce this with Chromium too. You can check for yourself at <https://wiki.shreyasminocha.me/programming/fish>. It's worth noting that my `Content-Security-Policy` is set to `"default-src 'self'; img-src *"`. The source is at [shreyasminocha/wiki on branch `test-hl-theme`](https://github.com/shreyasminocha/wiki/tree/8458b00a4bc7975bbf2a51f3af8aa58106a0a1b8).

PR #1261, opened 2020-06-22
Docker based convenience wrapper

pmjohann:
I wrote this wrapper for myself, added a short alias to my zshrc and found much easier to work with mdBook this way. Thought I would share, maybe other docker lovers find it useful as well!

Thanks for mdBook, keep up the good work guys!


ehuss:
Thanks for the PR!  Unfortunately although I can see how this could be useful, I think I would prefer to avoid adding this at this time.

PR #1265, opened 2020-06-25
Fix sidebar scrolling with part titles.

ehuss:
If a summary included section titles, then the automatic scrolling for the sidebar was broken (it was always scrolled to the top of the sidebar).


ehuss:
@mark-i-m I'm not sure if there was a reason to include the "active" class on the section titles.  I noticed that the sidebar scrolling was broken on the rustc-dev-guide.

mark-i-m:
@ehuss Oh, I didn't even know this was a feature. I probably added `active` to make it look right... my web-dev experience is pretty limited; I didn't realize it had an effect.

Issue #1275, opened 2020-07-08
No kotlin syntax highlighting despite highlight.js support?


maniyar1:
Running into a problem where Kotlin code snippets are not properly being highlighted. 
[Here is a test repository to demonstrate](https://github.com/maniyar1/test-book) (generated with current git version of mdBook). 

This is fixed by dropping in a [new version](cdnjs.cloudflare.com/ajax/libs/highlight.js/10.1.1/highlight.min.js) of highlight.js in the `theme/` folder, but this is strange since according to the [changelog](https://github.com/highlightjs/highlight.js/blob/master/CHANGES.md) Kotlin has been supported for a good while before the current version in mdBook (9.15).

In any case, it seems updating highlight.js is trivial, so I'd be totally open to PRing in the most recent version, just was wondering if anyone else had thoughts on this (especially since I know Kotlin isn't really a target language).

ehuss:
The default higlightjs uses the common languages and a few select extras. Kotlin isn't in the default set.  A PR to update to add Kotlin would be fine, just make sure not to remove any languages currently included.

maniyar1:
Ah, that's interesting and makes perfect sense. What's the criteria for being added? I doubt mdBook is being used that much for Kotlin code anyway, and the work-around is pretty simple. 
Looking at their bundling it looks like now Kotlin is in their "common" bundle/set (not sure about phrasing), so I guess it might make sense to add it in. Is there a running list of the languages currently supported?

ehuss:
The criteria has generally been when someone asks for a new language.

I don't remember exactly what the list is.  You can load the current [highlight.js](https://github.com/rust-lang/mdBook/blob/master/src/theme/highlight.js) and search for all the calls to `registerLanguage` to get the complete list. According to #942 it is `:common armasm d go handlebars haskell julia rust scala swift x86asm yaml`, but I don't remember if that is a current list.

PR #1277, opened 2020-07-09
Update highlight.js

maniyar1:
Updates to 10.1.1, with support for new languages added to the common set (#1275).
Built with:
common languages + D, handlebars, haskell, julia, scala, x86asm, armasm, yaml
Size difference:
New highlight.js: 132KiB
Old highlight.js: 84KiB

New highlight.js compressed: 48KiB
Old highlight.js compressed: 36KiB
So on average each request will be 12KiB more data (total payload is around 800KiB from my test).

I went through and double-checked that I didn't drop any languages, but a second pair of eyes couldn't hurt.

Full language list: 
```
    .properties
    C#
    Diff
    Json
    Less
    Nginx config
    Perl 
    Ruby
    Shell session
    YAML
    Apache
    C++
    Go
    Java
    Lua
    Objective-C
    Plain Text
    Rust
    Swift
    Bash
    C-like foundation
    HTML, XML
    JavaScript
    Makefile
    PHP
    Python
    SCSS
    TOML, also INI
    C
    CSS
    CoffeeScript
    HTTP
    Kotlin
    Markdown
    PHP Template
    Python REPL
    SQL
    Typescript
    D
    handlebars
    haskell
    julia
    scala
    x86asm
    armasm
    yaml
```


colman-humphrey:
It would be great to get R support if possible. Since this highlight update is (necessarily) replacing a block of code, would adding it to this PR be possible (instead of another block replacement)?

I think (could be wrong) that R is the most popular language not yet supported (after the above change)

maniyar1:
For me the easiest way would be to just regenerate the file, but it's pretty trivial to add R so I'll go ahead and add it in, surprised it's not in there already now that you mention it...

colman-humphrey:
Thanks @maniyar1 !

indiv0:
Would it be possible to add lisp support as well?

PR #1279, opened 2020-07-13
Sets favicon order for Chrome, modifies favicon.png

sl4m:
A couple of small changes I want to bring to attention (sorry for the lengthy description!)

1. Chrome does not seem to favour SVG favicons over PNG favicons. Rather, it loads the last favicon defined in the HTML. In Firefox, SVG favicons take precedence regardless of how it's defined in the HTML. One caveat I discovered when making this small change: I originally thought the PNG favicon was being loaded before my change. However, due to the size of the PNG favicon being 196x196, Chrome ignored it and loaded the SVG favicon instead. It seems to ignore PNG favicons larger than 32x32. When I fixed the PNG favicon to be 32x32, it unfortunately does not look great in dark mode:

Screenshot 2020-07-12 at 18 52 05

The PNG favicon change in this PR simply reduces the resolution to 32x32, but longer term solution should be to update the favicon to support both system modes (dark, light) since it's not something that can be manipulated via CSS `@media` query.

2. The `shortcut` link type seems to be non-conforming and should not be used anymore according to [MDN web docs][1]. Hence, it was removed as part of the PR.

Both browsers tested on the latest: Chrome 83.0.4103.116, Firefox 78.0.2. I reported the favicon issue to [Chromium][2]. We'll see what they have to say.

[1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types
[2]: https://bugs.chromium.org/p/chromium/issues/detail?id=1104663

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -20,8 +20,8 @@
         <meta name="viewport" content="width=device-width, initial-scale=1">
         <meta name="theme-color" content="#ffffff" />
 
+        <link rel="icon" href="{{ path_to_root }}favicon.png">
         <link rel="icon" href="{{ path_to_root }}favicon.svg">
-        <link rel="shortcut icon" href="{{ path_to_root }}favicon.png">
         <link rel="stylesheet" href="{{ path_to_root }}css/variables.css">
         <link rel="stylesheet" href="{{ path_to_root }}css/general.css">
         <link rel="stylesheet" href="{{ path_to_root }}css/chrome.css">


ehuss:
Hm, I'm not able to reproduce the issue. Which platform are you on? What is your display resolution?

It does seem like it is more correct to list the SVG last.  Usually browsers will pick the "best" one based on various criteria. If there are multiple "best" ones, it should pick the last.  However, in all of my tests it picks SVG as "best" regardless of order.

I also had no problems loading 196x196 png icons in Chrome.

I think the `shortcut` is for IE.  Probably not super important (I'm not even sure mdbook still works on IE), but does it cause any harm to leave it?

sl4m:
My env: macOS Catalina 10.15.5. Resolution according to macOS looks like 1920x1080.

It looks like I was incorrect about the PNG. Because of the dark on dark mode, I didn't see it. I have removed the PNG change.

Regarding SVG as being picked "best", have a look at this test I created for the Chromium issue: https://sl4m.github.io/chrome-favicon. Refreshing the page is best, so you're not seeing the cached favicon. In my tests, I do see Chrome picking the last favicon listed in `<head>`.

Does IE support PNG favicon other than 11?

PR #1282, opened 2020-07-19
Mention removed features in the changelog

JohnTitor:
I think it's worth mentioning in the changelog since it's one of breaking changes even if it was actually unused.


Dylan-DPC:
Thanks

PR #1284, opened 2020-07-25
Mention that uninstalled backend isn't marked optional.

### LINKED ISSUE #1283: What is the intended behavior when a backend isn't available?
Hi there! We are using `mdbook` for the Void Linux handbook. We have configured the `mdbook-linkcheck` backend to help with verifying contributions, since it keeps any dead links out of the handbook. However, when deploying the book, we don't include have `mdbook-linkcheck` available in the machine, and it simply complains about not finding the plugin, but still builds the whole book perfectly.

However, when using `mdbook serve` and `mdbook-linkcheck` isn't available, it errors out completely, with the error message below:

```
2020-07-24 00:42:26 [INFO] (mdbook::book): Book building has started
2020-07-24 00:42:26 [INFO] (mdbook::book): Running the html backend
2020-07-24 00:42:26 [INFO] (mdbook::book): Running the linkcheck backend
2020-07-24 00:42:26 [INFO] (mdbook::renderer): Invoking the "linkcheck" renderer
2020-07-24 00:42:26 [ERROR] (mdbook::renderer): The command `mdbook-linkcheck` wasn't found, is the `linkcheck` backend installed?
2020-07-24 00:42:26 [ERROR] (mdbook::utils): Error: Rendering failed
2020-07-24 00:42:26 [ERROR] (mdbook::utils): 	Caused By: Unable to start the backend
2020-07-24 00:42:26 [ERROR] (mdbook::utils): 	Caused By: No such file or directory (os error 2)
```

Ideally, I would like to be able to serve the handbook locally for testing purposes without having `mdbook-linkcheck` installed. I can try to fix this issue, but I don't know what the preferred behavior is. Would it be desirable to remove the failure from `mdbook serve`, so it acts like `mdbook build` and simply complains about the lacking backend, or should `mdbook serve` accept a flag to ignore missing backends and/or not use any backends beyond the HTML one?

ericonr:
If an uninstalled backend (command not found) isn't marked optional,
warn about it, since it causes mdbook commands to error out.

Fixes #1283

### DIFF
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -173,7 +173,9 @@ impl CmdRenderer {
                 return Ok(());
             } else {
                 error!(
-                    "The command `{}` wasn't found, is the `{}` backend installed?",
+                    "The command `{0}` wasn't found, is the \"{1}\" backend installed? \
+                    If you want to ignore this error when the \"{1}\" backend is not installed, \
+                    set `optional = true` in the `[output.{1}]` section of the book.toml configuration file.",
                     self.cmd, self.name
                 );
             }
--- a/book-example/src/for_developers/backends.md
+++ b/book-example/src/for_developers/backends.md
@@ -334,7 +334,9 @@ the usual `RUST_LOG` to control logging verbosity.
 If you enable a backend that isn't installed, the default behavior is to throw an error:
 
 ```text
-The command wasn't found, is the "wordcount" backend installed?
+The command `mdbook-wordcount` wasn't found, is the "wordcount" backend installed?
+If you want to ignore this error when the "wordcount" backend is not installed,
+set `optional = true` in the `[output.wordcount]` section of the book.toml configuration file.
 ```
 
 This behavior can be changed by marking the backend as optional.


src/renderer/mod.rs
```diff
@@ -173,7 +173,8 @@ impl CmdRenderer {
                 return Ok(());
             } else {
                 error!(
-                    "The command `{}` wasn't found, is the `{}` backend installed?",
+                    "The command `{}` wasn't found, is the `{}` backend installed? \
+                    It isn't marked as optional.",
```
ehuss, line 177: May I suggest something like:

```
If you want to ignore this error when the `{}` backend is not installed, 
set `optional = true` in the `[output.{}]` section of the book.toml configuration file.
```

ericonr: I felt it would be too verbose to have a longer message. If you prefer it, though, sure.

ehuss: Yea, I generally prefer the longer error message. It can help guide the user to a fix, and provide some more discovery and context.

ericonr: What do you think of this new version?

PR #1285, opened 2020-07-25
Handled UTF-8 BOM

### LINKED ISSUE #1155: Missing support of byte order mark
Markdown files encoded in UTF8+BOM can be rendered incorrectly, as `#` in the first line is not handled.

FrankHB:
Fixed #1155 .

### DIFF
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -265,7 +265,7 @@ fn load_chapter<P: AsRef<Path>>(
         })?;
 
         if content.as_bytes().starts_with(b"\xef\xbb\xbf") {
-            content = content[3..].to_string()
+            content.replace_range(..3, "");
         }
 
         let stripped = location
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -398,6 +398,29 @@ And here is some \
     }
 
     #[test]
+    fn load_a_single_chapter_with_utf8_bom_from_disk() {
+        let temp_dir = TempFileBuilder::new().prefix("book").tempdir().unwrap();
+
+        let chapter_path = temp_dir.path().join("chapter_1.md");
+        File::create(&chapter_path)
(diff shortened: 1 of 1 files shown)


src/book/book.rs
```diff
@@ -264,6 +264,10 @@ fn load_chapter<P: AsRef<Path>>(
             format!("Unable to read \"{}\" ({})", link.name, location.display())
         })?;
 
+        if content.as_bytes().starts_with(b"\xef\xbb\xbf") {
+            content = content[3..].to_string()
```
ehuss, line 268: Just to avoid the extra allocation, perhaps this could be `content.replace_range(..3, "");`

FrankHB:
Not sure where should be the test. Is an additional BOM in `tests/dummy_book/src/first/unicode.md` enough?

ehuss:
I'm a little concerned about having invisible bytes in a file. `unicode.md` is already quite subtle. But I think if you have an editor that can't handle a BOM properly, it will have a risk of corrupting the existing unicode, so it's probably fine.

My only other idea is to create a simple 1-chapter book in a temp directory like `book_with_a_reserved_filename_does_not_build` does.  Whichever seems easiest!

FrankHB:
Sorry, I don't get the point of the temp directory.

As of the subtleness... I largely agree. Nevertheless, I can use a hex editor to make sure only the BOM is inserted exactly. Anyway, the visible part of the file has already warn us about this. (Or one more visible warning about invisible bytes?)

ehuss:
Creating a book in Rust code (and writing it to a temp directory) would allow writing the content in a Rust string where the characters can be escaped and explicit. For example, `fs::write(chapter, "\u{feff}# Example chapter")`.

ehuss (approved):
Thanks!

PR #1290, opened 2020-07-27
theme: Update google analytics

HaoZeke:
The existing implementation no long collects any metrics.

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -228,20 +228,13 @@
 
         {{#if google_analytics}}
         <!-- Google Analytics Tag -->
+        <!-- Global site tag (gtag.js) - Google Analytics -->
+        <script async src="https://www.googletagmanager.com/gtag/js?id={{google_analytics}}"></script>
         <script type="text/javascript">
-            var localAddrs = ["localhost", "127.0.0.1", ""];
-
-            // make sure we don't activate google analytics if the developer is
-            // inspecting the book locally...
-            if (localAddrs.indexOf(document.location.hostname) === -1) {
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR!  Can you share any information on the change to the Global Site Tag, and why the old version does not work?  Is there any reason to *not* switch? I checked my test page, and by entering the Tracking ID, the current google analytics code seems to work fine.  Perhaps you have a blocker somewhere preventing it from working?

Also, this removes the check for localhost, which is intended to avoid hits during local testing. Is that not needed anymore? Any reason to remove it?

HaoZeke:
Ah, yes, I did later realize I had a blocker configured, but the `gtag` version is now the only recommended code for tracking site hits. The `localhost` rule isn't required in the new version (the js handles it).

Given all that, it might be OK to keep the current code for a while, but since it is no longer recommended, it might be depreciated later.

src/theme/index.hbs
```diff
@@ -224,20 +224,13 @@
 
         {{#if google_analytics}}
         <!-- Google Analytics Tag -->
+        <!-- Global site tag (gtag.js) - Google Analytics -->
+        <script async src="https://www.googletagmanager.com/gtag/js?id=UA-109503488-17"></script>
```
azerupi, line 228: It has been a long time since I have used Google analytics, but isn't `UA-109503488-17` a personal id that should not be here?

HaoZeke: Yup, dunno how I missed that, will replace it (script does replace it)

ehuss:

PR #1291, opened 2020-07-29
summary: turn SoftBreak events into spaces.

### LINKED ISSUE #1218: Sidebar doesn't add a space character when item is split in two lines
Hey there! We are using mdBook for the documentation for Void Linux, here: https://docs.voidlinux.org/.

If you look at the sidebar, under graphics drivers, there's a "NVIDIAOptimus" item. This item is specified in `SUMMARY.md` as:

```
- [NVIDIA
   Optimus](link.)
```

It is split in two lines due to our linting, where lines are constrained to 80 columns. The original source can be found in https://raw.githubusercontent.com/void-linux/void-docs/master/src/SUMMARY.md . GitHub formatting, when displaying this file ([here](https://github.com/void-linux/void-docs/blob/master/src/SUMMARY.md)), displays it correctly as "NVIDIA Optimus".

This seems to be a bug when creating the sidebar. I haven't noticed anything similar with normal files.

Thanks!

ericonr:
Summary items that had their name split into two lines ended up with the
last word of one line and the first word of the next line glued
together, since a space wasn't added.

Fixes #1218

### DIFF
--- a/src/book/summary.rs
+++ b/src/book/summary.rs
@@ -569,6 +569,7 @@ fn stringify_events(events: Vec<Event<'_>>) -> String {
         .into_iter()
         .filter_map(|t| match t {
             Event::Text(text) | Event::Code(text) => Some(text.into_string()),
+            Event::SoftBreak => Some(String::from(" ")),
             _ => None,
         })
         .collect()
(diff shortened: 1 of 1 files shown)


ericonr:
I feel like the change itself is mostly self explanatory, but I can leave a comment and/or create a test case, if desired.

ericonr:
I'm trying to add a test to it, in the end of the same file. Hope it's enough.

ericonr:
Test case added c:

Issue #1287, opened 2020-07-26
Can't include file with spaces in name


Evian-Zhang:
I found issue [Files with spaces are not found #527](https://github.com/rust-lang/mdBook/issues/527), which I had exactly the same problem.

I have a file `name with space.md`, and I want to include this file in `SUMMARY.md` as:

```markdown
- [name with sapce](./name with space.md)
```

and then I used `mdbook build`, and failed with
```
2020-07-26 10:06:23 [WARN] (mdbook::book::summary): Expected a start of a link, actually got Some(Text(Borrowed("[")))
2020-07-26 10:06:23 [ERROR] (mdbook::utils): Error: Summary parsing failed
2020-07-26 10:06:23 [ERROR] (mdbook::utils): 	Caused By: There was an error parsing the numbered chapters
2020-07-26 10:06:23 [ERROR] (mdbook::utils): 	Caused By: There was an error parsing the numbered chapters
2020-07-26 10:06:23 [ERROR] (mdbook::utils): 	Caused By: failed to parse SUMMARY.md line 4, column 3: The link items for nested chapters must only contain a hyperlink
```

I have tried:

```markdown
- [name with sapce]("./name with space.md")
- [name with sapce](./name+with+space.md)
- [name with sapce](./name%20with%20space.md)
- [name with sapce](./name\ with\ space.md)
```

as suggested in the issue #527, but none of those approaches succeeded. Method 1 and 4 failed with the same output, method 2 and 3 created a dummy `name+with+space.md` or `name%20with%20space.md` file, and the output html does not contain the content in my `name with space.md`.

I am working on macOS 10.15, with mdbook v0.4.1

magicdmer:
I meet the same problem , use win32 lastest relase

Issue #527, opened 2018-01-05
Files with spaces are not found


dns2utf8:
Hi all

I noticed that referenced files with spaces in the filename are not found.
However if I replace the spaces with `+` es mdbook finds the file.
Sadly the file is not accessable over http later.
Funny side effect reverting the `+` to spaces in the browser finds the correct file again.

Best,
Stefan

Michael-F-Bryan:
How exactly are you referencing these files? Can you provide an example?

It sounds like the issue is that markdown links are rendered as normal HTML links (the `<a>` tag), and browsers don't like having spaces in URLs.

dns2utf8:
I have the book in a weid state currently, but with this commit I get the error from the file in the way as it does translate the link into html but it does not work.
https://gitlab.com/dns2utf8/buch.feuerfestival.org/-/jobs/46822514

The CI throws a file not found error but I dont know why yet.

Michael-F-Bryan:
I just checked out a local version of your book and I believe I was able to reproduce the issue you're talking about.

First you add a dummy chapter to the book where the name contains a space:
```
$ echo "- [Another Chapter](./another chapter.md)" >> src/SUMMARY.md
$ echo "# Another Chapter" >> "src/another chapter.md"
```

Then you go to another chapter and try to link to it.

None of these work:

- `[Another Chapter](./another chapter.md)` - won't work because the rendered document being linked to is HTML
- `[Another Chapter](./another chapter.html)` - links containing spaces aren't valid markdown ([according to here](https://github.com/Microsoft/vscode/issues/24763))

However these work:

- `[Another Chapter](./another\ chapter.html)` - the space is escaped (valid markdown)
- `[Another Chapter](./another%20chapter.html)` - uses percent encoding
`[Another Chapter](./another+chapter.html)` - the `+` is also a valid replacement for space in a URL

It seems the issue is actually intended behaviour from the Common Mark spec. Markdown links aren't allowed to contain space characters (it introduces ambiguities with the [link title]), so you need to escape them with either `\`, `+`, or `%20`.

---

> The CI throws a file not found error but I dont know why yet.

The `master` version of `mdbook` has much better logging and error handling, so you get nicer error messages. Apparently you've put `video.css` as in your `book.toml`, but because additional CSS and JS files are resolved relative to the book's root it should actually be `src/video.css`.

```
Error: Unable to copy across additional CSS and JS
    Caused by: Copying video.css failed
[...]

PR #1293, opened 2020-08-03
allow space in SUMMARY.md's link destination

Evian-Zhang:
Related issues: #527 #1287 

As stated in [CommonMark Spec](https://spec.commonmark.org/0.28/#link-destination):

> A link destination consists of either
>
> *  a sequence of zero or more characters between an opening `<` and a closing `>` that contains no spaces, line breaks, or unescaped `<` or `>` characters, or
>
> * a nonempty sequence of characters that does not include ASCII space or control characters, and includes parentheses only if (a) they are backslash-escaped or (b) they are part of a balanced pair of unescaped parentheses. (Implementations may impose limits on parentheses nesting to avoid performance issues, but at least three levels of nesting should be supported.)

and [according to this](https://github.com/Microsoft/vscode/issues/24763) and MDN spec [Percent-encoding](https://developer.mozilla.org/en-US/docs/Glossary/percent-encoding), there are three ways to allow space in link destination:

```markdown
- [test1](./test%20link1.md)
- [test2](./test+link2.md)
- [test3](<./test link3.md>)
```

For now, mdBook only supports method 3, which is rarely used compared to method 1 and 2.

So I fix it and add a test.

### DIFF
--- a/src/book/summary.rs
+++ b/src/book/summary.rs
@@ -333,6 +333,7 @@ impl<'a> SummaryParser<'a> {
 
     /// Finishes parsing a link once the `Event::Start(Tag::Link(..))` has been opened.
     fn parse_link(&mut self, href: String) -> Link {
+        let href = href.replace("%20", " ").replace("+", " ");
         let link_content = collect_events!(self.stream, end Tag::Link(..));
         let name = stringify_events(link_content);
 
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR!  I would prefer to stick with the behavior from markdown where `%20` is converted to space.  The `+` transformation isn't something that is normally done for path components in URLs (or in markdown).  Would you mind removing the `+` transformation?

Also, it looks like rustfmt needs to be run.

Evian-Zhang:
> Thanks for the PR!  I would prefer to stick with the behavior from markdown where `%20` is converted to space.  The `+` transformation isn't something that is normally done for path components in URLs (or in markdown).  Would you mind removing the `+` transformation?
> 
> 
> 
> Also, it looks like rustfmt needs to be run.
> 
> 

Ok. I have removed the `+` transformation and ran rustfmt.

PR #1307, opened 2020-08-29
Prefer bundled version of Source Code Pro

### LINKED ISSUE #1304: Syntax highlighting doesn't work on macOS with Firefox 80
Most likely because of [this issue with Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1520157) (also see https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/Firefox.2080.2E0.20removes.20colors.20from.20the.20playground.20on.20MacOS.20.2E.2E.2E).

However, this issue does not occur with Rustdoc. [According](https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/Firefox.2080.2E0.20removes.20colors.20from.20the.20playground.20on.20MacOS.20.2E.2E.2E/near/208022585) to @ehuss, this is because:

> Yea, in rustdoc, it is intentionally using an older version of Source Code Pro to avoid this problem.

Perhaps mdBook can use the same trick as Rustdoc to avoid this issue?

camelid:
Fixes #1304.

This prevents an issue with Firefox 80 on macOS that prevents syntax
highlighting from working.

### DIFF
--- a/src/theme/fonts/fonts.css
+++ b/src/theme/fonts/fonts.css
@@ -96,6 +96,5 @@
   font-family: 'Source Code Pro';
   font-style: normal;
   font-weight: 500;
-  src: url('source-code-pro-v11-all-charsets-500.woff2') format('woff2'),
-       local('Source Code Pro Medium'), local('SourceCodePro-Medium');
+  src: url('source-code-pro-v11-all-charsets-500.woff2') format('woff2');
 }
--- a/src/theme/fonts/fonts.css
+++ b/src/theme/fonts/fonts.css
@@ -96,6 +96,6 @@
   font-family: 'Source Code Pro';
   font-style: normal;
   font-weight: 500;
-  src: local('Source Code Pro Medium'), local('SourceCodePro-Medium'),
-       url('source-code-pro-v11-all-charsets-500.woff2') format('woff2');
+  src: url('source-code-pro-v11-all-charsets-500.woff2') format('woff2'),
+       local('Source Code Pro Medium'), local('SourceCodePro-Medium');
 }


camelid:
## Before (master branch)

image

## After (this branch)

image

src/theme/fonts/fonts.css
```diff
@@ -96,6 +96,6 @@
   font-family: 'Source Code Pro';
   font-style: normal;
   font-weight: 500;
-  src: local('Source Code Pro Medium'), local('SourceCodePro-Medium'),
-       url('source-code-pro-v11-all-charsets-500.woff2') format('woff2');
+  src: url('source-code-pro-v11-all-charsets-500.woff2') format('woff2'),
+       local('Source Code Pro Medium'), local('SourceCodePro-Medium');
```
camelid, line 100: It may make sense to not even try the local version of SCP, which would also match Rustdoc's behavior: https://github.com/rust-lang/rust/blob/7fc048f0712ba515ca11fac204921b9ad8a0c5a3/src/librustdoc/html/static/rustdoc.css#L36-L43

```suggestion
  src: url('source-code-pro-v11-all-charsets-500.woff2') format('woff2');
```

camelid: It would perhaps be better to fall back to another font where syntax highlighting *does* work, rather than using a potentially buggy local version of SCP where syntax highlighting does not work.

camelid: Also see @nrc's comment here: https://github.com/rust-lang/rust/issues/24355#issuecomment-248443541

ehuss: I think it would be fine to completely remove the `local` setting here.  I believe they are processed in order, and would only use local if the first one doesn't load for some reason, which should be rare (and rarer still that someone has source code pro installed locally).

camelid: Well I'm guessing a decent number of programmers have SCP installed, but I agree that it's unlikely that the bundled version wouldn't load and yet the page would. Also it would just fall back to a different monospace font which is fine:

https://github.com/rust-lang/mdBook/blob/4b773024ae9872c80b7e866e03472b08523e33b3/src/theme/css/general.css#L23-L26

ehuss (approved):
Thanks!

camelid:
Thanks for reviewing! :)

PR #1309, opened 2020-08-31
Fix missing space before draft chapter titles

### LINKED ISSUE #1308: Draft chapters are missing a space between the number and title
Draft chapters don't display any space in between the chapter number and the name, like normal chapters do, shown below:

mdbook-missing-space

nerosnm:
Fixes #1308.

However, I'm not necessarily experienced enough with HTML/CSS to know if this is an acceptable solution or if it qualifies as a hack. Or maybe the `<div>` should be given a class? Happy to change this as needed.

### DIFF
--- a/src/renderer/html_handlebars/helpers/toc.rs
+++ b/src/renderer/html_handlebars/helpers/toc.rs
@@ -108,32 +108,32 @@ impl HelperDef for RenderToc {
             }
 
             // Link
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!  I am no expert on flexbox, either, so I don't really know if this is the most elegant way to fix it, but it looks fine on all the browsers I tried.

PR #1310, opened 2020-09-02
End .nojekyll file with newline

dtolnay:
See https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline.

Before:

```console
/path/to$  cat book/.nojekyll
This file makes sure that Github Pages doesn't process mdBook's output./path/to$  ▎
```

After:

```console
/path/to$  cat book/.nojekyll
This file makes sure that Github Pages doesn't process mdBook's output.
/path/to$  ▎
```


ehuss (approved):
Thanks!

PR #1311, opened 2020-09-02
Support emitting CNAME file for publishing at a custom domain

dtolnay:
This PR adds a `cname` configuration to \[output.html\] which is rendered into a CNAME file in the root of the generated site as required by GitHub Pages (see [*Managing a custom domain for your GitHub Pages site*](https://docs.github.com/en/github/working-with-github-pages/managing-a-custom-domain-for-your-github-pages-site)).

```toml
[output.html]
cname = "serde.rs"
```

The CNAME file is how GitHub knows which repo's site to serve when someone visits your domain. DNS records only point from domain (serde.rs) to domain (serde-rs.github.io), so the additional information is required to know that serde-rs.github.io/serde (i.e. the Pages site for github.com/serde-rs/serde) is the right repo to serve for that domain.


ehuss (approved):
Thanks!

PR #1312, opened 2020-09-03
Adds an example of git-repository-url in book.toml

IronicBadger:
#802 seemingly didn't remember to update the docs with this option, this PR address it.

### DIFF
--- a/book-example/src/format/config.md
+++ b/book-example/src/format/config.md
@@ -23,6 +23,7 @@ create-missing = false
 
 [output.html]
 additional-css = ["custom.css"]
+git-repository-url = "https://github.com/rust-lang/mdBook"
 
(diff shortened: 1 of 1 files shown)


ehuss:
I'm not sure I understand.  That example is just an small sample of some settings, it doesn't include all of them.  Why add this particular setting?  This setting is documented in https://rust-lang.github.io/mdBook/format/config.html#html-renderer-options, and at the end of that section there is a complete TOML example that includes `git-repository-url`.

IronicBadger:
Apparently reading is hard. I thought I'd checked thoroughly but clearly not. 

Feel free to close this one out.

ehuss:
Sure, no problem!

PR #1313, opened 2020-09-04
collect all test failures before failing

guswynn:
I found that iterating on `mdbook` test is fairly slow as any failure in any file halts the process
This prints the error with `error~`, marks a failure, and then moves on

I tested that the return code is 0 on a passing book (i had a working local version of the cargo book) and that its non-0 on a non-passing book (`dummy_book` in `tests` is good for this)


ehuss (approved):
Thanks!  I have kinda wanted this in the past, too.  I made a small wording change, which looked like maybe a typo.

PR #1323, opened 2020-09-09
Resolve full file paths when copying files

### LINKED ISSUE #1157: images will not be copied to 'book' if symbolic links are utilized in src folder
as we work on a multilingual project and #5 (=multilingual support) is still unsolved, i had to look for a workaround. that's why i worked with two independent trees, each containing: SUMMARY.md, config.tom, and a src folder, but tried to use **symbolic links** for the actual content, which we still maintain in an external repository.

this works for most of the content, but **images will not be copied to the generated `book` folder**.

the pages will be only include broken html links to the location, where the image files should be located.

it doesn't make a difference, if the images themself are realized as symbolic links or just the folders, where they are located. in both cases you will observe the same behavior.

this affects at least the last release of mdbook = v0.3.5

### LINKED ISSUE #1209: HTML renderer doesn't handle windows symlinks
In one of my books, I'd like to distribute a copy of a PDF. To make sure the document in my book doesn't get out of sync with the original I'm using a symlink.

I'm on a windows computer, so I created the symlink with the following:

```console
$ mklink "../../../Install Files/Documentation/some-document.pdf" "some-document.pdf"
```

My directory structure looks something like this:

- `docs/`
  - `book.toml`
  - `src/`
    - `SUMMARY.md`
    - `overview.md`
    - `some-document.pdf` -> `../../../Install Files/Documentation/some-document.pdf`

Now when I run the build server `mdbook serve --open` go to `http://localhost:3000/overview.html`, and click the link to `some-document.pdf` it will serve up a zero byte file... Which is kinda not helpful.

---

I believe what's happened is the HTML renderer copied the symlink as-is, meaning the output directory now contains a broken symlink. Because the output directory is meant to be copied around and standalone, we should be copying the actual file contents to the output directory.

One solution would be for the HTML renderer to run `std::fs::canonicalize()` first when copying files across (e.g. to resolve any symlinks), probably in [`mdbook::utils::copy_files_except_ext()`](https://github.com/rust-lang/mdBook/blob/11f95f76e6d054af294d46c946e32e98b4afb2e5/src/utils/fs.rs#L94). I'm not sure how much code depends on this function though.

iffyio:
Currently, the `copy_files` function doesn't support symlink files due to
its use of `DirEntry::metadata` - To avoid this issue, this patch
resolves the file path first before checking for metadata.

Fixes #1157
Fixes #1209

### DIFF
--- a/src/utils/fs.rs
+++ b/src/utils/fs.rs
@@ -110,7 +110,7 @@ pub fn copy_files_except_ext(
 
     for entry in fs::read_dir(from)? {
         let entry = entry?;
-        let metadata = entry.metadata()?;
+        let metadata = entry.path().metadata()?;
 
         // If the entry is a dir and the recursive option is enabled, call itself
         if metadata.is_dir() && recursive {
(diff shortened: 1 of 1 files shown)


azerupi (approved):

ehuss (approved):
Thanks for the PR!

This has the potential to be relatively risky, but I haven't been able to think of any specific problems. Windows users will need to enable Developer Mode, which I guess isn't too bad. There might be some risk that the user wants a symlink in the output directory, but since I believe that doesn't work today, this seems like a good/simple first step. I also think this will fail if there is a symlink cycle, but I believe most operating systems will detect that with an error.

PR #1330, opened 2020-09-15
Use embedded SVG instead of fonts for icons

notriddle:
The [downsides of icon fonts] are well-documented, and also, why ship all of the icons when it only uses 14?

[downsides of icon fonts]: https://speakerdeck.com/ninjanails/death-to-icon-fonts


ehuss:
Thanks for the PR!  I don't think this is something we can land right away, because I believe there are books that are using the FA icons, and this would break them. I'll try to review it when I get a chance, but merging may need to wait until the next major release.

notriddle:
That’s fine.

notriddle:
@ehuss Docs and inline fixup for fontawesome icons have been added to the HTML renderer, so that books using font-awesome will still be able to do so.

notriddle:
@ehuss Now that the fixups are included to allow books that already use font-awesome to continue doing so, and the documentation is written, is there anything else preventing this from being included in the 0.5.0 milestone?

uncenter:
Is there interest in reviving this pull request? I can try to bring these changes up to date with the latest in the repository... 👀

Dylan-DPC:
> Is there interest in reviving this pull request? I can try to bring these changes up to date with the latest in the repository... 👀

@uncenter yes if you can that would be helpful. Thanks

ehuss:
@notriddle I rebased and tried to update things. I think I mostly got everything going, but I noticed two issues:

* The spinner (for loading search index) doesn't animate.
* The spacing of the buttons on code blocks doesn't look right.

I ran out of time today to investigate that more. Do you offhand know what's up with the spinner animation?

notriddle:
To fix the spinner animation, this code needs added to the general.css file:

```css
.fa-spin {
	animation: rotating 2s linear infinite;
}

@keyframes rotating {
	from {
		transform: rotate(0deg);
	}
	to {
		transform: rotate(360deg);
	}
}
```

The fontawesome package happens to include this utility class. Since it's not there any more, we need to provide it ourselves.

WhyNotHugo:
Thank you so much! :pray: 

Finally such a huge accessibility issue has been fixed. I can't wait for this to trickle to different deployments and mdBooks start to actually be usable for everyone! <3

PR #1334, opened 2020-09-20
Add chapter number before the first page header

letientai299:
So that when reading the book-example, for example, we see
"3.1 SUMMARY.MD" instead of "SUMMARY.md", know how far we are without
opening the sidebar.


ehuss:
I think I would prefer for this to be an option (off by default). Also, the implementation here looks quite brittle. I think it would be best if it was handling the semantics at the markdown level instead of arbitrarily modifying whitespace.

letientai299:
@ehuss, I still learning Rust and try to understand the code base, will update this MR to address your points.

letientai299:
Closing since I don't really have a good solution for this feature.

PR #1335, opened 2020-09-22
Fix print icon.

ehuss:
I accidentally missed that the logic was reversed in #1169.

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -648,9 +648,9 @@ fn make_data(
         data.insert("playground_copyable".to_owned(), json!(true));
     }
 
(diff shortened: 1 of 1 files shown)


kalindkaria:
Sorry to bother you again.
As you mentioned in #1169  I pulled and checked but can still see the **printer** icon. Don't know I might be wrong.
Can you please check at your end once and confirm.
Thanks.

ehuss:
Hm, seems to work form me:

`output.html.print.enable = true` (default):
image

`output.html.print.enable = false`:
image

This is with d0deee90b04068ed949f524bb682a47fa26f2218.  Maybe check that you have the right version and the syntax in `book.toml` is correct?

kalindkaria:
Yes, really sorry to ask this.
I had my custom theme folder in `book.toml`. I disabled it and it worked well.
Thanks alot.
I think #1169 can be closed now.

PR #1336, opened 2020-09-23
Rename `book-example` to `guide`

### LINKED ISSUE #1319: Rename `book-example` to `guide`?
I know it functions as both, but it's not just "hello, world". What do you think?

camelid:
Closes #1319.

`book-example` is a bit of a strange name given that it's not just an
example.


Dylan-DPC:
Thanks for the PR :)

PR #1341, opened 2020-09-28
Allow preprocessors to pass generated resources (fix issue #1087)

sytsereitsma:
The problem it solves (#1087) is that some preprocessors actually generate files that need to be staged somewhere. I used to copy them directly to the book output dir, but the HBS rederer now removes the book output dir after the preprocessor has been executed (introduced with #985). Apart from that, copying the files to the book output dir will probably only work for the HBS renderer.

To overcome the above problem I have added a resource list to the Chapter struct. Renderers can read the extra resource data from there and process it as they see fit.

I do think the handles_book_resources test (hbs_renderer.rs:1131) may be a bit brittle. The render function does not really lend it self for unit testing.


sytsereitsma:
@ehuss This is the alternative solution using the book as single source of truth as @Michael-F-Bryan suggested ([PR 1144](https://github.com/rust-lang/mdBook/pull/1144))

sytsereitsma:
@ehuss A gentle reminder

PR #1343, opened 2020-09-29
Add expand rust macro button

sigmaSd:
context: https://github.com/Veykril/tlborm/issues/16

If this is an accepted feature, I'll try to figure out if I can add syntax highlighting to the result.

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -97,6 +97,46 @@ function playground_text(playground) {
         }
     }
 
+    function expand_rust_macro(code_block) {
+      var result_block = code_block.querySelector(".result");
+      if (!result_block) {
+          result_block = document.createElement('code');
+          result_block.className = 'result hljs language-bash';
+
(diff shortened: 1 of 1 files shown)


sigmaSd:
For highlighting , using this code on the console:
```js
Array
            .from(document.querySelectorAll('code:not(.editable)'))
            .forEach(function (block) { hljs.highlightBlock(block); });
```
The highlighting is incorrect, it greys out most of the code (image truncuated)
broken_highlight

ehuss:
I don't think this is something we can do using the playground, as we can't use arbitrary endpoints in their API.

I might suggest putting the examples in separate rust files, using `rustdoc_include` to embed them in the book, have a separate script which will run `cargo expand` on the examples, and use javascript/css to provide some kind of "expand" button to show those expanded results.

sigmaSd:
Understandable, thanks for the suggestion I'll ping the author about it.

Issue #1350, opened 2020-10-13
Add Bibliography setion to books with initial support for Bibtex-format


francisco-perez-sorrosal:
When you write books it's common to add a bibliography and reference citation entries from it. I think it would be nice to have support for this feature in mdBook, allowing to automatically add BibText files as a bibliography and add a referencing template system to allow referencing citation-keys in the bibliography through the markdown files.

francisco-perez-sorrosal:
Added an initial implementation. Added some tests to check the main paths and functionality. Also basic documentation in the mdBook guide. All the test were passing in my local machine. Please let me know if you're interested in this.

francisco-perez-sorrosal:
This can be done now through [this plugin/preprocessor](https://github.com/francisco-perez-sorrosal/mdbook-bib).
Crates:  https://crates.io/crates/mdbook-bib
Src/Binaries: https://github.com/francisco-perez-sorrosal/mdbook-bib/releases/tag/v0.0.3

pesho-ivanov:
I am surprised to see so little interest in this feature. References are central to developing a book. Is there maybe another solution that mdBook users have?

eliasboegel:
I'm also surprised. Its an essential component even in detailed software documentation, which was the primary use for mdbook as far as I am aware. The preprocessor suggested works, but it is very basic, does not support CSL styles and I was not even able to get some of the already existing style templates to work. In my opinion, references and citing is a feature so fundamental that it should be developed as part of mdbook.

PR #1351, opened 2020-10-13
Add Bibliography with Bibtex-format support

francisco-perez-sorrosal:
Allows adding a bibliography in [BibTex format](http://www.bibtex.org/Format/) to your book and reference the citation keys in your markdown files.

Instructions have been added to the guide.

This is related to issue #1350


ehuss:
Is this something that can exist as an external plugin?  I would prefer to minimize the maintenance burden on the project, and extensions like this seem like they would be better to be used as a separate addon.

francisco-perez-sorrosal:
Hi @ehuss. I understand your concern about minimizing the impact of a feature like this in the project structure. When I implemented it, I tried to impact minimally the project, by defining meaningful abstractions (e.g. `BibItem`, `Bibliography`...) according to the already present ones in the project regarding the structure of an mdBook (such as `Chapter` or `BookItem`...) Bibliographies are part of any book so the abstract concept is there.

I had no previous Rust experience, and I basically implemented this functionality in just a week starting from scratch with both Rust and the mdBook project. What do you mean by `external plugin`? I've been looking for `plugin` in the mdBook doc and the project structure and I can't find anything related to plugins in the mdBook project.

francisco-perez-sorrosal:
BTW I think issue #1371 is related also for some bibliography support too in mdBook.

ehuss:
> What do you mean by `external plugin`?

Ah, I mean a [preprocessor](https://rust-lang.github.io/mdBook/for_developers/preprocessors.html).  A preprocessor can make arbitrary changes to the book content. ([Plugins](https://github.com/rust-lang/mdBook/#3rd-party-plugins) sort of loosely refer to either preprocessors or renderers.)

francisco-perez-sorrosal:
This can be done now through [this plugin/preprocessor](https://github.com/francisco-perez-sorrosal/mdbook-bib).
Crates:  https://crates.io/crates/mdbook-bib
Src/Binaries: https://github.com/francisco-perez-sorrosal/mdbook-bib/releases/tag/v0.0.3

PR #1353, opened 2020-10-15
Enhanced contrast theme, for users with visual restrictions etc.

tim-seoss:
Now that I'm half way through my fifth decade, I find the default 'light' theme quite tiring for more than a few minutes.

Whilst reading mdBook material, I often find myself physically moving my face closer to the screen and/or zooming in more than I do on other similar material rendered with other frameworks.

I often use browser extensions to increase text contrast, but these don't cope well with mdBook's output (in particular, I haven't found one which works well with mdBook's syntax highlighting).

For my own use, I've created an enhanced contrast theme, based on the 'light' theme.  I've also asked the opinions of others whilst viewing, and have received universally positive feedback, so I hope you will consider this addition to mdBook.

Any feedback welcome!

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -207,6 +207,11 @@ impl HtmlHandlebars {
         write_file(destination, "highlight.css", &theme.highlight_css)?;
         write_file(destination, "tomorrow-night.css", &theme.tomorrow_night_css)?;
         write_file(destination, "ayu-highlight.css", &theme.ayu_highlight_css)?;
+        write_file(
+            destination,
+            "contrast-highlight.css",
+            &theme.contrast_highlight_css,
+        )?;
         write_file(destination, "highlight.js", &theme.highlight_js)?;
         write_file(destination, "clipboard.min.js", &theme.clipboard_js)?;
         write_file(
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -278,6 +278,7 @@ function playground_text(playground) {
     var themeColorMetaTag = document.querySelector('meta[name="theme-color"]');
     var stylesheets = {
         ayuHighlight: document.querySelector("[href$='ayu-highlight.css']"),
+        contrastHighlight: document.querySelector("[href$='contrast-highlight.css']"),
         tomorrowNight: document.querySelector("[href$='tomorrow-night.css']"),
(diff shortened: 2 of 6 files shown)


tim-seoss:
I've put a snapshot of https://github.com/rust-lang/book.git with the contrast theme added here: https://buttersideup.com/rustbook/ .  In that snapshot, contrast is not the default theme, so you will need to select it using the '🖌' icon.

tim-seoss:
Tweaked `--inline-code-color` based on feedback, and force-pushed.

pojntfx:
I'd like to second this. Both the default light and dark Navy theme fail the WAVE accessibility test ATM.

endgame:
After raising #1442, I was pointed at this PR. I would call the "contrast" theme almost strictly better than "light (default)". (After I found the theme change button, because I had no idea what the light-grey-on-white blob was actually representing.)

iamfrank:
I endorse this pull request. 

mdBook has some issues in the accessibility department - text contrast for readability being one of them. 
This makes mdBook a no-go on any publicly facing website, since web applications are required to adhere to WCAG standards in our part of Europe.

tim-seoss:
Although it's pretty close, in a couple of situations this theme fails to conform to the [W3C WCAG](https://www.w3.org/WAI/standards-guidelines/wcag/) 2.1 standard.  I'll revise this PR to fix that (without that revision, this shouldn't be merged, I think).

ehuss:
Hi @tim-seoss!  Sorry for the long silence.  I've actually been having private conversations with people about this PR.  Would you be willing to change it so that it just updates the default light scheme?  I would prefer to avoid adding new themes, and I think this one is better than the default, and I'd be happy to change it.

tim-seoss:
OK, no problem, thanks for reviewing this!

I'll try and get the PR updated tomorrow.

As of today's revision, the text contrast is WCAG compliant, but...

Now that my machine's blue light filter has come on during the evening, I've realised that the color contrast between the orange and magenta text (within code blocks) has dropped a bit further than I'd like, so I'll see if I can improve that.

I expect the tweak to be subtle, but because the plan is now to change the default theme, I'd really like to make sure it's spot-on, so I'll see if I can get an accessibility specialist to review it too.

tim-seoss:
I've created a new PR #1470 which supersedes this, in response to feedback here, and elsewhere.

PR #1355, opened 2020-10-16
Test cargo install in CI and bump MSRV

Alexendoo:
Currently trying to `cargo install mdbook` on 1.39 [fails to compile](https://github.com/Alexendoo/mdBook/runs/1265268352?check_suite_focus=true#step:6:189), for some reason it builds/tests just fine but when using `cargo install` it ignores the Cargo.lock, compiling `remove_dir_all v0.5.3` instead of `0.5.2`

This bumps the MSRV to 1.40 where `remove_dir_all v0.5.3` does compile and adds a test to catch it in the future

Ideally the MRSV could stay at 1.39, but I don't know what's going on with Cargo.lock being sometimes ignored

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,6 +40,8 @@ jobs:
       run: cargo test
     - name: Test no default
       run: cargo test --no-default-features
+    - name: Install mdbook
+      run: cargo install --path .
 
   rustfmt:
     name: Rustfmt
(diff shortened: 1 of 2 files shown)


ehuss:
`cargo install` needs the `--locked` flag to use `Cargo.lock`.  I would probably recommend adding that to the documentation instead.  I'm not sure what you are using `cargo install` for, but I would also recommend using the pre-built binaries if possible, which should download and install much faster.

Alexendoo:
Thanks, I wouldn't have thought `--locked` is opt in

I didn't run into this personally, it was someone on IRC confused they were getting an error message trying to update mdbook. Though I do also use `cargo install` for it (on up to date rust), as I don't want the headache of manually managing binaries

PR #1368, opened 2020-10-28
feat(html): cache bust static files by adding hashes to file names

### LINKED ISSUE #1254: Cache busting
I don't know much about the JS ecosystem so I'm not sure if there's already a great tool for this, but it would be nice to have the option of having a hash file name for my assets, including any custom additional assets. This could also probably be extended to also minify the assets. This would all be done during `mdbook build` and would generate the hashes of all the assets, rename the file (being able to choose between `<name>.<short_hash>.<ext>` and `<long_hash>.<ext>` may be useful) and replace any references to the original assets in the generated HTML.

### LINKED ISSUE #2547: sidebar not refreshing without clearing cache
### Question

Hi.

mdbook 0.4.44
Windows 10, Firefox and Ubuntu 24.04, Firefox (browser doesn't seem to make a difference)

The sidebar only refreshes on local browser (`http://127.0.0.1:3000` from `mdbuild serve`).
On browser accessing remote server, I need to clear the browser cache (Firefox,  CTRL+F5).
Chapter data refreshes correctly on both browsers, from local and remote server.

In other words - when I change my pages, the local version of `./book` (sidebar and chapter data) is shown correctly on the local browser when `mdbook serve` is active. I only need to refresh. All good. However when I copy './book' to the server, the same data changes only shows the new chapter data, not the new sidebar data. I need to clear the browser cache (Firefox, CTRL F5) to make the new sidebar data show up.

Did I miss something or is this a bug related to invalidating caches?

### Version

```text
mdbook 0.4.44
```

notriddle:
Closes #1254
Fixes #2547


ehuss:
Thanks!  Can you provide an overview on how this works?  

The documentation will also need to be updated.

notriddle:
I've added documentation for the new config option and handlebars helper. Is there any other particular thing that's confusing?

notriddle:
And fixed a Windows-specific bug in the test code.

notriddle:
Is this change okay? Is there some documentation that it still needs?

KOBA789:
Thank you for the great patch! @notriddle 
I need this feature too.
Could anyone review this PR?

siddarthkay:
would be really cool if this feature made it into MdBook

GuillaumeGomez (approved):

ehuss:
Thanks!

I pushed a few small changes.

I noticed that the paths for the no-js toc seem to be broken. Do [these paths](https://github.com/rust-lang/mdBook/blob/4941acdb8705728a5ea9d77987f31fdafefdcee1/src/theme/toc.html.hbs#L24-L38) need to be updated? Would it be possible to update the sidebar-nojs.goml test to verify those are loading?

notriddle:
Yes, they do need updated.

I've pushed another commit that fixes this problem, and that enables hashed files in the GUI test, so it now covers this case.

Issue #1322, opened 2020-09-09
Buttons still overlap code


jyn514:
Same as #389 but with an example from https://doc.rust-lang.org/book/ch10-00-generics.html#removing-duplication-by-extracting-a-function:

![image](https://user-images.githubusercontent.com/23638587/92293665-40368580-eef3-11ea-9fec-240ac5019d7a.png)

This screenshot was taken on iOS with Safari 13.

You can also reproduce by zooming to 200% on desktop (although I don't expect anyone's browsing the book that way):

![image](https://user-images.githubusercontent.com/23638587/92293694-783dc880-eef3-11ea-83a9-397c79db02ea.png)

serverwentdown:
What about instead of #1508, we add `padding-right` when the screen is too small so that there is enough padding to scroll left and right.

joshxyzhimself:
I think possible approach is move the code downward by 1 line? So buttons have their own line?

vwkd:
Another idea is to do it like GitHub. Hide the buttons by default. Show them on hover (desktop) and on single tap (mobile). Also give them a background.

Desktop

![desktop](https://user-images.githubusercontent.com/33468089/140620030-dbb7fcae-3f7d-4dfa-a538-8001e45c167b.gif)

Mobile

![mobile](https://user-images.githubusercontent.com/33468089/140620032-92574f9a-02e2-4c64-8fcc-fc3041fa4bf5.gif)

ehuss:
I posted #1806 with a proposed fix.

I would appreciate if you could take a look and see if you have any thoughts.  I posted a demo at https://ehuss.github.io/rust-book/.  If you could look through some examples, try different browsers and platforms, different color themes, etc.  If you have access to Android, that would particularly be nice to test since I don't have access to that.

joshxyzhimself:
![Screenshot_20220517-103540_Chrome](https://user-images.githubusercontent.com/60829711/168716798-34da45f1-ab4b-4fcb-8fc9-6cbb7952399c.png)

![Screenshot_20220517-103553_Chrome](https://user-images.githubusercontent.com/60829711/168716810-ca2081ab-6239-4854-a8cc-3b0a53438e28.png)

![Screenshot_20220517-103654_Chrome](https://user-images.githubusercontent.com/60829711/168716885-fb4620d7-80a4-4527-8c9c-1d28c7dd6377.png)

joshxyzhimself:
[...]

PR #1374, opened 2020-11-06
Fix stray spacing of code blocks containing injected main

dtolnay:
For example see the first code block in https://doc.rust-lang.org/1.47.0/book/ch03-02-data-types.html.

````markdown
```rust
let guess: u32 = "42".parse().expect("Not a number!");
```
````

<br>

### Before:

Screenshot from 2020-11-05 21-31-53

<br>

### After:

Screenshot from 2020-11-05 21-32-20

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -837,7 +837,7 @@ fn add_playground_pre(
                                 let (attrs, code) = partition_source(code);
 
                                 format!(
-                                    "\n# #![allow(unused)]\n{}#fn main() {{\n{}#}}",
+                                    "# #![allow(unused)]\n{}#fn main() {{\n{}#}}",
                                     attrs, code
                                 )
                                 .into()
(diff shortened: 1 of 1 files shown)


ehuss:
Hm, so I'm uncertain about this change. There's a problem that on narrower screens, the buttons will overlap the code (see #1322).  I think this could make the problem worse.  I would kinda like to have a solution for the overlap problem first before deciding what to do here.  

I'm uncertain if it makes sense to have a blank space on every example?  Some other docs sites do that (like [docusaurus](https://docusaurus.io/docs/en/commands)).  

Another option is to hide the buttons with a mouseover (like [bookdown](https://bookdown.org/yihui/bookdown/r-code.html)), but that doesn't work on mobile, and may not be discoverable and too magical.

Another option would be to hang the icons off the right side, but I don't know what that would look like, or if there would be enough space.

I don't really know what would be a good approach.

dtolnay:
Good call.

I since realized I do not want buttons anyway on my site (`​```rust,noplayground` &amp; `playground = { copyable = false }`) and without buttons the blank line does not appear, so I no longer need this -- feel free to close.

pickfire:
Is it possible to use a css based approach to do some padding on small screens?

Originally, I opened an issue in https://github.com/rust-unofficial/patterns/issues/155. I was searching for issues in this repo and came across this. I believe this would not look so good on desktop or even tablets, maybe very small screens with very long first line would be affected by this, I don't think I have seen any in the wild. Even if I did, it is less annoying than the extra space which leads to me searching issues in this repo.

Dylan-DPC:
closing this, we can discuss other approaches on the issue or in a new pr

dtolnay:
I filed #1433 to track in mdBook.

PR #1378, opened 2020-11-10
allow to serialize the "build" section

daynin:
Without this change it's impossible to setup build parameters through API like that:

```rust
use mdbook::MDBook;
use mdbook::config::Config;

let root_dir = "/path/to/book/root";

let mut cfg = Config::default();
cfg.book.title = Some("My Book".to_string());
cfg.book.authors.push("Michael-F-Bryan".to_string());

cfg.build.build_dir = PathBuf::from("output"); <---------------- we can set it, but it won't be serialized

MDBook::init(root_dir)
    .create_gitignore(true)
    .with_config(cfg)
    .build()
    .expect("Book generation failed");
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -352,6 +352,11 @@ impl Serialize for Config {
         let book_config = Value::try_from(&self.book).expect("should always be serializable");
         table.insert("book", book_config);
 
+        if self.build != BuildConfig::default() {
+            let build_config = Value::try_from(&self.build).expect("should always be serializable");
+            table.insert("build", build_config);
+        }
+
         if self.rust != RustConfig::default() {
             let rust_config = Value::try_from(&self.rust).expect("should always be serializable");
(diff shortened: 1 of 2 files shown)


ehuss:
Thanks!  A problem with this change is that `mdbook init` will start the `book.toml` file with a `[build]` section.  Can it maybe use the same check for `default` as the `[rust]` table uses?

daynin:
@ehuss yes, of course. I did it

ehuss:
Sorry, forgot to ask, can you add a test for this?  It looks like `run_mdbook_init_with_custom_book_and_src_locations` could just be updated to have an `assert_eq!` of the contents of `book.toml`, similar to how `base_mdbook_init_should_create_default_content` works.

daynin:
@ehuss done

ehuss (approved):
Thanks!

PR #1387, opened 2020-11-18
Update dependencies

tavianator:
I primarily wanted the pulldown-cmark update to be compatible with the latest pulldown-cmark-to-cmark, but I updated everything while I was at it.


Dylan-DPC:
Thanks for the PR. Looks like this bumps the msvr of the crate (just a note not a blocker)

PR #1389, opened 2020-11-25
Add chapter name to search result breadcrumbs

avitex:
Currently the search result breadcrumbs only contain the chapter's parents, and not the chapter name itself.
This PR adds the chapter name to the breadcrumbs for improved searchability.


ehuss (approved):
Thanks for the PR!  I agree the old breadcrumbs were a little strange, I'm uncertain why they work they way they do.

One minor oddity with this is that if it matches the top-level heading, then it will be duplicated.  For example, searching for "highlight" in mdbook's guide has a breadcrumb of "Format » Theme » Syntax highlighting » Syntax Highlighting".  I don't think this is too much of a problem, so we can probably tweak it later if it seems too annoying.

PR #1390, opened 2020-12-01
Guide: Add instructions for publishing via GitLab Pages

### LINKED ISSUE #813: GitLab guide for continuous integration

sburris0:
Closes #813

### DIFF
--- a/guide/src/continuous-integration.md
+++ b/guide/src/continuous-integration.md
@@ -133,12 +133,12 @@ stages:
 
 pages:
   stage: deploy
-  image: rust:latest
+  image: rust:alpine
   variables:
     CARGO_HOME: $CI_PROJECT_DIR/cargo
   before_script:
     - export PATH="$PATH:$CARGO_HOME/bin"
-    - mdbook --version || cargo install --debug mdbook
+    - mdbook --version || cargo install mdbook
   script:
         - mdbook build -d public
   only:
--- a/guide/src/continuous-integration.md
+++ b/guide/src/continuous-integration.md
@@ -124,3 +124,31 @@ deploy: book
 		git commit -m "deployed on $(shell date) by ${USER}" && \
 		git push origin gh-pages
 ```
+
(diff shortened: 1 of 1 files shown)


guide/src/continuous-integration.md
```diff
@@ -124,3 +124,31 @@ deploy: book
...
+    - deploy
+
+pages:
+  stage: deploy
+  image: rust:latest
+  variables:
+    CARGO_HOME: $CI_PROJECT_DIR/cargo
+  before_script:
+    - export PATH="$PATH:$CARGO_HOME/bin"
+    - mdbook --version || cargo install --debug mdbook
```
ehuss, line 141: Any particular reason to use `--debug` here?  On large books, it can be very slow (though I guess on CI it's not too important).  It's also quite a bit larger on disk.

FWIW, I usually suggest that CI installs the precompiled binaries (such as [here](https://github.com/rust-lang/rustup/blob/b7e79deea18b78497f91d218ef3051126f7c77a5/.github/workflows/deploy-docs.yaml#L16-L18)), as it can be substantially faster and avoids dealing with caching issues. I realize this is following the style of the other examples, just putting it out there.

sburris0: No, I've removed it.
I tried using an rust:alpine image and just using the package manager but got errors while trying to install.

ehuss (approved):
I don't know anything about GitLab CI, but I'll go ahead and merge.  I'm not sure how helpful these docs are since they tend to drift out of date, and people tend to set things up differently based on their needs.

Issue #1105, opened 2019-11-30
Pre-rendering math 


Shreyas4991:
As things stand now, rendering of mathematics is very slow and jarring. I believe one solution for this is to render math as MathML or SVG during the build process rather than handling it on the client side. Both Mathjax and KaTeX offer this ability, albeit in Javascript (Node). I assume there are rust equivalents for the same. I suggest that mathematics be rendered when running the build command.

Issue #651, opened 2018-03-27
Support Katex


Timmmm:
It would be good to support Katex as an alternative to MathJax. It's much faster, especially on mobile.

Comparison of speed and rendering differences [here](https://www.intmath.com/cg5/katex-mathjax-comparison.php).

clarcharr:
Another benefit of KaTeX is that it also allows pre-rendering, which can further improve speeds.

brendanzab:
This would be super cool. My [mathjax page](https://brendanzab.github.io/pikelet/appendix/theory.html) is getting super heavy with notation and takes forever to reflow! Pre-rendering would be much appreciated!

Michael-F-Bryan:
It sounds like using Katex could make the user experience nicer for people with loads of equations, so I'm all for the idea. Adding Katex support would probably require somebody to champion the feature seeing as it's touching a non-trivial amount of the code base. 

The code itself shouldn't be too much, but it requires tweaking the renderer and/or creating a pre-processor to update the source document with appropriate annotations.

I'm also happy to help mentor and or point people in the right direction :grin:

brendanzab:
How ight Katex be included/distributed with mdbook?

Michael-F-Bryan:
I'm not 100% sure. If it's a JS thing and used purely in the browser (i.e. after a book is rendered as HTML) then it should, in theory, be no different from JQuery or highlight.js.

I know @sorin-davidoi and @mattico have done some work with our JS/CSS stuff in the past, so they may be able to elaborate more.

brendanzab:
It's run as a preprocessing step, so would probably require the user had node installed.

dkotrada:
> How might Katex be included/distributed with the mdbook binary? Seeing as it is a JS thingy?

Is a node module https://katex.org/docs/node.html

jens1o:
is there anything I can work on to make this a reality?

leesei:
> > How might Katex be included/distributed with the mdbook binary? Seeing as it is a JS thingy?
> 
> Is a node module [katex.org/docs/node.html](https://katex.org/docs/node.html)

Well, it is indeed a JS thingy. Just like MathJax it is included via CDN.  
[Browser · KaTeX](https://katex.org/docs/browser.html)

shilangyu:
I am willing to help with this, can I get some pointers?

Recmo:
Unfortunately there is no easy way to inject HTML into the default template, but this worked for me:

[...]

PR #1392, opened 2020-12-03
Add Katex preprocessor to 3rd Party Plugins

lzanini:
I developed a preprocessor (see [here](https://github.com/lzanini/mdbook-katex)) pre-rendering LaTex equations to HTML, which I think would be helpful for books with lots of equations. See #651 and #1105.


ehuss (approved):
Thanks!

PR #1405, opened 2020-12-19
fix: readerer get theme dir path bug

### LINKED ISSUE #1404: [Renderer] - When the `theme` parameter is configured, `hbs_renderer` cannot get theme dir.
When the `theme` parameter is configured, `hbs_renderer` cannot get theme dir.
I found that there is no join `root_dir` when getting the `theme_dir`.I will fix it latter.

image

francis-du:
Fix #1404

Now, it works on my build in local.

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -452,7 +452,7 @@ impl Renderer for HtmlHandlebars {
         let mut handlebars = Handlebars::new();
 
         let theme_dir = match html_config.theme {
-            Some(ref theme) => theme.to_path_buf(),
+            Some(ref theme) => ctx.root.join(theme),
             None => ctx.root.join("theme"),
         };
 


ehuss:
Thanks for the PR!  Can you explain more about what the problem is?  This seems to change the meaning of the `theme` configuration variable, which I don't think we can change.  It is documented as a directory relative to the root of the book.

francis-du:
Assuming the root path is `/home/book`.

```rust
let theme_dir = match html_config.theme {
    Some(ref theme) => theme.to_path_buf(),
    None => ctx.root.join("theme"),
};

let theme = theme::Theme::new(theme_dir);
```

```rust
pub fn new<P: AsRef<Path>>(theme_dir: P) -> Self {
    let theme_dir = theme_dir.as_ref();
    let mut theme = Theme::default();

    println!("final theme dir => {}",&theme_dir.display());

    // If the theme directory doesn't exist there's no point continuing...
    if !theme_dir.exists() || !theme_dir.is_dir() {
        return theme; 
    }

  ...
}
```

If the theme directory is configured.
```
[output.html]
theme="theme"
default-theme = "ayu"
```
Beacuse of `theme_dir` does't join root path, the directory cannot be found here.
and can't load any files under `/home/book/src/theme`.

So I think  here need to join root path,and the theme must be under `src/theme`.
In this way, if I have a custom theme called `grogu`, I can directly configure 
theme name.eg: `theme="Grogu"`. So the theme can also be introduced as a submodule from git.

ehuss:
If you have a custom theme, it should be in a directory next to the `src` directory.  For example, `theme = "my-theme"`, in a directory `/home/book`, then the theme would be in the `/home/book/my-theme/` directory.  If you want to use a subdirectory, then the configuration would be `theme = "theme/my-theme"`.

francis-du:
> If you have a custom theme, it should be in a directory next to the `src` directory. For example, `theme = "my-theme"`, in a directory `/home/book`, then the theme would be in the `/home/book/my-theme/` directory. If you want to use a subdirectory, then the configuration would be `theme = "theme/my-theme"`.

If so, `theme_dir` is a relative path, it can't be found without join `root_dir`.You can try it.

ehuss:
Can you explain in more detail what is not working for you?  The following commands should create a book with a custom theme:

```bash
mdbook init foo
cd foo
cat << EOF >> book.toml
[output.html]
theme = "my-theme"
EOF
mkdir my-theme
echo "<html><body>{{{ content }}}</body></html>" > my-theme/index.hbs
mdbook build --open
```

francis-du:
When I configure the `theme` , then my custom style is not loaded .

ehuss:
Are you maybe running `mdbook` from a different directory and passing in the book directory?

In that case, it does seem to not work correctly.  I think it would be reasonable to fix that, but this PR would need to change to remove the `.join("theme")`.  I would also remove the call to `create_dir_all`, as I think it would be preferable to error out if it is misconfigured.  Finally, the call to `to_path_buf` doesn't seem to be necessary.

ehuss (approved):
Thanks, looks like there were some issues with a git merge, so I squashed the commits. Generally using a separate branch, and rebasing will avoid those kinds of issues.

francis-du:
@ehuss Thanks .

PR #1406, opened 2020-12-19
[Theme] - Add a new theme,name is Grogu.

francis-du:
This is from a custom theme of mine. [https://github.com/francis-du/mdbook-grogu-theme](https://github.com/francis-du/mdbook-grogu-theme)

The name `Grogu` comes from [`The Mandalorian`](https://disneyplusoriginals.disney.com/show/the-mandalorian), it is a `Baby Grogu` style theme.

I think it’s a very nice theme, I want to add it.

This is a preview image and here is a [live demo](https://grogu-theme.francis.run/)

image

### DIFF
--- a/src/theme/css/variables.css
+++ b/src/theme/css/variables.css
@@ -90,6 +90,46 @@
     --search-mark-bg: #355c7d;
 }
 
+.grogu {
+    --bg: #c6ebc9;
+    --fg: #4b5452;
+
+    --sidebar-bg: #70af85;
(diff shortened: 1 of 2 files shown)


ehuss:
Thanks for the PR.  Unfortunately at this time I don't think we will be adding any more themes, due to the maintenance cost. Of course you are welcome to use custom themes when publishing a book, and making that easier would be a nice addition.

PR #1408, opened 2020-12-21
feat(navigation): add support for vi bindings (h = left and l = right)

sassman:
This PR replaces #608 because it seems to be that the original author is gone.

Also, I propose to only handle the keys for rightArrow (l) and leftArrow (h) here, for the key bindings to navigate back and forth for whole sections in the sidebar I am planning to bring up a new PR.

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -539,6 +539,7 @@ function playground_text(playground) {
 
         switch (e.key) {
             case 'ArrowRight':
+            case 'l':
                 e.preventDefault();
                 var nextButton = document.querySelector('.nav-chapters.next');
                 if (nextButton) {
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR!  I'm a little reluctant to merge something like this because the existing keybindings have caused various issues with extensions and input fields (see #1020, #1244).  I'm uncertain, but maybe the keybindings should inspect the target similar to how the search bindings work?  Unfortunately, I don't fully understand the issues involved, so I'm not sure if that makes sense.

sassman:
@ehuss thanks for the feedback. I will have a closer look at those mentioned issues and see if your suggestion is applicable in that context.

sassman:
PS: just keep in mind that on almost every site even here on GitHub there are key bindings available. 
`?` for a help overlay, `/` for the search, `j` and `k` for navigation purpose. So I'd say that is some sort of best practice.

GuillaumeGomez:
I also kinda agree with @ehuss here. But just in case, let's nominate this for team discussion.

cc @rust-lang/mdbook

GuillaumeGomez:
Closing for the same reasons as https://github.com/rust-lang/mdBook/issues/3200. Thanks for your work in any case!

PR #1410, opened 2020-12-22
[guide/format/theme/index-hbs] `chapter_title` and `book_title` are inverted

maximeborges:
Looking into the code, we can confirm that the implementation is `{{ chapter_title }} - {{ book_title }}` and not `{{ book_title }} - {{ chapter_title }}` as written in the guide:
https://github.com/rust-lang/mdBook/blob/4c951d530d7cf225fcca18d6e9146b28bf8dcf10/src/renderer/html_handlebars/hbs_renderer.rs#L69


ehuss (approved):
Thanks!

PR #1418, opened 2020-12-31
Fix relative paths for renderer commands.

ehuss:
Custom renderers that explicitly set the `command` to a relative path was using the wrong base path.  On Windows, it was relative to the current directory, and on Unix it was relative to the renderer output directory (see https://github.com/rust-lang/rust/issues/37868 for this unfortunate behavior). Neither are correct, they should be relative to the book root.  This changes it so that it is relative to the book root, and has a legacy warning for the old behavior of relative to the output directory.  This uses absolute paths to avoid problems with `current_dir()`.

### DIFF
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -20,7 +20,7 @@ mod markdown_renderer;
 use shlex::Shlex;
 use std::fs;
 use std::io::{self, ErrorKind, Read};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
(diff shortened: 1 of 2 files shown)


ehuss:
Note to self:  The preprocessor custom command is relative to the current directory, but it should be relative to the book root.  Unfortunately the `Preprocessor::supports_renderer` API does not pass in a context to get the book root, so I don't think it is fixable without a breaking API change.  Will file an issue later about this.

PR #1425, opened 2021-01-04
Add ability to generate shell completions via clap

alexmaco:
Usual usage is something like 
```bash
mdbook completions bash > /path/to/completion/file/mdbook
```
where the shell is configured to load completions from that location.

Another possibility is to have the `completions` subcommand hidden by default, since it won't be as commonly used as the others/

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,8 +3,9 @@ extern crate clap;
 #[macro_use]
 extern crate log;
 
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Looks good, thanks!  I think it is fine to show the subcommand in the list of commands.

Issue #850, opened 2018-12-12
How do you keep table of contents open?


senalkularatne:
Is there a way to keep the table of contents (side menu) open by default? I think I would have to make changes to the `book.js` file or is there a way to do this through `book.toml` file?

dacut:
Related, from https://github.com/rust-lang/rust-by-example/issues/1294

> When skimming through later chapters (anything below 6.2 on my screen), I have to scroll the table of contents down to get to the items. Once I click on the item of interest, however, the table of contents resets to the top, making skimming through a chapter tedious.
>
> If I'm skimming through chapter 14 (Generics), for example, I have to scroll down, select a page, see it's not quite what I wanted, scroll back down, select another page, etc. If I were skimming sequentially, I could just use the left/right arrows; if I knew the exact text I'm looking for, I could search; but sometimes, there's something I'm looking for "that I remember seeing, but don't recall the exact words for, but remember seeing it in the Rust by Example book and it looked like X".
>
> Is there an option to the HTML renderer you could set to enable this so it doesn't refresh the TOC sidebar each time it jumps to a page?

PR #1452, opened 2021-01-23
Lower the width below which a screen is considered small

x3ro:
I've noticed that the point at which mdBook starts considering a screen as small (i.e. force-hides the sidebar after a reload) feel unnecessarily small. In my opinion, this makes mdBooks much more difficult to navigate on small screens or in small browser windows, given that after every click on a link the sidebar is hidden and has to be re-opened.

The original 1080px width seems to stem from 71689da6b151c922f0bb69c54f0bc6545d585652 which states that this is because of the sum of side bar width plus content width. This suggests the underlying assumption that the content becomes less readable once it shrinks below the default width.

In my opinion, this is not the case: at the proposed breakpoint of 900px, the reader still sees around 70 characters per line at default zoom level. According to [this thread](https://ux.stackexchange.com/a/108803/30751), that is still a good line width for reading on a screen.

This may also be related to #850, though I'm not entirely sure.

### DIFF
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -175,7 +175,7 @@ a > .hljs {
     right: var(--page-padding);
 }
 
-@media only screen and (max-width: 1080px) {
+@media only screen and (max-width: 900px) {
     .nav-wide-wrapper { display: none; }
     .nav-wrapper { display: block; }
 }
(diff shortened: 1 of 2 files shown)


ehuss:
Thanks for the PR, but I am going to close this as I don't think I want to accept this at this time.  Using a third of the screen for the sidebar just seems like too much. Many code examples are written expecting at least 80 columns to display, but in my testing this seems to cut off around 70, clipping the contents and requiring some awkward scrolling.

I do appreciate that the sidebar can be a little cumbersome and awkward when it is hidden, but I don't have any specific suggestions on how to improve that unfortunately.

PR #1466, opened 2021-02-17
Add options to resolve markdown symlinks

Evian-Zhang:
# Background

Since it is recommended that we put all markdown source files in a single directory `src`, problem occurs when I want to make a table of contents in the top-level `README.md` and `src/README.md` at same time.

My `src/` dir contains `README.md`, `a.md`, `b.md`, `SUMMARY.md`, and to avoid duplicating, the `README.md` is soft linked to top-level `README.md`. If I write

```markdown
My table of contents:

- [a](src/a.md)
- [b](src/b.md)
```

then the links in `src/README.md` is invalid; If I write

```markdown
My table of contents:

- [a](./a.md)
- [b](./b.md)
```

then the links in top-level `README.md` is invalid.

I know changing directory layout to be flat can handle this problem, but it is more acceptable if I organize all docs in a single directory. So I can

```shell
ln -s src src/src
```

then use `src/a.md` for markdown links. For top-level `README.md`, it links to real markdown files in `README.md`, and for `src/README.md`, it links to `a.md` in soft-linked `src` dir.

However, mdbook does not support soft-linked markdown files.

# Simplified problem

In `src/`, we have `SUMMARY.md`, `a.md`, `b.md`, `c.md`, where `c.md` is soft linked to `b.md`. In `SUMMARY.md`, we include `a.md` and `b.md`, so the generated html should be `a.html` and `b.html`. In `a.md`, we have a link to `c.md`. But in generated `a.html`, the link breaks.

# Solution

Before rendering files, we calculate a map, whose key is the canonicalized absolute path of markdown file, and value is the absolute path of generated html file.

When adjusting links, we get the path of generated html file of current markdown file, and of the link's target markdown file, from the pre-calculated map. Then we get the diffs of the two absolute path to get a relative path, then use this relative html path to replace the link's target markdown path.

However, this may lead a lot cost when calculating canonicalized absolute path, and the original link-adjust method (simply replace `.md` to `.html`) can work for most situations unless symlinks are involved. So I introduce an option `resolve-md-symlink` in the HTML config, and disable it by default.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -739,6 +739,7 @@ dependencies = [
  "memchr",
  "notify",
  "open",
+ "pathdiff",
  "pretty_assertions",
  "pulldown-cmark",
  "regex",
@@ -896,6 +897,12 @@ checksum = "53cdc5b785b7a58c5aad8216b3dfa114df64b0b06ae6e1501cef91df2fbdf8f9"
 dependencies = [
  "winapi 0.3.8",
 ]
+
+[[package]]
+name = "pathdiff"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "877630b3de15c0b64cc52f659345724fbf6bdad9bd9566699fc53688f3c34a34"
 
 [[package]]
 name = "percent-encoding"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ serde_json = "1.0"
 shlex = "0.1"
 tempfile = "3.0"
 toml = "0.5.1"
+pathdiff = "0.2.0"
 
 # Watch feature
 notify = { version = "4.0", optional = true }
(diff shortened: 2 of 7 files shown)


Yansongsongsong:
@ehuss I am sorry to bother you, but would you mind reviewing this issue? Thx a lot

ehuss:
Thanks for the PR, though I don't fully understand the motivation here.  Why do you want two copies of a README?  Is the README duplicating the index in SUMMARY?  You don't have to put everything in a `src` directory if you don't want to.

Evian-Zhang:
> Why do you want two copies of a README?  Is the README duplicating the index in SUMMARY?

@ehuss Thanks for reply.

No, README's content may differ from SUMMARY. We are writing an [introduction to linux x86-64 syscall](https://github.com/Evian-Zhang/introduction-to-linux-x86_64-syscall), and each article may contain multiple syscalls that have similar functionality. So SUMMARY contains links to the articles, and README provides links to the syscalls.

Evian-Zhang:
I come up with a new work around: move `SUMMARY.md` to root dir, and mark the root dir as src dir in `book.toml`, and in `SUMMARY.md`, directly links to files inside `src` dir. This could perfectly solve the problem.

Issue #1384, opened 2020-11-16
Support space as seperator between language and additional class in code block


tuyen-at-work:
### Use case

> Let say I want to add `h500` as a class for a code block.

### Editor

In VS Code, we can use space between language and additional class without loose syntax highlighting:

![image](https://user-images.githubusercontent.com/68000455/99209304-72b70580-27f5-11eb-8c76-6bc4d9b02cca.png)

But if I use comma instead of space between language and additional class, the highlighting doesn't work:

![image](https://user-images.githubusercontent.com/68000455/99209213-38e5ff00-27f5-11eb-97f2-9a951cd04314.png)

### Render

Let check what happen when two above code blocks got rendered:

The first case loose syntax highlighting:

![image](https://user-images.githubusercontent.com/68000455/99209622-24563680-27f6-11eb-8f3e-fd8ffb95abab.png)

The second case work nicely:

![image](https://user-images.githubusercontent.com/68000455/99209715-5ff10080-27f6-11eb-90f1-343edb4b702a.png)

### Current status

In current situation, if I want to use additional class, I have to choose:

- Either have syntax highlighting in VS Code, and loose it when rendered, or
- Don't have syntax highlighting in VS Code, and get it when rendered

### Feature request

Please support space as seperator between language and additional class in code block, so we can have syntax highlighting for both editor and rendered version.

ehuss:
This should be fixed by #1469.

PR #1469, opened 2021-02-21
Support space as seperator between language and additional class in c…

tuyen-at-work:
…ode block

This PR try to resolve this issue: https://github.com/rust-lang/mdBook/issues/1384

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -231,7 +231,7 @@ fn clean_codeblock_headers(event: Event<'_>) -> Event<'_> {
             let info: String = info
                 .chars()
                 .map(|x| match x {
-                    ' ' => ',',
+                    ' ' | '\t' => ',',
                     _ => x,
                 })
                 .filter(|ch| !ch.is_whitespace())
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks!  It looks like rustdoc will be making a similar change (https://github.com/rust-lang/rust/pull/78429).  Maybe this should match its behavior and also support the tab character?  (I don't think it should bother with the curly brace issue.)
Unfortunately, per https://github.com/rust-lang/rust/pull/78429#issuecomment-718340447, it doesn't seem like there is much agreement on how this should work, but spaces and tabs seem reasonable.  I'd also like to wait until the rustdoc changes land.

tuyen-at-work:
@ehuss as your comment, GitHub supports `,`, `space` and `\t`. VS Code supports `space` and `\t` but not `,`. They are the biggest guys in the market. So I think it is most flexible to support all 3 chars, like GitHub does. I think as long as GitHub supports it, other implementation will follow suits.

jyn514:
> Thanks! It looks like rustdoc will be making a similar change (rust-lang/rust#78429). Maybe this should match its behavior and also support the tab character?

Just to clarify - rustdoc already splits on spaces (in fact, it splits on almost anything that's not a letter). https://github.com/rust-lang/rust/pull/78429 is removing delimiters, not adding more.

ehuss (approved):
I went ahead and pushed a change to add tab to be consistent with rustdoc and GitHub.

PR #1470, opened 2021-02-21
Enhance text contrast of `light` theme to improve accessibility.

### LINKED ISSUE #1442: Stylesheet contrast issues
Some of the styling is quite low-contrast, which makes it harder to read and navigate than it could be:

* The `<h1>` of the currently-viewed page is light-on-light,
* I didn't notice the `<` and `>` previous/next page controls for a couple of minutes because they render light-on-light,
* Inline code (where syntax highlighting is not applied) is dark-on-dark.

I'm no designer so I hesitate to suggest specific fixes, but https://contrastchecker.com/ is good at least to check body text.

(Re-reporting scrive/nix-workshop#1 upstream.)

tim-seoss:
This PR aims to address #1442, and is a refinement of PR #1353 based on feedback there, and also on the WebAIM mailing list, in the replies to this post:

https://webaim.org/discussion/mail_message?id=45812

Principle differences from earlier revisions posted in #1353 are:
- The new colour scheme is now applied directly to the default `light` theme (instead of a separate and new `contrast` theme).
- The scroll bars are now slightly lower contrast (to make them less visually distracting/imposing, while still complying with WCAG v2.1).
- A subtle change to the syntax highlight text colours, to perform better on devices which have their "blue light filter" activated.
- The size of the title text (rendered at the top-centre of each page) is now 20% larger (this *doesn't* make the title bar steal area from other page elements, and seems unlikely to get the title ellipsised except when viewing on phone screens)

Of the changes, I'm least certain about the last one (title text size).  If people feel that it's now too big, an alternative would be to restore the size of the title text, and further increase its contrast (by introducing a new theme colour for it) instead.

Commit text follows:

The existing light theme has relatively low contrast between the text
(and other UI elements) and background (especially within code blocks).
This presents difficulties for people with reduced visual contrast
perception (common in older adults).

This patch makes changes to the default `light` theme to meet the
minimum contrast requirement of the v2.1 W3C WCAG (Web Content
Accessibility Guidelines)
https://www.w3.org/WAI/WCAG21/quickref/#contrast-minimum

The small size, and slender font used for the title text makes it hard
to read, even with the increased contrast colour scheme, so this patch
also increases the size of the title text font by 20%.

Closes #1442

### DIFF
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -93,7 +93,7 @@ a > .hljs {
 .menu-title {
     display: inline-block;
     font-weight: 200;
-    font-size: 2rem;
+    font-size: 2.4rem;
     line-height: var(--menu-bar-height);
     text-align: center;
     margin: 0;
--- a/src/theme/css/variables.css
+++ b/src/theme/css/variables.css
@@ -92,22 +92,22 @@
 
 .light {
     --bg: hsl(0, 0%, 100%);
-    --fg: #333333;
+    --fg: hsl(0, 0%, 0%);
 
     --sidebar-bg: #fafafa;
-    --sidebar-fg: #364149;
+    --sidebar-fg: hsl(0, 0%, 0%);
     --sidebar-non-existant: #aaaaaa;
-    --sidebar-active: #008cff;
+    --sidebar-active: #1f1fff;
     --sidebar-spacer: #f4f4f4;
 
-    --scrollbar: #cccccc;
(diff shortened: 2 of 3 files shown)


ehuss (approved):
Thanks for working on this, I appreciate the attention to detail here!

The only feedback that I might have is that the scroll bars might be a little distracting since they are now so visible.  However, I'm fine with trying this out and seeing if there is any other feedback.

PR #1471, opened 2021-02-21
Upgrade to shlex 1

fenhl:
This upgrades the [shlex](https://docs.rs/shlex) dependency from version 0.1 to 1. The breaking change in shlex 1 is a fix for [a correctness bug](https://github.com/comex/rust-shlex/issues/6). I haven't looked closely at how shlex is used in mdbook so it should probably be made sure that this doesn't break anything that relied on the incorrect behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1258,9 +1258,9 @@ dependencies = [
 
 [[package]]
 name = "shlex"
-version = "0.1.1"
+version = "1.0.0"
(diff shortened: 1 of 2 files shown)


Dylan-DPC:
It is also being used [here](https://github.com/rust-lang/mdBook/blob/4c951d530d7cf225fcca18d6e9146b28bf8dcf10/src/preprocess/cmd.rs#L74) so should be safe. 

Thanks a lot

PR #1477, opened 2021-02-28
Add background to `pre > .buttons`

camelid:
Before, the buttons would sometimes overlap with the text making it hard
to see the button icons. Adding a background prevents that, although it
does cover some of the code.

This is **draft** because the background covers some of the code.

### DIFF
--- a/src/theme/ayu-highlight.css
+++ b/src/theme/ayu-highlight.css
@@ -6,7 +6,7 @@ Original by Dempfi (https://github.com/dempfi/ayu)
 .hljs {
   display: block;
   overflow-x: auto;
-  background: #191f26;
+  background: var(--code-block-bg);
   color: #e6e1cf;
   padding: 0.5em;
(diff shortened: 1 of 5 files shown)


camelid:
As the images show, the buttons overlapped the text before—which makes the text hard to read but still possible to read—but with the background, some of the text is hidden. It seems like the best bet might be to somehow wrap the text so the buttons weren't on top of the text? I'm not sure how to do that though.

### Before

image

### After

image

camelid:
f26ce74 at least seems like a useful change, so should I split that into a separate PR?

Dylan-DPC:
@camelid 
>  at least seems like a useful change, so should I split that into a separate PR?

preferably yes. Makes it easier to change later. Thanks

camelid:
> preferably yes. Makes it easier to change later. Thanks

Opened #1484.

PR #1480, opened 2021-03-05
Improving documentation on implementation of preprocessors.

### LINKED ISSUE #975: Preprocessor in Python?
I'm trying to figure out how to write a preprocessor in Python, given that it's listed in the [example](https://rust-lang-nursery.github.io/mdBook/for_developers/preprocessors.html), but I can't even get started.

1) A simple script that read stdin and write to stdout without any changes

```
import sys

sys.stdout.writelines(sys.stdin)
sys.stdout.flush()
```

Results in

```
2019-07-08 10:35:25 [ERROR] (mdbook::utils): Error: Unable to parse the preprocessed book
2019-07-08 10:35:25 [ERROR] (mdbook::utils): 	Caused By: invalid type: map, expected a sequence at line 1 column 1
```

Clearly I'm missing something obvious.

2) Path to the script. I can call a script living in the same path as book.toml, but relative paths don't seem to work?

### LINKED ISSUE #1462: Use of non-rust preprocessors isn't clearly documented in the guide book
After reading the [chapter on preprocessors](https://github.com/rust-lang/mdBook/blob/4c951d530d7cf225fcca18d6e9146b28bf8dcf10/guide/src/for_developers/preprocessors.md) and learning that they were meant to operate in "typical unix fashion", I assumed I could just drop any stream transformer as a preprocessors. Come to find out, the executable must also respond to the subcommand `supports <foo>`, where `<foo>` is some format. 

My initial reaction is that this is not typical unix fashion, and that it requires an unnecessarily tight coupling between the preprocessor and mdBook. I wonder if it would be enough just to leave the designation of support up to the `renderer` key, rather than double checking with the tool whether it really can support what we've configured it to support :)

However, at minimum, adding this requirement into the docs should help future users.

Thanks for a very nice tool!

pauliyobo:
This is the starting point at an attempt to make the documentation more clearer to those who would like to implement a preprocessor but are stuck on how to.
I am sure this change is far from perfect, so I'd love to have advice on how it could be structured better.
Additionally, I was wondering whether the code shown should be moved in a file of it's own.
This should fix #1462 and resolve #975 
Thanks.

### DIFF
--- a/guide/src/for_developers/preprocessors.md
+++ b/guide/src/for_developers/preprocessors.md
@@ -34,8 +34,8 @@ command = "python3 /path/to/foo.py"
 renderer = ["html", "epub"]
 ```
 
-In typical unix style, all inputs to the plugin will be written to `stdin` as
-JSON and `mdbook` will read from `stdout` if it is expecting output.
+Once the preprocessor has been defined and the build process has started, MdBook will execute the command defined in the `preprocessor.foo.command` key passing the arguments support and the renderer name, monitoring the status code of the executed command.
+If the status code retrieved is 0, the library will be sending through stdin both the context and the book representation serialized in JSON format, and it'll be capturing the response from stdout, which will be the modified book which has to also be serialized in json format.
 
 The easiest way to get started is by creating your own implementation of the
 `Preprocessor` trait (e.g. in `lib.rs`) and then creating a shell binary which
@@ -80,7 +80,7 @@ without accidentally breaking the document.
 ```rust
 fn remove_emphasis(
     num_removed_items: &mut usize,
-    chapter: &mut Chapter,
+	    chapter: &mut Chapter,
 ) -> Result<String> {
     let mut buf = String::with_capacity(chapter.content.len());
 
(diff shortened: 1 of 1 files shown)


guide/src/for_developers/preprocessors.md
```diff
@@ -34,8 +34,8 @@ command = "python3 /path/to/foo.py"
 renderer = ["html", "epub"]
 ```
 
-In typical unix style, all inputs to the plugin will be written to `stdin` as
-JSON and `mdbook` will read from `stdout` if it is expecting output.
+Once the preprocessor has been defined and the build process has started, MdBook will execute the command defined in the `preprocessor.foo.command` key passing the arguments support and the renderer name, monitoring the status code of the executed command.
+If the status code retrieved is 0, the library will be sending through stdin both the context and the book representation serialized in JSON format, and it'll be capturing the response from stdout, which will be the modified book which has to also be serialized in json format.
```
ehuss, line 38: Generally as a style, I prefer active voice in present tense.  It can help to keep sentences shorter if possible.

Also, the name is usually written as `mdBook` with a lowercase `md`. 

I realize the existing documentation is a little inconsistent on these points.

I'd also like to see a little more detail on the exact behavior and data passed in.  (Sorry if this seems like I'm rewriting everything.)

``````suggestion
Once the preprocessor has been defined and the build process starts, mdBook executes the command defined in the `preprocessor.foo.command` key twice.
The first time it runs the preprocessor to determine if it supports the given renderer.
mdBook passes two arguments to the process: the first argument is the string `supports` and the second argument is the renderer name.
The preprocessor should exit with a status code 0 if it supports the given renderer, or return a non-zero exit code if it does not.

If the preprocessor supports the renderer, then mdbook runs it a second time, passing JSON data into stdin.
The JSON consists of an array of `[context, book]` where `context` is the serialized object [`PreprocessorContext`] and `book` is a [`Book`] object containing the content of the book.

The preprocessor should return the JSON format of the [`Book`] object to stdout, with any modifications it wishes to perform.
``````

ehuss:
> Additionally, I was wondering whether the code shown should be moved in a file of it's own.

I think it is fine to keep it inline like it is.

pauliyobo:
Thanks for the corrections.  I think I have a better understanding about how documentation should be written.

pauliyobo:
Hello
@ehuss I was wondering what was the status of this PR. I remember I had committed the changes you had suggested, but unfortunately I had a busy schedule until recently so couldn't check in to see the progress.
Is the current documentation alright with  what you had proposed, or do you think other additions should be made?
Thanks and I am really sorry for the delay.

ehuss:
Thanks for following up!
It looks like there are still some outstanding suggestions listed above.

pauliyobo:
@ehuss 
Alright. I am sorry for missing them, but I hope I've included everything this time :) .

ehuss (approved):
Thanks!  I pushed a few small fixes.

PR #1481, opened 2021-03-06
support google analytics 4 property

houqp:
Google analytics 4 tags has `G-` prefix and needs to be installed
differently, see:

https://support.google.com/analytics/answer/10089681?hl=en
https://support.google.com/analytics/answer/9304153?hl=en


houqp:
Here is an example of site built with GA-4 using this patch: https://roapi.github.io/docs

ehuss:

PR #1484, opened 2021-03-11
Add CSS variable for code block background color

camelid:
Now the code block background color can be used elsewhere without
worrying about the colors getting out of sync.

r? @Dylan-DPC

### DIFF
--- a/src/theme/ayu-highlight.css
+++ b/src/theme/ayu-highlight.css
@@ -6,7 +6,7 @@ Original by Dempfi (https://github.com/dempfi/ayu)
 .hljs {
   display: block;
   overflow-x: auto;
-  background: #191f26;
+  background: var(--code-block-bg);
   color: #e6e1cf;
   padding: 0.5em;
 }
--- a/src/theme/css/variables.css
+++ b/src/theme/css/variables.css
@@ -28,6 +28,7 @@
     --links: #0096cf;
 
     --inline-code-color: #ffb454;
+    --code-block-bg: #191f26;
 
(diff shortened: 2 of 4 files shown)


ehuss:
Thanks for the PR!  I'm a bit confused on the motivation here.  How can the colors get out of sync?  Each theme defines all the colors for highlighting in one file, why would just this one be moved to a variable in another file?

ISSOtm:
So that it can be used for custom styling, I believe. (Akin to [what I am doing here](https://github.com/ISSOtm/gb-asm-tutorial/blob/7fe0580af27a0aa55c1bc9f50174ae62727898dc/css/custom.css#L25).)

camelid:
> Thanks for the PR! I'm a bit confused on the motivation here. How can the colors get out of sync? Each theme defines all the colors for highlighting in one file, why would just this one be moved to a variable in another file?

It was a while ago, so I don't remember exactly, but I think it was because there might be multiple places in the CSS for a theme where you want to use the same code block background color. There's a variable for `inline-code` color, so it's like that, but for the background color of a code block.

The specific motivation was for #1477, but I should probably close that PR because that approach won't work. See https://github.com/rust-lang/mdBook/pull/1477#issuecomment-796386772 and the comments above it for more about why I opened this PR.

This PR is just a change that might make things a bit easier in the future, but it's possible it wouldn't be that useful. So you can close it if you want :)

GuillaumeGomez:
As per your comment @camelid, closing this PR. :)

PR #1485, opened 2021-03-12
Add page-break

Evian-Zhang:
This is a new version of old pr https://github.com/rust-lang/mdBook/pull/1300, since that one is out-dated.

This pr adds a `page-break` option, when printing the generated book to PDF, page breaks will be inserted between chapters, so that a new chapter will start at a new page. This option is default to be `true`. For implement details, please see content of old pr.


ehuss:
Thanks for following up.  There was a question in the previous PR:  Why make the page-break an option? Why not just have this on all the time?

Evian-Zhang:
@ehuss Fixed it.

ehuss (approved):
Thanks!

Issue #1486, opened 2021-03-12
Support for Right to Left


moaz-mokhtar:
EDITED

Very thankful for this tool. 
We need to add support to configure RTL languages (ex: Arabic) to be appeared professionally.

I did test `mdbook` for Arabic text as below:
![image](https://user-images.githubusercontent.com/5870208/111000776-56823a00-838b-11eb-9a07-38881dac8728.png)

---
### First fix
I did edit `html` tag in resulted `chapter_1.html` as below to support `rtl`:
```html
<html dir="rtl" lang="en" class="sidebar-visible no-js light">

```
It is appeared as below:
![image](https://user-images.githubusercontent.com/5870208/111001087-d4464580-838b-11eb-9d15-525e88029b79.png)

---
### Second fix
I did edit the linked stylesheet in `chapter_1.html` as below to right align.
```html
<link rel="stylesheet" type="text/css" href="font-awesome-rtl.css">
```
Appeared as below:
![image](https://user-images.githubusercontent.com/5870208/111002260-c8f41980-838d-11eb-9e49-f357f25a74de.png)

---

Still found issue:
- Sidebar is not Right aligned.

crawfxrd:
> Sidebar is not Right aligned

Seems to be because `src/theme/css/chrome.css` (`build/css/chrome.css`) only sets `padding-left` for `.chapter`.

moaz-mokhtar:
Thank you Tim, but actually I think we need to right alight the whole sidebar.

I tried to change padding-left to padding-right but same issue which only did change alight of the chapter name inside sidebar not the whole sidebar.

When `padding-lift` disabled, result is as below:
![image](https://user-images.githubusercontent.com/5870208/111050996-a24ee500-8458-11eb-9d01-3c26bf3dfd95.png)

When
```css
.chapter {
  list-style: none outside none;
  /* padding-left: 0; */
  padding-right: 0;
  line-height: 2.2em;
}
```
Result as below:
![image](https://user-images.githubusercontent.com/5870208/111051023-d62a0a80-8458-11eb-91c9-7de22c4da445.png)


So I think we need to fix the css properties to confirm RTL working properly.

crawfxrd:
Ah, I see. I thought you just meant the huge padding in the sidebar itself.

So more like this:

```diff
 .sidebar {
     ...
-    left:0;
+    right: 0;
     ...
 }
```

```diff
 @media only screen and (min-width: 620px) {
[...]

PR #1488, opened 2021-03-15
Use logical CSS properties for margin, padding, border

crawfxrd:
Replace phyiscal properties (top/bottom/left/right) with logical properties (start/end) that can be used in non-LTR contexts (e.g.,
content in Arabic or Hebrew).

Based on the [CSS Logical Properties and Values Level 1][css-logical] specification, currently an Editor's Draft.

Referencing MDN, all major browsers except Internet Explorer support the margin, padding, and border properties.

Ref: #1486 

[css-logical]: https://drafts.csswg.org/css-logical/

### DIFF
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -30,9 +30,9 @@ a > .hljs {
     display: flex;
     flex-wrap: wrap;
     background-color: var(--bg);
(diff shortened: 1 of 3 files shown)


crawfxrd:
Replaced by #1489

Issue #1500, opened 2021-04-15
Support wasm-pack POST method in rust-playground


titaneric:
integer32llc/rust-playground#374 mention support wasm rendering in rust-playground.
integer32llc/rust-playground#706 is the corresponding implementation

To support such feature in mdBook, we may
- Call their POST method and fetch result (JS and Wasm)
- Render the JS and Wasm in IFrame
- Differentiate the normal rust and wasm code, perhaps support new `wasm` className

Issue #350, opened 2017-06-23
Support alternative remote playpens for code execution


budziq:
Like:
- https://play.integer32.com/ for rust
- https://rust.godbolt.org/ - compiler explorer is not a playpen pe se but it's immensely useful. (Also for other languages)

Also take into consideration support for other languages:
- https://play.golang.org/
- https://code.sololearn.com/#cpp

See https://fiddles.io/ for extended list

agauniyal:
+1 for godbolt. I would like to present [wandbox](https://github.com/melpon/wandbox/blob/master/kennel2/API.rst) as common denominator here as well.

CjS77:
There's an easy workaround here:
mdbook init --theme

and edit book.js to point to your custom playpen/playground (Rust only, mind you)

vanceism7:
I'm really interested in getting this feature going - Mdbook is great but it'd be nice if we could hook into other playground code editors (like purescript, haskell, etc). I don't mind trying to implement this myself, does anyone have any suggestions as a good starting point?

julianfortune:
@vanceism7 Haskell folks have created a public API for executing code:
- https://play.haskell.org 
- https://github.com/haskell/play-haskell#api

Kleidukos:
Indeed, and we're also looking into integrating the Haskell playground into mdBook. :)

kwest-ng:
I'm quite interested in implementing this, for my own book. I could get started on this now, and I have a decent idea about how to do it, as the codebase is not terribly complex.  But I would prefer to have a reasonable idea that this is a desired change before committing too much time to it.

The rough idea I had was two-fold:

- Provide a clear JS API for interacting with other playgrounds, populated by the config from `book.toml`.
- Create configuration in `book.toml`, as well as defaults for rust that map to the current playground settings. Appropriate config params will be passed to the JS API.

The simplest generalized solution I can see for the JS API is to split the [`run_rust_code`](https://github.com/rust-lang/mdBook/blob/30e0e1d1021d406899c783447994de0a916643c3/crates/mdbook-html/front-end/js/book.js#L105) function into two parts:

- Identifying the language, reading the code, and editing the DOM with the resulting text (whether or not it was successful). This part will likely be the same for all languages, and can be used as a wrapper for calling the second part of the JS API. This code will be part of `mdBook`.
[...]

PR #1501, opened 2021-04-15
Support wasm-pack frontend in mdBook

titaneric:
This PR aim to solve #1500.

## Change of code

- Add new `run_wasm_pack_code` function in `theme/book.js` to fetch `wasm-pack` results and rendering them in Iframe
- Support `wasm` className in `theme/book.js` to differentiate the normal Rust and Wasm code
- Add new `theme/iframe.html` and `theme/wasm-entry.mjs` for Iframe
- Move `iframe.html` and `wasm-entry.mjs` to built directory by Rust

## Note

- This implementation is different than integer32llc/rust-playground#707
- It's possible to avoid new `iframe.html` and `wasm-entry.mjs` if we adapt their method, but the real implementation needs to be discussed.

## Demo

https://user-images.githubusercontent.com/12388235/114810242-3b697680-9dde-11eb-85f3-a6e746aaa0c1.mp4

## P.S.

This implementation is more like prototype than production, feel free to point out where I could do better or rewrite suitibly.

### DIFF
--- a/src/book/init.rs
+++ b/src/book/init.rs
@@ -156,13 +156,13 @@ impl BookBuilder {
 
         let mut highlight_js = File::create(themedir.join("highlight.js"))?;
         highlight_js.write_all(theme::HIGHLIGHT_JS)?;
-        
+
         let mut iframe = File::create(themedir.join("iframe.html"))?;
         iframe.write_all(theme::IFRAME)?;
 
         let mut wasm_entry_js = File::create(themedir.join("wasm-entry.mjs"))?;
         wasm_entry_js.write_all(theme::WASM_ENTRY_MJS)?;
-        
+
         Ok(())
     }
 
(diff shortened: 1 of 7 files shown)


devkadirselcuk (approved):

titaneric:
Hi, I have already served my own rust-playground with `wasm-pack` feature, perhaps it's ready to be reviewed now.

ehuss:
Sorry, I'm maybe a little confused on what this is doing.  Has the corresponding PR been accepted by integer32?  Have they committed to allowing that endpoint to be used by public tools?  I probably won't review until they commit to it.  If you're proposing to use a new service, I unfortunately probably won't accept that.  I think I would prefer making the backend more configurable (#350), so that books can define their own (maybe through some kind of theme files?), but that would need some design work.

titaneric:
Yes, #350 would be the better solution when it comes to plenty of available backends.
I do propose a new service to let user to try `wasm-pack` on the mdBook, and the real backend in `rust-playground` is still under PR and haven't been reviewed yet. Perhaps I could try #350 first and let someone could use my own service.
Thanks for your comment and  I may closed it until #350 is been implemented.

Issue #1433, opened 2021-01-06
Stray spacing of code blocks containing injected main


dtolnay:
For example see the first code block in https://doc.rust-lang.org/1.47.0/book/ch03-02-data-types.html.

````markdown
```rust
let guess: u32 = "42".parse().expect("Not a number!");
```
````

<br>

#### Screenshot:

![Screenshot from 2020-11-05 21-31-53](https://user-images.githubusercontent.com/1940490/98330028-6689a780-1fae-11eb-8e85-c617e0197b92.png)

dtolnay:
#### Desired:

![Screenshot from 2020-11-05 21-32-20](https://user-images.githubusercontent.com/1940490/98330033-6ab5c500-1fae-11eb-8f63-fdb4fe8a3849.png)

<br>

#1374 has one fix, but it isn't perfect because a long first line on a narrow screen can cause the buttons to overlap the code when they didn't before. (There is a separate bug about this: #1322.)

pickfire:
But I see that mdbook examples itself does not have the extra spacing.

![image](https://user-images.githubusercontent.com/4687791/103844159-3eb0e000-50d4-11eb-83fb-82de25a6d35f.png)

pickfire:
Oh, the newline only appears on some block.

![Screen Shot 2021-01-07 at 10 47 32](https://user-images.githubusercontent.com/4687791/103844844-c77c4b80-50d5-11eb-8189-87eda427c01e.png)

What if we put it the buttons closer to the borders on mobile? So even if it did overlap, it overlap half of it.

Or maybe we can put the buttons outside of the box in mobile.

dtolnay:
> But I see that mdbook examples itself does not have the extra spacing.

Correct. That code block does not have an injected `main`. The `main` is handwritten.

https://github.com/rust-lang/mdBook/blob/3d55375f618594d62707eab18b98d32c52b4c9b7/guide/src/format/mdbook.md#L21-L28

ehuss:
I posted #1806 with a proposed fix.

I would appreciate if you could take a look and see if you have any thoughts.  I posted a demo at https://ehuss.github.io/rust-book/.  If you could look through some examples, try different browsers and platforms, different color themes, etc.  If you have access to Android, that would particularly be nice to test since I don't have access to that.

dtolnay:
Android/Firefox:

[...]

PR #1508, opened 2021-04-23
Add spacing between buttons and code

MatheusRich:
## Problem

Sometimes the buttons in code listings are above the code itself, making it hard to read. Here's some examples:

image
image

## Solution

I've added a vertical padding to help separate the code from the buttons:

image
image

I could've added just a padding-top, but I thought it was ugly:

image

If that's not the right file for css changes, let me know!

### DIFF
--- a/src/theme/ayu-highlight.css
+++ b/src/theme/ayu-highlight.css
@@ -8,7 +8,7 @@ Original by Dempfi (https://github.com/dempfi/ayu)
   overflow-x: auto;
   background: #191f26;
   color: #e6e1cf;
-  padding: 0.5em;
+  padding: 2em 0.5em;
 }
 
 .hljs-comment,
--- a/src/theme/ayu-highlight.css
+++ b/src/theme/ayu-highlight.css
@@ -11,12 +11,6 @@ Original by Dempfi (https://github.com/dempfi/ayu)
   padding: 0.5em;
 }
 
-@media (max-width: 640px) {
-  .hljs {
-    padding-top: 2em;
-  }
-}
-
 .hljs-comment,
 .hljs-quote {
   color: #5c6773;
--- a/src/theme/ayu-highlight.css
+++ b/src/theme/ayu-highlight.css
@@ -8,7 +8,13 @@ Original by Dempfi (https://github.com/dempfi/ayu)
   overflow-x: auto;
   background: #191f26;
   color: #e6e1cf;
-  padding: 2em 0.5em;
+  padding: 0.5em;
+}
+
+@media (max-width: 640px) {
+  .hljs {
+    padding-top: 2em;
+  }
 }
 
 .hljs-comment,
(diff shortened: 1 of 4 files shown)


ehuss:
Hi, thanks for the PR!  There is a bit of uncertainty about exactly what the behavior should be here.  There is more discussion in https://github.com/rust-lang/mdBook/pull/1374#issuecomment-725007417 where there is a PR that does the opposite where it collapses the spacing.   Issues #1322 and #1433 also contain some more discussion.

One issue with this PR is that it adds too much spacing for rust code blocks that have an injected main.

image

I'm wondering about maybe making the spacing depend on the screen width?  Like on a narrow screen, it would have the increased spacing unconditionally?  That could still have some overlap issues if the first line is very long, but might be something to try.

MatheusRich:
We could add a media query so the spacing would be applied only on small screens. What do you think?

ehuss:
A media query seems worth trying.  I'm not sure how to detect the "injected main" case, or if that should be changed (I don't remember why they are different).

MatheusRich:
@ehuss  I've added padding-top with media queries. See if that fits what you're thinking of.

serverwentdown:
This change is pretty good with media queries!

serverwentdown (changes requested):
I think `theme` is the wrong place to put the padding tweaks. Maybe place the change [here](https://github.com/rust-lang/mdBook/blob/f5598b2eeed677e518adb32c111b76d7e23045f2/src/theme/css/chrome.css#L205) instead.

serverwentdown:
Now the padding inserted is outside the box, maybe you can try this:
```
pre > .button ~ .hljs {
padding-top: 2em;
}
```

MatheusRich:
@serverwentdown yeah, that seems to work!

serverwentdown:
My bad, `.hljs` would break if the syntax highlighter changes, so we shouldn't depend on that class. `code` should be better:
```
pre > .buttons ~ code {
padding-top: 2em;
}
```

(The above reads as "Any `pre` element containing a direct child `.button` that also contains a subsequent `code` element. Pick that `code` element.")

MatheusRich:
Is there anything we can do to move this forward?

ehuss:
I think I would like to go in a slightly different direction here.  I posted #1806 with a proposed fix.

I would appreciate if you could take a look and see if you have any thoughts.  I posted a demo at https://ehuss.github.io/rust-book/.  If you could look through some examples, try different browsers and platforms, different color themes, etc.  If you have access to Android, that would particularly be nice to test since I don't have access to that.

ehuss:
I'm going to close as I think this should mostly be resolved by #1806. If there are further ideas to improve the layout, feel free to open an issue with some screenshots. I appreciated the help and discussion here, hopefully the primary concern is resolved.

PR #1509, opened 2021-04-24
Remove duplicate "name" attribute on input

GuillaumeGomez:
I discovered this issue while working on https://github.com/rust-lang/rust/pull/84480.


ehuss (approved):
Thanks!  I'm not entirely certain this won't break anything, but it seems to work. The URL (which uses the key `search`) seems to always be handled manually in javascript.  I have no idea what browsers' behavior is with multiple name fields.

GuillaumeGomez:
From what I could read, the last "apparition" of the attribute overwrite the previous one.

PR #1511, opened 2021-04-26
Use semver crate to validate version compatibility in nop-preprocessor example

joshrotenberg:
The comment in `handle_preprocessing` suggests that `semver` could be used to check the compatibility between a book and the current version of mdbook. 

Make dreams come true.


ehuss (approved):
Thanks!

PR #1515, opened 2021-05-02
Allow alternate config file

### LINKED ISSUE #1154: Allow alternatives to `book.toml` on the command-line
For various reasons, I need a different .toml file depending on whether I am building for HTML or PDF (with mdbook-latex). It would be nice if I could specify which TOML file to use during `mdbook build`.

(I realize as I ask this that I can just create different directories, each with their own `book.toml` file pointing to the common source to get the behavior that I want. So perhaps this feature isn't needed.)

joshrotenberg:
## Summary 

This PR adds the ability to load a book with an explicit configuration file, both in the library and the command line tool. The changes preserve backward compatibility for both cases.

Fixes #1154. 

### Library

* adds `MDBook::load_with_config_file<P: Into<PathBuf>>(book_root: P, config_file: P) -> Result<MDBook>`
* `load` and `load_with_config_file` now use the same function to load the book with a caveat: if `load` is called and no configuration file is found, uses `Config::default()` (same as current behavior), and if `load_with_config_file` is called and the supplied config file isn't found, `bail!` 
* add more documentation with examples

### Command line

* adds the `-c, --config-file` option to `build`, `serve`, `test`, and `watch`.  When specified, `mdbook` will load the given configuration file instead of `book.toml` or the default `Config::default()`. 

### Tests

* adds `test/load.rs` to explicitly test various load methods
* adds the `test/cmd` suite of tests for testing `mdbook` using the [assert_cmd](https://crates.io/crates/assert_cmd) crate.

### TODO (still a Draft)

* add more `mdbook` tests to the new test suite
* if this PR is too big, it can be split into smaller PRs, just ask!


joshrotenberg:
Splitting this into two PRs, one for the [library](https://github.com/rust-lang/mdBook/pull/1518) and, if accepted, I'll open one for the command line.

PR #1518, opened 2021-05-03
Add MDBook::load_with_config_file

joshrotenberg:
Adds `MDBook::load_with_config_file` to load a book from its root and explicitly specify a configuration file to use. This is already possible with multiple steps; this just wraps up the work of loading the `Config` first and then loading the book with `MDBook::load_with_config`. A subsequent PR will add the ability to optionally specify a configuration explicitly on the command line using `mdbook`.

Also adds a dedicated test namespace for explicitly testing the various methods of loading a book and configuration.

### DIFF
--- a/src/book/mod.rs
+++ b/src/book/mod.rs
@@ -47,7 +47,7 @@ pub struct MDBook {
 }
 
 impl MDBook {
-  /// Load a book from its root directory on disk. If a `book.toml` file is present in the
+    /// Load a book from its root directory on disk. If a `book.toml` file is present in the
     /// root directory, the configuration is loaded from that. Otherwise, the
     /// default `Config` is used.
     ///
--- a/src/book/mod.rs
+++ b/src/book/mod.rs
@@ -30,6 +30,8 @@ use crate::utils;
 
 use crate::config::{Config, RustEdition};
 
+const DEFAULT_CONFIG_FILE_NAME: &str = "book.toml";
+
 /// The object used to manage and build a book.
 pub struct MDBook {
     /// The book's root directory.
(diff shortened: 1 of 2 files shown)


ehuss:
I'm a little uneasy making this change.  Can you say more about the scenario where the default `book.toml` doesn't work? I generally strive to keep tools easy to use without too many options, as those complexities tend to leak out into various places.  For example, cargo only allows manifests to be called `Cargo.toml`, and that is intentional.

joshrotenberg:
The main driver for this is #1154 which would allow someone to maintain separate `book.toml` files for the same book for various reasons.

The intent was to add this to the API, and then in another PR expose the functionality in the cli with an option (like `-c, --config-file`). It could, of course, only be available via the cli (by just using `MDBook::load_with_config`). 

That said, it is more of a nice to have than anything else, so I'm happy to shelve it.

bryfry:
As a user of mdbook I'd like the ability to do one of two things: 

 - override any book.toml configuration setting from the cli via specific flags
 - provide an alternative configuration file

IMO it seems like exposing this flag for an alternative configuration file applies a relief valve for all of the possible configuration changes that might be desired at execution time without implementation burden for every possible flag.

joshrotenberg:
> * override any book.toml configuration setting from the cli via specific flags

You can do this to some degree with environment variables: https://rust-lang.github.io/mdBook/format/config.html#environment-variables

bryfry:
Thanks, @joshrotenberg that feature was very helpful for my off by one change to the config.  

```
MDBOOK_OUTPUT__HTML__PRINT__ENABLE=true mdbook build
```

For more drastic changes it might be cumbersome to make many changes to a config via many environment variables

PR #1529, opened 2021-05-11
theme: Account for menu bar when scrolling to target

roberth:
When navigating to a page with a fragment, the browser will scroll
such that the element with that id is at the top of the viewport,
but that's also where the menubar is.

`scroll-padding-top` tells the browser to scroll just a little bit
less, so that the target isn't obscured by the menu bar.


ehuss:
Thanks for the PR!  There's already a `scroll-margin-top` in `general.css`.  Is that not working for you?  Are you perhaps running Safari 14?

PR #1531, opened 2021-05-14
Add before_body_end handlebars partial

xrmx:
Add a partial just before the body closing tag to add custom html code.
The use case is for adding javascript snippet that involves more than
just importing external scripts, e.g. web analytics.


xrmx:
Rebased and updated tests

PR #1539, opened 2021-05-19
Report book.toml parse error when invalid fields are found

### LINKED ISSUE #1522: Misleading book.toml related error message
Just on the subject of #1521, if I wrongly define the `book` section such as below, mdBook returns an inaccurate error message:
```toml
[book]
authors = ["Rafael Lúcio"]
src = "docs/src"
language = ["en", "pt-br"]
multilingual = true
title = "CET961 - Engenharia Assistida por Computador"
```
```bash
$ mdbook build
2021-05-05 10:41:52 [ERROR] (mdbook::utils): Error: Couldn't open SUMMARY.md in "/home/rafael/Documents/Repos/CAE/src" directory
2021-05-05 10:41:52 [ERROR] (mdbook::utils):    Caused By: No such file or directory (os error 2)
```
It cannot open `SUMMARY.md` at `src` because it is defined to be at `docs/src`.

If I set the language correctly it runs perfectly. For some reason a syntax error at the `language` field makes mdbook not read correctly the `src` field.

`mdbook build` here is just and example. This same message error does appear if the `clean`, `watch` or `serve` subcommands are executed.

joshrotenberg:
Prior to this change, any type errors parsing the `book`, `rust` or `build` sections of the `book.toml` would be silently ignored and the default for the given section would be used instead. 

This change adds an explicit check when parsing the above sections, and if an error is encountered, the error is reported rather than using the default configuration. For example, if the value of `language` in `[book]` is a list rather than a string:

```toml
[book]
language = ["en", "pt-br"]
# ...
```

The result will be something like: 

```bash
❯ mdbook build                         
2021-05-19 10:16:43 [ERROR] (mdbook::utils): Error: Invalid configuration file
2021-05-19 10:16:43 [ERROR] (mdbook::utils):    Caused By: invalid type: sequence, expected a string for key `language`
```

Fixes #1522.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -321,16 +321,11 @@ impl<'de> Deserialize<'de> for Config {
             }
         };
 
-        let book = if let Some(book) = table.remove("book") {
-            match book.try_into() {
-                Ok(b) => b,
-                Err(e) => {
-                    return Err(D::Error::custom(e));
-                }
-            }
-        } else {
-            BookConfig::default()
-        };
+        let book: BookConfig = table
+            .remove("book")
+            .map(|book| book.try_into().map_err(D::Error::custom))
+            .transpose()?
+            .unwrap_or_default();
 
         let build = if let Some(build) = table.remove("build") {
             match build.try_into() {
--- a/src/config.rs
+++ b/src/config.rs
@@ -327,27 +327,17 @@ impl<'de> Deserialize<'de> for Config {
             .transpose()?
             .unwrap_or_default();
 
-        let build = if let Some(build) = table.remove("build") {
-            match build.try_into() {
-                Ok(b) => b,
-                Err(e) => {
-                    return Err(D::Error::custom(e));
-                }
-            }
-        } else {
-            BuildConfig::default()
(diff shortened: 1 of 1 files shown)


ehuss (commented):
Thanks, this will be great to fix!  I don't know why the original just ignored the errors.

src/config.rs
```diff
@@ -310,30 +311,48 @@ impl<'de> Deserialize<'de> for Config {
...
+        let book = if let Some(book) = table.remove("book") {
+            match book.try_into() {
+                Ok(b) => b,
+                Err(e) => {
+                    return Err(D::Error::custom(e));
+                }
+            }
+        } else {
+            BookConfig::default()
+        };
```
ehuss, line 333: I think this can be kept to something closer to original like this:

```suggestion
        let book: BookConfig = table
            .remove("book")
            .map(|book| book.try_into().map_err(D::Error::custom))
            .transpose()?
            .unwrap_or_default();
```

ehuss (approved):
Thanks!

PR #1542, opened 2021-05-23
Apply max width to video as well as images

ISSOtm:
Since videos are essentially animated images, this same max width makes sense for both.


ehuss (approved):
Seems reasonable, thanks!

PR #1544, opened 2021-05-24
Restructure guide configuration section

joshrotenberg:
The guide's configuration documentation is rather long and has some well defined individual parts. This change just restructures it to be four separate sections plus an index. No actual content changes aside from the top level markdown heading and the README.md.


ehuss:
Thanks for the PR!  There appears to be a few broken links after moving the files.  Also, can you add a redirect in book.toml for the old file location?

guide/book.toml
```diff
@@ -25,3 +25,6 @@ boost-hierarchy = 2
 boost-paragraph = 1
 expand = true
 heading-split-level = 2
+
+[output.html.redirect]
+"/format/config.html" = "/format/configuration/index.html"
```
ehuss, line 30: I don't think this will work, the path needs to be relative.  I think it needs to be something like this:

```suggestion
"/format/config.html" = "configuration/index.html"
```

ehuss (approved):
Thanks!

Issue #747, opened 2018-07-28
How to disable playpen feature?


killercup:
I'm currently writing a book on CLI applications in Rust (<https://github.com/rust-lang-nursery/cli-wg/pull/47>). Executing the code snippets in playpen will sadly not be very useful.

Is there a way to disable the playpen feature globally?

mattico:
Apparently not! This should be easy to fix.

badboy:
This should be covered by https://github.com/rust-lang-nursery/mdBook/pull/744:

```
```rust,noplaypen
fn main(){
    println!("Don't run it.");
}
\```
```

ehuss:
`noplaypen` needs documentation.

Also, @killercup, is that good enough?

detly:
Could this be a global thing, set in `book.toml`? I had hoped that eg. `[output.html.playground]` would let me control it, but alas no.

pineapplehunter:
I thought a global flag in `book.toml` would be nice as @detly said. So I tried implementing in a draft PR #1546. Would this help?

adpaco-aws:
@pineapplehunter a global flag would be helpful. Does #1546 work? If it does, why not mark as ready for review?

schungx:
Yes, I keep having to put `no_run` in all my code examples just to disable the playpen.

pineapplehunter:
@adpaco-aws Thanks.
I'll check my code again and mark it as ready for review.

ehuss:
Closing as resolved by #1546.

PR #1546, opened 2021-05-25
Config to toggle the run button on codeblocks

pineapplehunter:
This PR is for #747.
This adds a flag to the config that toggles the run button on codeblocks.

In `book.toml` you can write like below to disable all run buttons on codeblocks.

```toml
[output.html.playground]
runnable = false
```


pineapplehunter:
If this needs more tests or docs, I'm happy to help!

ehuss:
Thanks!  Can you also include a small addition to https://github.com/rust-lang/mdBook/blob/master/guide/src/format/mdbook.md#rust-playground about this config option?

pineapplehunter:
Updated the doc!

ehuss (approved):
Thanks!

PR #1548, opened 2021-05-25
Mention missing file creation in build/watch/serve's docs

### LINKED ISSUE #1246: `mdbook serve` creates files linked from summary if they do not exist
I'm relatively sure that this is a "feature" and not a bug, but the behaviour surprised me.

I created a `SUMMARY.md` like the following:

```md
# Summary

- [Chapter 1](./chapter_1.md)
- [Chapter 2](#)
```

With the `#` uri as a placeholder since I hadn't yet created the file.
When I ran `mdbook serve` I was surprised to discover that the `src/#` file existed!

I guess this may be the behaviour described in the User Guide here:

> When a SUMMARY.md file already exists, the init command will first parse it and generate the missing files according to the paths used in the SUMMARY.md. This allows you to think and create the whole structure of your book and then let mdBook generate it for you.

Should this also happen when running `mdbook serve`?

* If so, it should probably be documented.
* If not, should the behaviour be toggle-able via a CLI flag?

ISSOtm:
Fixes #1246

### DIFF
--- a/guide/src/cli/build.md
+++ b/guide/src/cli/build.md
@@ -7,7 +7,8 @@ mdbook build
 ```
 
 It will try to parse your `SUMMARY.md` file to understand the structure of your
-book and fetch the corresponding files.
+book and fetch the corresponding files. Note that files mentioned in `SUMMARY.md`
+but not present will be created.
 
 The rendered output will maintain the same directory structure as the source for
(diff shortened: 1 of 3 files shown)


Dylan-DPC:
thanks

PR #1550, opened 2021-05-27
update handlebars to 4.0

sunng87:


ehuss:
Thanks!  Can you provide some more information about what has changed?  Is there anything in the `.hbs` syntax that is different that might break existing users?

sunng87:
@ehuss Sorry for late response. A changelog is available [here](https://github.com/sunng87/handlebars-rust/blob/master/CHANGELOG.md). There are some minor break changes with subexpressions and template inheritance for template developer. But these should not affect current mdbook use cases.

ehuss (approved):
Ok, thanks!

tuyen-at-work:
hi @ehuss  @sunng87 
one nice feature in handlebars 3.x is I can nest things like below:

```text
{{#include {{#include env.txt:1}}.PublishSettings}}
```

Let says I have **env.txt** file like this:

```text
dev-env-01

something else
```

Then above include would become:

```text
{{#include dev-env-01.PublishSettings}}
```

I used it for my environment list and it really helpful. Few days a go I see this break and look like because we update to Handlebars 4.x

PR #1554, opened 2021-05-28
Use the configured book src directory for the edit url template path

### LINKED ISSUE #1543: Edit URL leads to 404 if you've specified a custom src directory
It looks like the code that generates the URL that users navigate to when clicking on the "Suggest an edit" button (added in https://github.com/rust-lang/mdBook/pull/1506) assumes that your documentation is stored in a directory called `src`.

https://github.com/rust-lang/mdBook/blob/7e7e779ef7ae29ac051858b7e8c27843880c2366/src/renderer/html_handlebars/hbs_renderer.rs#L40-L50

As I've customised the directory where my documentation files are stored by using the `src` config option (to be `docs/` instead), I end up getting a 404 when I click the "Suggest an edit" menu item button.

When I click the button, I'm lead to https://github.com/anoadragon453/examplerepo/edit/develop/src/README.md instead of [https://github.com/anoadragon453/examplerepo/edit/develop/**docs**/README.md](https://github.com/anoadragon453/examplerepo/edit/develop/src/README.md).

I believe the above code just needs to read from the value of the `src` config option, and default to `"src"` if it's not set.

joshrotenberg:
The "Suggest an edit" URL used `src` as the top level path to the file to edit regardless of whether the book.toml `src` config was set resulting in an incorrect link to the file. This change uses the actual configured `src` to build the path instead.

Fixes #1543.

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -1,5 +1,5 @@
 use crate::book::{Book, BookItem};
-use crate::config::{Config, HtmlConfig, Playground, RustEdition};
+use crate::config::{BookConfig, Config, HtmlConfig, Playground, RustEdition};
 use crate::errors::*;
 use crate::renderer::html_handlebars::helpers;
 use crate::renderer::{RenderContext, Renderer};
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks!

PR #1555, opened 2021-05-28
Use std::panic::set_hook to exit serve thread if serving fails

### LINKED ISSUE #1545: mdbook should exit if serve can't serve
If, for example, the address is already in use, an error is given but the tool doesn't exit:

```bash
❯ mdbook serve                                                                                                130 ↵
2021-05-24 13:23:22 [INFO] (mdbook::book): Book building has started
2021-05-24 13:23:22 [INFO] (mdbook::book): Running the html backend
2021-05-24 13:23:22 [INFO] (mdbook::cmd::serve): Serving on: http://localhost:3000
thread '<unnamed>' panicked at 'error binding to [::1]:3000: error creating server listener: Address already in use (os error 48)', /Users/joshrotenberg/.cargo/registry/src/github.com-1ecc6299db9ec823/warp-0.2.5/src/server.rs:213:27
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2021-05-24 13:23:22 [INFO] (mdbook::cmd::watch): Listening for changes...
```

joshrotenberg:
`mdbook serve` starts the warp server in a thread and then continues on to watch for changes. However, if the thread panics, for example, because the address is already in use, `mdbook` doesn't exit. 

This change will explicitly call `std::process::exit(1)` from the thread if `warp::serve` panics.

Fixes #1545.

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -161,5 +161,12 @@ async fn serve(
     let fallback_route = warp::fs::file(build_dir.join(file_404))
         .map(|reply| warp::reply::with_status(reply, warp::http::StatusCode::NOT_FOUND));
     let routes = livereload.or(book_route).or(fallback_route);
+
+    std::panic::set_hook(Box::new(move |panic_info| {
+        // exit if serve panics
+        error!("Unable to serve: {}", panic_info);
+        std::process::exit(1);
+    }));
+
     warp::serve(routes).run(address).await;
 }


ehuss (approved):
Thanks!

I'm a bit surprised by this behavior in `warp`, it seems like it would be better to return a Result somewhere somehow.

ehuss:
Oh, I see this is in Draft status, are you intending to make more changes?

joshrotenberg:
Moved to ready. I was debating whether this was a good excuse to start some integration tests for the binary but that's probably better on its own.

joshrotenberg:
> I'm a bit surprised by this behavior in `warp`, it seems like it would be better to return a Result somewhere somehow.

Yeah, same. I dug around the docs and in some example code to see how others might handle it. My guess is that the usual use case is that you run the server in the main thread, so a panic would result in an exit anyway.

ehuss:
Ah, sounds good, thanks!

Issue #1556, opened 2021-05-30
mdbook init force flag documentation is wrong


joshrotenberg:
The documentation for the force flag for the `init` subcommand states "Skips confirmation prompts", but it actually only affects whether a prompt is given if the `--theme` flag is present and the theme directory exists. It has no effect on the prompts for creating the `.gitignore` or getting the book name (which, ok, isn't technically a "confirmation" prompt). 

The easiest fix is to just update the documentation. Another fix might be to make the `.gitignore` prompt also respect the flag so that the "confirmation" part is technically right. It might also be helpful to add a flag for creating the `.gitignore` and the book title too to round things out and make book initialization fully prompt free if someone desires.

Something like this. If `--force` is specified, in addition to the current behavior, init the book with the defaults for `--with-gitignore` and `--with-title`.

```bash
mdbook-init v0.4.8
Creates the boilerplate structure and files for a new book

USAGE:
    mdbook init [FLAGS] [dir]

FLAGS:
        --force             Skips confirmation prompts
    -h, --help              Prints help information
        --theme             Copies the default theme into your source folder
    -V, --version           Prints version information
        --with-gitignore    Creates a .gitignore file [default: true]
        --with-title        Set the book title [default: "my book"]

ARGS:
    <dir>    Directory to create the book in
             (Defaults to the Current Directory when omitted)
```

ehuss:
Hm, looks like this was changed as part of #491, but there doesn't seem to be an explanation why.  I would say the original intent is that it avoids all prompts.  I agree that `--force` should probably not do any prompts.

joshrotenberg:
Thoughts on adding the extra flags as well to make it possible to supply an answer for book name and .gitignore creation as part of the fix here? Or do you prefer to just make force do what it says and use defaults for those (and maybe call them out in the documentation string)?

ehuss:
Preferably have separate PRs.

As for the flags, how about `--title=...` and `--ignore=git` or `--ignore=none`?

joshrotenberg:
> As for the flags, how about `--title=...` and `--ignore=git` or `--ignore=none`?

I went with `--gitignore` as a boolean. If we wanted to support other `.ignore` file types, though, `--ignore=<type>` is probably better ... but the only other one I know is `.hgignore`.

The next step would be to set defaults and check `--force` first for both args:

* if `--force`, set default
* if `is_present`, set from command line value
* otherwise prompt

ISSOtm:
[...]

PR #1559, opened 2021-06-01
Add --title option and --gitignore flag to mdbook init

joshrotenberg:
Adds two new flags to `mdbook init`; `--title` and `--gitignore`.

These flags optionally replace the two currently interactive prompts so that a book can be initialized without any prompts if desired. 

See #1556.

### DIFF
--- a/src/cmd/init.rs
+++ b/src/cmd/init.rs
@@ -62,7 +62,7 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
     if let Some(ignore) = args.value_of("ignore") {
         match ignore {
             "git" => builder.create_gitignore(true),
-            _ => builder.create_gitignore(false)
+            _ => builder.create_gitignore(false),
         };
     } else {
         println!("\nDo you want a .gitignore to be created? (y/n)");
--- a/src/cmd/init.rs
+++ b/src/cmd/init.rs
@@ -1,5 +1,5 @@
 use crate::get_book_dir;
-use clap::{App, ArgMatches, SubCommand};
+use clap::{App, Arg, ArgMatches, SubCommand};
 use mdbook::config;
 use mdbook::errors::Result;
 use mdbook::MDBook;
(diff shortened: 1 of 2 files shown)


msathis (approved):
LGTM

src/cmd/init.rs
```diff
@@ -18,14 +18,27 @@ pub fn make_subcommand<'a, 'b>() -> App<'a, 'b> {
         )
         .arg_from_usage("--theme 'Copies the default theme into your source folder'")
         .arg_from_usage("--force 'Skips confirmation prompts'")
+        .arg(
+            Arg::with_name("title")
+                .short("t")
```
ehuss, line 23: I would probably drop the short flags for now.  Unless an option is extremely common, I find it is generally more readable to use the long form.

src/cmd/init.rs
```diff
@@ -45,13 +58,20 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
         }
     }
 
-    println!("\nDo you want a .gitignore to be created? (y/n)");
-
-    if confirm() {
+    if args.is_present("gitignore") {
```
ehuss, line 61: There doesn't appear to be a way to pass "no" on the command-line for gitignore.  Perhaps the flag should take a value?  Maybe something like `--ignore=git` or `--ignore=none`?

src/cmd/init.rs
joshrotenberg: Removed the short values for ignore and title.

src/cmd/init.rs
joshrotenberg: The arg is now `--ingore` and it either takes the value `git` (creates a .gitignore), `none` doesn't create a .gitignore, or if neither is present, the user is prompted.

mont1985 (approved):

ehuss (approved):
Nice, thanks!

I pushed a small change to update the CLI docs, too.

PR #1560, opened 2021-06-01
clippy: PathBuf to Path fixes.

joshrotenberg:
Fixes all instances of `warning: writing `&PathBuf` instead of `&Path` involves a new object where a slice will do`


ehuss (approved):
Thanks!  I'm not sure if that might be a breaking api change, but in practice probably won't affect anyone.

Issue #1517, opened 2021-05-03
Mixing `-` and `*` lists causes omissions


djmitche:
https://github.com/taskchampion/taskchampion/pull/220/files#diff-6e490221db20aaa056f2ed2c63443806b309ec63182541ddea9dd0b84e7eb501R4 changed SUMMARY.md:

```markdown
- [Welcome to TaskChampion](./welcome.md)
  - [Installation](./installation.md)
  * [Using the Task Command](./using-task-command.md)
```
to
```markdown
- [Welcome to TaskChampion](./welcome.md)
  * [Installation](./installation.md)
  * [Using the Task Command](./using-task-command.md)
```

before this change, the `installation` chapter did not appear in the TOC, nor did it get generated into HTML.

I suspect there's an issue with mixing `-` and `*`.  I admit it's a weird way to write a list, but the docs do say that either one can be used.  At least, it'd be nice if this was flagged as an error in `mdbook test`.

joshrotenberg:
Quick guess here ... the [parser](https://crates.io/crates/pulldown-cmark) is following the [CommonMark spec](https://spec.commonmark.org/0.29/). The list section of the spec [says](https://spec.commonmark.org/0.29/#lists):

> A list is a sequence of one or more list items of the same type.

The [same type](https://spec.commonmark.org/0.29/#of-the-same-type) part is key, because it means that a list item that starts with a different delimiter is technically the start of a new list. See [Example 271](https://spec.commonmark.org/0.29/#example-271):

So this markdown ...

```markdown
- foo
- bar
+ baz
```

Results in this html:

```html
<ul>
<li>foo</li>
<li>bar</li>
</ul>
<ul>
<li>baz</li>
</ul>
```

My assumption is that in the mdbook implementation, we (correctly) see the end of the list because the delimiter changes and end the parsing of the indented section.

joshrotenberg:
> I'd expect to either see it at the top level, or in a new anonymous indented section, or an error message.

Given that the SUMMARY.md is intended to have a stricter than normal syntax I think an error is the right way to go. In the meantime I added a note to the documentation.

frewsxcv:
Does our commonmark parser inform us that there is a mixing of symbols?

PR #1566, opened 2021-06-05
Documentation clarification on markdown list markers

joshrotenberg:
The markdown list markers in a list in the `SUMMARY.md` must be of the same type to avoid creating a new list, which results in the remaining list items not being rendered. This should arguably produce an error or be handled in some other way, but in lieu of that, here is some documentation. 

This clarifies the issue seen in #1517.

### DIFF
--- a/guide/src/format/summary.md
+++ b/guide/src/format/summary.md
@@ -54,8 +54,8 @@ to be ignored at best, or may cause an error when attempting to build the book.
 
    - [Another Chapter](relative/path/to/markdown4.md)
    ```
-   Numbered chapters can be denoted with either `-` or `*`.
-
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR!  I'm wondering, since this is an issue that should be extremely rare, could the clarification perhaps be a little shorter?  It takes up a significant amount of space, and I'd like to retain as much clarity here.  Perhaps something like this?

> Numbered chapters can be denoted with either `-` or `*` (do not mix delimiters).

joshrotenberg:
You're right. I shortened it.

ehuss (approved):
Thanks!

PR #1569, opened 2021-06-06
Add 'email' to list of fields where `s` is not a shortcut for search

daniel-vainsencher:


src/theme/searcher/searcher.js
```diff
@@ -316,7 +316,7 @@ window.search = window.search || {};
     
     // Eventhandler for keyevents on `document`
     function globalKeyHandler(e) {
-        if (e.altKey || e.ctrlKey || e.metaKey || e.shiftKey || e.target.type === 'textarea' || e.target.type === 'text') { return; }
+        if (e.altKey || e.ctrlKey || e.metaKey || e.shiftKey || e.target.type === 'textarea' || e.target.type === 'text' || e.target.type === 'email') { return; }
```
ehuss, line 319: Instead of wack-a-mole on the on the input types, perhaps something like this would work?

```javascript
!hasFocus() && /^(?:input|select|textarea)$/i.test(e.target.nodeName)
```

notriddle: `<input type=checkbox>` and `submit` do not accept arbitrary keyboard input, and should probably allow the search shortcut.

sspaeti: `!hasFocus() && /^(?:input|select|textarea)$/i.test(e.target.nodeName)` worked for me. I have a sign-up and login form, which otherwise I couldn't type `s`. When adding the above, it works perfectly fine. I would love to get this merged to avoid building myself every time.

sspaeti:
@daniel-vainsencher @ehuss Could we add ehuss suggestions and merge it afterward? WDYT?

ehuss:
I'm going to close as I believe this will be resolved by #2311. Thanks!

PR #1557, opened 2021-05-30
Some clippy fixes

xrmx:

PR #1572, opened 2021-06-08
Revert "book: use non_exhaustive attribute for struct Book"

### LINKED ISSUE #1571: 0.4.9 contains a breaking change compared to 0.4.8 for preprocessors
When using preprocessors, they should work if they are compiled for a semver-compatible mdbook version. However, if I compile a preprocessor using 0.4.8 and try to use it with 0.4.9, parsing stdin fails. This is due to the change in the commit c1b2bec7d7a56909f695f103d316453dab68798e. The `__non_exhaustive` field of `Book` was part of the json output passed to preprocessors:

```json
[
  {processor-context...},
  {
    "sections": [...],
    "__non_exhaustive": null
  }
]
```

With c1b2bec7d7a56909f695f103d316453dab68798e, that field was removed and replaced with the `#[non_exhaustive]` annotation. Thus, the `__non_exhaustive` field isn't serialized anymore but still expected by preprocessors of earlier versions. Thus, 0.4.9 breaks all preprocessors compiled for 0.4.x where x < 9.

joshrotenberg:
This reverts commit c1b2bec7d7a56909f695f103d316453dab68798e due to a breaking change. 

Fixes #1571. 
See #1557.

### DIFF
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -74,10 +74,10 @@ fn create_missing(src_dir: &Path, summary: &Summary) -> Result<()> {
 /// [`iter()`]: #method.iter
 /// [`for_each_mut()`]: #method.for_each_mut
 #[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize)]
-#[non_exhaustive]
 pub struct Book {
     /// The sections in this book.
     pub sections: Vec<BookItem>,
+    __non_exhaustive: (),
 }
 
 impl Book {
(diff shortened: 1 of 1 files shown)


oberien:
Would it make sense to revert the commit on v0.4.9 directly, yank 0.4.9 and publish 0.4.10 with the revert as only difference?

joshrotenberg:
> Would it make sense to revert the commit on v0.4.9 directly, yank 0.4.9 and publish 0.4.10 with the revert as only difference?

Possibly, yeah. I just did this quickly because I was already in there, but I'll let @ehuss decide the right way to handle it.

PR #1576, opened 2021-06-11
MyPaths are not required and fail

kana4:
Inputting paths are not necessary and will break ci, we can just leave paths out. We also changed 'master' to main since anyone reading this is probably going to be making mains

### DIFF
--- a/guide/src/continuous-integration.md
+++ b/guide/src/continuous-integration.md
@@ -26,7 +26,7 @@ before_script:
   - cargo install-update -a
 
 script:
-  - mdbook build && mdbook test # Edit in case of custom book path: mdbook build path/to/mybook && mdbook test path/to/mybook
+  - mdbook build && mdbook test # In case of custom book path: mdbook build path/to/mybook && mdbook test path/to/mybook
 ```
 
 ## Deploying Your Book to GitHub Pages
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR!

I'm a bit curious, why would local paths break CI?  Normally CI runs processes from the root of the git repo, and if the book isn't in the root, you need to pass the path in.

kana4:
Hi thanks for the speedy reply, If the book isn't in the root then yes you'd need to pass in the path, but if it is in the root like in the default config then passing a path in will cause redundancy. It was a misunderstanding on my part, I originally thought mypath meant the name of my repo, which caused me to have $HOME/myrepo/myrepo/book rather than $HOME/myrepo/book as I had just following the mdbook tutorial. The edits here just make it so that the user can copy paste in if they made a tutorial book.

ehuss:
Hm, ok.  Maybe it would be best to include comments on the lines where the user may need to make changes?

guide/src/continuous-integration.md
```diff
@@ -50,10 +50,10 @@ deploy:
   provider: pages
   skip-cleanup: true
   github-token: $GITHUB_TOKEN
-  local-dir: path/to/mybook/book
+  local-dir: book
   keep-history: false
   on:
-    branch: master
+    branch: main
```
ehuss, line 56: The text above still references the "master" branch.

kana4: Done

kana4:
Done

ehuss (approved):
Thanks!

PR #1578, opened 2021-06-14
Add feature to enable/disable opening the browser (enables building for wasm32-wasi)

lachlansneff:
When the `watch`, `serve`, and `open` features are disabled, `mdbook` can build for `wasm32-wasi`.

I'm personally partial to distributing tools like mdbook through wasm, so they can run without worrying about the os/architecture of the host machine.

This could be extended to build wasm binaries within the release CI action.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,6 @@ handlebars = "4.0"
 lazy_static = "1.0"
 log = "0.4"
 memchr = "2.0"
-open = "1.1"
 pulldown-cmark = "0.7.0"
 regex = "1.0.0"
 serde = "1.0"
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks!  I think this will be useful for library users, too, to reduce compile time since it isn't used by the library.

I'm undecided if I want to merge this now, though, since this is technically a breaking change. If someone runs something like `cargo install mdbook --no-default-features --features watch`, this would remove the `open` capability.  My instinct is that nobody does that, so the risk should be extremely small. Hm...

lachlansneff:
Definitely didn't realize this was going to be breaking. Well, let me know what you decide.

ehuss:
I'm going to close as `open` is no longer used, and I believe `opener` should not be a direct problem for cross-compiling.

PR #1581, opened 2021-06-19
Fix inconsistent between bash version and rust version of the sample

tuyen-at-work:
In the bash version, `y` have value 6, while rust version it has value 7


ehuss (approved):
Thanks!

PR #1586, opened 2021-06-25
Clarify the library path documentation

### LINKED ISSUE #1585: Unclear documentation for `mdbook test -L`
After reading your documentation, I was led to an incorrect understanding of how to use the `--library-path` option for `mdbook test`.

What it says:

> The `--library-path` (`-L`) option allows you to add directories to the library
> search path used by `rustdoc` when it builds and tests the examples. Multiple
> directories can be specified with multiple options (`-L foo -L bar`) or with a
> comma-delimited list (`-L foo,bar`).

What I thought it meant:

> Pass the cargo workspace as an argument. (e.g. `-L project/`)

What it actually meant:

> Pass the compilation artifacts as an argument. (e.g. `-L project/target/debug/deps/`)

How I found out what it actually meant:

- I looked through the code, until I determined that it repasses the argument to `rustdoc`.
- I looked at the `rustdoc` guide, which contained an example that corrected my misconception.

It looks like the person who wrote [this issue](https://github.com/rust-lang/mdBook/issues/394) had the same misconception that I did.

joshrotenberg:
Fixes #1585

### DIFF
--- a/guide/src/cli/test.md
+++ b/guide/src/cli/test.md
@@ -44,7 +44,7 @@ The `--library-path` (`-L`) option allows you to add directories to the library
 search path used by `rustdoc` when it builds and tests the examples. Multiple
 directories can be specified with multiple options (`-L foo -L bar`) or with a
 comma-delimited list (`-L foo,bar`). The path should point to the Cargo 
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!  I'm not sure if this will be a little misleading, since it might be implying it only works with cargo, but that is probably the common case.

Issue #1398, opened 2020-12-12
real ToC implementation and folder structures


finzzz:
Hello, thanks for making this app. I have been looking for gitbook alternative for a while. Apparently, I am using vscode to view and edit md files. In vscode, I can see folder pane and outline pane. The folder pane is to categorize the content, while the outline pane shows ToC based on md header. I think this is more dynamic structures because I don't need to edit summary.md as in your app. 

Looking forward for v1.

ehuss:
I'm not sure I fully understand, but it sounds like this is asking for something like #1016, so I'm going to close as a duplicate of that.

Issue #1016, opened 2019-08-22
feature request: automatically generate SUMMARY.md from the list of src/*.md files


wusunlab:
It would be nice if there is an option to generate `SUMMARY.md` automatically from the list of markdown files under `src` (except `SUMMARY.md` itself).  The generated `SUMMARY.md` would map the directory structure.  For example, if I have chapters and sections organized in subfolders,

```
.
├── book.toml
└── src
    ├── SUMMARY.md
    ├── chapter_1
    │   ├── section_1_1.md
    │   ├── section_1_2.md
    │   └── section_1_3.md
    ├── chapter_1.md
    ├── chapter_2
    │    └── section_2_1.md
    └── chapter_2.md
```

the generated `SUMMARY.md` could be look like

```markdown
# SUMMARY

- [Chapter 1 title](./chapter_1.md)
    - [Section 1.1 title](./chapter_1/section_1_1.md)
    - [Section 1.2 title](./chapter_1/section_1_2.md)
    - [Section 1.3 title](./chapter_1/section_1_3.md)
- [Chapter 2 title](./chapter_2.md)
    - [Section 2.1 title](./chapter_2/section_2_1.md)
```

The advantage of this feature would be to save the effort of manually checking if every chapter/section is included in `SUMMARY.md`.

mrec:
Are you proposing that titles are generated from the slugged filenames? That feels a bit limiting.

A possible alternative might be to take the ordering from filenames, but titles from the top-level headings in each file.

wusunlab:
Hi, sorry if it wasn't clear. I was suggesting that the chapter or section title could be parsed from the top-level heading of each markdown file.  I was able to prototype this in a crude way in Python. So I think this could be doable in Rust as well, though I'm not familiar with markdown parsing.

segfaultsourcery:
Hi, I've been wanting this as well, but I've had trouble figuring out what the title should be for the folders. By title I mean the text used for the links in the summary.

I tried getting around the problem by assuming there's a file in each folder called `landing.md`. It's in the same spirit of `index.html`, and it'll be the first file linked to in any folder. The first line in this file determines the name of the folder. For all other .md files, the first line is always used as the title.

This is a PoC combining my concerns with what I understood yours to be. Can you try it out and give your input?

https://github.com/segfaultsourcery/auto-generate-summary

PsiACE:
I like this feature, which provides a pretty good basis for migration.

[...]

PR #1588, opened 2021-06-26
Automatically generate the book's summary from the sources directory structure.

timothee-haudebourg:
Hi,

For now, even for simple books, the `SUMMARY.md` file is required. When the book's structure matches the `src` directory structure I find it quite annoying to have to manually maintain a copy of the structure in the `SUMMARY.md` file. It slows down the early process of writing, when the book's structure is not quite figured out yet, files are moving around, chapter titles are still changing, etc. and every time we have to manually reflect the changes in the summary file.

I believe this is what motivated issue #1016 (and #1398 I think?).

## What is in the pull request

I've introduced a new command line flag `--auto-summary` to automatically generate the book's summary from the sources directory structure. A `Summary` instance is created by scanning the `src` directory and creating a chapter for every file/directory.
For directory chapters, the inner `README.md` file is used as content. For instance the following structure:
```
1-foo/
    README.md
    1-a.md
    2-b.md
2-bar/
    README.md
    1-c.md
    2-subchapter/
       README.md
       1-d.md
```
creates the following summary (here in markdown for the sake of the example even though no file is created):
```markdown
- [Foo](1-foo/README.md)
  - [A](1-foo/1-a.md)
  - [B](1-foo/2-b.md)
- [Bar](2-bar/README.md)
  - [C](2-bar/1-c.md)
  - [Subchapter](2-bar/2-subchapter/README.md)
    - [D](2-bar/2-subchapter/1-d.md)  
```
Files are indexed in alphabetical order, that's why I'm prefixing every entry name with a number so they are added in the correct order. The title of each chapter is extracted from the markdown file.

### Implementation details

In practice, the title of each chapter is unknown during the summary creation stage. This means that the `name` field of `book::summary::Link` is now an `Option<String>`. In the `book::book::load_chapter` function, if the link name is `None`, then the title is extracted from the chapter's content.

The downside to this is that I had to introduce a new function `read_chapter_title` that extracts the title from a markdown document using `pulldown_cmark`. Even though I stop parsing right when the title is found, `pulldown_cmark` is implemented in such a way that the entire file is pre-parsed before I can extract the title. So in total, each chapter is parsed 3 times instead of 2. Also I'm not sure how to use this crate so `read_chapter_title` may need to be revised to be more reliable when the title contains formatting.

## What is not in the PR but could be

For now only link summary items are added but I imagine we could extract more structure information from the source directory.
For instance in the following directory:
```
group/
    1-foo.md
    2-bar.md
```
Since the `group` directory is not a chapter (it does not contain a `README.md` file), we could instead add the `1-foo.md` and `2-bar.md` chapters with a separator between them. We could also handle unnumbered prefix and suffix chapters inside specials `prefix` and `suffix` directories or something.

As mentioned before, to keep everything in the correct order, I personally prefix every file/directory name with a number (even when I manually write a `SUMMARY.md`). That number appears in the final chapter's URL. This is not really a problem, but adding an option to ignore it could be nice. Maybe for another PR though.

Finally, we could add a command to actually generate the `SUMMARY.md` file from the directory structure. Again, I feel like it would be for another PR (in fact, it could even replace this one if you're not happy with the idea of generating books without a summary file).

### DIFF
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::VecDeque;
 use std::fmt::{self, Display, Formatter};
 use std::fs::{self, File};
@@ -11,19 +12,27 @@ use crate::errors::*;
 /// Load a book into memory from its `src/` directory.
 pub fn load_book<P: AsRef<Path>>(src_dir: P, cfg: &BuildConfig) -> Result<Book> {
     let src_dir = src_dir.as_ref();
-    let summary_md = src_dir.join("SUMMARY.md");
 
-    let mut summary_content = String::new();
-    File::open(&summary_md)
-        .with_context(|| format!("Couldn't open SUMMARY.md in {:?} directory", src_dir))?
-        .read_to_string(&mut summary_content)?;
+    let summary = if !cfg.auto_summary {
+        let summary_md = src_dir.join("SUMMARY.md");
 
-    let summary = parse_summary(&summary_content)
-        .with_context(|| format!("Summary parsing failed for file={:?}", summary_md))?;
+        let mut summary_content = String::new();
+        File::open(&summary_md)
+            .with_context(|| format!("Couldn't open SUMMARY.md in {:?} directory", src_dir))?
+            .read_to_string(&mut summary_content)?;
 
-    if cfg.create_missing {
-        create_missing(src_dir, &summary).with_context(|| "Unable to create missing chapters")?;
-    }
+        let summary = parse_summary(&summary_content)
+            .with_context(|| format!("Summary parsing failed for file={:?}", summary_md))?;
+
+        if cfg.create_missing {
+            create_missing(src_dir, &summary)
+                .with_context(|| "Unable to create missing chapters")?;
+        }
+
+        summary
+    } else {
+        Summary::from_sources(src_dir)?
+    };
 
     load_book_from_disk(&summary, src_dir)
 }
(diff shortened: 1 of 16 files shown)


joshrotenberg:
I haven't looked through your changes yet, but there is something similar out there already: https://github.com/knightflower1989/mdbook-auto-gen-summary

In general, though, given that project, these changes and the various other issues you linked plus other related issues I've seen working on the project it seems like some kind of more flexible, possibly auto-generated solution for defining a book's structure is something that the project needs.

mont1985 (approved):

BodeSpezial:
I think the problem is that the path where it starts from is `src`. So the Readme in src fails the `include_path` function but unlike the other parts `src` will never be included.

I think the simplest option is to just check if there is a Readme in src and include it.

timothee-haudebourg:
Actually it is not really clear where the `src/README.md` should be placed in the summary. Should we add it as a prefix chapter or numbered chapter? If we add it as a numbered chapter, should it be placed before the other chapters of the `src` directory?

BodeSpezial:
I think I wouldn't number it. Just a prefix/preface/prologue chapter before all others.
And if numbered I'd probably set it to `0`.

timothee-haudebourg:
@joshrotenberg you seem to be one of the core maintainers. Is there any news on this, or the related issues?

aitorpazos:
Really looking forward for a feature like this, if maintainers feel the code is correct I'd like to see it merged :)

PR #1591, opened 2021-07-04
Update renderers.md

fritsstegmann:
I was doing some work on the epub plugin repo and when trying to build an epub I came across what looked like some broken references.


ehuss (approved):
Thanks!

PR #1592, opened 2021-07-04
Update README.md

fritsstegmann:
I was working on the epub plugin repo and when I tried to build an epub I came across what looked like a broken reference.


ehuss:
Can you say more about why you think it is a broken fragment?  The following URL seems to work for me: https://rust-lang.github.io/mdBook/format/configuration/renderers.html#html-renderer-options

fritsstegmann:
Ah sorry, my mistake, it seems that the epub generator does not add the ids to the h tags, looks like I've got more work to do on my side.

PR #1596, opened 2021-07-04
Rust Edition 2021 Support

### LINKED ISSUE #1593: Rust edition 2021 support
mdbook needs to be updated to support the 2021 edition configuration for Rust.  There are several things to do here:

* Update [`RustEdition`](https://github.com/rust-lang/mdBook/blob/b3c0b013501b43468852f4fb648e7a80495f8cd4/src/config.rs#L469-L476) and everything that references it (including the javascript code for the playground).  In hindsight this probably should have been a free-form string (so it wouldn't need updating each time), but that can't change now. Perhaps that would be good to change for 0.5.
* Update documentation.
* Unfortunately until the edition is stabilized, it requires the `-Zunstable-options` flag to be passed to rustdoc. I'm thinking maybe just pass that for now, and then remove it when it is stabilized?

joshrotenberg:
Updates to include the upcoming 2021 edition. 

Fixes #1593.

### DIFF
--- a/guide/src/format/configuration/general.md
+++ b/guide/src/format/configuration/general.md
@@ -63,8 +63,8 @@ Options for the Rust language, relevant to running tests and playground
 integration.
 
 - **edition**: Rust edition to use by default for the code snippets. Default
-  is "2015". Individual code blocks can be controlled with the `edition2015`
-  or `edition2018` annotations, such as:
+  is "2015". Individual code blocks can be controlled with the `edition2015`, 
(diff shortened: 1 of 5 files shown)


ehuss (approved):
Looks good, thanks!  

One minor issue is that specifying the edition in the example (like ```` ```rustc,edition2021 ````) will fail due to being unstable.  We can probably work around it, though.  I've been wondering if supporting something like RUSTDOCFLAGS to support that might help, but I'm reluctant to add that just for this.

PR #1599, opened 2021-07-07
feat(playground): show "No output" on playgrounds that print nothing

### LINKED ISSUE #1594: Play button should print something if there is no output.
When examples have a "Play" button, clicking it will say "Running...", and then nothing apparent to the user happens.  Most examples don't actually emit any output.  I'm wondering if it might be better, if there is no output from the example, to display some text to indicate it finished and that there is no output?  I'm not sure how confusing this actually is for people, but it doesn't seem like a great user experience to me.

notriddle:
Fixes #1594

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -133,7 +133,15 @@ function playground_text(playground) {
             body: JSON.stringify(params)
         })
         .then(response => response.json())
-        .then(response => result_block.innerText = response.result)
+        .then(response => {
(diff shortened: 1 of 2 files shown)


joshrotenberg (approved):
This looks right to my otherwise JS inexperienced eyes 👀 .

ehuss (approved):
Looks good, thanks!

Issue #1568, opened 2021-06-05
Create command line tool integration tests


joshrotenberg:
The mdbook API has test coverage but the command line tool, which is probably more often used than the API, could probably benefit from some integration (and possibly unit) testing so that changes can be made with more confidence.

[rustup](https://github.com/rust-lang/rustup/tree/master/tests) has some examples. [assert_cmd](https://crates.io/crates/assert_cmd) is also easy to use.

bbros-dev:
Our 2c.  From experience the more robust e2e test for a CLI is to actually run it as a user would.  There are alternatives, but [bats](https://github.com/bats-core/bats-core) or some such setup is actually where you want to sink any time - in our experience.

szabgab:
Some ideas for improvements:
* Check more details in the existing tests.
* Test with custom plugins.
* Maybe also test with existing 3rd party plugins.

PR #1604, opened 2021-07-13
Add CLI tests

notriddle:
Part of #1568

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,6 +42,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "28ae2b3dec75a406790005a200b1bd89785afc02517a00ca99ecfe093ee9e6cf"
 
 [[package]]
+name = "assert_cmd"
+version = "1.0.7"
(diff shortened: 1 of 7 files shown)


notriddle:
Finally, I managed to make a version that actually works correctly under Windows.

joshrotenberg:
This is great, thanks for kicking it off!

Obviously there is a lot to test, and doing it in small chunks makes sense. One suggestion I have would be to make a `test/cli` top level module, and then separate each test file to match the `cli` files, so like:

```
tests/cli/mod.rs  - any shared test stuff can go here
tests/cli/build.rs - all tests for the build command
...
```

notriddle:
@joshrotenberg

I can do that. I don't really like the result, because you're required to have a top-level `.rs` file for the test harness to actually run the tests. But it is doable.

ehuss (approved):
Thanks!  I went ahead and rebased to resolve the conflict.  I'm slightly concerned about adding so many dependencies, but I don't think there is an easy way around that.

@joshrotenberg I'm not sure if you had any more feedback on this, but I'm going to go ahead and merge. If you had anything else to add, please let me know.

PR #1607, opened 2021-07-21
Allow specifying preprocessor order

### LINKED ISSUE #1172: [feature] I'd like to specify the order of preprocessors
I'm building a preprocessor, but it requires that I work on the markdown file that already has the links resolved, i.e. it should have the `{{ #playpen }}`, `{{ #include }}`, and `{{ #rustdoc_include }}` already resolved. I'd like to work on Markdown files that have those `include`d files in the Markdown.

Very simply it would be awesome if I could just do:
```toml
[preprocessor.links]
[preprocessor.custom]
command = "cargo run --manifest-path=../../Cargo.toml --bin mdbook-custom"
```

But since this creates a toml table, it would require either:

1. a `depends` key in the `custom` preprocessor, like so:
```toml
[preprocessor.custom]
command = "cargo run --manifest-path=../../Cargo.toml --bin mdbook-custom"
depends-on = "links"
```
Note that in this hypothetical solution explicitly marking the `[preprocessor.links]` would not be needed.

2. Making the `preprocessor` into a list.

```toml
[[preprocessor]]
name = "links"
[[preprocessor]]
name = "custom"
command = "cargo run --manifest-path=../../Cargo.toml --bin mdbook-custom"
```

And after this configuration, each preprocessor would work on the book that has already been preprocessed by the preprocessors defined before it (if it is not the first).

If this is already possible, or a workaround exists please let me know (thx).

ISSOtm:
Fixes #1172

### DIFF
--- a/guide/src/format/configuration/preprocessors.md
+++ b/guide/src/format/configuration/preprocessors.md
@@ -56,3 +56,25 @@ be overridden by adding a `command` field.
 [preprocessor.random]
 command = "python random.py"
 ```
+
+### Require A Certain Order
+
+The order in which preprocessors are run is not guaranteed, but you can request some to run before or after others.
+For example, suppose you want your `linenos` preprocessor to process lines that may have been `{{#include}}`d; then you want it to run after the built-in `links` preprocessor, which you can require using the `before` and/or `after` fields.
+
+```toml
+[preprocessor.linenos]
+after = [ "links" ]
+```
+
+or
+
+```toml
+[preprocessor.links]
+before = [ "linenos" ]
+```
+
+It would be possible, though redundant, to specify both of the above in the same config file.
+
+`mdbook` will detect any infinite loops and error out.
+Note that order of preprocessors besides what is specified using `before` and `after` is not guaranteed, and may change within the same run of `mdbook`.
--- a/guide/src/format/configuration/preprocessors.md
+++ b/guide/src/format/configuration/preprocessors.md
@@ -59,8 +59,8 @@ command = "python random.py"
 
 ### Require A Certain Order
 
-The order in which preprocessors are run is not guaranteed, but you can request some to run before or after others.
-For example, suppose you want your `linenos` preprocessor to process lines that may have been `{{#include}}`d; then you want it to run after the built-in `links` preprocessor, which you can require using the `before` and/or `after` fields.
+The order in which preprocessors are run can be controlled with the `before` and `after` fields.
+For example, suppose you want your `linenos` preprocessor to process lines that may have been `{{#include}}`d; then you want it to run after the built-in `links` preprocessor, which you can require using either the `before` or `after` field:
 
 ```toml
 [preprocessor.linenos]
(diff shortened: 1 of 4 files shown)


ISSOtm:
I've determined that the order in which preprocessors are run is non-deterministic (blame [`topological-sort`](https://docs.rs/topological-sort/0.1.0/topological_sort)), and that makes one of the tests randomly fail since it expects an exact order.

The question then is: is this acceptable?

ehuss:
Thanks!  I think it would be ideal if the order would be at least stable.  Would it be possible to sort the preprocessors by name?

ISSOtm:
Sure, I'll see about stabilizing ties.

ISSOtm:
Should it be made a contract in the documentation, or should we specify that ties are guaranteed to be identically resolved across runs but not across versions?

ehuss:
I think mentioning that they are otherwise sorted by name would be fine.
I wouldn't worry about ignoring case, just doing a normal string comparison would be fine.

ISSOtm:
So sticking with the default string sorting should be enough, right?

ehuss:
Yea, the default string sort should be good.

ISSOtm:
Okay, I updated the test to expect `index` before `links`, and the documentation using your almost exact wording. Ready for review!

guide/src/format/configuration/preprocessors.md
```diff
@@ -56,3 +56,25 @@ be overridden by adding a `command` field.
 [preprocessor.random]
 command = "python random.py"
 ```
+
+### Require A Certain Order
+
+The order in which preprocessors are run is not guaranteed, but you can request some to run before or after others.
```
ehuss, line 62: I think this will be guaranteeing a certain order, since they will be sorted.  Perhaps this introduction could be changed to something like this?

```suggestion
The order in which preprocessors are run can be controlled with the `before` and `after` fields.
```

src/book/mod.rs
```diff
@@ -386,36 +382,102 @@ fn is_default_preprocessor(pre: &dyn Preprocessor) -> bool {
...
-                )),
+        for (name, table) in preprocessor_table.iter() {
+            preprocessor_names.insert(name.to_string());
+
+            let exists = |name| {
+                (config.build.use_default_preprocessors && DEFAULT_PREPROCESSORS.contains(&name))
+                    || preprocessor_table.contains_key(name)
+            };
+
+            if let Some(before) = table.get("before").and_then(Value::as_array) {
```
ehuss, line 404: Can this return an error if the field is not an array?  I suspect it might be a likely mistake a user can make (`before = "links"`), and currently it seems to ignore it.

src/book/mod.rs
```diff
@@ -386,36 +382,102 @@ fn is_default_preprocessor(pre: &dyn Preprocessor) -> bool {
...
-                "links" => preprocessors.push(Box::new(LinkPreprocessor::new())),
-                "index" => preprocessors.push(Box::new(IndexPreprocessor::new())),
-                name => preprocessors.push(interpret_custom_preprocessor(
-                    name,
-                    &preprocessor_table[name],
-                )),
+        for (name, table) in preprocessor_table.iter() {
+            preprocessor_names.insert(name.to_string());
+
+            let exists = |name| {
```
ehuss, line 399: I'm curious, what's the motivation for ignoring preprocessors that don't exist? I would expect those to be an error, since it likely indicates the user made a typo or mistake.

src/book/mod.rs
ISSOtm: Right, my bad.

src/book/mod.rs
ISSOtm: The use case was if the user was toggling a preprocessor on and off (e.g. if troubleshooting), so that they wouldn't have to also update teh dependency fields all the time. Should this be changed to print a warning instead?

ehuss: Ah, a warning sounds good!

ISSOtm:
Done!

ehuss (approved):
Thanks!

PR #1613, opened 2021-07-26
Don't highlight `inline code` blocks in headers with output.html. playpen(playgroud).editable=true

### LINKED ISSUE #1580: Code spans in headers are highlighted with editable playpen
If `output.html.playpen.editable = true` is set, then any headers with a code span end up with the span highlighted as a link.  This is not intended, and I'm not sure why it is happening. This was previously fixed in #1162, but for some reason that didn't fix it with editable code blocks.

Example:

```
## This has a `code` span
```

Is rendered as:

image

when it should be:

image

FWYongxing:
Don't highlight `inline code` blocks in headers with output.html. playpen(playgroud).editable=true

Fixes #1580

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -162,12 +162,13 @@ function playground_text(playground) {
     if (window.ace) {
         // language-rust class needs to be removed for editable
         // blocks or highlightjs will capture events
-        Array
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

PR #1617, opened 2021-07-30
Fix some x overflows

lf-:
This PR fixes long inline code doing an x overflow.

There's another x-overflow on this page, where the table with the heading  "punycode + encoding" also overflows slightly. I have fixed this one as well.

Spotted on
https://rust-lang.github.io/rfcs/2603-rust-symbol-name-mangling-v0.html

### DIFF
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -28,7 +28,7 @@ code {
 
 /* make long words/inline code not x overflow */
 main {
-    overflow-wrap: anywhere;
+    overflow-wrap: break-word;
 }
 
 /* make wide tables scroll if they overflow */
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -26,9 +26,9 @@ code {
     font-size: 0.875em; /* please adjust the ace font size accordingly in editor.js */
 }
 
-/* make long inline code not x overflow */
-:not(pre) > code {
-    word-break: break-word;
+/* make long words/inline code not x overflow */
+main {
+    overflow-wrap: anywhere;
 }
 
 /* make wide tables scroll if they overflow */
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -31,6 +31,11 @@ code {
     word-break: break-word;
 }
 
+/* make wide tables scroll if they overflow */
+.table-wrapper {
+    overflow-x: auto;
+}
+
 /* Don't change font size in headers. */
 h1 code, h2 code, h3 code, h4 code, h5 code, h6 code {
     font-size: unset;
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -26,6 +26,11 @@ code {
     font-size: 0.875em; /* please adjust the ace font size accordingly in editor.js */
 }
 
+/* make long inline code not x overflow */
+:not(pre) > code {
+    word-break: break-word;
+}
+
 /* Don't change font size in headers. */
 h1 code, h2 code, h3 code, h4 code, h5 code, h6 code {
     font-size: unset;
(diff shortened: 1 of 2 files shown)


ehuss:
Thanks, I have a few questions, though:

What does the `overflow-x` do specifically?  What circumstances are there when the tables overflow? Can you show screenshots of the differences?

Can you say more about why this uses a div wrapper?  Is it a concern about `<table>` being used by other things? 

Are there any concerns about using `word-break: break-word;`?  According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#values) it is deprecated.  I'm not sure what deprecated even means in this case.  Would it maybe be better to use `overflow-wrap: anywhere`?

lf-:
Thanks for the review! Looking deeper into these questions has been useful for me as a CSS non-expert.

> What does the `overflow-x` do specifically? What circumstances are there when the tables overflow? Can you show screenshots of the differences?

Tables overflow like this:
image

The conditions in which this happen are when there are long cells (I stuffed some extra long text into the cell to get it to overflow a little harder) or generally more width than fits on the screen.

This is what the `overflow-x: scroll` on the wrapper div does: it makes the table scrollable horizontally rather than making the entire page scrollable horizontally:

image

The reason it needs to be on a wrapper div with the `overflow-x: auto` on it — or at least my best guess of it based on [reading MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-x) — is that `table` elements are `display: table` rather than `display: block` (indeed overriding them to `display: block` has it scroll the table itself. It appears, however, that the element then doesn't end up centered with the `margin: 0 auto` on it, so I think we probably need the wrapper?). [This SO answer](https://stackoverflow.com/a/35329296/2350164) claims that `display: table` in some parent is required for table cell rendering to work properly.

---

> Are there any concerns about using `word-break: break-word;`? According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#values) it is deprecated. I'm not sure what deprecated even means in this case. Would it maybe be better to use `overflow-wrap: anywhere`?

I somehow missed the deprecation notice, my bad (from further googling, it was deprecated as of CSS 3). I can confirm that `overflow-wrap: anywhere` gets the same effect.

From putting some long stuff into a paragraph, I realized we probably should enable this everywhere on the app since it's possible to cause x overflows with body text too. I am going to do that:

image

`overflow-wrap: always` on `main`:

image

Some code, overflowing without word breaking configured:

image

With wrap turned on:
image

YJDoc2:
@lf- @ehuss  please take a look :
Following tests are done on the test_book, which I have requested to be added in PR #1662, it can be found in my fork at https://github.com/YJDoc2/mdBook/tree/test-book


I think rather than `overflow-wrap: anywhere` , `overflow-wrap: break-word` will be a better choice :
I checked this on Linux + Firefox + (Desktop mode / Mobile mode)  and Linux + Brave (chromium based) + (desktop/mobile) mode, following results are consistent on both.
### Original
This is without the changes in this PR
Original table desktop
original-table-desktop


Original Table mobile
original-table-mobile

As you can see this makes the whole page scroll in mobile, rather than the table (evident by the page title being off screen), which was the issue.

### After this PR
The issue is due to `overflow-wrap: anywhere`, this also breaks the words which should not be broken, and the tables don't scroll
For example, on desktop mode :
PR-desktop
this breaks words `col2` and `val2`, which definitely should not have been broken on desktop.
On mobile, it is worse :
PR-mobile

Here, to prevent overflow, it breaks every word into characters and wraps them around.

## With suggested changes

Now with using 
```css
main {
    overflow-wrap: break-word;
}
```
(table wrapper class introduced in this PR is kept as in this PR)
we get :
on desktop :
suggested-desktop

and on mobile :
suggested-mobile

where it scrolls the table, rather than scrolling the whole page, as evident by the page title staying in view, and the columns being scrolled.

ehuss:
@lf- Would you be willing to update this using `break-word`?

lf-:
Done and rebased.

lf-:
@ehuss if you missed the notif

PR #1628, opened 2021-08-17
增加用户指定 summary 文件的功能（now support to appoint the summary file）

CC2130:
add -s or --summary Option in subcommand(build):
```
    -s, --summary <summary-file>    Book's summary file
                                    Relative paths are interpreted relative to the book's src root directory.
                                    If omitted, use the default `SUMMARY.md` as summary file
```
allow user to appoint a custom `summary` file rather than always use the `SUMMARY.md`


ISSOtm:
Wouldn't it be better to make this a configuration option? If you want to use a different summary file, you'll likely want to use it at every run; and you could still [override the configuration using environment variables](https://rust-lang.github.io/mdBook/format/configuration/environment-variables.html).

ehuss:
FWIW, I think I would prefer to not make it an option or configurable, unless there is a strong reason to.

PR #1632, opened 2021-08-24
Fix preprocessor example

tsoutsman:
Hi, I was looking into building my own preprocessor and copied the example from the repo. When I ran Clippy, there were 2 changes suggested, which I assume should be changed in the example.


ehuss (approved):
Thanks!

PR #1634, opened 2021-08-26
Add information about a new backend `mdbook-man`

vv9k:
This PR adds links to a new backend that I'm working on: [mdbook-man](https://github.com/vv9k/mdbook-man). It generates manual pages from the book that can later be viewed with tools like `man`.


ehuss (approved):
Thanks!

PR #1637, opened 2021-08-31
Include chapters with no headers in the search index

notriddle:


ehuss (approved):
Thanks!  In the future, it would be helpful to have a PR description that discusses the nature of the change (what is changing and why).

PR #1638, opened 2021-09-01
serve/watch: open browser in separate thread

johannst:
When running 'serve --open' / 'watch --open', start the browser in a
separate OS thread to start watching for changes immediately in case the
watch feature is enabled.

See issue (#1639).

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -100,9 +100,11 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
     let serving_url = format!("http://{}", address);
     info!("Serving on: {}", serving_url);
 
-    if open_browser {
-        open(serving_url);
-    }
+    let open_thread = if open_browser {
+        Some(std::thread::spawn(|| open(serving_url)))
+    } else {
+        None
+    };
 
     #[cfg(feature = "watch")]
     watch::trigger_on_change(&book, move |paths, book_dir| {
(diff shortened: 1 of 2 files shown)


ehuss:
Hm, so I haven't noticed this problem, and I'm wondering if it is specific to some configurations of xdg-open.  I'm wondering if it would make sense to switch to the [`opener`](https://crates.io/crates/opener) crate which just [spawns the xdg-open in the background](https://github.com/Seeker14491/opener/blob/86e0a154d29a87fbde1dd2548ccb216f078951ee/opener/src/linux_and_more.rs#L52-L59) as opposed to `open` which for some reason [waits for the process to finish](https://github.com/Byron/open-rs/blob/066a591823ddebb2904959b6395bc945c22ba213/src/lib.rs#L187).

johannst:
I looked into it again and observed the following (linux, firefox):
1. If the browser process is started due to `--open` then `xdg-open` will block waiting for the browser process to terminate (as you linked in the [`open`](https://github.com/Byron/open-rs/blob/066a591823ddebb2904959b6395bc945c22ba213/src/lib.rs#L187) crate). I mean that makes sense it waits for the child process to terminate.
2. If a browser process is already running then `xdg-open` will return after opening a new tab in the running browser instance (the new firefox child process terminates after opening the tab).

So to support 1) we should either use a crate to just spawn the process (as you mentioned opener) or handle it by ourself.
But if we always want non-blocking behavior I don't see much value in doing ourself. In that case we could just use a crate that already does the thing we want :)

johannst:
@ehuss, please feel free to close this PR in case you agree that using a crate which allows non-blocking `xdg-open` is preferred over doing it ourselves here.

ehuss:
Sure, yea.  I'd be happy to merge a PR that switched to `opener`.

PR #1647, opened 2021-09-08
Fix vertical alignment of chapter navigation arrows

raviqqe:
# Description

This PR fixes vertical alignment of chapter navigation arrows.

Before:

Screenshot 2021-09-07 19 11 51

After:

Screenshot 2021-09-07 19 11 39

### DIFF
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -159,11 +159,14 @@ a > .hljs {
 
 .mobile-nav-chapters {
     font-size: 2.5em;
-    text-align: center;
     text-decoration: none;
     width: 90px;
+    height: 55px;
     border-radius: 5px;
     background-color: var(--sidebar-bg);
+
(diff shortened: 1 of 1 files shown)


src/theme/css/chrome.css
```diff
@@ -159,11 +159,14 @@ a > .hljs {
 
 .mobile-nav-chapters {
     font-size: 2.5em;
-    text-align: center;
     text-decoration: none;
     width: 90px;
+    height: 55px;
```
raviqqe, line 164: This height is based on the rendering result on the Chrome browser without these changes. We need to add it because CSS grids seem to ignore `line-height` which were previously respected.

GuillaumeGomez:
The layout and display both changed in the last 5 years (!!), so I'll close this PR as it's very outdated now. Thanks for your contribution in any case!

PR #1651, opened 2021-09-14
Only switch to themes on buttons that have the `theme` class

### LINKED ISSUE #1649: Clicking the border of the theme popup produces broken theme
Clicking the border of the theme popup has an effect, and apparently selects "no theme", removing all theme styles.

This probably shouldn't be happening.

notriddle:
Fixes #1649


ehuss (approved):
Thanks!

PR #1652, opened 2021-09-17
Using syntect as a higlighting backend rather than highlight.js

notriddle:
This version of the syntect highlighting backend fixes two of the complains coming from [version 2](https://github.com/rust-lang/mdBook/pull/1624):

* It fixes the bugs. This version takes advantage of syntect's Scope Stack API to insert the "boring" span in the perfect spot regardless of how the scopes are nested, with very little groveling around the generated HTML using regexes (it does run one simple post-processing step to replace `<span class="">` with `<span class="boring">`, because we would have to actually patch the syntax definition to enable a span that's actually named "boring").
* It moves the playground handling step into the pulldown-cmark converter, instead of trying to do it by groveling around the generated HTML using regexes, which was one of the TODOs on the iteration 2 PR.

Here's the result of running all of these versions using hyperfine. These aren't directly comparable to v2's, since I'm using a different setup than @ThePuzzlemaker

Command | Mean [s] | Min [s] | Max [s]
-- | -- | -- | --
Syntect V3 | 1.727 s ±  0.053 s | 1.673 s | 1.833 s
Syntect V2 | 2.934 s ±  0.085 s | 2.805 s | 3.133 s
HLJS | 1.214 s ±  0.022 s | 1.190 s | 1.262 s

## To-dos

- [x] Consider detecting "runs" of boring lines, and wrapping these whole chunks in a boring span instead of wrapping each line in its own. This thing is already extremely complex and subtle, and adding that would make it even harder to understand, but the generated HTML gets pretty bloated under the current approach.
- [x] Make sure CSS files are free of any possible bugs from syntect's sublime theme to CSS generator (carried over from v2).
- [x] Update documentation accordingly (carried over from v2).
- [x] Consider supporting both highlight.js and syntect, with config options to choose between them. (not gonna do it unless someone figures out how to make boring lines work with a current version of hljs, which it seems like they haven't)


notriddle:
The color schemes aren't actually the same. Is that a bug?

Old version:

image

New version:

image

ThePuzzlemaker:
That's a remnant of the differences between the original (hljs) color schemes and the new (`.tmTheme`) color schemes converted using [`syntect::html::css_for_theme_with_class_style`](https://docs.rs/syntect/4.6.0/syntect/html/fn.css_for_theme_with_class_style.html). I tried to find ones that were the closest to the original, but clearly there were some differences. If we want them to be the exact same, we'll need to make our own. It'd probably be easiest to tweak the generated CSS (which I already did with some simple fixes) to be more like the originals, rather than to tweak the `.tmTheme`s.

Otherwise, this looks good and thanks for figuring this out! :P

notriddle:
@ThePuzzlemaker I've tweaked the themes to make them closer to before, and also wrote the documentation for generating these themes from their tmTheme source files.

notriddle:
@ThePuzzlemaker where did you get the Handlebars syntax from? Whatever source it came from should really be documented.

ThePuzzlemaker:
It was from https://github.com/daaain/Handlebars/blob/master/grammars/Handlebars.tmLanguage, converted using Sublime Text 3 to a `.sublime-syntax` file. Must've forgotten to write that down :sweat_smile:

notriddle:
Okay, @ThePuzzlemaker. It's added.

src/theme/syntaxes/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+Handlebars.sublime-syntax is from https://github.com/daaain/Handlebars/blob/master/grammars/Handlebars.tmLanguage (it was converted to `sublime-syntax` using Sublime Text).
+
+TOML.sublime-syntax is from https://github.com/jasonwilliams/sublime_toml_highlighting/blob/0f14b7caf3c775a5d18551a9563a9debdb10c9a9/TOML.sublime-syntax
+
+# Note
+
+This folder is not copied over to the book directory when using `mdbook init`, nor is it indexed at runtime. All of the files in this folder are scraped by build.rs.
+
+To make build.rs run again without running `cargo clean`, touch the run `touch build.rs`.
```
ISSOtm, line 11:
```suggestion
To make build.rs run again without running `cargo clean`, touch the file: `touch build.rs`.
```

src/theme/syntaxes/README.md
```diff
@@ -0,0 +1,12 @@
+Console.sublime-syntax was written for mdBook.
+
+Handlebars.sublime-syntax is from https://github.com/daaain/Handlebars/blob/master/grammars/Handlebars.tmLanguage (it was converted to `sublime-syntax` using Sublime Text).
```
ISSOtm, line 3: This is under the MIT license, shouldn't it be pasted somewhere?

ISSOtm:
Nice! :+1:

This is very interesting for [us](https://github.com/gbdev), since we have [Pan Docs](https://gbdev.io/pandocs), which contains some [RGBASM](https://rgbds.gbdev.io/docs/master/rgbasm.5) assembly [snippets](https://gbdev.io/pandocs/Reducing_Power_Consumption.html); so we could test this with [the existing syntax definition](https://github.com/ISSOtm/sublime-RGBDS). That'd be some testing with a 3rd-party grammar, seeing if it beings up any bugs.

ISSOtm:
What needs to be done for this to be merged?

jonaprieto:
What is the status of this? are there any plans to merge this soon?

KFearsoff (commented):
Let's get this to be merged!

I've successfully built this PR locally after merging master changes. I'll submit it shortly. It passes all tests, but I haven't verified by hand that it looks correct.

KFearsoff:
> If you'd like any help, please @ me and I'll look through your revised PR.

Thank you! I definitely wouldn't mind any help as long as it helps this PR get through!

Here's the current diff I have: https://github.com/KFearsoff/mdBook/compare/notriddle/use-syntect-v3~2...KFearsoff:mdBook:notriddle/use-syntect-v3

One issue I noticed is that a lot of grammars in the test book are not formatted, so we should maybe change the docs (or the book, or the default grammars we build). The colorscheme would also be nice to fix I guess, though our options are quite limited (perhaps there's a better default colorscheme that can function as a default?)

Right now, I'm working on one more patch that would fix the merge conflicts with the master.

SichangHe:
@KFearsoff, ping.

KFearsoff:
So what are the action points now?

jeaye:
Happy holidays, folks! Is there any word on getting this merged? I see a lot of friction going on in the tickets related to highlight.js, wanting to update, but this PR continually being referenced as the next step.

notriddle:
Do we even want to do it this way, or would arborium make more sense nowadays?

jeaye:
> Do we even want to do it this way, or would arborium make more sense nowadays?

I would take _any_ build-time highlighting back-end. Currently, syntect has the most momentum, due to your work (and patience). If your branch is easily buildable, perhaps I can just use that rather than prompting for the merge.

PR #1656, opened 2021-09-26
switch from open to opener

### LINKED ISSUE #1639: serve/watch with --open blocks change watcher
In case the `watch` feature is enabled and we  run one of the following commands
```bash
mdbook serve --open
mdbook watch --open
```
listening for file changes is blocked until the browser is closed.

Should we open the browser in a separate thread and start listening for changes immediately?

See PR (#1638) for example.

johannst:
By default, `opener` launches the subprocess without waiting for its
completion, compared to `open` which waits for its completion.

This is helpful in case the `watch` feature is enabled and one of the
following commands `watch | serve --open` is used. If this command would
open the browser, listening for changes would be blocked by the browser.

This fixes #1639.


ehuss (approved):
Thanks!

PR #1657, opened 2021-09-29
Documentation: CI: GitLab Pages Example

lclc:
Replace 'only' with 'rules'.
'only' is not actively developed anymore with GitLab (see https://docs.gitlab.com/ee/ci/yaml/#only--except)

+ Minor formatting changes.


ehuss (approved):
Thanks!

PR #1659, opened 2021-10-03
Fix unreadable code blocks chrome.css

pickfire:
I am not sure cases `.hljs` is used so I can't specifically enable this so I added a few other whitelisted stuff.

Before

image

After

image

Screenshots taken from helix editor. https://github.com/helix-editor/helix although the styles are a bit custom but these parts are not customized.

### DIFF
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -193,7 +193,7 @@ a > .hljs {
     border-radius: 3px;
 }
 
-:not(pre):not(a) > .hljs {
+:not(pre):not(a):not(td):not(p) > .hljs {
     color: var(--inline-code-color);
     overflow-x: initial;
 }


ehuss:
Thanks for the PR!  Can you provide some background here?  What does the original markdown look like?  Why is it being rendered in such a strange way? Why add `td` and `p` to this rule?  

I wouldn't expect this change to be needed.  Inline code is intended to be highlighted except for exceptionally rare circumstances (like `a` tags).

pickfire:
The blockquote parts seemed to have the same background and foreground.

```markdown
## Normal mode                                                                  

### Movement                                                                    

> NOTE: Unlike vim, `f`, `F`, `t` and `T` are not confined to the current line. 
| Key          | Description                                        | Command                     |
| -----        | -----------                                        | -------                     |                                                             
| `h`, `Left`  | Move left                                          | `move_char_left`            |                                                             
| `j`, `Down`  | Move down                                          | `move_char_right`           |                                                             
| `k`, `Up`    | Move up                                            | `move_line_up`              |
```

Not sure why is it being rendered in such as strange way. So I had to add `td` and `p` to those `.hljs` that was being generated here to fix the issue.

ehuss:
I don't see that when built locally:

image

It looks like they have a [custom theme](https://github.com/helix-editor/helix/tree/a930f99179c4e0706254b243a4c0672089cb59ed/book/theme).  Perhaps there is an issue in their theme? Maybe it is using an out-of-date copy?

PR #1670, opened 2021-10-14
Anonymize IP for Google analytics

kkaempf:
This patch anonymizes IPs tracked by Google analytics
to address privacy concerns.

Signed-off-by: Klaus Kämpf <kkaempf@suse.de>


ehuss:

PR #1675, opened 2021-10-24
Deprecate google-analytics

### LINKED ISSUE #1635: Google Analytics only work with old Universal Analytics tag propery id
It seems that the new google analytics property id does not work with the current js.

## To reproduce

You can start a new mdbook and check for tags, and use google tag assistant recording To debug the issue...

Updating the js bundle should be enough of a fix

ehuss:
This puts a deprecation notice for the `output.html.google-analytics` field, with the intent to remove it in the 0.5 release. The current code has been outdated for a while, and doesn't provide very much flexibility.  I also prefer not to endorse any particular service, instead letting the user pick whichever they want.  Placing the code in the `<head>` tag is the recommended approach, and the `head.hbs` extension provides a means to do so.

Closes #1635

### DIFF
--- a/guide/src/format/configuration/renderers.md
+++ b/guide/src/format/configuration/renderers.md
@@ -21,8 +21,8 @@ The following configuration options are available:
 - **mathjax-support:** Adds support for [MathJax](../mathjax.md). Defaults to
   `false`.
 - **copy-fonts:** Copies fonts.css and respective font files to the output directory and use them in the default theme. Defaults to `true`.
-- **google-analytics:** If you use Google Analytics, this option lets you enable
-  it by simply specifying your ID in the configuration file.
+- **google-analytics:** This field has been deprecated and will be removed in a future release.
+  Use the `theme/head.hbs` file to add the appropriate Google Analytics code instead.
 - **additional-css:** If you need to slightly change the appearance of your book
   without overwriting the whole style, you can specify a set of stylesheets that
   will be loaded after the default ones where you can surgically change the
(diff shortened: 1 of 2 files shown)


ashutoshmjain:
Quick question. - Do I need to be on a specific version of mdbook to be able to use GA4 ? Asking cuz I received this notice from google ..

`Universal Analytics will no longer process new data in standard properties beginning July 1, 2023. Prepare now by setting up and switching over to a Google Analytics 4 property.`

ashutoshmjain:
My current version is 4.12  ( as available in nixpkgs)

schungx:
If you use the new script snippet from Google, it'd work both for GA4 as well as the old UA accounts.

However, beware - GA4 is missing a _LOT_ of functionalities, especially filters.  So you may actually want to send simultaneously to both accounts for a while.

ashutoshmjain:
Stephen @schungx 3- Thanks for the response  .  Could you please link me to the  google script you mentioned above .

schungx:
This is how I have it in `theme/head.hbs`

```html
<!-- Global site tag (gtag.js) - Google Analytics -->
<script async src="https://www.googletagmanager.com/gtag/js?id=G-XXXXXXXXXX"></script>
<script>
    var localAddrs = ["localhost", "127.0.0.1", ""];

    // Make sure we don't activate Google Analytics if the developer is inspecting the book locally...
    if (localAddrs.indexOf(document.location.hostname) === -1) {
        window.dataLayer = window.dataLayer || [];
        function gtag() { dataLayer.push(arguments); }
        gtag('js', new Date());

        gtag('config', 'G-XXXXXXXXX');
    }
</script>
```

OBITOONDEADO1MZ (approved):

Issue #1222, opened 2020-05-13
Preprocessor cannot add folders/files directly to build directory


paul-nelson-baker:
Hello!

I'm implementing a [preprocessor that generates plantuml diagrams](https://github.com/paul-nelson-baker/mdbook-plantuml-renderer/blob/2ef1cb8b3dd2258f4f96bf902ca6e7a6faacf18b/src/main.rs). It extracts codeblocks with the plantuml language, renders the SVG, and then places a markdown element to reference the image.

While I write my mdbook with `mdbook serve` I'm finding that I must place these rendered files in the `src` directory in order for them to appear in the `book` directory. If my preprocessor tries to put these files into the `book` path directly, they're immediately nuked.

Is there a way to programmatically tell mdbook that these dynamically created files are valid so they're not removed?

auser:
I would absolutely love the ability to update the `book.toml` within a preprocessor. @ehuss commented on my issue #1687 

I am able to update it within a preprocessor, just cannot find a way to reassign the current context to a new context address.

PR #1682, opened 2021-11-07
Handle changes in external files.

zvavybir:
I noticed that if you include files which are outside of the root directory of the book, than changes don't trigger rebuilds with the watch (and serve) subcommand.  I'm not really familiar with this code base, so I'm aware that it's probably sub optimal.


ehuss:
Thanks for the PR!  Sorry, though, I'm having a hard time understanding what this is doing or what the context is.  Is there an issue this is related to, such as #1222 maybe?  I think changes to the preprocessor API will require a bit of design work and discussion before heading to a PR.  There is also some more recent discussion at #1689.

zvavybir:
One of the files ([`src/english.md`](https://github.com/zvavybir/legacylisten/blob/24062b6ec728fab11226c95ace036c39260dc1c1/doc/src/english.md)) in my current project just includes some outer file:
```
{{#include ../../README.md}}
```
Usually `mdbook serve` automatically reloads if one of the files gets edited, but since the `README.md` is not in the book's root directory, it doesn't when I edit it.  This PR tries to fix this by allowing preprocessors to notify mdbook that `serve` has to look at some additional files.  Hope that clarifies it a bit.
I think this problem is unrelated to #1222 and was not mentioned in #1689, but has to be considered for the last one, since – as far as I understand it – style and script files will be outside of the root directory too.

Concerning the "design work and discussion", I think this PR would be a acceptable solution, but is ultimately just a proposal.

ehuss:
I'm going to close as this probably isn't the direction I think it should go for this issue.

Issue #1053, opened 2019-10-05
Strip HTML from ids


ad-si:
This markdown

```
## 2019-10-04 - <small>[0.2.2.0]</small>
```
yields this HTML

```
<h2>
  <a 
    class="header" 
    href="#2019-10-04---smalla-hrefhttpsgithubcomad-sitasklitereleasestagv02200220asmall" 
    id="2019-10-04---smalla-hrefhttpsgithubcomad-sitasklitereleasestagv02200220asmall"
  >
    2019-10-04 - <small></small>
  </a>
  <small>
    <a href="https://github.com/ad-si/TaskLite/releases/tag/v0.2.2.0">
        0.2.2.0
    </a>
  </small>
</h2>

```

I think the HTML should be stripped before creating the `id` and the `href`, so that it's something like:
```
#2019-10-04---v02200220
```

PR #1683, opened 2021-11-09
Strip HTML tags from anchor IDs

abdnh:
Fixes https://github.com/rust-lang/mdBook/issues/1053

This strips HTML tags before generating anchor IDs.


ehuss (approved):
👍 Thanks!

PR #1685, opened 2021-11-13
Prevent iOS Safari from enlarging text in landscape orientation

### LINKED ISSUE #1684: Text is enlarged on iOS Safari landscape orientation
Text is rendered in oddly big size on iOS Safari.

## Cause

 `-webkit-text-size-adjust` added in PR #740 -- which fixed #184 -- seems gone, even though it is still relevant.

[Mozilla CSS reference doc](https://developer.mozilla.org/en-US/docs/Web/CSS/text-size-adjust#basic_disabling_usage) suggests putting both `-webkit-text-size-adjust` and `text-size-adjust`. It also says Safari only supports vendor prefixed "-webkit-text-size-adjust" not "text-size-adjust". This applies to Firefox for iOS also.

iPadOS Safari seems not to suffer from this issue, according to these threads. -> [apple developers](https://developer.apple.com/forums/thread/123373#653685022) [webkit](https://trac.webkit.org/changeset/261940/webkit)

## Fix suggestion

We should bring back missing `-webkit-text-size-adjust: none`

igxactly:
closes #1684

### before this patch

Text gets bigger on landscape orientation.

### after this patch

Text keeps same size with portrait orientation.

## referernces
  - https://developer.mozilla.org/en-US/docs/Web/CSS/text-size-adjust#basic_disabling_usage
  - https://trac.webkit.org/changeset/261940/webkit

### DIFF
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -12,6 +12,7 @@ html {
     color: var(--fg);
     background-color: var(--bg);
     text-size-adjust: none;
+    -webkit-text-size-adjust: none;
 }
 
 body {


ehuss (approved):
Thank you!

PR #1691, opened 2021-11-20
Change book title to link to homepage when clicked on

rsapkf:


ehuss:
I'm not sure this change can be made. The menu bar is currently clickable and used to scroll to the top of the page.  Also, `/` can't be used as a destination as not all books live in the root.  Since the sidebar already provides a means of navigation, I'm going to close this out. But thanks for the PR!

Issue #1713, opened 2021-12-28
mdBook fails to open the web browser upon `mdbook build --open`, if the first link in the summary is a draft chapter (or no links exist in the summary)


LikeLakers2:
(Hope you don't mind, but I'm going to use a template similar to Godot Engine's issue template. Makes it easier to read the issue, in my opinion. :D )

**mdBook version:**
mdbook v0.4.14, installed via `cargo install mdbook`, compiled via the `stable-x86_64-pc-windows-msvc` toolchain (currently `rustc 1.57.0 (f1edd0429 2021-11-29)` as of this issue posting)

**Issue description:**
mdBook fails to open the web browser upon building the book, when building using `mdbook build --open`, if one of the following conditions are met:

* The first link in `SUMMARY.md` is a draft chapter (e.g. `[Chapter 1]()`)
* The `SUMMARY.md` file is devoid of links.

If either of these conditions are met, then mdBook returns an error upon trying to `mdbook build --open`:
```
[INFO] (mdbook): Opening web browser
[ERROR] (mdbook): Error opening web browser: IO error
```
with no further detail on what this IO error is.

I would expect mdBook to cover these cases, most likely in the following ways:

* If the first link in `SUMMARY.md` is a draft chapter, then I would expect it to look to the next chapter, and the next, and the next, etc... until it finds a valid link, which it opens the web browser to.
* If mdBook cannot find a valid link, I would expect it to return an error saying as such. For example, "Could not find a valid book page to open to!"

**Steps to reproduce:**
1. In the `SUMMARY.md` file of any book (even a freshly-created book), add a new draft chapter above all other chapters, such as `[Chapter 1]()`.
2. Attempt to build with the `--open` parameter, asking mdBook to open the resulting book afterwards.
3. Observe that mdBook will return a vague "IO error".

**Minimal reproduction project:**
[mdbook-issue-1713.zip](https://github.com/rust-lang/mdBook/files/7786196/mdbook-issue-1713.zip)

joshrotenberg:
My guess after looking for a couple minutes: with `--open` the build command opens the path explicitly to the book's `index.html` but if the chapter is a draft that file never gets created.

LikeLakers2:
@joshrotenberg I don't think that covers the whole error, though.

From some testing, it appears I missed a little bit of how mdBook works when doing `mdbook build --open`:

1. If any link to `index.md` exists in the summary, it opens the browser to that.
[...]

PR #1714, opened 2021-12-29
Check for the index.html file before trying to opener::open it

joshrotenberg:
See #1713.

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -2,6 +2,7 @@ use crate::{get_book_dir, open};
 use clap::{App, ArgMatches, SubCommand};
 use mdbook::errors::Result;
 use mdbook::MDBook;
+use std::path::Path;
 
 // Create clap subcommand arguments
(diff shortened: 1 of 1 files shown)


ehuss:
I think an error would be fine, but if you want to try opening the first non-draft chapter, I think something like the following could work:

```rust
if let Some(first) = book.iter().filter_map(|item| match item {
    BookItem::Chapter(Chapter{path: Some(path), ..}) => Some(path),
    _ => None
}).next() {
    let path = book.build_dir_for("html").join(first).with_extension("html");
    dbg!(path);
}
```

Just keep in mind that the `open` option is used in multiple commands (build, serve, watch).

ehuss:
Just checking in to see if you are still interested in completing this?

joshrotenberg:
Sorry, it fell off my list. I'll pick it back up this week.

ehuss (approved):
Thanks!

Issue #1258, opened 2020-06-20
How to embed absolute links when serving from a subdirectory in a domain?


schungx:
Say, I have a link going to another page and I need it starting from the root:

`[some url link](/path/to/page.md)`

This gets compiled to a link to the correct page:

`<a href="/path/to/page.html">some url link</a>`

However, this doesn't work if I'm serving from a subdirectory:

`http://example.com/my_book/path/to/page.html`

The absolute url will skip over the subdirectory `my_book` and point to `http://example.com/path/to/page.html` which doesn't exist.

How can we specify that links should go to an absolute path based upon the root location of the _book_?

schungx:
Right now, I solved it by using https://github.com/avitex/mdbook-tera which is a pre-processor.

I added a variable in front of every URL link.  The variable maps to the subdirectory name.

I am hoping for a simpler, not so hacky solution though.

mimoo:
see https://github.com/rust-lang/mdBook/issues/1764

engeir:
> Right now, I solved it by using https://github.com/avitex/mdbook-tera which is a pre-processor.
> 
> I added a variable in front of every URL link. The variable maps to the subdirectory name.
> 
> I am hoping for a simpler, not so hacky solution though.

@schungx Would you mind sharing exactly how you solved it in this way?

schungx:
> [@schungx](https://github.com/schungx) Would you mind sharing exactly how you solved it in this way?

Oops, sorry, just realize that I have not responded to this.  So, this is how I did it (one year too  late):

* Install `mdbook-tera`

* Define a Tera variable in a context file, e.g. `root_url = /my/book`

* In book pages, for links, use `{{ root_url }}/path/to/doc.md`

* Bob's your uncle

You need this unless your root is really the book's root.  In many sites it isn't.

schungx:
Seeing that version 0.5 is coming out...  I wonder if there is now a solution to this issue?

PR #1716, opened 2021-12-31
Use chapter.source_path for partial includes, not chapter.path

petersooley:
Partial includes are found using relative paths, but relies on the logical `path` variable rather than the actual `source_path`. 

The `Chapter::source_path` property [was introduced](https://github.com/rust-lang/mdBook/commit/7aff98a8596bd7d52a54f7729f210b53566b9f8c) to render URLs differently from how they might be on the filesystem. It's a good change that allows preprocessors to do more with the rendered URLs, but that almost always breaks the `links` preprocessor. This PR switches to using `source_path` when looking up relative partial paths.

### DIFF
--- a/src/preprocess/links.rs
+++ b/src/preprocess/links.rs
@@ -46,7 +46,7 @@ impl Preprocessor for LinkPreprocessor {
 
         book.for_each_mut(|section: &mut BookItem| {
             if let BookItem::Chapter(ref mut ch) = *section {
-                if let Some(ref chapter_path) = ch.path {
+                if let Some(ref chapter_path) = ch.source_path {
                     let base = chapter_path
                         .parent()
                         .map(|dir| src_dir.join(dir))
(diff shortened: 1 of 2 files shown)


ehuss:
Thanks!  Can you say more about what situation involves changing the `path` in such a way that breaks the links preprocessor?

Also, can you add a test for this?

psoots:
I have a preprocessor [mdbook-fs-summary](https://crates.io/crates/mdbook-fs-summary) that uses filename prefixes and indicators to generate the summary from the filesystem, but I'd like to strip these indicators from the resulting URLs. It generates `BookItem::Chapter(Chapter)` where `path` captures the intended URL path and `source_path` as the location of the file.

Other preprocessors offering similar behavior of wanting to alter the URL without losing access to the underlying filesystem paths would benefit from this. Like for #1258.

As far as tests go, I've added a new test to validate that the `Chapter::path` can be maintained differently from `Chapter::source_path` which is ultimately the API I'm inferring with this PR. Turns out this currently fails as there are still places in the overall project that use `Chapter::path` when it should probably use `Chapter::source_path`. 

I can continue pursuing this to get to a point where this test passes. Please let me know if this is the wrong direction.

Issue #1564, opened 2021-06-05
`{{#include ...}}` with indented


kynthus:
First, this ticket isn't a bug, it's more like a feature request.  

`{{#include ...}}` currently does not consider indentation when replaced.  
For example, replace some code of a control flow, struct, enum, and implementation from another file, and dealing with a code block that is indented with bullets or sequences, the second and subsequent lines are not indented.  
As a result, the HTML layout and code indentation may be corrupted.  

## Example code
````markdown
# sample.md

- Example without including files.  

  ```rust
  struct MyStruct;

  impl MyStruct {
      pub fn myfn() -> () {
          println!("MyStruct#myfn()");
      }
  }

  pub fn main() -> () {
      println!("main()");
  }
  ```

- Example using including files.  

  ```rust
  struct MyStruct;

  impl MyStruct {
      {{#include myfn.rs}}
  }

  {{#include main.rs}}
  ```
````

```rust
// myfn.rs
pub fn myfn() -> () {
    println!("MyStruct#myfn()");
}
```

```rust
// main.rs
pub fn main() -> () {
    println!("main()");
}
```

Personally, I expected that both cases with and without `{{#include}}` would be output with the same layout.  
However, the inclusion of `myfn.rs` and` main.rs` does not indent the second and subsequent lines, so the results are different as shown below.  

## Output HTML
-	Without including files.  
	```html
	<p>Example without including files.</p>
	<pre><pre class="playground"><code class="language-rust">struct MyStruct;

	impl MyStruct {
	    pub fn myfn() -&gt; () {
	        println!(&quot;MyStruct#myfn()&quot;);
	    }
	}

	pub fn main() -&gt; () {
	    println!(&quot;main()&quot;);
	}
	</code></pre></pre>
	```

-	Using including files.  
	```html
	<p>Example using including files.</p>
	<pre><pre class="playground"><code class="language-rust">
	<span class="boring">#![allow(unused)]
	</span><span class="boring">fn main() {
	</span>struct MyStruct;

	impl MyStruct {
	    pub fn myfn() -&gt; () {
	  println!(&quot;MyStruct#myfn()&quot;);
	<span class="boring">}
	</span></code></pre></pre>
	</li>
	</ul>
	<p>}
	}</p>
	<p>pub fn main() -&gt; () {
	println!(&quot;main()&quot;);
	}</p>
	<pre><code></code></pre>
	```

[...]

PR #1718, opened 2022-01-02
Keep indent when using include in code block

### LINKED ISSUE #1626: Maintain leading whitespace when including portions of a file
Given this anchor:
```
-- ANCHOR: eqNonEmpty  
instance Eq (NonEmpty a) where
  eq (NonEmpty x xs) (NonEmpty y ys) = x == y && xs == ys    
-- ANCHOR_END: eqNonEmpty
```

when included as:
```
## Exercises
1. In the the code below:
    ```haskell
    {{#include ../exercises/chapter6/test/no-peeking/Solutions.purs:eqNonEmpty}}
    ```
    we ...
```

is expanded to:
```
## Exercises
1. In the the code below:
    ```haskell
    instance Eq (NonEmpty a) where
  eq (NonEmpty x xs) (NonEmpty y ys) = x == y && xs == ys    
    ```
    we ...
```

and rendered incorrectly as:

<img src="https://user-images.githubusercontent.com/3578681/129512990-ddf2c7a7-8877-43f0-9e82-208f5f65ee17.png" width="50%">

We'd ideally like it to be rendered as:

<img src="https://user-images.githubusercontent.com/3578681/129513050-5f887a7f-93ce-4920-bae0-a9a827246ee4.png" width="50%">

which requires this markdown source:
```
## Exercises
1. In the the code below:
    ```haskell
    instance Eq (NonEmpty a) where
      eq (NonEmpty x xs) (NonEmpty y ys) = x == y && xs == ys    
    ```
    we ...
```

Things would behave as expected if any indentation before an `{{#include ...}}` block (4 spaces in this case) was also applied to _every_ line of the included file (and not just the first line of that file).

https://rust-lang.github.io/mdBook/format/mdbook.html#including-portions-of-a-file

mdbook version: `v0.4.12`

<!-- TRIAGEBOT_START -->

<!-- TRIAGEBOT_ASSIGN_START -->

<!-- TRIAGEBOT_ASSIGN_END -->
<!-- TRIAGEBOT_END -->

rchaser53:
I tried to fix issue 1626. mdBook ignores the indent in code block created by #include. I think this is the reason why issue 1626 happened. So I implemented to add the offset to the code block created by #include.
fix: https://github.com/rust-lang/mdBook/issues/1626


rchaser53:
I think this PR also fixes #1564. Could you review this PR? @ehuss
Many users suffer from this bug.
スクリーンショット 2022-01-05 13 47 03

onbjerg:
Writing a book currently where I ran into the issue that is resolved in this PR. Are there any blockers for this to get merged?

ISSOtm (approved):

src/preprocess/links.rs
```diff
@@ -89,19 +89,26 @@ where
...
-                            depth + 1,
-                            chapter_title,
-                        ));
+                        let v =
+                            replace_all(&new_content, rel_path, source, depth + 1, chapter_title);
+                        let lines = v.split('\n').into_iter().collect::<Vec<&str>>();
+                        // no need to add offset for the first line
+                        replaced.push_str(lines[0]);
+                        for line in lines.iter().skip(1) {
+                            replaced.push_str(&format!("\n{}{}", &offset, &line));
```
mgeisler, line 110: You could consider changing this to

```rust
                            replaced.push('\n');
                            replaced.push_str(&offset);
                            replaced.push_str(&line);
```
since you're simply concatenating strings.

src/preprocess/links.rs
```diff
@@ -89,19 +89,26 @@ where
...
-                        replaced.push_str(&replace_all(
-                            &new_content,
-                            rel_path,
-                            source,
-                            depth + 1,
-                            chapter_title,
-                        ));
+                        let v =
+                            replace_all(&new_content, rel_path, source, depth + 1, chapter_title);
+                        let lines = v.split('\n').into_iter().collect::<Vec<&str>>();
```
mgeisler, line 106: Since you're iterating over the lines one-by-one, you could iterate directly over the result of `split`. So you would save the `collect()` call be a tiny bit more efficient.

GuillaumeGomez:
It was fixed in https://github.com/rust-lang/mdBook/pull/3196. Thanks for your work in any case!

PR #1719, opened 2022-01-02
Update some deps

klensy:
replace `chrono` by `time` v0.3 crate to mitigate RUSTSEC-2020-0159
`headers` v0.3.4 -> v0.3.5 to drop `time` 0.1 dep. Now only `xml5ever` depends on it, with unreleased patch
`pretty_assertions` v0.6 -> v0.7
`clap` v2.33.3 -> v2.34.0 to update to `ansi_term` 0.12
`env_logger` v0.7 -> v0.8 drops `quick-error` 1.2.3


klensy:
Err, MSRV noted nowhere on repo page, but only checked on PR. Maybe add MSRV badge and/or note about it somewhere?

ehuss:
Ah, the documentation hasn't been published, yet, but the minimum version is documented here: https://github.com/rust-lang/mdBook/blob/master/guide/src/guide/installation.md

Can you determine what the new minimum is?  It's fine to update it as long as it isn't too new.

klensy:
Hm, didn't looked at [this](https://github.com/rust-lang/mdBook/blob/master/guide/src/guide/installation.md), looked around title page and https://github.com/rust-lang/mdBook/blob/master/CONTRIBUTING.md

With `time` 0.3.2 builds with rustc 1.48.0
With `time` 0.3.3+ (0.3.5 was used initially in this PR) builds with rustc 1.51.0

ehuss:
I think 1.51 should be fine, which is over 9 months old.

Can you also add a comment in `main.yml` to point to the documentation to update so they don't get out of sync?

ehuss:
This looks like it has some test errors.  Are you still interested in looking into it?

Issue #1190, opened 2020-04-14
Option for page breaks while printing


bamhm182:
##  What is Happening

When printing, multiple chapters can print on one page. This makes it a little difficult to find the beginning of new chapters as in some cases, the chapter may start at the bottom of a page.

[no_breaks.pdf](https://github.com/rust-lang/mdBook/files/4474913/no_breaks.pdf)

## What is Expected

As in real printed books, when a book is printed, the start of a new chapter means the start of a new page.

[breaks.pdf](https://github.com/rust-lang/mdBook/files/4474912/breaks.pdf)

## Proposed Resolution

The code `<p style="page-break-before: always"/>` can be added to the end of every chapter while on the `print.html` page in order to allow for the next chapter to be printed on a new page.

I realize this may not be ideal for all cases, so I also propose an option like `[page-break-between-chapters] = true` be available in the `book.toml`

morberg:
As a temporary workaround I've added:

```css
h1 {
    page-break-before: always;
}
```

to `print.css`.

Kinrany:
Is there a way to revert this? For a "book" with small chapters it's better to print on the same page.

I looked at the theme files, but the only mention of `page-break-before` is in `highlight.js`.

ehuss:
Sure, I think that can be added.  I'm a bit surprised that is something that would be desired, but if you'd like, you can post a PR that reverts https://github.com/rust-lang/mdBook/pull/1485/commits/0eb23efd44c96d79c96161648a6c286dbc7424c1 to add the option back in.

Kinrany:
My use case is using mdbook as lazy way to have a personal webpage and a CV in one 🙂

dmorawetz:
> Sure, I think that can be added. I'm a bit surprised that is something that would be desired, but if you'd like, you can post a PR that reverts [0eb23ef](https://github.com/rust-lang/mdBook/commit/0eb23efd44c96d79c96161648a6c286dbc7424c1) to add the option back in.

@ehuss I posted a PR to revert the change, #1728. Can you please have a look?

bitdivine:
Just curious: It feels to me as if CSS is the right place to specify things like line breaks, no?

PR #1728, opened 2022-01-17
Make page-break configurable

dmorawetz:
As discussed in #1190, the option to disable page-breaks between chapters is useful for some specific use-cases. I use mdBook for printable cheat sheets, and therefore want to keep the printable version as compact as possible.


ehuss (approved):
Thanks!

PR #1731, opened 2022-01-19
Port mdBook to clap3

epage:
This includes resolving the deprecation warnings, one warning per commit.  This was a fairly straightforward conversion, sticking to simpler parts of clap.  `App::arg_from_usage` was probably the most involved.  The replacement, `arg!()`, doesn't seem to support kebab-case in value names.

I tested by
- Adding the `App::debug_assert()` test
- Inspecting before/after of `--help` for each subcommand

I'm assuming there is some end-to-end tests to at least catch egregious cases.

Differences I noticed (all cosmetic)
- Colored help is now on all subcommands (`ColoredHelp` was set only on the parent before rather than with `global_setting`, now its the default)
- Combining of the Flags and Options sections (`UnifiedHelpMessage` is now the only behavior)
- Some spacing is slightly different so more lines of text wrap.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,7 +31,8 @@ jobs:
             rust: stable
           - build: msrv
             os: ubuntu-latest
-            rust: 1.46.0
+            # sync MSRV with docs: guide/src/guide/installation.md
+            rust: 1.54.0
     steps:
     - uses: actions/checkout@master
     - name: Install Rust
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -185,17 +185,27 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "2.33.3"
+version = "3.0.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "37e58ac78573c40708d45522f0d80fa2f01cc4f9b4e2bf749807255454312002"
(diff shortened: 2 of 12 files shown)


epage:
mdBook's MSRV is at 1.46.0.  clap is at 1.54.  We are open to experimenting with lowering the MSRV (https://github.com/clap-rs/clap/issues/3267) though it isn't a priority for us.

Should we up mdBook's MSRV or hold off on this until either (1) clap's lowers or (2) mdBook is ready to bump?

ehuss:
I think we can probably bump the MSRV.  We don't have a fixed policy, but I think 8 months is probably ok.  Does 3.1.6 still work on 1.54?

ehuss:
Can this also update https://github.com/rust-lang/mdBook/blob/46345b8e4925e2cbd43d58f5508c1d7fdb601035/guide/src/guide/installation.md?plain=1#L23?

ehuss:
Oh, and maybe add a comment in main.yml to keep them in sync (that was added in #1719, but that PR seemed to have stall).

epage:
Updated this with the MSRV bump with the comment linking the locations.

And yes, clap is still on 1.54.  We had a request to extend our MSRV to a 6 month period.  We are doing this on a trial basis

ehuss (approved):
Thanks!

PR #1732, opened 2022-01-23
Update semver dev-dependency in nop-preprocessor example to 1.0

dtolnay:
The API involved in the example is the following, none of which has changed between 0.11 and 1.0.

https://github.com/rust-lang/mdBook/blob/336640633b1741de644e971a1a2cd8b30c595368/examples/nop-preprocessor.rs#L6

https://github.com/rust-lang/mdBook/blob/336640633b1741de644e971a1a2cd8b30c595368/examples/nop-preprocessor.rs#L37-L40

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1454,21 +1454,9 @@ dependencies = [
 
 [[package]]
 name = "semver"
-version = "0.11.0"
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thank you!

PR #1741, opened 2022-02-09
add a --chapter option to mdbook test.

clovett:
Sometimes when working on large books it is handy to be able to run mdbook on a single chapter of a book.  For example:
````
mdbook test --chapter build

mdbook test --chapter build
2022-02-09 13:48:17 [INFO] (mdbook::book): Skipping chapter 'Introduction'...
2022-02-09 13:48:17 [INFO] (mdbook::book): Skipping chapter 'Installation'...
2022-02-09 13:48:17 [INFO] (mdbook::book): Skipping chapter 'Reading Books'...
2022-02-09 13:48:17 [INFO] (mdbook::book): Skipping chapter 'Creating a Book'...
2022-02-09 13:48:17 [INFO] (mdbook::book): Skipping chapter 'Command Line Tool'...
2022-02-09 13:48:17 [INFO] (mdbook::book): Skipping chapter 'init'...
2022-02-09 13:48:17 [INFO] (mdbook::book): Testing chapter 'build': "d:\\git\\clovett\\mdBook\\guide\\src\\cli/build.md"
2022-02-09 13:48:17 [INFO] (mdbook::book): Skipping chapter 'watch'...
2022-02-09 13:48:17 [INFO] (mdbook::book): Skipping chapter 'serve'...
2022-02-09 13:48:17 [INFO] (mdbook::book): Skipping chapter 'test'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'clean'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'completions'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'Format'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'SUMMARY.md'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'Configuration'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'General'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'Preprocessors'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'Renderers'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'Environment Variables'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'Theme'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'index.hbs'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'Syntax highlighting'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'Editor'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'MathJax Support'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'mdBook-specific features'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'Markdown'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'Continuous Integration'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'For Developers'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'Preprocessors'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'Alternative Backends'...
2022-02-09 13:48:18 [INFO] (mdbook::book): Skipping chapter 'Contributors'...
```


src/book/mod.rs
```diff
@@ -270,8 +271,13 @@ impl MDBook {
                     _ => continue,
                 };
 
+                if !chapter.is_empty() && ch.name != chapter {
```
ehuss, line 274: Can this be an `Option<&str>` instead of using an empty string as a sentinel?

src/book/mod.rs
```diff
@@ -270,8 +271,13 @@ impl MDBook {
...
+                    info!("Skipping chapter '{}'...", ch.name);
```
ehuss, line 275: I'm think it might be better to make this `debug!`.  In a larger book, it will flood the console with a bunch of skipping messages that obscure what it is doing.

src/theme/index.hbs
```diff
@@ -67,11 +67,11 @@
                 var theme = localStorage.getItem('mdbook-theme');
                 var sidebar = localStorage.getItem('mdbook-sidebar');
 
-                if (theme.startsWith('"') && theme.endsWith('"')) {
+                if (theme && theme.startsWith('"') && theme.endsWith('"')) {
                     localStorage.setItem('mdbook-theme', theme.slice(1, theme.length - 1));
                 }
 
-                if (sidebar.startsWith('"') && sidebar.endsWith('"')) {
+                if (sidebar && sidebar.startsWith('"') && sidebar.endsWith('"')) {
```
ehuss, line 74: These changes seem unrelated to the PR?

ehuss (commented):
Thanks!  This seems like it could be useful.

I'd like to suggest that perhaps the `-c` flag could take either a chapter name or file path.  In most cases, I know the file path but not the chapter name.  File paths also support command-line completion, whereas chapter names need to be typed exactly (and deal with shell quoting and such).

Can you also update the documentation at https://github.com/rust-lang/mdBook/blob/master/guide/src/cli/test.md?

src/theme/index.hbs
lovettchris: they were null reference bugs I ran into while testing this PR.

src/book/mod.rs
lovettchris: Done.

src/book/mod.rs
lovettchris: Done.

lovettchris:
> I'd like to suggest that perhaps the `-c` flag could take either a chapter name or file path. In most cases, I know the file path but not the chapter name. File paths also support command-line completion, whereas chapter names need to be typed exactly (and deal with shell quoting and such).

Done.

ehuss:
@clovett Just checking in if you are still interested in this. It looks like there are some test failures.

lovettchris:
I am still interested in this PR, but I'm kinda stumped by the test failures.  Clearly the test output changes, and I tried to fix that, but I'm not sure why it didn't work.

ehuss:
The PR changes the output to use `chapter_path` instead of `path`.  `path` is absolute whereas `chapter_path` is relative. That means that the pattern `Testing chapter [^:]*: "([^"]+)[\\/]README.md` needs to be updated, as there is no leading slash in front of the `README.md` file.

lovettchris:
> The PR changes the output to use `chapter_path` instead of `path`. `path` is absolute whereas `chapter_path` is relative. That means that the pattern `Testing chapter [^:]*: "([^"]+)[\\/]README.md` needs to be updated, as there is no leading slash in front of the `README.md` file.

Ah, thanks for the hint - I think I got it now.

clovett:
@ehuss - yay the tests are passing.

src/book/mod.rs
```diff
@@ -244,8 +244,9 @@ impl MDBook {
         self
     }
 
-    /// Run `rustdoc` tests on the book, linking against the provided libraries.
-    pub fn test(&mut self, library_paths: Vec<&str>) -> Result<()> {
+    /// Run `rustdoc` tests on the book, linking against the provided libraries
+    /// and optionally testing only a signal named chapter while skipping the others.
```
ehuss, line 248:
```suggestion
    /// and optionally testing only a single named chapter while skipping the others.
```

src/book/mod.rs
```diff
@@ -244,8 +244,9 @@ impl MDBook {
...
+    pub fn test(&mut self, library_paths: Vec<&str>, chapter: Option<&str>) -> Result<()> {
```
ehuss, line 249: Unfortunately we can't change the API as that would be a semver breaking change.

I think you will need to add a new method, and have the old `test` call it.

src/book/mod.rs
```diff
@@ -270,8 +271,15 @@ impl MDBook {
                     _ => continue,
                 };
 
-                let path = self.source_dir().join(&chapter_path);
-                info!("Testing file: {:?}", path);
+                if chapter.is_some()
+                    && ch.name != chapter.unwrap()
+                    && chapter_path.to_str().unwrap() != chapter.unwrap()
```
ehuss, line 276: I'd prefer to avoid using unwrap here.  Perhaps more like this:

```suggestion
                if let Some(chapter) = chapter {
                    if ch.name != chapter && chapter_path.to_str() != Some(chapter) {
```

However, this logic seems wrong. It seems to require that the `--chapter` value passed in equals *both* the chapter name and its filename. I would expect it to match one or the other. Perhaps something more like?

```rust
!(ch.name == chapter || chapter_path.to_str() == Some(chapter))
```

src/theme/index.hbs
ehuss: I would prefer unrelated changes to not be included.

ehuss (commented):
I think this will need some tests.

Also, can you add some kind of error message if the `--chapter` value doesn't match anything?

src/theme/index.hbs
lovettchris: this is a bit different as I couldn't get the tests to pass without this.

src/book/mod.rs
lovettchris: fixed

src/book/mod.rs
lovettchris: done, I added "test_chapter", but instead of duplicating the code I just make test call test_chapter.

src/book/mod.rs
lovettchris: it works, notice this is the inverse of the test `!(ch.name == chapter || chapter_path == chapter)`

lovettchris:
> Can you also update the documentation at https://github.com/rust-lang/mdBook/blob/master/guide/src/cli/test.md?

I added:


#### --chapter

The `--chapter` (`-d`) option allows you to test a specific chapter of the
book using the chapter name or the relative path to the chapter.

src/book/mod.rs
```diff
@@ -270,8 +282,18 @@ impl MDBook {
...
 
-                let path = self.source_dir().join(&chapter_path);
-                info!("Testing file: {:?}", path);
+                let cp = match chapter_path.to_str() {
+                    Some(s) => s,
+                    None => "",
+                };
+
+                if chapter_name != "" && ch.name != chapter_name && cp != chapter_name {
+                    info!("Skipping chapter '{}'...", ch.name);
```
ehuss, line 291: This all seemed to have reverted back to using an empty string as a sentinel instead of an Option. Is there any particular reason for that?

src/book/mod.rs
```diff
@@ -246,6 +246,11 @@ impl MDBook {
 
     /// Run `rustdoc` tests on the book, linking against the provided libraries.
     pub fn test(&mut self, library_paths: Vec<&str>) -> Result<()> {
+        self.test_chapter(library_paths, None)
+    }
+
+    /// Run `rustdoc` tests on a specific chapter of the book, linking against the provided libraries.
```
ehuss, line 252: This is a little unclear to me on what happens if `chapter` is None. Can it include a little more explanation?

src/theme/index.hbs
ehuss: Which tests aren't passing? What kind of errors are you seeing? I don't believe there are any javascript tests, so it is not clear to me how this could have any impact on anything.

src/book/mod.rs
lovettchris: Sure thing, I added clarifying comments `// test_chapter with chapter:None will run all tests.` and on the `test_chapter` function itself the doc comment: `/// If `chapter` is `None`, all tests will be run.`

src/theme/index.hbs
lovettchris: Hmmm, trying to repro it I can't find it any more and I see there's a `try{ } catch (e) { }` around this whole thing, so it's possible I caught the exception in the debugger and mistook it for a bug when it was expecting the try/catch block to catch it and continue.  So I reverted these changes.

src/book/mod.rs
lovettchris: sorry this might be my Rust ignorance, I couldn't figure out how to compare `chapter : Option<&str>` and `ch.name: String`.  Rust complains: "can't compare `std::string::String` with `std::option::Option<&str>`".  So, moving them both to &str solved the problem, but I guess I could instead convert `ch.name` to `Option<&str>` using:
```
let cn: Option<&str> = Some(&ch.name);
```
I pushed this change to see if this is what you prefer...

lovettchris:
@ehuss I think I resolved all the excellent issues you raised.

src/book/mod.rs
```diff
@@ -270,8 +279,16 @@ impl MDBook {
...
 
-                let path = self.source_dir().join(&chapter_path);
-                info!("Testing file: {:?}", path);
+                let cp = chapter_path.to_str();
+                let cn: Option<&str> = Some(&ch.name);
+
+                if chapter.is_some() && cn != chapter && cp != chapter {
+                    info!("Skipping chapter '{}'...", ch.name);
+                    continue;
+                };
```
ehuss, line 288: Instead of using local variables like this, I still think it would be clearer to use something like the suggestion I posted before. The local variables create an indirection where you then have to refer to other parts of the code to figure out what they mean. Using the values directly I think makes it a little clearer as to what is being checked. 

Also, after testing this out a bit, I feel like the output is overwhelmed with the "skipping chapter" text. I'm thinking it might be better to reduce that to a `debug` message to reduce the noise.

Perhaps something like this:

```suggestion
                if let Some(chapter) = chapter {
                    if ch.name != chapter && chapter_path.to_str() != Some(chapter) {
                        debug!("Skipping chapter '{}'...", ch.name);
                        continue;
                    }
                }
```

src/book/mod.rs
```diff
@@ -311,6 +328,10 @@ impl MDBook {
         if failed {
             bail!("One or more tests failed");
         }
+
+        if chapter.is_some() && !chapter_found {
+            bail!(format!("Chapter not found: {:?}", chapter));
+        }
```
ehuss, line 334: This produces some strange output. Generally it is recommended to not use debug displays for user output. Also, the `bail!` macro can take formatted strings directly, there is no need to use `format!`. I think it might be better to extract the name through a pattern and use a display impl. Something like this:

```suggestion
        if let Some(chapter) = chapter {
            if !chapter_found {
                bail!("Chapter not found: {}", chapter);
            }
        }
```

ehuss (commented):
Thanks! Just some code style suggestions.

src/book/mod.rs
lovettchris: The `skipping chapter` prompts are how you find the chapter names so you can write `mdbook test -c ?` for example then you find the chapter name you want to test and run the specific command.  But I could make it so that it looks for the specific string "?" so it only prints in that case.

src/book/mod.rs
lovettchris: Much better thanks, I merged your fix.  I was just copying the use of `bail` from a few lines earlier: `bail!("One or more tests failed");`, so unless there is a better error formatter available, I'll leave the bail as you have written it.

src/book/mod.rs
lovettchris: I pushed the if-check looking for command line arg "?" instead and I added that to the help text to make it discoverable.

lovettchris: Oh, and I see now how you are using `if let Some` to remove the need for local variables, so I pushed that fix too.

ehuss (approved):
Thanks!

Issue #720, opened 2018-07-04
Search does not handle headers with special formatting.


ehuss:
When headers have special formatting, the search results do not have the correct # anchor links.

Example:
```### The `[profile.*]` sections```

has an anchor hash of `#the-profile-sections` but the search index uses a hash of `#the--profile--sections` (notice the extra dashes).

It also does not handle duplicate header names.

I looked at it briefly.  The issue is that the anchor id is generated from the HTML content ([here](https://github.com/rust-lang-nursery/mdBook/blob/6287e6a44fc5a4039e859f949bfaf7ac7108272b/src/renderer/html_handlebars/hbs_renderer.rs#L517-L541)), and has special logic for striping out certain tags such as `<code>`.  However, the search code re-parses the content, and replaces all tags with a single space ([here](https://github.com/rust-lang-nursery/mdBook/blob/6287e6a44fc5a4039e859f949bfaf7ac7108272b/src/renderer/html_handlebars/search.rs#L116-L124)).  

I'm not sure what the best solution is.  Some options I can think of:
* Don't insert a space in the header for `Event`s.  I'm not sure which scenarios that would cause problems, but the common inline tags (code, strong, em) are normally surrounded by spaces already.
* Check for `Event`s that match the same list in [`id_from_content()`](https://github.com/rust-lang-nursery/mdBook/blob/6287e6a44fc5a4039e859f949bfaf7ac7108272b/src/utils/mod.rs#L46-L69) and not insert a space in `header` in those situations.
* Collapse duplicate whitespace in `normalize_id` into a single whitespace.

mattico:
Hello! Thanks for the detailed issue.

It seems to me that we should generate the appropriate anchor for a section in one place and pass it around so that we don't end up with different anchors. This is somewhat related to the anchor changes proposed in #707, perhaps an extension of that method could solve this issue.

I should be able to work on this soon, but would also be happy to mentor if you're interested in solving it.

ehuss:
Note: The example listed above was fixed some time before 0.3.0.  I think it was due to a change in pulldown_cmark which trimed whitespace from the text strings.  However, the way the IDs are generated are still different codepaths with subtle differences.  The biggest remaining issue is that duplicate IDs are not handled.

Issue #1579, opened 2021-06-15
headings differing only in case are linked wrongly in search results


simonmichael:
https://github.com/badboy/mdbook-toc/issues/16 seems to be an mdBook issue. At https://hledger.org/hledger.html I have "tags" and "Tags" sections. This is almost working now, except that in https://hledger.org/hledger.html?search=tags, both results link to the "tags" section. [The first result, "Tags", should link to <https://hledger.org/hledger.html?highlight=tags#tags-1>]

[cc <https://github.com/simonmichael/hledger_site/issues/33>]

ehuss:
This is mentioned in #720 (duplicate header names don't use the same logic).  It would be nice if they could use the same code.

ehuss:
Closing as resolved by #1749.

PR #1749, opened 2022-02-18
search: fix anchor ids for duplicate headers

tommilligan:
Closes: #720, #1579

Currently, if one page contains two sections with the same header, the generated anchor IDs will be uniquely enumerated:

```html
<h3 id="my-title">
  <a class="header" href="#my-title">My Title</a>
</h3>
...
<h3 id="my-title-1">
  <a class="header" href="#my-title-1">My Title</a>
</h3>
```

However, **search** doesn't match the way it generates ids from header content, and doesn't enumerate them. This leads to all search results for a given section header, linking to the first section header with that content on the page.

---

This PR factors out unique id generation for headers to a new function, `unique_id_from_content`. The existing function `id_from_content` is publicly exported and so cannot be deleted - it has been marked as deprecated from the next patch version `0.4.16`.

Matching tests have been added, and I have verified that this works for my specific use case locally.

Let me know if there's anything that needs changing - thanks very much for maintaining this project!


ehuss (commented):
Very nice, thanks!

PR #1755, opened 2022-02-23
Add --verbose logging and --extern flags to test subcommand

rickwebiii:
To run tests against an external crate, I find I need to specify the --extern argument to pass down to rustdoc. This changeset adds this.

I also added --verbose to print the full rustdoc command line and pass --verbose to the rustdoc child command.


rickwebiii:
New commit, ran `cargo fmt` and added a warning if edition not specified in book.toml.

I (as I would expect most users to be) was baffled by `missing crate x` errors despite correct -L and --extern flags. After manually playing with rustdoc I found I could repro the errors if I elided or set the edition to 2015. The default edition for rustdoc is 2015, which almost nobody uses anymore.

My missing crate errors were because my code didn't call `extern crate x;`, which was required in Rust 2015, but hasn't been required since 2018.

rickwebiii:
Bump rust to stable 1.58 to use `.get_args()` on `std::process::Command` for `--verbose` flag.

PR #1759, opened 2022-02-26
Open up support for alternative playground languages

vanceism7:
This PR opens up the ability for the playground to run code for languages other than rust. 

It adds an option to set which language is "playground-able" in `book.toml`, and then uses that config option to choose whether the html `pre` blocks and friends should be wrapped around that code. Overall, it's a pretty minor change, but this allows others languages with playgrounds (e.g: purescript) to have active code blocks in mdbook. 

(It looks like the remainder of the heavy lifting to implement the playground for other languages rests in the hands of an alternative `book.js` via `mdbook init --theme`.)

Related to #350 and purescript-contrib/purescript-book#91

Im happy to make changes if anything seems wrong. Thanks!

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -631,7 +631,7 @@ pub struct Playground {
     /// Display line numbers on playground snippets. Default: `false`.
     pub line_numbers: bool,
     /// Set's the language the playground will work with
-    /// TODO: Turn this into an array when there's support for multiple languages
+    /// TODO: Use an array when there's support for multiple languages simultaneously
     pub language: String,
 }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -630,6 +630,9 @@ pub struct Playground {
     pub copy_js: bool,
     /// Display line numbers on playground snippets. Default: `false`.
     pub line_numbers: bool,
+    /// Set's the language the playground will work with
+    /// TODO: Turn this into an array when there's support for multiple languages
+    pub language: String,
 }
 
 impl Default for Playground {
@@ -639,6 +642,7 @@ impl Default for Playground {
             copyable: true,
             copy_js: true,
             line_numbers: false,
+            language: "rust".to_string(),
         }
     }
 }
@@ -748,6 +752,7 @@ mod tests {
         [output.html.playground]
         editable = true
         editor = "ace"
+        language = "rust"
 
         [output.html.redirect]
         "index.html" = "overview.html"
@@ -781,6 +786,7 @@ mod tests {
             copyable: true,
             copy_js: true,
             line_numbers: false,
+            language: "rust".to_string(),
         };
         let html_should_be = HtmlConfig {
             curly_quotes: true,
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -824,62 +824,76 @@ fn add_playground_pre(
             let text = &caps[1];
             let classes = &caps[2];
             let code = &caps[3];
+            let lang_class = format!("language-{}", playground_config.language);
 
-            if classes.contains("language-rust") {
-                if (!classes.contains("ignore")
-                    && !classes.contains("noplayground")
-                    && !classes.contains("noplaypen"))
-                    || classes.contains("mdbook-runnable")
-                {
-                    let contains_e2015 = classes.contains("edition2015");
-                    let contains_e2018 = classes.contains("edition2018");
-                    let contains_e2021 = classes.contains("edition2021");
-                    let edition_class = if contains_e2015 || contains_e2018 || contains_e2021 {
-                        // the user forced edition, we should not overwrite it
-                        ""
-                    } else {
-                        match edition {
-                            Some(RustEdition::E2015) => " edition2015",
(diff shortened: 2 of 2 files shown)


mgeisler:
I'm just a random user who is also interested in this :smiley: 

I think this looks like a nice first step, but would it not be necessary to also override the hard-coded `play.rust-lang.org` strings in `book.js`? Or perhaps you wanted to make this as a follow-up change?

mgeisler:
> Having to manually modify book.js isnt ideal, but with this PR, atleast it gives determined users the ability to actually do it successfully

Okay, thanks, that makes sense to me!

hhstore:
- This PR is great.
- Can a set of other languages ​​be built in? Such as golang, typescript and dart. This is convenient to use. (Of course, it is also allowed to specify custom playgrounds)
- `mdbook` is a very handy project for writing tutorials for beginners.
-I just want to use `mdbook` to write some `flutter/golang` tutorials. Use playground to run examples.

> some online playgrounds:

- [x] golang: https://go.dev/play/
- [x] dart: https://dartpad.dev/?
- [x] typescript: https://www.typescriptlang.org/play
- [x] more others: https://fiddles.io/

vanceism7:
@ehuss Hey, I know this PR is a bit out of date now, but just wondering if we could get a review on this. Thanks!

Dylan-DPC:
@vanceism7 can you resolve the conflicts? we can look at getting it merged after that

vanceism7:
@Dylan-DPC Updated. Thanks!

vanceism7:
As mdbook currently is, one can create their own `book.js` using `mdbook init --theme, and modify it to use a different playground url, but mdbook will still forcibly disable the ability to run any code snippet that isn't explicitly rust. The intention of this PR is that we don't disable code snippets just because they're not using rust. It doesn't explicitly enable support for alternative playgrounds, it just opens up the ability to do it, instead of making it forbidden as is currently the case. 

I think ideally, it'd be nice if we could add another config field alongside the `language` field, that allows the user to specify what playground url they want to use, and have that propagate within book.js without needing to create a custom theme. But IIRC, `book.js` also has a lot of rust specific stuff in it, so I don't think enabling alternative backend support would be that trivial. This atleast gets the ball rolling in that direction though, and technically adds the ability for anyone already using mdbook to make an alternative language work.

ehuss:
If you are already modifying the javascript, would it be possible to just change the query to modify the code elements with your language (instead of looking for `.playground`)? 

I would prefer to not add an option that is as low-level as "add `playground` class to this one language".  Ideally I'd like to see a general strategy of how alternate languages could be supported. It is OK to make incremental progress towards that goal. 

Does https://try.purescript.org/ have a relatively simple API for sending a blob of code and getting the output?

vanceism7:
Hmm, I see what you're saying. I think in a world where each mdbook site only supports one runnable language, this change makes sense in the sense that `book.js` sort of standardizes itself on looking for code snippets with the `.playground` class, rather than having every language write queries for random classes in its own custom `book.js`. But if we wanted to have a general way for mdbook to run multiple languages at once, this change probably isn't very satisfactory currently. 

If we wanted to run multiple languages at once, you'd probably want to change the `languages` config field to an array of strings, and have users list all the languages they want to be runnable. Then maybe you'd tag each code snippet with `playground-{your-language}`. But then again, maybe that's not even needed because you can just query for `language-{your-language}`. 

Maybe the right way to general alternative backends is to (somehow) get a list of runnable languages to call `querySelectorAll` on, and then for each language in that list, we have an action we run that hits that languages playground and handles the response for it.

So that we go from this:
```javascript
(function codeSnippets() {
    ...

    var playgrounds = Array.from(document.querySelectorAll(".playground"));
    if (playgrounds.length > 0) {
        fetch_with_timeout("https://play.rust-lang.org/meta/crates", {
            ...
        })
        .then(response => response.json())
        .then(response => {
            // get list of crates available in the rust playground
            let playground_crates = response.crates.map(item => item["id"]);
            playgrounds.forEach(block => handle_crate_list_update(block, playground_crates));
        });
    }
```

to something like this (pseudo code)
```javascript
(function codeSnippets() {
    ...
    foreach( var playgroundClass in playgroundsList ) {
      var playgrounds = Array.from(document.querySelectorAll(playgroundClass));
      if (playgrounds.length > 0) {
        var env = playgroundRunners[playgroundClass].initializeRunner(); // Something like grabbing rust crates could go here
        playgrounds.forEach(block => playgroundRunners[playgroundClass].eval(block, env);
    }
```

Then to get alternative backends in, we'd probably just need contributions of js modules from the community that handle connecting to that languages playground. Something like a `Rust.js`, `Purescript.js` and whatever else could exist as small modules that just export the two functions `initializeRunner` and `eval`. Or the user just throws `<mylanguage>.js` into their own book and the book theme will read it automatically. 


Well, thinking that all through, we probably want to close this PR now 😅 

> Does https://try.purescript.org/ have a relatively simple API for sending a blob of code and getting the output?
Yea, try purescript has a pretty simple api. Its actually `compile.purescript.org`. I think you just `POST` to it with the body being a blob of code and it spits back a response at you
https://github.com/purescript/trypurescript#server-api

Thanks for taking the time!

PR #1763, opened 2022-03-09
update regex

Dylan-DPC:
Update regex to `1.5.5` due to [security advisory](https://blog.rust-lang.org/2022/03/08/cve-2022-24713.html)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1390,9 +1390,9 @@ dependencies = [
 
 [[package]]
 name = "regex"
-version = "1.5.4"
+version = "1.5.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


ehuss:
It looks like this updated the entire Cargo.lock.  Was that intended?

Dylan-DPC:
yeah i ran a complete `cargo update` though i guess i could just update regex

ehuss:
I think a full update sounds fine as long as it isn't an API break, or it doesn't update MSRV too much.  May want to update the title if that's what you want to do.

Dylan-DPC:
changed it to only regex. Will do the others in a separate pr

ehuss (approved):
Sounds good, thanks!

PR #1772, opened 2022-03-19
add option to enable inline math $…$

SichangHe:
# add option to enable inline math `$…$`

We only support `\\(…\\)` for inline math right now. This PR adds an option to enable using `$…$` for inline math.

## usage

add this following to `book.toml` to enable this feature:

```toml
[output.html]
mathjax-support = true
inline-math-dollar-sign = true
```

### unwanted usage

If the user enables this option without enabling MathJax support, nothing happens:

```toml
[output.html]
inline-math-dollar-sign = true
```

## benefits

This will be helpful for writers that want to use a lot of inline math in their books, especially if it is a Mathematics book.

```text
In equation [1], $x$ is …, $y$ is …
```

Looks much more friendly than:

```text
In equation [1], \\(x\\) is …, \\(y\\) is …
```

It is also much easier to type.

Some extensions (for example, Markdown All in One for VSCode) gives a keyboard shortcut to add two `$$` and put your cursor in the middle. This is very convenient.

## drawback

If this option is enabled, the user will need to escape all their regular dollar signs:

```test
\$
```

Therefore, this option is set to `false` by default.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -493,6 +493,8 @@ pub struct HtmlConfig {
     pub curly_quotes: bool,
     /// Should mathjax be enabled?
     pub mathjax_support: bool,
+    /// Should mathjax enable `$…$` inline math? This option has no effect if mathjax is not enabled.
+    pub inline_math_dollar_sign: bool,
     /// Whether to fonts.css and respective font files to the output directory.
     pub copy_fonts: bool,
     /// An optional google analytics code.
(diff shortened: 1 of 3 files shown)


SichangHe:
This makes no sense to me now because anyone that writes a lot of math with mdbook should probably use the [mdbook-katex](https://github.com/lzanini/mdbook-katex) preprocesser unless mdbook has proper `\` support in math. The MathJax support in mdbook is extremely clunky because all the `\` needs to be written as `\\`, which makes it incompatible with any other markdown math renderer and makes the code look bad.

The takeaway is probably that we need a "math" section in the mdbook book that tells the user that they should opt in mdbook-katex for a better math support, or we should include an easy way to use their features.

PR #1777, opened 2022-03-31
tests/cli: ignore user's RUST_LOG= environment variable in tests

trofi:
nixpkgs build system sets `RUST_LOG=` (empty value) by default.
This switches `mdBook` into warnings+ mode (instead of info+).

This causes the following tests to fail:

    $ RUST_LOG= cargo test --test cli_tests
    ...
    cli::test::mdbook_cli_can_correctly_test_a_passing_book
    cli::test::mdbook_cli_detects_book_with_failing_tests
    cli::build::mdbook_cli_dummy_book_generates_index_html

The change drops RUST_LOG= entry.

### DIFF
--- a/tests/cli/build.rs
+++ b/tests/cli/build.rs
@@ -1,6 +1,5 @@
+use crate::cli::cmd::mdbook_cmd;
 use crate::dummy_book::DummyBook;
-
-use assert_cmd::Command;
 
 #[test]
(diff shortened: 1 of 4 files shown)


trofi:
Applied `cargo fmt`.

trofi:
Moved environment helper setting out to `mdbook_cmd()`.

ehuss (approved):
Thanks!

PR #1791, opened 2022-04-24
bail! in render() if specified theme directory does not exist

### LINKED ISSUE #1756: Theme directory not being read
Hi. I'm trying to customize the Ayu theme, add additional fonts, and update the favicon on my mdbook instance.
However, it seems that the files aren't copied in place, and instead mdbook just spawns the default fonts and CSS files.

My directory structure looks like this:
```
├── book.toml
├── src
└── theme
    ├── book.js
    ├── css
    │   ├── chrome.css
    │   ├── general.css
    │   ├── print.css
    │   └── variables.css
    ├── favicon.png
    ├── favicon.svg
    ├── highlight.css
    ├── highlight.js
    └── index.hbs
```

I made this with `mdbook init --theme`, and then subsequently I'm trying to change files inside.
Not even changing the favicon results in a change, but rather the output directory contains stock files, let alone
changing something in the CSS files.

My `book.toml`:

```
[book]
authors = ["P"]
language = "en"
multilingual = false
src = "src"
title = "Book"

[output.html]
theme = "ayu"
default-theme = "ayu"
preferred-dark-theme = "ayu"

[output.katex]

[preprocessor.katex]
```

I'm building the book with `mdbook build` and viewing the HTML files with Firefox.
Does anyone know what I'm doing wrong here?

clarkfw:
This PR fixes #1756.

With the following config,
```toml
#book.toml

[output.html]
theme = "./xxx"
```
if `./xxx` directory does not exist, an error message will be logged:
```log
#log

2022-04-25 02:19:12 [INFO] (mdbook::book): Book building has started
2022-04-25 02:19:12 [INFO] (mdbook::book): Running the html backend
2022-04-25 02:19:12 [ERROR] (mdbook::config): theme dir "./xxx" does not exist
2022-04-25 02:19:12 [INFO] (mdbook::cmd::serve): Serving on: http://localhost:3000
2022-04-25 02:19:12 [INFO] (warp::server): Server::run; addr=127.0.0.1:3000
2022-04-25 02:19:12 [INFO] (warp::server): listening on http://127.0.0.1:3000 
2022-04-25 02:19:12 [ERROR] (mdbook::config): theme dir "./xxx" does not exist
2022-04-25 02:19:12 [INFO] (mdbook::cmd::watch): Listening for changes...
```

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -481,7 +481,13 @@ impl Renderer for HtmlHandlebars {
         let mut handlebars = Handlebars::new();
 
         let theme_dir = match html_config.theme {
-            Some(ref theme) => ctx.root.join(theme),
+            Some(ref theme) => {
+                let dir = ctx.root.join(theme);
+                if !dir.is_dir() {
+                    bail!("theme dir {} does not exist", dir.display());
+                }
+                dir
+            }
             None => ctx.root.join("theme"),
         };
 
--- a/tests/rendered_output.rs
+++ b/tests/rendered_output.rs
@@ -657,6 +657,57 @@ fn summary_with_markdown_formatting() {
     );
 }
 
+/// Ensure building fails if `[output.html].theme` points to a non-existent directory
+#[test]
+fn failure_on_missing_theme_directory() {
+    // 1. Using default theme should work
+    let temp = DummyBook::new().build().unwrap();
+    let book_toml = r#"
+        [book]
+        title = "implicit"
+        src = "src"
+        "#;
+
+    write_file(temp.path(), "book.toml", book_toml.as_bytes()).unwrap();
+    let md = MDBook::load(temp.path()).unwrap();
+    let got = md.build();
+    assert!(got.is_ok());
+
+    // 2. Pointing to a normal directory should work
+    let temp = DummyBook::new().build().unwrap();
+    let created = fs::create_dir(temp.path().join("theme-directory"));
+    assert!(created.is_ok());
+    let book_toml = r#"
+        [book]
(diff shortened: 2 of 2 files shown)


ehuss:
Thanks!  I think I would expect mdbook to exit with an nonzero status if there is an error. Maybe instead of adding a check to the `theme_dir` method, it could just check if it exists in the `render` method, and `bail!` if it doesn't exist?

clarkfw:
> Thanks! I think I would expect mdbook to exit with an nonzero status if there is an error. Maybe instead of adding a check to the `theme_dir` method, it could just check if it exists in the `render` method, and `bail!` if it doesn't exist?

Yes, this looks much better. I've changed the content of commit.

With the following config,
```toml
#book.toml

[output.html]
theme = "./xxx"
```

if `./xxx` directory does not exist, build will fail:
```
$ mdbook build
2022-04-26 20:33:37 [INFO] (mdbook::book): Book building has started
2022-04-26 20:33:37 [INFO] (mdbook::book): Running the html backend
2022-04-26 20:33:37 [ERROR] (mdbook::utils): Error: Rendering failed
2022-04-26 20:33:37 [ERROR] (mdbook::utils): 	Caused By: theme dir /Users/clark/projects/site/./xxx does not exist
```

ehuss:
Thanks, can you add a test for this?  (I know it is a bit trivial, but I think it is good to keep test coverage of error cases.)

clarkfw:
Thanks for the reminder! I've added a test function. Since it's my first time to write a test for mdBook, let me know if anything is missing.

ehuss (approved):
Looks good, thanks!

There is some risk that this will cause some unexpected errors for some projects, but I don't expect that to affect too many people.

Issue #1288, opened 2020-07-26
we need footer.hbs


zhangguanzhang:
When multiple people collaborate to write a document, we hope to generate information such as the modified person, the generation date, and the modification date in the footer location. Especially the github-page page. We really need this feature to trace the document

jakubgs:
It would be great to have a configurable footer to display things like the license of the content.

PR #1793, opened 2022-04-28
Adding a footer partial integration

hamflx:
Add footer.hbs support #1288

### DIFF
--- a/guide/src/format/theme/README.md
+++ b/guide/src/format/theme/README.md
@@ -14,6 +14,7 @@ Here are the files you can override:
 - **_index.hbs_** is the handlebars template.
 - **_head.hbs_** is appended to the HTML `<head>` section.
 - **_header.hbs_** content is appended on top of every book page.
+- **_footer.hbs_** content is appended at the bottom of every book page.
 - **_css/_** contains the CSS files for styling the book.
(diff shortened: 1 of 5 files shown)


ehuss:
Thanks!  Can you say more about your use case?  I think it would be good to discuss the particular location for where the footer is located.  At first glance it is not clear how the footer should interact with the content, navigation buttons, and sidebar. I think below the buttons is probably OK, but I want to check why you want to add this, and are there considerations for other projects.

hamflx:
I added the disqus comment section to footer.hbs. Also for websites in China you need to put icp record information at the bottom of the website.

image

ljrxzr:
> 添加 footer.hbs 支持#1288

Hey, I noticed your need. You are trying to place the comment system and government information at the end of the article. I don't think this needs a separate footer file. Now, you just need to make some settings in index.hbs, like I did.

image

image

hamflx:
Thank you for your advice.

> caused the login button to be obscured

I fixed the problem by inserting the dom element of the comment system into the <main> tag.

> you just need to make some settings in index.hbs

I didn't redefine index.hbbs for my blog, it was too much of a hassle.

zhangguanzhang:
@ehuss PTAL

PR #1800, opened 2022-05-05
Add '--is-dark-theme' css variable to support dynamic css-only dark theme image modifications

stevenengler:
This PR adds a `--is-dark-theme` variable to each theme. This is useful when you have images or diagrams that you want to look nice in both light and dark themes. This is kind of a hack and doesn't help in situations where you want to switch between different dark-theme and light-theme images, but on the other hand this is a very simple way to help support dynamic light/dark theme images with very little change. Anyways, just a suggestion :)

For example, in the markdown:

```html
<img class="color-adapting-image" src="assets/my-diagram.svg">
```

Then in `book.toml`:

```toml
[output.html]
additional-css = ["custom.css"]
```

And in `custom.css`:

```css
img.color-adapting-image,
figure.color-adapting-image img {
    filter: invert(calc(85% * var(--is-dark-theme))) hue-rotate(calc(180deg * var(--is-dark-theme)));
}
```

theme-example

### DIFF
--- a/src/theme/css/variables.css
+++ b/src/theme/css/variables.css
@@ -48,6 +48,8 @@
     --searchresults-border-color: #888;
     --searchresults-li-bg: #252932;
     --search-mark-bg: #e3b171;
+
+    --is-dark-theme: 1;
 }
 
 .coal {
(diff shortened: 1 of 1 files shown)


stevenengler:
This is straightforward to do with only the custom css by adding:

```css
.ayu, .coal, .navy {
    --is-dark-theme: 1;
}

.light, .rust {
    --is-dark-theme: 0;
}

@media (prefers-color-scheme: dark) {
    .light.no-js {
        --is-dark-theme: 1;
    }
}
```

PR #1801, opened 2022-05-06
init regexes via lazy_static

klensy:
This reduces total allocated bytes (tracked by DHAT) for `mdbook build`. For example, for [`rustbook`](https://github.com/rust-lang/book) from ~663mb to ~548mb


ehuss (approved):
Thanks!  It might be good to consider using once_cell instead in the future (or std::lazy if it ever gets stabilized).

PR #1806, opened 2022-05-17
Make code buttons appear on hover (or tap on mobile)

### LINKED ISSUE #1322: Buttons still overlap code

### LINKED ISSUE #1433: Stray spacing of code blocks containing injected main

ehuss:
This changes it so that code buttons will appear on hover (similar to how many other sites like GitHub do it). This also has various tweaks to the overall spacing to give some more breathing room.

For implicit main, it no longer adds a blank line at top.

Closes #1322
Closes #1433

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -880,11 +880,8 @@ fn add_playground_pre(
                                 // we need to inject our own main
                                 let (attrs, code) = partition_source(code);
 
-                                format!(
-                                    "\n# #![allow(unused)]\n{}#fn main() {{\n{}#}}",
-                                    attrs, code
-                                )
-                                .into()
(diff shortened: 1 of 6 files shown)


ehuss:
Note that the ACE editor still has overlap (see comments for explanation).

I'd still consider hiding the buttons with a hover rule.  It would also need a tap handler for mobile.  I figure that is something that could possibly be added later. I kinda like the balanced vertical whitespace in this change regardless.

PR #1807, opened 2022-05-17
Add copy-custom-fonts parameter to HtmlConfig to enable custom fonts

emirror-de:
The first PR #1799 has been accidentially closed.

This PR introduces a new parameter to the `HtmlConfig` struct named `copy_custom_fonts`. When set to true, the fonts/fonts.css file is scanned for @font-face rules that contain a src: url(font-file.name) attribute. The files that have been found are copied into the book folder on build/serve.

If `copy_fonts` is also enabled, the build in `fonts/fonts.css` is extended by the custom one. In addition to that, the build in font files are also copied to the output folder.

## Example usage:

Create a `fonts/` folder in your book root. Then copy the fonts you want to use into this folder (in this example we are assuming `Lato-Regular.ttf`)

Create a custom fonts file `fonts/fonts.css`
```css
@font-face {
     font-family: "Lato";
     font-style: normal;
     font-weight: normal;
     src: url('Lato-Regular.ttf');
}
```

Setup your `book.toml` that it contains the following parameters:
```toml
[output.html]
copy-fonts = true
copy-custom-fonts = true
```

Adjust your `theme/css/general.css` file according to your needs, for example:
```css
html {
    font-family: "Lato", sans-serif;
}
```

It adds a new dependency `lewp-css` to `Cargo.toml` for simple parsing of the `fonts/fonts.css` file.
It also updates the `msrv` to `1.56.0` because `lewp-css` requires edition 2021.
Furthermore, the `clap` dependency gets updated.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,7 +32,7 @@ jobs:
           - build: msrv
             os: ubuntu-latest
             # sync MSRV with docs: guide/src/guide/installation.md
-            rust: 1.54.0
+            rust: 1.56.0
     steps:
     - uses: actions/checkout@master
     - name: Install Rust
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -194,15 +194,15 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "3.0.10"
+version = "3.1.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7a30c3bf9ff12dfe5dae53f0a96e0febcd18420d1c0e7fad77796d9d5c4b5375"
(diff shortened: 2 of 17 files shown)


ISSOtm:
AIUI, non-processed files under `src/` get copied as-is, so couldn't you simply create `src/fonts/Lato-Regular.ttf` and reference that (`fonts/Lato-Regular.ttf`)?

emirror-de:
Thanks for your response! You are right, this works indeed.
But if it comes to project structure, the font files in my opinion do not belong to the source code of the project. As a result, they should go into a separate folder to have clean separation. Would that make sense?

ehuss:
Thanks for the PR!  I think I would like to go with a different solution posted in #1807. My goal is to keep the config as simple as possible. Any feedback on that approach is welcome, though.

PR #1810, opened 2022-05-23
Update to elasticlunr-rs 3.0.0

mattico:
I plan to release elasticlunr-rs 3.0.0 soon. None of the API changes affect mdBook.

Just checking that CI passes for now.


ehuss:
Thanks for helping with the update!  Is there anything we need to know about this update? From the [changelog](https://github.com/mattico/elasticlunr-rs/blob/master/CHANGELOG.md) I don't see anything, but just wondering if there are any changes the user may experience with this?

mattico:
Nope! There are some minor changes to other languages, but mdBook only uses English which should behave identically.

ehuss (approved):
Sounds good, thanks!

Issue #1340, opened 2020-09-27
Please allow table of contents entries that don't have markdown pages.


Lokathor:
eg:

```md
# Summary

- [Toolchain Setup](toolchain_setup.md)
- [Project Setup](project_setup.md)
- [Volatile](volatile.md)
- [Bit Packing](bit_packing.md)
- Video
  - [Display Control](display_control.md)
```

This would allow for better TOC organization without having to have useless blank pages.

marco-m-pix4d:
@Lokathor 

I believe you can obtain what you want with a `Part Title` (it took me a while to understand that the `#` was not parsed as markdown:

```
- [Bit Packing](bit_packing.md)
# Video
  - [Display Control](display_control.md)
```

Lokathor:
Very neat, I didn't know a person could do that.

However, it seems that style makes the numbering not include the part entries.

marco-m-pix4d:
> However, it seems that style makes the numbering not include the part entries.

Yes. I think that the only way to have subnumbering is to use a blank page as you mention.

To help people stumbling on this thread:

## Approach 1 (requires blank page)

```
- [Bit Packing](bit_packing.md)
- [Video](blank.md)
  - [Display Control](display_control.md)
  - [Foo](foo.md)
```

### Generates:

```
1. Bit Packing
2. Video
   2.1 Display Control
   2.2 Foo
```

## Approach 2 (no blank page)

```
- [Bit Packing](bit_packing.md)
# Video
  - [Display Control](display_control.md)
  - [Foo](foo.md)
```

### Generates:

```
1. Bit Packing
Video
2. Display Control
3. Foo
```

shonfeder:
It's a hack, but seems like you could use draft pages to achieve this?

```
- [Video]()
  - [Display Control](display_control.md)
  - [Foo](foo.md)
```

Then `Video` will be greyed out, but not require a blank page.

ISSOtm:
This was discussed in #1820 as well (the PR closed just above). See https://github.com/rust-lang/mdBook/pull/1820#issuecomment-1255663374 for a rationale against using draft chapters for this. Special syntax (just an entry without a link?) would thus be warranted, and IMO make more sense.

PR #1820, opened 2022-06-18
Allow folding draft chapters with children by interacting with their names

ISSOtm:
This PR allows links to draft chapters that have children elements to function like the "arrow" to toggle collapsing children items when clicked. This includes styling them as "real" links to highlight their interactivity, and turning the arrow when clicked.

This has been requested by some of my readers for the "Graphics" link in https://eldred.fr/gb-asm-tutorial.

This is implemented by outputting links with the `toggle` class instead of `div`s for chapters fulfilling the aforementioned criteria, removing the styling and adding the toggle functionality.

I did not add any documentation, as this is purely a UI change, and this part of the UI is not documented anywhere in the first place.

### DIFF
--- a/src/renderer/html_handlebars/helpers/toc.rs
+++ b/src/renderer/html_handlebars/helpers/toc.rs
@@ -116,8 +116,12 @@ impl HelperDef for RenderToc {
                 continue;
             }
 
+            let is_foldable = item.get("has_sub_items").map_or(false, |flag| {
+                fold_enable && flag.parse::<bool>().unwrap_or_default()
+            });
+
             // Link
-            let path_exists: bool;
+            let has_link: bool;
             match item.get("path") {
                 Some(path) if !path.is_empty() => {
                     out.write("<a href=\"")?;
@@ -139,11 +143,15 @@ impl HelperDef for RenderToc {
                     }
 
                     out.write(">")?;
-                    path_exists = true;
+                    has_link = true;
+                }
+                _ if is_foldable => {
+                    out.write("<a class=\"toggle\">")?;
+                    has_link = true;
                 }
                 _ => {
                     out.write("<div>")?;
-                    path_exists = false;
+                    has_link = false;
                 }
             }
 
@@ -160,18 +168,15 @@ impl HelperDef for RenderToc {
                 out.write(&bracket_escape(name))?
             }
 
-            if path_exists {
+            if has_link {
                 out.write("</a>")?;
             } else {
                 out.write("</div>")?;
             }
 
             // Render expand/collapse toggle
-            if let Some(flag) = item.get("has_sub_items") {
-                let has_sub_items = flag.parse::<bool>().unwrap_or_default();
-                if fold_enable && has_sub_items {
-                    out.write("<a class=\"toggle\"><div>❱</div></a>")?;
-                }
+            if is_foldable {
+                out.write("<a class=\"toggle\"><div>❱</div></a>")?;
             }
             out.write("</li>")?;
         }
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -446,15 +446,18 @@ ul#searchresults span.teaser em {
 
 .chapter li > a.toggle {
     cursor: pointer;
+}
+
+.chapter li > a.toggle:not(:first-child) {
     display: block;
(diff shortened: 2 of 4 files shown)


schungx:
I solve this problem with a custom CSS:

```css
.chapter lii {
    color: inherit;
}
```

ISSOtm:
This change is not about presentation as much as functionality: "disabled" chapters that have folded children can be clicked on, and doing so expands the list. Since they then have functionality, they are not styled as "disabled" anymore, but that's not the main focus of the PR.

src/renderer/html_handlebars/helpers/toc.rs
```diff
@@ -132,6 +136,9 @@ impl HelperDef for RenderToc {
 
                 out.write(">")?;
                 true
+            } else if is_foldable {
+                out.write("<a class=\"toggle\">")?;
```
anzome, line 140: In general it isn't a good idea to wrap something into link (<a>) if that isn't an actual link but only the interactive element, browsers have the tons of tricky behavior around the links and well as the search engins, it could be hard to predict all the unintended impact caused by the empty link. Simple <span> with the same class, or with reasonable modifications could be ok for this case.

ISSOtm: Well, this only mimics the behaviour of the same line just below. This should be changed separately, then.

Rangi42: If it's not appropriate to use an `<a>` link for non-hyperlink elements, that should also apply to the toggle-able chevrons which are currently using `<a class="toggle">`. The chevron, *and* its adjacent text, could use a `<button>` if preferred (or at least an aria `role="button"`).

ISSOtm: I don't think we'd want any of the default styling that should come with `<button>`, so I'd personally prefer `role="button"`. (Though again, this would have to be changed separately, so I will only add it to this PR if upstream is updated first.)

ISSOtm:
Bump

ehuss:
@ISSOtm Can you provide a more detailed description of what this does, how this works, what it looks like, and what the use case is?

Also, this will need:
* Documentation
* A test of some kind
* Update the [`test_book`](https://github.com/rust-lang/mdBook/tree/master/test_book) with an example of what this does.

ISSOtm:
I updated the OP regarding the description and documentation; I'll update the test book when I get back near a computer.

As for a test, I'm not sure what should be tested.

ISSOtm:
Rebased, and pushed the addition to `test_book` (oops).

ehuss:
I'm concerned that this stretches the draft chapters beyond what the intended behavior is. I think people will abuse this to implement non-chapter organization (that is, not using them as drafts at all).  Not only could that be confusing from a UX standpoint, but also things are broken in that situation (such as next/prev navigation).

If your only intent is to make the parent draft titles clickable, I think that's fine to do as long as it retains the draft styling.

schungx:
> I think people will abuse this to implement non-chapter organization

Incidentally, this is _exactly_ what I plan to use it for...  because this type of functionality is not available otherwise.

ISSOtm:
I find your concern valid, but I must also second @schungx: this is a functionality that is otherwise lacking, and exactly the use case that I am aiming to fill. (See [the "Graphics" group here](https://eldred.fr/gb-asm-tutorial/part1/tiles.html) for an example.)

An alternative would be to allow non-link list items in the `SUMMARY.md` (and probably error out if they don't have any children), and transfer this functionality to them.

ehuss:
Ah. In that case, I think something like this should probably go through more of a design process before progressing with a PR.

I think to move forward with something like this, there will need to be quite a bit more investment to figure out the right design. Some ideas of things to investigate:

* How do other documentation tools with a navigation sidebar or TOC handle sections or chapters without content?
* Semantically what is a better way to represent these kinds of non-content chapters? I think calling them "draft" does not correctly match the new behavior.
* What should the syntax be in SUMMARY to differentiate draft chapters from non-content entries?
* What will it take to make sure everything is functional? What besides next/prev is broken?

ehuss:
I'm going to close in favor of discussing a proposed change in #1340. Thanks!

PR #1825, opened 2022-06-20
Update warp to 0.3.2

### LINKED ISSUE #1797: serve command does not implement HTTP HEAD
Just a small thing I noticed: The HTTP server run with `mdbook serve` does not properly implement HTTP HEAD requests, which my url handler script tests for 404 (and some other specific error codes like 403 and 502) and redirects in order to save myself from opening them in various applications (like a media player or a browser as the fallback).

I will ~~probably~~ exempt localhost URLs from this check regardless, since I can hardly think of a situation where checking for HEAD responses makes sense there, but maybe this is something you're interested in anyway.

joshrotenberg:
This updates `warp` to the latest release, which happens to fix #1797 by making HEAD requests automatically part of `warp::fs::dir` routes:

With warp 0.3.1:

```
# cd guide && mdbook serve ...

$ curl --head localhost:3000/index.html
HTTP/1.1 404 Not Found
content-length: 14598
content-type: text/html
accept-ranges: bytes
last-modified: Mon, 20 Jun 2022 18:53:48 GMT
date: Mon, 20 Jun 2022 18:53:52 GMT
```

With warp 0.3.2:

```
# cd guide && mdbook serve ...

$ curl --head localhost:3000/index.html
HTTP/1.1 200 OK
content-length: 17487
content-type: text/html
accept-ranges: bytes
last-modified: Mon, 20 Jun 2022 18:56:24 GMT
date: Mon, 20 Jun 2022 18:56:28 GMT
```
See https://github.com/seanmonstar/warp/blob/master/CHANGELOG.md#v032-november-9-2021

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -383,24 +383,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "futures"
-version = "0.3.16"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1adc00f486adfc9ce99f77d717836f0c5aa84965eb0b4f051f4e83f7cab53f8b"
-dependencies = [
- "futures-channel",
- "futures-core",
- "futures-io",
- "futures-sink",
- "futures-task",
- "futures-util",
-]
-
-[[package]]
 name = "futures-channel"
-version = "0.3.16"
+version = "0.3.21"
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks!

PR #1829, opened 2022-06-22
Fix up "index page" functionality

### LINKED ISSUE #1814: index.html fails to be created silently when there is no prefix chapter
When building a book that has no prefix chapter before other titles are defined in `SUMMARY.md` the `index.html` file isn't built.

Steps to reproduce:
```bash
cd /tmp
mdbook init --ignore git --title 'My Book' my-book
cd my-book
sed -i '3i# Part 1' src/SUMMARY.md
cat src/SUMMARY.md 
# # Summary
# 
# # Part 1
# - [Chapter 1](./chapter_1.md)
mdbook build
ls book/index.html
# ls: cannot access 'book/index.html': No such file or directory
```

The docs say:

> Before the main numbered chapters, prefix chapters **can** be added...

My understanding is that prefix chapters are thus optional, but mdbook doesn't render correctly without them.

ISSOtm:
- Fixes #1814 by only marking the first non-draft chapter as the "index" page, and adds a test for that
- Makes the link to the first chapter active in the "index" page, improving consistency (and making the aforementioned test pass without hacks)

Additionally, I would've liked to document that "index page" functionality, but I couldn't find any place for it in the documentation.

### DIFF
--- a/tests/rendered_output.rs
+++ b/tests/rendered_output.rs
@@ -15,7 +15,7 @@ use select::predicate::{Class, Name, Predicate};
 use std::collections::HashMap;
 use std::ffi::OsStr;
 use std::fs;
-use std::io::{Read, Write};
+use std::io::Write;
 use std::path::{Component, Path, PathBuf};
 use std::str::FromStr;
 use tempfile::Builder as TempFileBuilder;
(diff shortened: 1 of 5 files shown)


tests/rendered_output.rs
```diff
@@ -467,6 +467,76 @@ fn by_default_mdbook_use_index_preprocessor_to_convert_readme_to_index() {
...
+    let temp = DummyBook::new().build().unwrap();
+    let mut cfg = Config::default();
+    cfg.set("book.src", "index_html_test")
+        .expect("Couldn't set config.book.src to \"index_html_test\"");
+    let md = MDBook::load_with_config(temp.path(), cfg).unwrap();
+    md.build().unwrap();
+
+    let chapter_path = temp.path().join("book").join("chapter_1.html");
+    let index_path = temp.path().join("book").join("index.html");
+    // In theory, just reading the entire files into memory and comparing is sufficient for *testing*,
```
ehuss, line 481: I think this could use `pretty_assertions::assert_eq` to get a highlighted context diff.  I think it would be good to change this to just read the files into memory and use `pretty_assertions`. Would that work?

ehuss (commented):
I *think* this should be fine to land.  It seems a little strange to not have a prefix chapter like an introduction.  Another option would be to create an index.html that redirects to the first chapter, though that's probably not necessary.

tests/rendered_output.rs
ISSOtm: I agree that performance is not really a concern as far as failure is concerned, but I was kind of worried for a moment that it was frozen.

The real concern I had was that the entire file gets printed on a single line, which is not all too handy. But this is no longer the case with `pretty_assertions` v1.2, *and* it's fast, so indeed, no reason not to.

tests/rendered_output.rs
```diff
@@ -467,6 +467,34 @@ fn by_default_mdbook_use_index_preprocessor_to_convert_readme_to_index() {
...
+    let read_file = |path: &str| {
+        let mut buf = String::new();
+        fs::File::open(book_path.join(path))
+            .with_context(|| format!("Failed to read {}", path))
+            .unwrap()
+            .read_to_string(&mut buf)
+            .unwrap();
+        buf
+    };
+    pretty_assertions::assert_eq!(read_file("chapter_1.html"), read_file("index.html"));
```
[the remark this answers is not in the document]
ISSOtm: Ah, right, my bad on the comment. As for the simplification, I guess I'm still getting used to some of the Rust style ^^'

ehuss (approved):
Thanks!

PR #1830, opened 2022-06-22
Always open index page with `serve --open`

ISSOtm:
This fixes an issue where `mdbook serve --open` may land on a 404 page if the file name is changed. This is easily reproduced by running `mdbook serve --open guide` on mdBook itself, which will open `http://localhost:3000/README.html`, which 404's thanks to the `index` preproc.

Since mdBook seems to always generate an `index.html` page for the first chapter anyway, it seems that "trimming the fat" is also more correct.

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -1,8 +1,7 @@
-use crate::{first_chapter, get_book_dir, open};
+use crate::{get_book_dir, open};
 use clap::{arg, App, Arg, ArgMatches};
 use mdbook::errors::Result;
 use mdbook::MDBook;
-use std::path::Path;
 
 // Create clap subcommand arguments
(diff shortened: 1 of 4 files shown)


ehuss:
Thanks!  This was intentionally changed in #1714 to avoid problems with draft chapters. Can you find a solution that fixes both issues?

ehuss:
Can you also update `build.rs` and `watch.rs`?

ISSOtm:
What needs to be changed?

ehuss:
They both have `--open` flags with code that needs to be updated.  More or less, this should revert #1714.

ISSOtm:
Done!

ehuss (approved):
Thanks!

PR #1832, opened 2022-06-26
Fix Clippy lints

ISSOtm:
Also remove `#[allow(clippy::*)]`s where possible

### DIFF
--- a/src/book/mod.rs
+++ b/src/book/mod.rs
@@ -386,7 +386,7 @@ fn determine_renderers(config: &Config) -> Vec<Box<dyn Renderer>> {
     renderers
 }
 
-const DEFAULT_PREPROCESSORS: &[&'static str] = &["links", "index"];
+const DEFAULT_PREPROCESSORS: &[&str] = &["links", "index"];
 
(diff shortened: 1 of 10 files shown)


src/book/book.rs
```diff
@@ -77,10 +77,10 @@ fn create_missing(src_dir: &Path, summary: &Summary) -> Result<()> {
 /// [`iter()`]: #method.iter
 /// [`for_each_mut()`]: #method.for_each_mut
 #[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize)]
+#[non_exhaustive]
```
ehuss, line 80: This is a breaking change and cannot be done in a point release.

ISSOtm: I'll revert it, then; but, should the lint be kept so we can remember for the next major release, or should it be suppressed?

ehuss: I filed #1835 to track considering it in a point release. I don't have much of an opinion on allowing it, as I don't use clippy. I'm not particularly happy with lots of `allow(clippy)` stuff, but if you'd like to add it, that would be fine.

PR #1836, opened 2022-06-29
Avoid empty last line in editable code blocks

mgeisler:
Before, a code block would always end with a final newline. The newline was added unconditionally by `hide_lines`.

When the code block is syntax highlighted by highlight.js, this is not a problem, no empty line is added for a final trailing `\n` character. However, when the code block is editable and thus handled by the ACE editor, a trailing newline _is_ significant. I believe this issue is most closely described by https://github.com/ajaxorg/ace/issues/2083 in the upstream repository.

The effect of the way ACE handles newlines is that a code block like

    <pre>
      Some code
    </pre>

will create an editor with _two_ lines, not just one.

By trimming trailing whitespace, we ensure that we don’t accidentally create more lines in the ACE editor than necessary.

# Before

# After

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -903,13 +903,16 @@ fn hide_lines(content: &str) -> String {
     }
 
     let mut result = String::with_capacity(content.len());
-    for line in content.lines() {
+    let mut lines = content.lines().peekable();
+    while let Some(line) = lines.next() {
+        // Don't include newline on the last line.
+        let newline = if lines.peek().is_none() { "" } else { "\n" };
         if let Some(caps) = BORING_LINES_REGEX.captures(line) {
             if &caps[2] == "#" {
                 result += &caps[1];
                 result += &caps[2];
                 result += &caps[3];
-                result += "\n";
+                result += newline;
                 continue;
             } else if &caps[2] != "!" && &caps[2] != "[" {
                 result += "<span class=\"boring\">";
@@ -918,13 +921,13 @@ fn hide_lines(content: &str) -> String {
                     result += &caps[2];
                 }
                 result += &caps[3];
-                result += "\n";
+                result += newline;
                 result += "</span>";
                 continue;
             }
         }
         result += line;
-        result += "\n";
+        result += newline;
     }
     result
 }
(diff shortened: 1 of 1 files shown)


mgeisler:
@ehuss or others, would you be able to review this? I would love to help make mdBook look more uniform.

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -903,28 +903,42 @@ fn hide_lines(content: &str) -> String {
...
-    for line in content.lines() {
+    let mut lines = content.lines();
+    let last_line = lines.next_back();
+    let mut handle_line = |line: &str, newline: &str| {
         if let Some(caps) = BORING_LINES_REGEX.captures(line) {
             if &caps[2] == "#" {
                 result += &caps[1];
                 result += &caps[2];
                 result += &caps[3];
                 result += "\n";
```
ehuss, line 914: Is there a reason this newline isn't treated like the others?

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -903,28 +903,42 @@ fn hide_lines(content: &str) -> String {
     }
 
     let mut result = String::with_capacity(content.len());
-    for line in content.lines() {
+    let mut lines = content.lines();
+    let last_line = lines.next_back();
+    let mut handle_line = |line: &str, newline: &str| {
```
ehuss, line 908: Instead of using a closure, can this be simplified a bit by using a peekable iterator? Perhaps something like this:

```rust
    let mut lines = content.lines().peekable();
    while let Some(line) = lines.next() {
        // Don't include newline on the last line.
        let newline = if lines.peek().is_none() { "" } else { "\n" };
```

mgeisler: Thanks, that's a nice idea!

src/renderer/html_handlebars/hbs_renderer.rs
mgeisler: Oh, sorry about that — there's no reason to treat it differently.

ehuss (approved):
Thanks!

mgeisler:
> Thanks!

Thanks for the review!

PR #1844, opened 2022-07-04
Upgrade clap

wendajiang:
This updates clap and clap_complete from version 3 to version 4.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,8 @@ rust-version = "1.60"
 [dependencies]
 anyhow = "1.0.28"
 chrono = "0.4"
-clap = { version = "3.0", features = ["cargo"] }
-clap_complete = "3.0"
+clap = { version = "4.0.29", features = ["cargo", "wrap_help"] }
+clap_complete = "4.0.6"
 once_cell = "1"
 env_logger = "0.10.0"
 handlebars = "4.0"
--- a/src/cmd/test.rs
+++ b/src/cmd/test.rs
@@ -1,63 +1,54 @@
+use super::command_prelude::*;
 use crate::get_book_dir;
-use clap::{arg, App, Arg, ArgMatches};
+use clap::builder::NonEmptyStringValueParser;
+use clap::{Arg, ArgAction, ArgMatches, Command};
 use mdbook::errors::Result;
 use mdbook::MDBook;
+use std::path::PathBuf;
 
 // Create clap subcommand arguments
-pub fn make_subcommand<'help>() -> App<'help> {
-    App::new("test")
+pub fn make_subcommand() -> Command {
+    Command::new("test")
         .about("Tests that a book's Rust code samples compile")
+        // FIXME: --dest-dir is unused by the test command, it should be removed
+        .arg_dest_dir()
+        .arg_root_dir()
         .arg(
-            Arg::new("dest-dir")
-                .short('d')
-                .long("dest-dir")
-                .value_name("dest-dir")
-                .help(
-                    "Output directory for the book{n}\
-                    Relative paths are interpreted relative to the book's root directory.{n}\
-                    If omitted, mdBook uses build.build-dir from book.toml or defaults to `./book`.",
-                ),
-        ).arg(
             Arg::new("chapter")
                 .short('c')
                 .long("chapter")
-                .value_name("chapter")
-                .help(
-                    "Only test the specified chapter{n}\
-                    Where the name of the chapter is defined in the SUMMARY.md file.{n}\
-                    Use the special name \"?\" to the list of chapter names."
-                )
+                .value_name("chapter"),
         )
-        .arg(arg!([dir]
-            "Root directory for the book{n}\
-            (Defaults to the Current Directory when omitted)"
-        ))
-        .arg(Arg::new("library-path")
-            .short('L')
-            .long("library-path")
-            .value_name("dir")
-            .takes_value(true)
-            .use_delimiter(true)
-            .require_delimiter(true)
-            .multiple_values(true)
-            .multiple_occurrences(true)
-            .forbid_empty_values(true)
-            .help("A comma-separated list of directories to add to {n}the crate search path when building tests"))
+        .arg(
+            Arg::new("library-path")
+                .short('L')
+                .long("library-path")
+                .value_name("dir")
+                .value_delimiter(',')
+                .num_args(1..)
+                .value_parser(NonEmptyStringValueParser::new())
+                .action(ArgAction::Append)
+                .help(
+                    "A comma-separated list of directories to add to the crate \
+                    search path when building tests",
+                ),
+        )
 }
 
 // test command implementation
 pub fn execute(args: &ArgMatches) -> Result<()> {
     let library_paths: Vec<&str> = args
-        .values_of("library-path")
-        .map(std::iter::Iterator::collect)
+        .get_many("library-path")
+        .map(|it| it.map(String::as_str).collect())
         .unwrap_or_default();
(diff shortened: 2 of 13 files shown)


wendajiang:
Emm, if msrv is edition 2021, clap can upgrade to 3.2.8. So temporially make this PR draft.

mgeisler:
Hi @wendajiang, I'm hoping we can move to the 2021 edition soon via #1887.

wendajiang:
> Hi @wendajiang, I'm hoping we can move to the 2021 edition soon via #1887.

When it's merged intto master, I'm pleasure to make this PR active. Thanks.

mgeisler:
Hi @wendajiang the PR has been merged, so you can now depend on Rust 2021 crates :tada:

wendajiang:
> Hi @wendajiang the PR has been merged, so you can now depend on Rust 2021 crates 🎉

I have reopen this PR. And use once_cell instead of lazy_static.

ehuss:
Can you remove the clippy changes? Or at least move them to a separate PR? Most of them seem arbitrary, and some seem wrong, and they seemed to be mixed with the clap changes.

This also seems to update all of the dependencies. I would prefer to be intentional about updating individual dependencies. For example, if you could have a separate PR updating clap, and a separate one changing lazy_static, that would be great.

I think updating to 1.57 should probably be fine. It looks like the installation docs weren't updated, though.

wendajiang:
> Can you remove the clippy changes? Or at least move them to a separate PR? Most of them seem arbitrary, and some seem wrong, and they seemed to be mixed with the clap changes.
> 
> This also seems to update all of the dependencies. I would prefer to be intentional about updating individual dependencies. For example, if you could have a separate PR updating clap, and a separate one changing lazy_static, that would be great.
> 
> I think updating to 1.57 should probably be fine. It looks like the installation docs weren't updated, though.

I have revert the clippy fix, however, it's almost the [suggestion](https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq).

And I have put the once_cell instead of lazy_static into this https://github.com/rust-lang/mdBook/pull/1894

Dylan-DPC:
Given clap 4.0 is out i think it is better to submit a new PR that updates it to that version directly

wendajiang:
> Given clap 4.0 is out i think it is better to submit a new PR that updates it to that version directly

I have upgrade the clap to 4.0.

ehuss:
I think 1.60 should be fine.

Can you remove all the clippy changes?

Also, I'm a little confused, but this seems to revert #1894?

ehuss:
> Actually, there is not the clippy changes. I remove from the title.

From looking at the diff at https://github.com/rust-lang/mdBook/pull/1844/files, there seems to be quite a few changes to unrelated files. For example, the changes to the `tests` directory, changes to `src/config.rs`, `src/book/book.rs`, `src/renderers`, etc.

wendajiang:
> > Actually, there is not the clippy changes. I remove from the title.
> 
> From looking at the diff at https://github.com/rust-lang/mdBook/pull/1844/files, there seems to be quite a few changes to unrelated files. For example, the changes to the `tests` directory, changes to `src/config.rs`, `src/book/book.rs`, `src/renderers`, etc.

I think it's very little tuning and not change the semantics. :)

src/cmd/test.rs
```diff
@@ -36,27 +36,25 @@ pub fn make_subcommand<'help>() -> App<'help> {
...
-    let library_paths: Vec<&str> = args
-        .values_of("library-path")
-        .map(std::iter::Iterator::collect)
-        .unwrap_or_default();
-    let chapter: Option<&str> = args.value_of("chapter");
+    let library_paths: Vec<&str> = if let Some(it) = args.get_many::<String>("library-path") {
+        it.map(|s| s.as_str()).collect()
+    } else {
+        vec![]
+    };
```
ehuss, line 50: I think this can keep the style of the original code by something like:

```suggestion
    let library_paths: Vec<&str> = args
        .get_many("library-path")
        .map(|it| it.map(String::as_str).collect())
        .unwrap_or_default();
```

src/cmd/watch.rs
```diff
@@ -37,13 +37,13 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
     let mut book = MDBook::load(&book_dir)?;
 
     let update_config = |book: &mut MDBook| {
-        if let Some(dest_dir) = args.value_of("dest-dir") {
+        if let Some(dest_dir) = args.get_one::<String>("dest-dir").map(|s| s.as_str()) {
```
ehuss, line 40: ditto about map and PathBuf

src/main.rs
```diff
@@ -36,12 +36,13 @@ fn main() {
         Some(("test", sub_matches)) => cmd::test::execute(sub_matches),
         Some(("completions", sub_matches)) => (|| {
             let shell: Shell = sub_matches
-                .value_of("shell")
+                .get_one::<String>("shell")
+                .map(|s| s.as_str())
```
ehuss, line 40: This doesn't work and it currently panics. I recommend looking at the clap_complete documentation for how to set this up.

src/main.rs
```diff
@@ -77,12 +78,12 @@ fn create_clap_app() -> App<'static> {
...
         .subcommand(cmd::clean::make_subcommand())
         .subcommand(
-            App::new("completions")
+            Command::new("completions")
                 .about("Generate shell completions for your shell to stdout")
                 .arg(
                     Arg::new("shell")
-                        .takes_value(true)
-                        .possible_values(Shell::possible_values())
+                        .num_args(1)
```
ehuss, line 85: I don't think `num_args` is necessary here.

src/main.rs
```diff
@@ -124,7 +125,7 @@ fn init_logger() {
 }
 
 fn get_book_dir(args: &ArgMatches) -> PathBuf {
-    if let Some(dir) = args.value_of("dir") {
+    if let Some(dir) = args.get_one::<String>("dir") {
```
ehuss, line 128: I don't think this is necessary to go through the roundtrip of making a String. I think it can do something like `get_one::<PathBuf>`, and then remove the call to `Path::new`.

ehuss (commented):
This seems to do a full update to `Cargo.lock`. Can you make sure that this only updates clap and not all of the other dependencies?  I have asked this previously, but it seems like it was added back in again?

I'd also appreciate it if you could spend a little time and test out all of the commands and arguments, and make sure they are working correctly.

Cargo.toml
```diff
@@ -18,11 +18,11 @@ description = "Creates a book from markdown files"
 [dependencies]
 anyhow = "1.0.28"
 chrono = "0.4"
-clap = { version = "3.0", features = ["cargo"] }
-clap_complete = "3.0"
-once_cell = "1"
+clap = { version = "4", features = ["cargo"] }
+clap_complete = "4"
```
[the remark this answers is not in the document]
wendajiang: Thanks. I will check it and smoke test all subcommand.

ehuss:
@wendajiang Just checking in to see if you've had a chance to test things to make sure there aren't any regressions?

wendajiang:
I test all the command and subcommand as the smoking test. And the unit test is ok. Maybe you can suggest some other test ?

ehuss:
> I test all the command and subcommand as the smoking test. And the unit test is ok. Maybe you can suggest some other test ?

I meant to run all the subcommands and verify every option works as it did before. With the most recent version posted, almost every command was now panicking due to some downcast problems.

I went through and reworked things a bit to fix the issues and clean up the argument handling a little. I've rebased, squashed the commits, and updated to the latest version.

Issue #1835, opened 2022-06-27
0.5: Switch to non_exhaustive attribute


ehuss:
In 0.5, it might be worthwhile to switch to using `#[non_exhaustive]` instead of using hidden `__non_exhaustive` members.

FauconFan:
Maybe we can close this then! Maybe I can do it with rustbot, but I don't know how...

ehuss:
Closing as resolved by https://github.com/rust-lang/mdBook/pull/2779.

PR #1848, opened 2022-07-09
remove __non_exhaustive members, add non_exhaustive attribute instead

FauconFan:
As stated in #1835, I deleted every non_exhaustive members and added the `#[non_exhaustive]` attribute.  
There are two more structs that may benefit from this: the Chapter struct and the Theme struct. The remaining structs have either a private member either all public members and doesn't deserve the attribute (to my opinion).
I could have put this attribute to enums as well, but this and adding attribute to the two structs above would be a breaking change. [The page reference for non_exhaustive attribute](https://doc.rust-lang.org/reference/attributes/type_system.html). So I let this up to review.

### DIFF
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -77,10 +77,10 @@ fn create_missing(src_dir: &Path, summary: &Summary) -> Result<()> {
 /// [`iter()`]: #method.iter
 /// [`for_each_mut()`]: #method.for_each_mut
 #[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize)]
+#[non_exhaustive]
 pub struct Book {
     /// The sections in this book.
     pub sections: Vec<BookItem>,
-    __non_exhaustive: (),
 }
 
 impl Book {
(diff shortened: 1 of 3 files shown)


Dylan-DPC:
Thanks

ehuss:
Sorry, I'm maybe a bit confused. This is a breaking change, and #1835 explicitly says this should be done for the 0.5 release. Unless I'm misunderstanding what the intent was here, can you please revert this change?

FauconFan:
Well, maybe i'm wrong, but adding the #[non_exhaustive] attribute is not a breaking change if there were already private members right ?
I was talking about breaking change for other structs / enums.

ehuss:
It is unfortunately exposed to preprocessors.  #1571 contains some more information.

FauconFan:
Oh yes indeed, sorry for that, it makes sense, now !

FauconFan:
Well maybe it makes sense to create a new branch for release 0.5.0, then ?

Dylan-DPC:
Ah my bad. 

Nah i wouldn't recommend having a another branch  - having 2 active branches makes it annoying to maintain and keep up to date. I would prefer holding this or having an open pr till 0.5 lands

PR #1849, opened 2022-07-09
authors -> author in user guide

### LINKED ISSUE #1804: Mismatch in documentation
At the beginning of [this page](https://rust-lang.github.io/mdBook/format/configuration/general.html), the field used to represent author is `book.author` (requires a string)

https://github.com/rust-lang/mdBook/blob/bf258eeb9b245b9c329e147dd1884307c5ef7644/guide/src/format/configuration/general.md?plain=1#L7-L14

but then changed to `book.authors` (requires a list):

https://github.com/rust-lang/mdBook/blob/bf258eeb9b245b9c329e147dd1884307c5ef7644/guide/src/format/configuration/general.md?plain=1#L37-L43

and it should be the latter one, according to the config parser

https://github.com/rust-lang/mdBook/blob/981b79b3b39d80b61da2a10622d6cf611a23686a/src/config.rs#L403-L420

(please point out if I have misunderstood)

FauconFan:
It looks like this was the only place author was used in the user guide.

Closes #1804

### DIFF
--- a/guide/src/format/configuration/general.md
+++ b/guide/src/format/configuration/general.md
@@ -7,7 +7,7 @@ Here is an example of what a ***book.toml*** file might look like:
 ```toml
 [book]
 title = "Example book"
-author = "John Doe"
+authors = ["John Doe"]
 description = "The example book covers examples."
 
 [rust]


ehuss (approved):
Thanks!

PR #1853, opened 2022-07-11
Add Fig completion support

nihaals:
Fig completion generation provided by [`clap_complete_fig`](https://crates.io/crates/clap_complete_fig).

This adds [`clap_complete_command`](https://crates.io/crates/clap_complete_command) as a dependency to make supporting both `clap_complete`'s shells and `clap_complete_fig` simpler and keep a similar API to just using `clap_complete` (full disclosure, I'm the maintainer of this crate).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,7 +32,7 @@ jobs:
           - build: msrv
             os: ubuntu-latest
             # sync MSRV with docs: guide/src/guide/installation.md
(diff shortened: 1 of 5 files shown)


ehuss:
Thanks for the PR! I think I'm going to close as I would prefer to stick with `clap_complete` for now. If and when https://github.com/clap-rs/clap/discussions/3151 settles to some solution, we can reevaluate fig support.

PR #1851, opened 2022-07-09
Revert "remove __non_exhaustive members, add non_exhaustive attribute instead"

FauconFan:
Reverts rust-lang/mdBook#1848

Fixing a breaking change looks like #1571

PR #1854, opened 2022-07-12
Release 0.5.0 - #[non_exhaustive] attribute

FauconFan:
This follows multiple discussions: #1848 #1851 #1835.

As stated, this PR could be accepted when the release 0.5.0 will be landed.
I've added the non_exhaustive attribute on the Chapter and Theme structs, since this is a breaking change anyway.
I checked if adding non_exhaustive attribute on enums is a good practice, but in my research I have found that is not really very well handled by serde: Link [here](https://github.com/serde-rs/serde/issues/1991). I think this PR is good as it is.


ehuss:
Closing as resolved by https://github.com/rust-lang/mdBook/pull/2779. Thanks!

PR #1855, opened 2022-07-12
Fix code padding in headings

stevenengler:
In #1806, extra padding was added to code blocks, but this adds the padding to headings as well. I think this was only meant to affect code blocks in the main content. If spacing was intended in the headings, a margin would probably be better. This PR just limits the scope of the change from #1806.

With this PR:

### Before
1657647912_grim

### After
1657648007_grim

### DIFF
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -252,7 +252,7 @@ pre > .buttons button {
         padding: 0.3rem 1rem;
     }
 }
-code {
+pre > code {
     padding: 1rem;
 }
 


src/theme/css/chrome.css
```diff
@@ -252,7 +252,7 @@ pre > .buttons button {
         padding: 0.3rem 1rem;
     }
 }
-code {
+p code {
```
ehuss, line 255: This doesn't look quite right to me. This padding should only be applied to code blocks which have the form:

```
<pre><code>…</code></pre>
```

`p code` would apply to almost everything *except* code blocks.

It seems to me that the rule should be something like:

```suggestion
pre > code {
```

(which also matches the other nearby rules.)

ehuss (commented):
Thanks, apologies for the oversight.

src/theme/css/chrome.css
stevenengler: Ah, sorry I misunderstood the original PR. I've changed it to your suggestion, and amended it to the old commit since it's small.

ehuss (approved):
Thanks!

Issue #1601, opened 2021-07-10
Add option to remove indentation of included file snippets


HoLLy-HaCKeR:
I am using mdBook to write documentation about a code framework and include code snippets from an example project with the `{{#include file.rs:component}}` syntax. Because the code is written in C#, there is typically a lot more indented than with rust code meaning that when I include a single method, it will have at least 2 levels of indentation. Obviously it looks rather silly when all code in your documentation has 2 extra tabs :)

I want to propose adding an extension to the `#include` syntax to remove this extra indentation. I'm not sure yet how exactly this would look, but one way is to add a `:-no-indent` suffix to the filename that indicates that extra indentation should be removed (eg. `{{@include file.rs:-no-indent}}` or `{{@include file.rs:component:-no-indent}}`). Then, the preprocessor would look at every non-empty or non-whitespace line to find the minimum amount of whitespace, and remove that from all included lines.

Before (note that the 4th line is completely empty):
```cs
		public void MyMethod()
		{
			int x = 1;

			Console.WriteLine(x);
		}
```

After:
```cs
public void MyMethod()
{
	int x = 1;

	Console.WriteLine(x);
}
```

ISSOtm:
This sounds like it might be more useful to add to code blocks are than to this preproc?

PR #1565, opened 2021-06-05
Align when `{{#include}}` is indented

kynthus:
Fixes #1564  
Considering the impact on embedding simple code, it may be necessary to discuss whether to make it optional or introduce it, so I made it a draft.

PR #1856, opened 2022-07-14
add option -no-indent to remove indentation

Salka1988:
In this PR, I have added support for the new `LinkType::IncludeNoIndent`. This feature can be utilized with the `Anchor` syntax, such as `{{#include file.rs:anchor_name:-no-indent}}`, or with the `Range` syntax, `{{#include file.rs::20:-no-indent}}`.

Parsing is done in the same way as with the `Range(LineRange)` type. This means that if we use `{{#include file.rs::20: 30}}`, the command will not work until the whitespace between 20 and 30 is removed, resulting in 20:30. The same applies to the `-no-indent` command, `{{#include file.rs::20: -no-indent}}` will not work, while `{{#include file.rs::20:-no-indent}}` does.

The parsing process begins in the `parse_include_path` function where we check for the presence of the `-no-indent` command. Based on this, we return `LinkType::IncludeNoIndent` or only `LinkType::Include`.

If we have returned `LinkType::IncludeNoIndent`, the `take_format_remove_indent` function will find the part of the code closest to the left margin and move it to the left by that much space. The format of the code remains the same, with only the indentation being removed.

`{{#include file.rs::20:30}}`

Screenshot from 2023-01-23 14-16-07

`{{#include file.rs::20:30:-no-indent}}`

Screenshot from 2023-01-23 14-15-47

`{{#include ../../../examples/providers/src/lib.rs:setup_test_blockchain}}`

Screenshot from 2023-01-23 14-16-51

`{{#include ../../../examples/providers/src/lib.rs:setup_test_blockchain:-no-indent}}`

Screenshot from 2023-01-23 14-17-14

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,4 @@
+#![feature(str_split_as_str)]
 //! # mdBook
 //!
 //! **mdBook** is a tool for rendering a collection of markdown documents into
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,3 @@
-#![feature(str_split_as_str)]
 //! # mdBook
 //!
 //! **mdBook** is a tool for rendering a collection of markdown documents into
--- a/src/preprocess/links.rs
+++ b/src/preprocess/links.rs
@@ -1,6 +1,6 @@
 use crate::errors::*;
 use crate::utils::{
-    take_anchored_lines, take_lines, take_rustdoc_include_anchored_lines,
+    take_anchored_lines, take_lines, take_remove_indent, take_rustdoc_include_anchored_lines,
     take_rustdoc_include_lines,
 };
 use regex::{CaptureMatches, Captures, Regex};
@@ -134,6 +134,7 @@ where
 enum LinkType<'a> {
     Escaped,
     Include(PathBuf, RangeOrAnchor),
+    IncludeNoIndent(PathBuf, RangeOrAnchor),
     Playground(PathBuf, Vec<&'a str>),
     RustdocInclude(PathBuf, RangeOrAnchor),
     Title(&'a str),
@@ -205,6 +206,7 @@ impl<'a> LinkType<'a> {
         match self {
             LinkType::Escaped => None,
             LinkType::Include(p, _) => Some(return_relative_path(base, &p)),
+            LinkType::IncludeNoIndent(p, _) => Some(return_relative_path(base, &p)),
             LinkType::Playground(p, _) => Some(return_relative_path(base, &p)),
             LinkType::RustdocInclude(p, _) => Some(return_relative_path(base, &p)),
             LinkType::Title(_) => None,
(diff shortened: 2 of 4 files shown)


Salka1988:
https://github.com/rust-lang/mdBook/issues/1601

Salka1988:
[ehuss](https://github.com/ehuss) Hello, Hello :)

sanity:
Hi, I'm also having this problem - did this get merged?

sanity:
> > Hi, I'm also having this problem - did this get merged?
> > Is the idea to remove all indentation from code snippets or only "extra" indentation? Hopefully, the latter.
> 
> No.

Do you mean that it will remove all indentation, ie.

```
   fn test() {
      cat.foo();
   }
```

Becomes:

```
fn test() {
cat.foo();
}
```

Or (what I'm hoping):

```
fn test() {
   cat.foo();
}
```

Salka1988:
> > > Hi, I'm also having this problem - did this get merged?
> > > Is the idea to remove all indentation from code snippets or only "extra" indentation? Hopefully, the latter.
> > 
> > 
> > No.
> 
> Do you mean that it will remove all indentation, ie.
> 
> ```
>    fn test() {
>       cat.foo();
>    }
> ```
> 
> Becomes:
> 
> ```
> fn test() {
> cat.foo();
> }
> ```
> 
> Or (what I'm hoping):
> 
> ```
> fn test() {
>    cat.foo();
> }
> ```

Second option. What you need too :)

rachitnigam:
Just pinging this thread to say that I've also run into this problem and would find this PR really useful when merged

sanity:
Do we expect this to get merged soon? If not I may create a fork to use in the meantime, this would solve some headaches.

sanity:
That's fair, I'll try to find some time.

ehuss:
I would prefer to avoid adding an option for this. I would expect there to be one true behavior.

There's also PRs and issues about indentation (#1564, #1565, #1718).  Whatever solution is taken, I would like to see a clear description of the issue and why a particular solution was chosen. This particular PR doesn't have much of a description.

Another thing to consider is the behavior around hidden lines.

Salka1988:
@ehuss @sanity @Dylan-DPC  I added some description! Can you take a look?

ehuss:
Can you respond to the questions raised in https://github.com/rust-lang/mdBook/pull/1856#issuecomment-1348532278? Would it be possible to avoid requiring an option to determine the indentation behavior? Why choose this behavior over the other PRs?

Salka1988:
I think that in these PR `indent` commands are not optional. What will we do if the user likes the way MdBook currently formats the text?
That's why I added a new option that is easy to use.
There is an option is to expand `Include(PathBuf, RangeOrAnchor, "/ indent/no_indent /")`.

mattburgess:
Just to add my opinion as a user here. The behaviour described in #1564 and #1626 is clearly incorrect so shouldn't require an option on the part of the user to get correct behaviour. The PR at #1718 implements exactly the solution that #1626 describes. So, if it were up to me, (which it clearly isn't :smile: ) I'd do the following:

* Close #1626 as a duplicate of #1564; the latter is just a subset of the case described in the former
* Close this PR in favour of #1718; as mentioned that PR requires no options and also includes test cases.

julio4:
Any progress on this PR?

Issue #1860, opened 2022-07-22
"`local_ctx` does not live long enough" compilation error on `rustc 1.64.0-nightly (62b272d25 2022-07-21)`


mitchmindtree:
First noticed downstream in one of our CI passes [here](https://github.com/FuelLabs/sway/pull/2359#issuecomment-1192097008), it appears the latest nightly has trouble building `mdbook`.

This includes all versions at least from `0.4.19` through to the current HEAD of `master`.

```
error[E0597]: `local_ctx` does not live long enough
   --> src/renderer/html_handlebars/helpers/navigation.rs:154:25
    |
154 |             t.render(r, &local_ctx, &mut local_rc, out)
    |                         ^^^^^^^^^^ borrowed value does not live long enough
155 |         })?;
    |         -
    |         |
    |         `local_ctx` dropped here while still borrowed
    |         borrow might be used here, when `local_rc` is dropped and runs the destructor for type `handlebars::RenderContext<'_, '_>`
    |
    = note: values in a scope are dropped in the opposite order they are defined

For more information about this error, try `rustc --explain E0597`.
error: could not compile `mdbook` due to previous error
```

The issue appears to have been introduced in the latest nightly (`rustc 1.64.0-nightly (62b272d25 2022-07-21)`) as the previous nightly (`rustc 1.64.0-nightly (d68e7ebc3 2022-07-20)`) builds successfully.

Yet to dig in to what's going on here, but just thought I'd open an issue to track this first.

mitchmindtree:
I've opened #1861 that appears to fix this.

I've also opened an issue upstream https://github.com/rust-lang/rust/issues/99591.

andrewdavidmackenzie:
Just ran into this too, on linux (Ubuntu) only it seems

ehuss:
Closing fixed by #1860.

PR #1861, opened 2022-07-22
Workaround rust nightly borrowcheck error (#1860)

mitchmindtree:
Surprisingly, this fixes the error filed at #1860!

This seems suspicious, perhaps indicative of a bug in Rust's non-lexical
lifetime handling on nightly?

The lifetimes in the `handlebars::Renderable::render` method signature
are quite complicated, and its unclear to me whether or not Rust is
catching some new safety edge-case that wasn't previously handled
correctly...

Possibly related to `drop` order, which I *think* is related to the
order of binding statements?

### DIFF
--- a/src/renderer/html_handlebars/helpers/navigation.rs
+++ b/src/renderer/html_handlebars/helpers/navigation.rs
@@ -149,8 +149,8 @@ fn render(
     _h.template()
         .ok_or_else(|| RenderError::new("Error with the handlebars template"))
         .and_then(|t| {
-            let mut local_rc = rc.clone();
             let local_ctx = Context::wraps(&context)?;
+            let mut local_rc = rc.clone();
             t.render(r, &local_ctx, &mut local_rc, out)
         })?;
 


andrewdavidmackenzie:
Do you plan to release a v0.4.21 on crates.io?

ehuss (approved):
Thanks so much for the quick fix!

Yea, this fix looks correct. I don't think it is an issue with the borrow checker. What happens is that the handlebars `render` function has this signature:

```rust
fn render<'reg: 'rc, 'rc>(
    &'reg self,
    registry: &'reg Registry<'reg>,
    context: &'rc Context,
    rc: &mut RenderContext<'reg, 'rc>,
    out: &mut dyn Output,
) -> Result<(), RenderError>;
```

Based on that signature, `RenderContext` is mutable and has a `'rc` lifetime, which means in theory it could receive references from `&'rc Context`. If `local_ctx` is dropped before `local_rc`, then in theory `local_rc` could contain a dangling reference into something from `local_ctx`. 

I don't think that happens. The `RenderContext` is mutable for other reasons.

I don't have time right now to do a release, but I will make one later today.

PR #1862, opened 2022-07-22
Deploy {x86_64,aarch64}-unknown-linux-musl binary

### LINKED ISSUE #1779: Prebuilt `x86_64-unknown-linux-gnu` binaries produce loader errors on systems with older glibc
Hi. I'm attempting to add support for `mdbook` to the [`taiki-e/install-action`](https://github.com/taiki-e/install-action) github action, which downloads prebuilt binaries of developer tools. The work is happening in https://github.com/taiki-e/install-action/pull/4. In theory, this should be fairly straightforward for mdBook, because you already include prebuilt binaries in the release.

Unfortunately, the build of the prebuilt binaries is done against `ubuntu-latest`. This means it will link against a very recent version of glibc. This will cause the loader to emit a runtime error from on systems which only have an older glibc. For example, [this run of CI](https://github.com/taiki-e/install-action/runs/5810893419?check_suite_focus=true) produces error output like the following:

```
mdbook: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found (required by mdbook)
mdbook: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by mdbook)
```

I believe there are a few options for how to address this. All of these should be changes only to your CI pipeline.

1. Use `ubuntu-18.04` to perform the build of the `x86_64-unknown-linux-gnu` binary. This should solve the issue for us, and should not break when used on systems with newer versions of glibc. This is probably the option I would take, since it seems like the most painless.

2. Provide a binary which statically links against libc, such as one made against the `x86_64-unknown-linux-musl` target. This could be done either instead of or in addition to the `x86_64-unknown-linux-gnu` binary.

    This may or may not be possible for `mdbook`, but if it is possible, I believe it would produce a binary that works on any linux system, regardless of libc version.

3. Do nothing. It's possible it's not worth the trouble for you to provide binaries that run on Linuxes with older versions of glibc, and that the CI test done in `taiki-e/install-action` is overly aggressive.

Thank you.

P.S. Sorry if I've misunderstood something, or if this is intentional somehow.

gifnksm:
Added functionality to `.github/workflows/deploy.yml` and associated scripts to deploy pre-built binaries for `x86_64-unknown-linux-musl` and `aarch64-unknown-linux-musl`.

The `x86_64-unknown-linux-musl` executable is statically linked, so it will work with older versions of glibc or the Alpine Linux container environment.
I also added the `aarch64-unknown-linux-musl` executable because I want mdBook to run in aarch64 Linux environments (using rust-lld as the linker makes cross-compilation easier).

This fixes #1779

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -9,19 +9,35 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        target:
+          - aarch64-unknown-linux-musl
+          - x86_64-unknown-linux-gnu
+          - x86_64-unknown-linux-musl
+          - x86_64-apple-darwin
+          - x86_64-pc-windows-msvc
+        include:
+          - target: aarch64-unknown-linux-musl
+            os: ubuntu-latest
+          - target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
+          - target: x86_64-unknown-linux-musl
+            os: ubuntu-latest
+          - target: x86_64-apple-darwin
+            os: macos-latest
+          - target: x86_64-pc-windows-msvc
+            os: windows-latest
     steps:
     - uses: actions/checkout@master
     - name: Install hub
       run: ci/install-hub.sh ${{ matrix.os }}
       shell: bash
     - name: Install Rust
-      run: ci/install-rust.sh stable
+      run: ci/install-rust.sh stable ${{ matrix.target }}
       shell: bash
     - name: Build and deploy artifacts
       env:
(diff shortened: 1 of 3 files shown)


gifnksm:
I tested this workflow in my repository and it works as expected: https://github.com/gifnksm/mdBook/releases/tag/test-release

gifnksm:
The CI failure does not seem to be related to the changes made in this pull request (see #1861).

ehuss (approved):
Thanks for the PR!

I'm fairly reluctant to add non-tier-1 targets, since there is risk that they will not operate correctly. However, I think the risk for these particular targets is not very high. Just a forewarning that if they cause any issues, they may be removed in the future.

PR #1866, opened 2022-07-28
Bump MSRV to 1.56 to get access to Rust 2021 crates

mgeisler:
I noticed in #1864 that mdBook does not yet use the Rust 2021 edition.
This 2021 edition was made stable in the 1.56 release on 2021-10-21.
The previous MSRV (1.54) was released on 2021-07-29.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,7 +32,7 @@ jobs:
           - build: msrv
             os: ubuntu-latest
             # sync MSRV with docs: guide/src/guide/installation.md
-            rust: 1.54.0
+            rust: 1.56.0
     steps:
     - uses: actions/checkout@master
     - name: Install Rust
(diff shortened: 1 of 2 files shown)


ehuss:
Is there something specific that needs this to be updated?  

I'm a bit confused to the relationship to #1062, as that is 3 years old.

There is #1831 to update the actual edition, but the PR currently has some failures. If someone else wants to revive it, I think that would be fine.  But otherwise I'm not sure if I'd bother updating the version without a need.

mgeisler:
> Is there something specific that needs this to be updated?
> 
> I'm a bit confused to the relationship to #1062, as that is 3 years old.

Hi @ehuss, I'm sorry, I don't know what I was copying before... I wanted to refer to #1864 and ended up copying something else. I've updated it now in the comment and will fix it in the commit too.

Thanks for the pointer to https://github.com/rust-lang/mdBook/pull/1831, I will be happy to see what needs to be fixed there.

mgeisler:
The same change (and more) is being made in https://github.com/rust-lang/mdBook/pull/1831.

mgeisler:
The edition was eventually bumped in https://github.com/rust-lang/mdBook/pull/1887.

PR #1873, opened 2022-08-13
[+] Add the ability to config the hostname and port in book.toml

tuyen-at-work:
This PR add the ability to specify the default hostname and port for serve the book in book.toml file.

## Why I need this feature?

Sometime I need to run multiple mdBook instances, but in order to do so, I need to set the port explicitly, otherwise new instances will try to reuse the same port with existing instance, which will be failed.

This PR allow to set default hostname/port for a book, so user doesn't need to set port every time they serve book in above use case. If user would like to override the port via command line, they still able to do so.

Other benefit is that when you run the book again and again, it will keep using old host/port you already choosen, without the need to remember it. By doing so, you will also have better browser history, i.e. you can filter them easier, avoid the case when at different time you run different books but share the same address. The consistent also help in case you have some browser externsion and you want to apply some filter rules to specific books only (i.e. inject some CSS/JS,...).

### DIFF
--- a/guide/src/format/configuration/general.md
+++ b/guide/src/format/configuration/general.md
@@ -17,6 +17,10 @@ edition = "2018"
 build-dir = "my-example-book"
 create-missing = false
 
+[serve]
+hostname = "localhost"
+port = "3000"
+
 [preprocessor.index]
 
 [preprocessor.links]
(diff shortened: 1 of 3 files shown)


mgeisler:
> This PR allow to set default hostname/port for a book, so user doesn't need to set port every time they serve book in above use case.

I'm just a random mdbook user, but to me, it seems like the hostname is per-machine configuration by its nature. This means, that if you put a `book.toml` file somewhere with a hard-coded hostname, I will most likely not be able to `mdbook serve` the book on my machine.

I don't know if mdbook can read environment variables, but they might be a better place to put per-machine configuration.

GuillaumeGomez:
The port can be overloaded with the CLI, so I don't think it's worth adding a new config for that. Closing then.

PR #1881, opened 2022-08-29
Remove unneeded `type` attribute for `<script>`

GuillaumeGomez:
I saw it when working https://github.com/rust-lang/rust/pull/101166: the default value of the `type` attribute of `<script>` is `text/javascript`, so no need to set it manually.


Dylan-DPC:
Thanks

PR #1884, opened 2022-09-01
Add support for watching additional directories

willcrichton:
When using mdBook preprocessors like [mdbook-quiz](https://github.com/willcrichton/mdbook-quiz), the book depends on files outside the source directory, such as TOML files describing quizzes. It would be very convenient as an author for the book to automatically rebuild when these files change. However, the `watch` and `serve` commands (via `trigger_on_change`) only watch a fixed set of directories.

Probably the ideal solution would be for preprocessors to register files to be watched, similar to how extensible build systems like [esbuild](https://esbuild.github.io/plugins/#on-resolve-results) manage plugins. However, this would be a big change with a lot of work. A much cheaper solution is to allow mdbook CLI users to pass desired directories-to-watch on the command line.

Therefore this PR adds a new flag to `watch` and `serve`, `--extra-watch-dirs`. For example, a user could write:

```
mdbook serve --extra-watch-dirs quizzes,snippets
```

The `--extra-watch-dirs` flag takes as input a comma-separated list of directories. Each of these directories is added to the file watcher, so the book is regenerated when a file in these directories is changed.

### DIFF
--- a/guide/src/format/configuration/general.md
+++ b/guide/src/format/configuration/general.md
@@ -87,6 +87,7 @@ This controls the build process of your book.
 build-dir = "book"                # the directory where the output is placed
 create-missing = true             # whether or not to create missing pages
 use-default-preprocessors = true  # use the default preprocessors
+extra-watch-dirs = []             # directories to watch for triggering builds
 ```
 
 - **build-dir:** The directory to put the rendered book in. By default this is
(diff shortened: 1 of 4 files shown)


ehuss:
Might it be better to add this to `book.toml` instead of a CLI flag? That way, all users will get the benefit of proper `watch` behavior without needing to know a specific CLI flag to pass?

willcrichton:
Ah yeah that's a good suggestion. Plus it makes the implementation simpler! Now the configuration looks like:

```toml
[build]
extra-watch-dirs = ["quizzes", "snippets"]
```

ehuss:
Thanks!  It looks like CI is failing?

Also, when adding new config options, the documentation needs to be updated, too.

willcrichton:
Ah, didn't realize that CI was using an older version of Rust. That's been fixed. I just added documentation.

ehuss (approved):
Thanks!

PR #1887, opened 2022-09-09
Require Rust 2021 edition

mgeisler:
This allows us to clean up and simplify the code.

The changes were originally written by @Dylan-DPC in #1831. This PR fixes a few build errors in that PR, thanks to the helpful comments made by @ISSOtm. I've also rebased the changes onto the latest `master`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,7 +32,7 @@ jobs:
           - build: msrv
             os: ubuntu-latest
             # sync MSRV with docs: guide/src/guide/installation.md
-            rust: 1.54.0
+            rust: 1.56.0
     steps:
     - uses: actions/checkout@master
     - name: Install Rust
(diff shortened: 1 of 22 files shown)


ISSOtm (approved):

Dylan-DPC:
@ehuss do we want to do this in a new major version?

.github/workflows/main.yml
```diff
@@ -31,8 +31,7 @@ jobs:
             rust: stable
           - build: msrv
             os: ubuntu-latest
-            # sync MSRV with docs: guide/src/guide/installation.md
-            rust: 1.54.0
+            rust: 1.56.0
```
ehuss, line 34: It looks like there was a comment removed from here. Can we make sure it is retained?

ehuss:
> @ehuss do we want to do this in a new major version?

No, I think a normal release is fine.

.github/workflows/main.yml
mgeisler: Definitely, it should be fixed now.

ehuss (approved):
Thanks!

PR #1889, opened 2022-09-10
Simplify the use of `Option::and_then`

mgeisler:
I found a few places where `Option::and_then` could be simplified (in my opinion) with `?` or with `match`.


ehuss (approved):
Seems reasonable, thanks!

mgeisler:
Thanks! :)

PR #1894, opened 2022-09-20
feat: use once_cell instead of lazy_static

wendajiang:


ehuss:
Thanks!  This seems to have several other changes, so it's not quite clear to me:
* Updates `assert_cmd`. Is there a reason this is updated at the same time?
* It looks like all dependencies have been updated in Cargo.lock?
* It looks like the msrv documentation hasn't been updated.

wendajiang:
> Thanks! This seems to have several other changes, so it's not quite clear to me:
> 
> * Updates `assert_cmd`. Is there a reason this is updated at the same time?
> * It looks like all dependencies have been updated in Cargo.lock?
> * It looks like the msrv documentation hasn't been updated.

` assert_cmd ` is out-of-habit use the newest version. And the dependencies is from this https://github.com/rust-lang/mdBook/pull/1844 and both I revert the dependencies update.

ehuss (approved):
Thanks!

Issue #1574, opened 2021-06-09
Add compatibility tests


ehuss:
I think it would be good to have some tests that validate the preprocessor and renderer interfaces remain compatible with plugins built with older versions. That could help catch regressions like #1571 and others.

earv1:
Hey there, I'm interested in your project as a first time contributor. I started looking at this, but it seems like this might already be solved by #1897. Is there anything still missing?

ehuss:
That doesn't quite do what I was thinking about for this issue. That has a hard-coded input that applies to the current preprocessor. The intent here is that the *current* mdbook's output can be consumed and processed by an old preprocessor (and renderer).

I'm not sure how the test could be best structured, but essentially I was thinking that there would be a preprocessor and renderer that just assert that their input matches a hard-coded value. The test would run mdbook against that preprocessor/renderer combination, and if either fail, then the test fails. That would help detect if mdbook starts generating different structures.

I'm not sure how maintainable that is, I haven't thought about it too much.

PR #1897, opened 2022-09-24
Add test code to show preprocessor developers what the interface data looks like

noritada:
This PR adds test code to the nop-preprocessor example.

If you have better test data or test methods, I would appreciate your suggestions.

### Purpose

The purpose of this PR is showing preprocessor developers what the interface (input) data looks like and how to test the preprocessing results.  Currently, the documentation states that JSON data consisting of a two-element array `[context, book]` is passed from the mdbook, but developers need to actually dump the input themselves to get the actual input data.

### Consideration

Since this PR hard-codes an example of current interface data, it is necessary to keep up with the changes if the interface JSON format changes in the future.

This may not be the best solution to compatibility testing (#1574), but it may be part of it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,6 +39,8 @@ jobs:
       run: bash ci/install-rust.sh ${{ matrix.rust }}
     - name: Build and run tests
       run: cargo test
+    - name: Test examples
+      run: cargo test --examples
     - name: Test no default
       run: cargo test --no-default-features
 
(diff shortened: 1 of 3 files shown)


ehuss:
Thanks! I think the extra step of running `cargo test --examples` can be skipped by adding this to Cargo.toml:

```toml
[[example]]
name = "nop-preprocessor"
test = true
```

That way, users running locally can still rely on just running `cargo test` to cover all of the tests.

noritada:
@ehuss Thank you for the suggestion! I agreed and have made changes to use the configuration so that users don't forget to run the test.

ehuss (approved):
Thanks!

Issue #1813, opened 2022-06-01
Feature Suggestion: Keyboard Keys Rendering


epomatti:
It would be nice to have keyboards key markup like this:

<kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>Space</kbd>

ISSOtm:
Can/should we use the same style entirely, or should we come up with something similar?

epomatti:
hmmm what is "same style"?

ISSOtm:
The style applied to `<kbd>` elements on GitHub. I'm not sure if we would be able to copy their CSS; if not, I'm no good designer, so I would have trouble coming up with something similar but different.

epomatti:
Maintainability comes first then, simpler would be a better choice.

ISSOtm:
My question isn't so much about maintainability as it is about copyright.

brettchalupa:
I'd like to take on this issue if it's still available.

PR #1906, opened 2022-10-06
Add styles for <kbd> elements

brettchalupa:
Allows for special styles to call them out since they're different than normal text and different than code. They can make use of styles they inherit for font style and weight.

Notes on changes:

- Added new CSS variables for reused elements
- The font-* rules are separate for each aspect so that they can inherit bold/italic/etc
- I picked the color vars based on what looks good, not necessarily based on logical name reuse

Closes https://github.com/rust-lang/mdBook/issues/1813

## Screenies

image
Screen Shot 2022-10-06 at 4 07 32 PM
Screen Shot 2022-10-06 at 4 07 28 PM
Screen Shot 2022-10-06 at 4 07 24 PM
Screen Shot 2022-10-06 at 4 07 20 PM

And here's how it looks in a list with bold styles applied:

Screen Shot 2022-10-06 at 4 07 13 PM

### DIFF
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -22,8 +22,8 @@ body {
 }
 
 code {
-    font-family: "Source Code Pro", Consolas, "Ubuntu Mono", Menlo, "DejaVu Sans Mono", monospace, monospace !important;
-    font-size: 0.875em; /* please adjust the ace font size accordingly in editor.js */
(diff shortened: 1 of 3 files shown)


ISSOtm (approved):

ehuss (approved):
Thanks!

Issue #1187, opened 2020-04-13
Use an optional .mdbookignore file to specify files to not copy to the output directory


andrewdavidmackenzie:
Current Behaviour
===
Currently almost all files in the source directory are copied to the output directory, often copying unneeded or undesired files, which maybe uploaded to gh-pages or otherwise "published"

Proposed Change
===
Allow an optional “.mdbookignore” file (like .gitignore) to allow specifying patterns of files to ignore and hence NOT get copied during the build.

Implementation
===
If you provide comments, and the general idea seems acceptable, I would prepare a PR for this.

I would use the existing "ignore" crate from the ripgrep project, and using the provided method to override the name of the ignore file from ".gitignore" to ".mdbookignore"

ehuss:
I might prefer just making it a field in the config, like `exclude = ["src/foo/*"]`.

But a question is, why add files to the src directory, but don't want them as part of the book? Why not just put those files somewhere else?

andrewdavidmackenzie:
On the why: It's due to the structure of my book. 

My book has a lot of it's content in a docs subfolder, BUT, it also includes README.md and other markdown files from multiple other sub-folders (it's a workspace rust project with multiple crates) so the docs files are mixed in with the code tree structure.

With the fix to avoid the recursive directory copy you merged recently I can now have my book.toml and SUMMARY.md in the root folder of the project, and SUMMARY.md includes entries for those multiple crates in sub-directories.

This leads to all sorts of files being copied into the output folder, such as:
```.idea, .iml, .git, .sh, Cargo.toml and .lock, .json, *.rs from source folders, etc```

I currently have a make target that deletes all the extra before doing the gh-pages deploy, cutting down the size of the upload considerably.

GH Project: [flow](https://github.com/andrewdavidmackenzie/flow)
GH Pages Deployed [here](http://andrewdavidmackenzie.github.io/flow/)

**Comment by williamd1k0** [comment]
> But a question is, why add files to the src directory, but don't want them as part of the book? Why not just put those files somewhere else?

In my case, I'm using some helper files in `{{#include file:anchor}}`. I just need parts of these files to be expanded, I don't need them in the output directory.

> I might prefer just making it a field in the config, like exclude = ["src/foo/*"].

That would be enough.

**Comment by andrewdavidmackenzie** [comment]
To give my own specific example of the files that are copied and undesirable, here is the additiional make target I have had to add to cleanup the output directory before all of these are uploaded to gh_pages:

```
[...]

Issue #1156, opened 2020-03-03
Feature: suppress copying files by configuration


FrankHB:
I find no options to ignore specific files to be built. Similar problems for `watch`/`serve` commands were mentioned in #831, and the improvement of the `[build]` configuration was also suggested there. But I think this is still a different issue better to be tracked separately.

A workaround to this problem is to specify `src` to a subdirectory. This can work, but far from perfect. It is not pleasant for transition from a repository of documents (consisting of plenty of `.md` files in the root directory) already checked in, as moving the files from the root to the subdirectory leads to much noise in the VCS history.

(Moreover, relying on `.gitignore` is problematic. What if there is only `.hgignore`, for example?)

FrankHB:
To be more specific, [Butbucket](https://bitbucket.org) serves the project wikis as Mercurial repositories, but the Markdown features are limited. So, I decide to rearrange the documents in some other forms (hopefully with HTML output). However, at least the original `.md` documents still work without any changes online. If `mdbook` does not copy files (like my local `.hg` directory) by default, I can just start to work by simply adding `book.toml` and `SUMMARY.md`. This is not possible without some dirty hacks, and the workaround mentioned above does not work well because it breaks the file layout and I have to edit some internal links to make sure it can still work as a wiki. Currently I work it around by symbolic links in a separate `src` directory out of the tree, which is still a bit annoying for maintenance.

FrankHB:
Since the default behavior [may be expected](https://github.com/rust-lang/mdBook/issues/441), I wish the configuration opt-in. This should not break any existing usage.

tv42:
I would love to use `src = "."` and not have unnecessary things like my build script or .gitignore copied into the output directory.

Bergmann89:
Any updates on this issue? Would love to have this feature too! :)

PR #1908, opened 2022-10-13
Ignore files listed in `.mdbookignore` during build

Bergmann89:
Implementation according to issue #1187 and #1156

### DIFF
--- a/guide/src/format/configuration/renderers.md
+++ b/guide/src/format/configuration/renderers.md
@@ -322,7 +322,19 @@ This will generate an HTML page which will automatically redirect to the given l
 
 When fragment redirects are specified, the page must use JavaScript to redirect to the correct location. This is useful if you rename or move a section header. Fragment redirects work with existing pages and deleted pages.
 
-## Markdown renderer
+### `.mdbookignore`
+
+You can use a `.mdbookignore` file to exclude files from the build process.
+The file is placed in the `src` directory of your book and has the same format as
+[`.gitignore`](https://git-scm.com/docs/gitignore) files.
+
+For example:
+```
+*.rs
+/target/
+```
+
+## Markdown Renderer
 
 The Markdown renderer will run preprocessors and then output the resulting
 Markdown. This is mostly useful for debugging preprocessors, especially in
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1029,6 +1029,7 @@ name = "mdbook-core"
 version = "0.5.2"
 dependencies = [
  "anyhow",
+ "ignore",
  "regex",
  "serde",
  "serde_json",
@@ -1070,6 +1071,7 @@ dependencies = [
  "handlebars",
  "hex",
  "html5ever 0.38.0",
+ "ignore",
  "indexmap",
  "mdbook-core",
  "mdbook-markdown",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,17 +89,17 @@ rust-version.workspace = true
 
 [dependencies]
 anyhow.workspace = true
-clap.workspace = true
 clap_complete.workspace = true
+clap.workspace = true
+ignore.workspace = true
 mdbook-core.workspace = true
 mdbook-driver.workspace = true
 mdbook-html.workspace = true
 opener.workspace = true
-tracing.workspace = true
 tracing-subscriber.workspace = true
+tracing.workspace = true
 
 # Watch feature
-ignore = { workspace = true, optional = true }
 notify = { workspace = true, optional = true }
 notify-debouncer-mini = { workspace = true, optional = true }
 pathdiff = { workspace = true, optional = true }
@@ -125,7 +125,7 @@ walkdir.workspace = true
 
 [features]
 default = ["watch", "serve", "search"]
-watch = ["dep:notify", "dep:notify-debouncer-mini", "dep:ignore", "dep:pathdiff", "dep:walkdir"]
+watch = ["dep:notify", "dep:notify-debouncer-mini", "dep:pathdiff", "dep:walkdir"]
 serve = ["dep:futures-util", "dep:tokio", "dep:axum", "dep:tower-http"]
 search = ["mdbook-html/search"]
 
--- a/crates/mdbook-core/Cargo.toml
+++ b/crates/mdbook-core/Cargo.toml
@@ -9,6 +9,7 @@ rust-version.workspace = true
 
 [dependencies]
 anyhow.workspace = true
+ignore.workspace = true
 regex.workspace = true
 serde.workspace = true
 serde_json.workspace = true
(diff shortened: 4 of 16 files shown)


ehuss:
Thanks for the PR!  Per the comment at https://github.com/rust-lang/mdBook/issues/1187#issuecomment-612970155, can this be a setting in `book.toml` instead?

Bergmann89:
I can add this to the config too, but I would like to have this feature as ignore file.

schungx:
An ignore file has the benefit of not polluting the main config file and so it can really be done on a user-by-user basis.

The pitfall is... of course... yet another file.

Bergmann89:
Just realized, that the `gitignore` crate has a bug :/ The following patterns will not ignore any file, but it should ignore everything except the `mdbook-plantuml-img` directory:

```
/*
!mdbook-plantuml-img
```

`gitignore` was not maintained since two years. Maybe it's better to have an own implementation for the ignore files or use the config only. If we add this to the config, it would be nice to have a setting for `include` and a `exclude`.

Bergmann89:
The [`ignore`](https://docs.rs/ignore/0.4.18/ignore/gitignore/struct.Gitignore.html) crate work as expected.

Bergmann89:
I've pushed a new version that uses the `ignore` crate instead of `gitignore`.

With the `ignore` crate we can do some further improvements: 
- Remove the custom implementation of file walk `copy_files_except_ext` with [`ignore::Walk`](https://docs.rs/ignore/latest/ignore/struct.Walk.html) which will have the benefit, that also ignore files in the sub directories will be evaluated.
- Use [`ignore::Walk`](https://docs.rs/ignore/latest/ignore/struct.Walk.html) to evaluate all `.gitignore` files (not only in the root directory) to filter files in the `watch` command.

ehuss:
Ok, seems fair to have a separate file. I personally prefer to keep things in one place, and to not have too many files, just to keep things simple (there's one place to describe the config).

I think switching to `ignore` sounds great.

Can you add some documentation and tests?

Bergmann89:
Any remarks? Or can we merge this? :)

guide/src/format/configuration/renderers.md
```diff
@@ -279,6 +279,16 @@ The value can be any valid URI the browser should navigate to (e.g. `https://rus
 This will generate an HTML page which will automatically redirect to the given location.
 Note that the source location does not support `#` anchor redirects.
 
+### `[.mdbookignore]`
+
+You can use a `.mdbookignore` file to exclude files from the build process. The file is places in the `src` directory of your book and has the format known from [`.gitignore`](https://git-scm.com/docs/gitignore) files.
```
ehuss, line 284:
```suggestion
You can use a `.mdbookignore` file to exclude files from the build process. The file is placed in the `src` directory of your book and has the same format as [`.gitignore`](https://git-scm.com/docs/gitignore) files.
```

Also, can you split it to one sentence per line to make diffs easier to manage?

guide/src/format/configuration/renderers.md
```diff
@@ -279,6 +279,16 @@ The value can be any valid URI the browser should navigate to (e.g. `https://rus
 This will generate an HTML page which will automatically redirect to the given location.
 Note that the source location does not support `#` anchor redirects.
 
+### `[.mdbookignore]`
```
ehuss, line 282: I'm not sure why this is in square brackets. The other sections are like that because they are describing TOML sections of `book.toml`.

```suggestion
### `.mdbookignore`
```

ehuss (commented):
Overall I'm finding the difference between .gitignore and .mdbookignore a little confusing. 

I'm guessing that `watch` does not ignore .mdbookignore files for a reason? Can you say more about why that choice was made?

Bergmann89:
> Overall I'm finding the difference between .gitignore and .mdbookignore a little confusing.

What exactly? We can also add this to the config if you want to, but I need any solution to ignore specific files processed by mdbook.

> I'm guessing that `watch` does not ignore .mdbookignore files for a reason? Can you say more about why that choice was made?

Should not. If this is the case, this is a mistake on my side. I will check this.

I'm currently really busy in my own projects. I will resolve all of your findings once I have some more time for this. This may take a couple of weeks.

Greetings,
Bergmann89.

ehuss:
> What exactly? We can also add this to the config if you want to, but I need any solution to ignore specific files processed by mdbook.

There is a subtle difference between them (as indicated by the confusion around `watch`). `gitignore` files are copied, but not watched. `mdbookignore` files are not copied but are watched.  The documentation doesn't make the differences clear to me, or explain why one would use one over the other. I might suggest in places where gitignore is mentioned in the documentation to make links to the mdbookignore section ("gitignore will only affect which files are watched for changes. To avoid copying certain source files to the output, see [mdbookignore]" or something like that).

I'm also wondering if copying should also avoid git ignored files. However, I think that will require opening the git repo and verifying which files are actually ignored similar to how Cargo itself works, but that is a pretty big hairy mess.

Bergmann89:
Sorry, I had no time to make this ready yet. Feel free to fix the findings above :)

src/utils/fs.rs
```diff
@@ -94,13 +95,13 @@ pub fn copy_files_except_ext(
     to: &Path,
     recursive: bool,
     avoid_dir: Option<&PathBuf>,
-    ext_blacklist: &[&str],
```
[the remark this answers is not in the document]
Bergmann89: Done. I've added a new public function `copy_files_except_ignored` that is called by `copy_files_except_ext` to not have code duplication.

src/book/mod.rs
```diff
@@ -856,4 +857,21 @@ mod tests {
         let got = preprocessor_should_run(&BoolPreprocessor(should_be), &html, &cfg);
         assert_eq!(got, should_be);
     }
+
+    #[test]
+    fn build_test_book() {
```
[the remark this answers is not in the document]
Bergmann89: Fixed.

src/utils/fs.rs
```diff
@@ -128,6 +129,13 @@ pub fn copy_files_except_ext(
                 }
             }
 
+            if let Some(ignore) = ignore {
+                let path = entry.path();
+                if ignore.matched(&path, path.is_dir()).is_ignore() {
+                    continue;
+                }
+            }
```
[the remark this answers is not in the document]
Bergmann89: Fixed.

guide/src/format/configuration/renderers.md
Bergmann89: Fixed.

guide/src/format/configuration/renderers.md
Bergmann89: Fixed.

Bergmann89:
@ehuss I (finally) had some time to resolve the findings and rebased to the current master. Might there be some time for another review? :)

Bergmann89:
Friendly reminder to get this merged. No merge conflicts and checks are passing :smiley:

JL710:
@ehuss another friendly reminder that this PR would like to get merged. 😅

jhult:
I think this PR could resolve #2246 as well (and is more configurable than #2247 and #2991).

PR #1911, opened 2022-10-14
fix copying invisible text

yoyomo:
I noticed this bug whenever I copied any code from a book, in rust, it would copy a 
```
#![allow(unused_variables)]
fn main() {
...
}
```
or
```
#![allow(unused)]
fn main() {
...
}
```

This fixes that, by copying the `innerText`, instead of the `textContent`, which included invisible text (`display: none`, or `.boring` class)


Dylan-DPC:
Thanks

PR #1912, opened 2022-10-16
Prefere .bookignore when build too

podsvirov:
Reuse watch command experience (#1044) and process .bookignore (as .gitignore) if prezent.

*MOTIVATION*: I'm just type `src = "."` in [book.toml](https://github.com/podsvirov/mdbook-pikchr/blob/v0.1.2/book.toml) and understand that something is wrong...

*SOLUTION*: I'm make this PR and add [.bookignore](https://github.com/podsvirov/mdbook-pikchr/blob/master/.bookignore) file and life is good, because `mdbook build` no more copy `target` to `book`! :-)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,10 +34,10 @@ shlex = "1"
 tempfile = "3.0"
 toml = "0.5.1"
 topological-sort = "0.1.0"
+gitignore = "1.0"
 
 # Watch feature
 notify = { version = "4.0", optional = true }
-gitignore = { version = "1.0", optional = true }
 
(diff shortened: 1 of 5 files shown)


brettchalupa:
These efforts seem similar to https://github.com/rust-lang/mdBook/pull/1908, which is adding support for an `.mdbookignore` file option too.

podsvirov:
My solution is very small and not change any dependencies, but looks like the `ignore` crate more popular and use it will add some improvements.

I'm fine with any solution that allows to filter files at build time.

Issue #984, opened 2019-07-20
Internal links to README.md are broken


lily-commure:
`README.md` gets translated to `index.html`, but internal links (except for those in `SUMMARY.md`) don't get translated.  So they point to a non-existent `README.html` path.  mdBook should either generate both `index.html` and `README.html` or translate internal links so that they point to `index.html` instead.

WofWca:
Ok, here's an example:
If you want to add a link to some page, you can type `[Some page](some_dir/some_page.md)`, but if you try to link README this way, the link will be broken.

nihaals:
Repro steps
```
docker run -it --rm rust:slim bash
# cargo install mdbook
# mkdir /app
# cd /app
# mdbook init
...
# printf "[Test](./test/README.md)\n[Test 2](./test.md)" > src/SUMMARY.md
# printf "[Test](./test/README.md)" > src/test.md
# mkdir src/test
# printf "# Test" > src/test/README.md
# mdbook build
# cat book/test.html | grep README.html
<p><a href="./test/README.html">Test</a></p>
# ls book/test
index.html
```

YushiOMOTE:
duplicated issue? https://github.com/rust-lang/mdBook/issues/1268

ehuss:
Yea, both issues are essentially the same.  I'll keep this one, but there is some more context and discussion in #1268 that is worth reading.

sanmai-NL:
@umdstu You can upvote #1921

bitdivine:
Opinion: Mapping README.md to index.html seems fine, however _if_ that is done then I think any links to `README.md` also need to be handled.  The options seem to be:
* Update the links
* Include a README.html in the output that is a copy of index.html or that redirects to index.html

carlocorradini:
Any update? 😥

cmyam:
After disabling the preprocessor, `README.md` generates both the expected `README.html` and an additional `index.html` copy for preview purposes.  
﻿
I believe this is a good practice and should be the default behavior when preprocessors have limited support for link handling.

CCJK123:
For anyone who encounters this issue in the future, the current possible workarounds seem to be the following:


| No. | Description                                             | Example                                                                                                                                                         | Remarks                                                                                                                                                                                                                         |
| --- | ------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
[...]

PR #1921, opened 2022-11-08
Fix relative links to README.md files

### LINKED ISSUE #1920: Bug: relative links to 'README.md' files don't resolve
Linking to other markdown files in the book's source generally works as expected. The exception are links to 'README.md' files. Clicking on such links in the HTML version of the book results in a 404 (Document not found) error.

Steps to reproduce:

1. Create a book's source as follows:

    `src/book.toml`:
    ```toml
    [book]
    authors = []
    language = "en"
    multilingual = false
    src = "src"
    ```

    `src/SUMMARY.md`:
    ```markdown
    # Summary

    - [Chapter 1](./chapter_1/README.md)
    - [Chapter 2](./chapter_2/README.md)
        - [Chapter 2.1](./chapter_2/chapter_2.1.md)
    ```

    `src/chapter_1/README.md`:
    ```markdown
    # Chapter 1

    - Link to [Chapter 2](../chapter_2/README.md)
    - Link to [Chapter 2.1](../chapter_2/chapter_2.1.md)
    ```

    `src/chapter_2/README.md`:
    ```markdown
    # Chapter 2
    ```

    `src/chapter_2/chapter_2.1.md`:
    ```markdown
    # Chapter 2.1
    ```

2. Build and serve the book: `mdbook serve`

3. Open the book in your browser: `http://localhost:3000`

4. Click on the links in Chapter 1: while the link to Chapter 2.1 works fine, the link to Chapter 2 doesn't.

chrstnst:
Fixes #1920

I suggest adding to the 'index' preprocessor a call to swap the filenames in links in the chapters' contents.

### DIFF
--- a/src/preprocess/index.rs
+++ b/src/preprocess/index.rs
@@ -40,6 +40,8 @@ impl Preprocessor for IndexPreprocessor {
                         path.set_file_name("index.md");
                     }
                 }
+
+                ch.content = convert_links_to_readme_files(&ch.content);
             }
         });
 
@@ -78,6 +80,22 @@ fn is_readme_file<P: AsRef<Path>>(path: P) -> bool {
     )
 }
 
+fn convert_links_to_readme_files(s: &str) -> String {
+    static README_LINK: Lazy<Regex> = Lazy::new(|| {
+        Regex::new(r"(?i)(?P<before>\[.+\]\([^:]*)readme(?P<after>\.md(?:.{0}|#.*)\))").unwrap()
+    });
+
+    README_LINK
+        .replace_all(s, |caps: &regex::Captures<'_>| {
+            format!(
+                "{before}index{after}",
+                before = &caps["before"],
+                after = &caps["after"],
+            )
+        })
+        .to_string()
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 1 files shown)


chrstnst:
Fixes #984

sanmai-NL:
@ehuss Is this change easy to review and accept?

ehuss:
I would prefer to not use regular expressions to translate the links. Can the translation be done in `adjust_links` or somewhere like that?

CaydenPierce:
Would love to see this merged.

sanmai-NL:
@CaydenPierce would you be able to make the requested improvement to the original PR, using a more robust/performant matching algorithm than regular expression matching?

pnax:
Hi, I've been reading the discussions on this topic and I'm also interested in seeing mdBook incorporate a solution for this issue.

> Can the translation be done in adjust_links or somewhere like that?

I had a look and came up with something that extends the existing regular expression in `adjust_links`. However this would apply to all links, even if the 'index' preprocessor has been disabled. I'm sharing it in case this can help moving forward on this.

```diff
diff --git a/src/utils/mod.rs b/src/utils/mod.rs
index 9156916..d0cda17 100644
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -95,7 +95,7 @@ pub fn unique_id_from_content(content: &str, id_counter: &mut HashMap<String, us
 fn adjust_links<'a>(event: Event<'a>, path: Option<&Path>) -> Event<'a> {
     static SCHEME_LINK: Lazy<Regex> = Lazy::new(|| Regex::new(r"^[a-z][a-z0-9+.-]*:").unwrap());
     static MD_LINK: Lazy<Regex> =
-        Lazy::new(|| Regex::new(r"(?P<link>.*)\.md(?P<anchor>#.*)?").unwrap());
+        Lazy::new(|| Regex::new(r"(?P<link>.*?)(?P<readme>README)?\.md(?P<anchor>#.*)?").unwrap());
 
     fn fix<'a>(dest: CowStr<'a>, path: Option<&Path>) -> CowStr<'a> {
         if dest.starts_with('#') {
@@ -126,7 +126,16 @@ fn adjust_links<'a>(event: Event<'a>, path: Option<&Path>) -> Event<'a> {
             }
 
             if let Some(caps) = MD_LINK.captures(&dest) {
-                fixed_link.push_str(&caps["link"]);
+                let link = &caps["link"];
+                fixed_link.push_str(link);
+                // "Links to README.md will be converted to index.html"
+                if let Some(readme) = caps.name("readme") {
+                    if link.is_empty() || link.ends_with('/') {
+                        fixed_link.push_str("index");
+                    } else {
+                        fixed_link.push_str(readme.as_str());
+                    }
+                }
                 fixed_link.push_str(".html");
                 if let Some(anchor) = caps.name("anchor") {
                     fixed_link.push_str(anchor.as_str());
```

Several questions I can think of:
1. How to deal with the fact that the 'index' preprocessor can be run or not? Maybe the logic should live in `src/preprocess/index.rs` and make calls to `adjust_links`.
2. "_I would prefer to not use regular expressions to translate the links._": Is it ok to extend the regex in `adjust_links`?

bitdivine:
Maybe `README.md` could be mapped to `index.html` AND `README.html`.  This way the links don't need to be updated.  Just an alternative as changing the links seems to be complicated.  I don't believe that having `README.md` should be a problem for users.  After all, users can have either paths ending in `/` or in `/index.html`, so why not `/README.html` as well?  It might not look as cool as having URLs end in `/` but it is simple and robust.

sanmai-NL:
> Maybe `README.md` could be mapped to `index.html` AND `README.html`.  This way the links don't need to be updated.  Just an alternative as changing the links seems to be complicated.  I don't believe that having `README.md` should be a problem for users.  After all, users can have either paths ending in `/` or in `/index.html`, so why not `/README.html` as well?  It might not look as cool as having URLs end in `/` but it is simple and robust.

That would work great, but the canonical address should be recorded in the HTML to prevent duplicate content for indexers like online search engines.

gluax:
Hey all, what's the status on getting this fix in?

PR #1924, opened 2022-11-12
Better HTML5 conformance

averms:
Four minor changes to the templates to improve HTML5 conformance. Here is the commit log

```
State canonical URL using <link> not <meta>.

The Google SEO docs describe using the link element to specify the canonical
page in a redirect. They don't mention anything about <meta>.

----

Fix formatting of content attr in <meta http-equiv=refresh>

According to the HTML specification[1], there needs to be a space after the
semicolon and the URL must be unquoted.

[1]: https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-http-equiv-refresh

----

Don't generate obsolete charset attrs in <script>.

According to the HTML specification[1], the charset attribute is obsolete in
script elements.

[1]: https://html.spec.whatwg.org/multipage/obsolete.html#HTMLScriptElement-partial

----

Don't generate redundant <meta http-equiv=content-type>.

Quoting from the HTML specification[1]:

> A document must not contain both a meta element with an http-equiv attribute
> in the Encoding declaration state and a meta element with the charset
> attribute present.

So we remove the <meta> with the http-equiv attribute from our template.

[1]: https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-http-equiv-content-type
```

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -15,7 +15,6 @@
         <!-- Custom HTML head -->
         {{> head}}
 
-        <meta content="text/html; charset=utf-8" http-equiv="Content-Type">
         <meta name="description" content="{{ description }}">
         <meta name="viewport" content="width=device-width, initial-scale=1">
         <meta name="theme-color" content="#ffffff" />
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks!

Issue #1919, opened 2022-11-06
HTML errors


coveritytest:
The generated HTML has still a few errors, see: https://validator.w3.org/nu/?doc=https%3A%2F%2Frust-lang.github.io%2FmdBook%2Fcli%2Fclean.html

ulivz:
Could you provide your documentation source code?

PR #1925, opened 2022-11-13
Fixed html errors caused by redundant `meta[http-equiv]`

ulivz:
Quoting from the HTML specification<sup>[1]</sup> about "Specifying the document's character encoding":

  A document must not contain both a `meta` element with an `http-equiv` attribute
  in the `Encoding declaration state` and a `meta` element with the `charset`
  attribute present.

Closes: #1919

[1]: https://html.spec.whatwg.org/dev/semantics.html#charset

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -15,7 +15,6 @@
         <!-- Custom HTML head -->
         {{> head}}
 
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks!  Closing as this is superseded by #1924.

PR #1929, opened 2022-11-19
Update docs link in Cargo.toml to HTTPS version

Benjins:
Currently, the documentation link is specified in `Cargo.toml` as:

```
documentation = "http://rust-lang.github.io/mdBook/index.html"
```

which propagates to [crates.io](https://crates.io/crates/mdbook), and if users click on the docs link there they get the non-TLS version. Not likely to cause major issues, but still best practice to use the HTTPS version since it's there


ehuss (approved):
Thanks!

Issue #1234, opened 2020-05-21
Break out the core book from everything else


Michael-F-Bryan:
In the same vein as #1211 and #1076, a *massive* pain point when developing plugins is that you need to import `mdbook` (which is fine) and this pulls in half of crates.io as transitive dependencies (which is not).

My thoughts are we could have a `mdbook-core` crate containing the bare minimum needed to load a book from disk, plus common traits:

- the `Book` type and things it contains (`BookItem`, `Chapter`, etc.)
- the `SUMMARY.md` parser and related code for loading a book from disk
- the `Renderer` trait
- the `Preprocessor` trait
- the `CmdRenderer` - because it only requires `std` and we'll already be pulling in `serde`

This then opens the way for making the HTML renderer *just another renderer* (e.g. by installing a `mdbook-html` binary alongside `mdbook`).

It's quite a big change in terms of work and backwards compatibility (although we could probably re-export items from `mdbook-core` in `mdbook`?) and I'm happy with doing most of the grunt work, but thought I'd get some thoughts before making changes. 

Does this seem like a reasonable idea? Or would it just be a bunch of busy work which is going to break a lot of in-flight PRs and waste time?

ehuss:
I think it would be great to have a more stripped-down core library. I'm uncertain if it is necessary to make it a separate crate. My concern is that adds some maintenance overhead. Would it be OK to just use features more aggressively? I assume the main dependencies to avoid are:

- clap (I didn't know this was always built!)
- handlebars
- env_logger?
- pulldown-cmark? Is this possible?
- regex
- tempfile, shlex, other small ones

That seems relatively easy to make those optional, default features.

Michael-F-Bryan:
I think the biggest offenders are all the extras used by the binary, and the HTML renderer. For example, a library doesn't actually need to instantiate a logger (`env_logger`), parse arguments (`clap`), or spin up a web server.

From what I can tell the only thing you really need for the `Book` and configuration are pulldown-cmark (which is really cheap because it's just text processing) and serde (expensive, but configuration is coupled to it). Likewise, the `Renderer` and `Preprocessor` traits a result just dumb objects, possibly with a `#[derive(Serialize, Deserialize)]` or two.

> I'm uncertain if it is necessary to make it a separate crate. My concern is that adds some maintenance overhead

I'm not sold on this... We use workspaces a lot to keep compile times down and maintain separation in work projects and in practice it's no different from putting things into modules, they're just different folders on disc. 

[...]

PR #1937, opened 2022-11-25
Move binary-only dependencies under a feature

willcrichton:
### Problem
When mdBook is used as a library for preprocessors, preprocessors need far less functionality than for the main mdbook binary. However, even with `default-features = false`, mdbook currently has 91 transitive dependencies. 

### Solution
This PR reduces the number of transitive dependencies from 91 to 46 with `default-features = false`, while maintaining enough API surface for preprocessors to function. It introduces a new default `"full"` feature that activates several dependencies like `chrono` and `handlebars`. This feature then gates the compilation of several modules and functions.

To reduce the usage of `#[cfg(feature = "full")]`, this PR factors out the `MDBook` data structure from the `book` module into a new module `manage` (better names welcome), while leaving the `Book` data structure (essential for preprocessors) inside the `book` module. Specifically the set of renamings are:
* `book/book.rs` -> `book/mod.rs`
* `book/mod.rs` -> `manage/mod.rs`
* `book/init.rs` -> `manage/init.rs`

### Alternatives
In theory, a cleaner alternative would be to factor the monolithic `mdbook` project into a Cargo workspace: one crate for the core API that would be used by preprocessors, and one crate for building the mdbook binary. However, this is a much larger change that requires messing with CI, publishing multiple packages, etc. so this PR seemed like a simpler way to achieve the same desired outcome for now.

### Notes
Note there's still some TODOs for documentation --- I will add the docs if it is decided to go ahead with this PR.


ehuss:
Thanks for taking a look at this. I've been thinking of going the multiple-package route in the 0.5 release.

Unfortunately I don't think a change like this can be made in 0.4 since it is a breaking change.

Would you be interested in helping come up with an outline of what it might look like split into separate packages?

willcrichton:
Yes I would be interested! You can treat this PR as a proof-of-concept for one possible factoring.

Is that discussion happening anywhere in particular?

ehuss:
#1234. I haven't given it too much thought recently. I think it would be good to identify the use cases ("preprocessor user", "renderer user", non-CLI driver (like [rustbook](https://github.com/rust-lang/rust/tree/master/src/tools/rustbook)), and try to minimize those. I was thinking maybe something like the "preprocessor user" just gets some basic structs (`Preprocessor`, `SerializedBook` or something like that) with serde impls, so it can be as lightweight as possible. That would likely entail doing a little survey of existing extensions to see what they use (and why they use it).

PR #1399, opened 2020-12-14
Show path in the error message (#1396)

u7693:
With this pull request, in the case of the example in #1396, the error message will look like this:

```
2020-12-14 15:31:55 [INFO] (mdbook::book): Book building has started
2020-12-14 15:31:55 [INFO] (mdbook::book): Running the html backend
2020-12-14 15:31:55 [ERROR] (mdbook::utils): Error: Rendering failed
2020-12-14 15:31:55 [ERROR] (mdbook::utils):    Caused By: Failed to read "/home/u7693/Workspace/book/src/not-found"
2020-12-14 15:31:55 [ERROR] (mdbook::utils):    Caused By: No such file or directory (os error 2)
```

Fixes #1396

Issue #1396, opened 2020-12-07
Rendering failed when book directory contains broken symlink


soareschen:
When the book src directory contains broken symlink, mdbook will fail to build the book with obscure error message `No such file or directory`.

Steps to reproduce:

- Create a broken symlink in the book src directory, e.g. `ln -s /path/not-found`.

- Running `mdbook` will produce the following error:

```bash
$ mdbook build
[INFO] (mdbook::book): Book building has started
[INFO] (mdbook::book): Running the html backend
[ERROR] (mdbook::utils): Error: Rendering failed
[ERROR] (mdbook::utils):    Caused By: No such file or directory (os error 2)
```

As shown above, the error message does not give any clue that the error is caused by broken symlink, making it difficult to debug.

Tested with mdbook v0.4.4 on Linux.

maowen:
With the current implementation on the `master` branch, mdBook still fails to build when the book directory contains a broken symlink although there is now a helpful error message that provides helpful feedback and makes it more clear that the broken symlink caused the build to fail.

I recently submitted an MR (#1944) that I believe provides a full resolution to the original issue. Would it make sense to re-open this issue or should I create a new one instead?

PR #1944, opened 2022-12-05
Build even when broken symlinks are present

maowen:
Better resolution to #1396.

The previous resolution to #1396 (#1399) provides better context when reporting the error but didn't resolve the issue because the mdBok build still fails if there is a broken symlink in the book source tree.

I have need for the mdBook builds to succeed even when broken symlinks are present (e.g. symlinks in source code tree that will be used to create a installable package).

This MR gracefully ignores symlinks that are broken and provides a helpful `info!` level log message. A unit test was added to verify the new behavior.

### DIFF
--- a/src/utils/fs.rs
+++ b/src/utils/fs.rs
@@ -1,5 +1,5 @@
 use crate::errors::*;
-use log::{debug, trace};
+use log::{debug, info, trace};
 use std::convert::Into;
 use std::fs::{self, File};
 use std::io::Write;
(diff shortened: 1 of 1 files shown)


maowen:
Would someone be willing to give me feedback on this PR? This solves a problem that I am having with mdBook and I'd like to get it merged if appropriate.

ehuss:
I'm concerned about broken symlinks leading to a broken build or broken links. Would something like ignoring those paths (#1908) work for you?

xscottxbrownx:
@maowen 

I've run into an issue where I create a valid symlink (or so it seems, as it creates the file in the folder I told it to - shows up in green, meaning it's new - and hovering on it says it is a Symbolic Link)

but when trying to run `mdbook serve` it errors and has a weird path...


Command shows blue on left, meaning it worked:
Screenshot 2023-04-05 at 2 58 11 PM


New file made, says it's a Symbolic Link:
Screenshot 2023-04-05 at 3 07 46 PM


Try to run `mdbook serve` and looks like it's trying to use path as if `.` is a folder:
Screenshot 2023-04-05 at 2 55 26 PM



So, I deleted the symlink with `unlink`. Then, I attempted the command `ln -s ./setup/languages.md miscellaneous/languages/md` which resulted in the exact same thing.


**Thoughts or ideas?**

PR #1946, opened 2022-12-06
fix: random `Array` ref in `src/theme/book.js`

### LINKED ISSUE #1945: What is this Array for?
https://github.com/rust-lang/mdBook/blob/master/src/theme/book.js#L169-L172
```js
        Array 👈🤔
        code_nodes
            .filter(function (node) {return !node.classList.contains("editable"); })
            .forEach(function (block) { hljs.highlightBlock(block); });
```

LePichu:
Removes `Array` class ref as it's needed and was a mistake 
Closes #1945


Dylan-DPC:
Thanks

Issue #1949, opened 2022-12-12
Hidden code blocks no longer sent to rust Playground?


cfsamson:
I've had a lot of errors lately when re-building some of my mdbooks on the CI server. On mdbook version 0.4.15 all code snipptes run as expected but the ones with parts of code hidden by `#` seem to fail by not sending the part of the code which is hidden to the playground when using the latest version (0.4.22). If you expand the hidden parts, everything works as expected.

ehuss:
This seems to be a regression due to #1911.  @yoyomo can you look at posting a fix?

ehuss:
Fixed by #1950.

PR #1950, opened 2022-12-13
#1949 update for hidden only on clipboard

yoyomo:
fix for https://github.com/rust-lang/mdBook/issues/1949


ehuss (approved):
Thanks for the quick fix!

I fixed a bug, there was a missing `return` statement that prevented it from working. I also squashed the commits.

PR #1955, opened 2022-12-15
Switch to older ubuntu image

### LINKED ISSUE #1954: Precompiled `mdbook` gnu executable requires newer glibc version (v0.4.23)
With the latest `v0.4.23` release the precompiled [mdbook-v0.4.23-x86_64-unknown-linux-gnu.tar.gz](https://github.com/rust-lang/mdBook/releases/download/v0.4.23/mdbook-v0.4.23-x86_64-unknown-linux-gnu.tar.gz) binary doesn't run on the github runner of `ubuntu-20.04` anymore. As it fails with the following error:

```sh
mdbook --version
mdbook: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by mdbook)
```
Further workflow logs her: https://github.com/jontze/action-mdbook/actions/runs/3702275493/jobs/6272375311. It still works on the latest ubuntu-22.04 runner.

I observed this behavior especially in context of the current ongoing github `ubuntu-latest` runner rollout: https://github.com/actions/runner-images/issues/6399 . So `ubuntu-latest` will slowly switch from `20.04` to `22.04`.

With the latest mdBook release, the build of the precompiled binaries are now executed on `ubuntu-22.04`.

Other projects where affected by this as well and their selected solution was to build the precompiled binaries on `ubuntu-20.04` instead of `ubuntu-latest`.

https://github.com/rust-lang/mdBook/blob/eb77083d23c768521a0024c6317b8d295198a501/.github/workflows/deploy.yml#L21-L22

For example the `nushell` project had the same issue:
- https://github.com/nushell/nushell/issues/7282
- https://github.com/nushell/nushell/pull/7290

So we might choose this way as well for the mdBook builds. However, I'm not quite sure if this is the way to go, or if there are better solutions to resolve this.

ehuss:
This switches from 22.04 to 20.04 in order to keep the precompiled binaries compatible with older versions of linux due to glibc version errors.

Fixes #1954

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -17,11 +17,11 @@ jobs:
           - x86_64-pc-windows-msvc
         include:
           - target: aarch64-unknown-linux-musl
-            os: ubuntu-latest
+            os: ubuntu-20.04
           - target: x86_64-unknown-linux-gnu
-            os: ubuntu-latest
+            os: ubuntu-20.04
           - target: x86_64-unknown-linux-musl
-            os: ubuntu-latest
+            os: ubuntu-20.04
           - target: x86_64-apple-darwin
             os: macos-latest
           - target: x86_64-pc-windows-msvc
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,7 +30,7 @@ jobs:
             os: windows-latest
             rust: stable
           - build: msrv
-            os: ubuntu-latest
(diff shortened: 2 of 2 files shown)


niaiwomattt:
You should upgrade the glibc version to the latest, or match ubuntu 22.04. Otherwise, in many cases, mdbook is not available, which brings great problems to users.

Issue #1483, opened 2021-03-08
Scrollbars are weird in Safari


arniu:
<img width="1280" alt="图片" src="https://user-images.githubusercontent.com/1286674/110286020-294a2a80-801f-11eb-9c1e-c278b67c8a8a.png">

ehuss:
Can you say more what is "weird" about it?

arniu:
The scrollbars are always present.

julianfortune:
The scroll bars:
- Are blocky instead of round
- Are far wider than normal
- Have the same background as the main content in the sidebar which looks incredibly janky
- Do not disappear when not scrolling
- Displace the context horizontally, messing up the alignment of the page (see red annotations)

### Firefox

<img width="1641" alt="Screen Shot 2022-12-21 at 5 47 03 PM" src="https://user-images.githubusercontent.com/25210657/209037264-10b63034-405f-426b-b06a-0da38a148869.png">

### Safari

<img width="1728" alt="Screen Shot 2022-12-21 at 5 47 14 PM" src="https://user-images.githubusercontent.com/25210657/209037281-efd36897-86ad-46e1-be51-45db3aa182e9.png">


The visual weirdness of the scrollbars is super janky and off-putting, and this visual glitch is sufficient to make me not want to use mdBook. I'll see if I can make a minimal reproducible example or something later

julianfortune:
<img width="1174" alt="Screenshot 2022-12-21 at 9 53 48 PM" src="https://user-images.githubusercontent.com/25210657/209067132-6fe44cc4-eaff-472c-bf38-171179543b57.png">

Yeah, so deleting all the chrome CSS for `::-webkit-scrollbar` fixes the issue alright, but also has the effect of making the sidebar scrollbar appear darkened.

noritada:
I had looked into this problem some time ago and found a solution to thin the scrollbars for the time being, so I sent a PR. However, this PR may not be a complete solution, since it does not turn off the scrollbars when they are not needed.

julianfortune:
I just noticed the Jekyll theme Just the Docs (with the dark color theme) renders the side bar with a dark scroll bar—it appears this is the expected behavior for scroll areas within a page in Safari and not an issue that needs to be corrected.

<img width="1205" alt="just-the-docs-scrollbar" src="https://user-images.githubusercontent.com/25210657/211093052-6c8cfa0a-0e0e-403d-814b-f625277d42e5.png">

julianfortune:
[...]

PR #1961, opened 2022-12-22
Use default scrollbars on webkit browsers

noritada:
This PR thins scrollbars in Chrome and Safari to make them less assertive.
This is part of #1483, but may not completely close the issue since this does not hide the scrollbars when unused.

I will appreciate your feedback!
Many thanks.

### DIFF
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -2,14 +2,6 @@
 
 @import 'variables.css';
 
-::-webkit-scrollbar {
-    background: var(--bg);
-    width: 8px;
-}
-::-webkit-scrollbar-thumb {
-    background: var(--scrollbar);
-    border-radius: 4px;
-}
 html {
     scrollbar-color: var(--scrollbar) var(--bg);
 }
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -4,9 +4,11 @@
 
 ::-webkit-scrollbar {
     background: var(--bg);
+    width: 8px;
 }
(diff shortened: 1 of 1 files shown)


arniu:
How about to remove all the `::-webkit-scrollbar` selectors as  https://github.com/rust-lang/mdBook/issues/1483#issuecomment-1362451893 ?

noritada:
@arniu 
Thank you for comments. Yes, it is another option I considered.

If making the scrollbar disappear when it is not needed is more important than the color, then removing the `::-webkit-scrollbar` would be a good idea. On the other hand, if the project respect the current color scheme of the display, we could simply make it thinner.

I tried the latter for now, but I think the former approach is also good. It is a matter of what the project consider important.

ehuss:
I'm inclined to just remove the webkit scrollbar styling altogether. I don't think the color differences are too important, and I think it is worthwhile to stick with the OS and browser default behavior. There also may be accessibility concerns about making the scrollbar too narrow.  I also don't see other projects styling the scrollbar.  I think it would be fine to keep the Firefox coloring since it doesn't affect any other behavior on that browser.

noritada:
@ehuss 
Thank you for comments.

Regarding the scrollbars, it looks like we are almost in agreement to remove the WebKit-specific style specification, so I have updated the PR to be in line with that. Could you please check it again?

My personal opinion is that it is better to respect browser and OS defaults as much as possible, so that conflicts are less likely to arise in the future when changes are made to the browser and OS defaults.

ehuss (approved):
Thanks!

PR #1964, opened 2022-12-25
include site_url on all pages

davidhewitt:
This PR adjusts the result of setting the `site-url` such that if it is set, the `<base>` tag will be rendered on all pages, not just the 404 page.

Based off the documentation for `site-url`, it sounded to me like including `<base>` on all pages was probably intended, just omitted by mistake?

> site-url: The url where the book will be hosted. This is required to ensure navigation links and script/css imports in the 404 file work correctly, even when accessing urls in subdirectories. Defaults to /. If site-url is set, make sure to use document relative links for your assets, meaning they should not start with /.

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -614,6 +614,9 @@ fn make_data(
         "book_title".to_owned(),
         json!(config.book.title.clone().unwrap_or_default()),
     );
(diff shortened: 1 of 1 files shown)


davidhewitt:
Actually sorry, I don't think this addition on its own is sufficient; it'll break loadng of assets for pages in subdirectories.

To implement this correctly then all links on all pages would need to be rendered relative to the base, but that's a much bigger change.

Issue #1969, opened 2022-12-29
`mdbook watch` hijacks the terminal


amab8901:
When I run `mdbook build --open`, then it lets me browse the documentation without having to keep the terminal busy/occupied. Just browse the docs and keep doing other stuff in the terminal.

But when I run `mdbook watch --open`, then the terminal becomes unavailable while the docs are open. If I run `ctrl+C` in the terminal, it will automatically close the documentation.

Please modify `mdbook watch` so that it can serve the documents without hijacking the terminal.

mgeisler:
Hi there.

> Please modify `mdbook watch` so that it can serve the documents without hijacking the terminal.

How should this work and what should `mdbook watch` do differently than `mdbook build`?

amab8901:
> How should this work and what should mdbook watch do differently than mdbook build?

It was a while ago, but let's imagine this scenario: you have built a book. You just want to read it without having to recompile it redundantly. `mdbook build` and `mdbook watch` will both rebuild what has already been built, if I remember correctly. I can't promise that I remember all the details correctly since it was some time ago.

amab8901:
I think it was mostly a minor issue. I don't remember all the details and it was a while ago. Let's close it for now. If someone else encounters this issue, then can file an issue report and PR and whatnot.

mgeisler:
> `mdbook build` and `mdbook watch` will both rebuild what has already been built, if I remember correctly. I can't promise that I remember all the details correctly since it was some time ago.

You do remember correctly :smile: I'm just a user of `mdbook` and in my experience, it only takes a few seconds to rebuild the book.

amab8901:
> You do remember correctly smile I'm just a user of mdbook and in my experience, it only takes a few seconds to rebuild the book. 

ok let's fix that issue then (i.e. someone else other than me :smile: )

PR #1970, opened 2022-12-29
Add `open` command

amab8901:
Addresses https://github.com/rust-lang/mdBook/issues/1969.

This is intended to allow the user to browse the documents with the same convenience as `mdbook build --open`, but without building a book redundantly when it's already built. And without hijacking the terminal like `mdbook watch` and `mdbook serve`. 

This is a rough first iteration. I have likely missed some details (I'm new to the codebase).

### DIFF
--- a/src/cmd/open.rs
+++ b/src/cmd/open.rs
@@ -1,0 +1,30 @@
+use super::command_prelude::*;
+use crate::{get_book_dir, open};
+use mdbook::errors::Result;
+use mdbook::MDBook;
+use std::path::PathBuf;
+
+// Create clap subcommand arguments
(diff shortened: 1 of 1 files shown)


mgeisler:
> And without hijacking the terminal like `mdbook watch` and `mdbook serve`.

I asked on #1969, but maybe it's better here: why how just open the generated files directly? You have them right there on your disk, so they're only a

```
your-browser book/index.html
```

away :slightly_smiling_face:

amab8901:
I think it was mostly a minor issue. I don't remember all the details and it was a while ago. Let's close it for now. If someone else encounters this issue, then can file an issue report and PR and whatnot.

Issue #1622, opened 2021-08-02
Figure out how to use newer highlightjs


ehuss:
highlightjs 11 removed a feature called "HTML merging".  This is a feature we are depending on to handle "boring" lines, which get wrapped with `<span class="boring">`.  HLJS 11 now completely strips HTML out of a code block.

There is discussion at https://github.com/highlightjs/highlight.js/issues/2889 about writing a plugin to add this functionality back in, but I do not understand the plugin API well enough to write that (the current example there does not work). 

Either we'll need to figure out how to get that working, or we'll need to find a new highlighting engine.  #1494 is a potential route, though I am concerned about the significant performance loss.

notriddle:
#1652 seems to perform a lot better than the earlier syntect benchmarks.

sanmai-NL:
I have good experiences with [Prism.js](https://prismjs.com/). 
Lightweight and very functional.

ISSOtm:
I wanted to start working on this, but I don't want the effort to be wasted in case #1652 is merged instead. Should I go ahead regardless?

joshgoebel:
What is a boring line?  How does one know which lines are boring *Before* the highlighting has happened?

PR #1976, opened 2023-01-02
Add documentation on how to add a highlight.js for additional syntax highlighting support

DrTobe:
This adds a few lines on how a custom version of `highlight.js` can be obtained and made available to `mdbook`. With that information available, all issues about adding support for a specific language should be safe to be closed because `mdbook` in its default setup probably does not need to support every possible language.

I have seen that there are ongoing efforts to move away from `highlight.js` but I guess this information may nevertheless be useful in the meantime.

### DIFF
--- a/guide/src/format/theme/syntax-highlighting.md
+++ b/guide/src/format/theme/syntax-highlighting.md
@@ -63,12 +63,17 @@ your own `highlight.js` file:
 - xml
 - yaml
 
+If you want to support another set of languages for syntax highlighting,
+you can download a "Custom package" from their website, rename the
+`highlight.min.js` to `highlight.js` and move it into the `theme` folder of your
+book.
+
 ## Custom theme
 Like the rest of the theme, the files used for syntax highlighting can be
 overridden with your own.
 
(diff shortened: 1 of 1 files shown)


max-sixty:
Would be good to merge this!

ehuss:
Unfortunately I don't think just using the highlight.js website will completely work. mdBook is incompatible with newer versions of highlight.js (such as #1622). https://github.com/rust-lang/mdBook/blob/master/CONTRIBUTING.md#updating-highlightjs contains some instructions for building it from source, where you can check out a version from the 10.x series.

DrTobe:
I just proposed to add this because it took me quite some time to figure this out and it solved our specific problem. So I still think some kind of documentation like this might be valuable but you are right that we should not include it like that if it is not 100% correct. So what specific approach do you propose?

ehuss:
I'm not sure, but one suggestion might be to copy the first four steps from the contributing guide to this page instead of recommending going to the website.

GuillaumeGomez:
I agree with @ehuss that instructions should be more detailed (in particular for the `highlight.js` version). Feel free to reopen it with updated docs.

PR #1980, opened 2023-01-09
Use md file's header to link name while summary file's title is blank

wkevin:
You can leave title blank ( `[ ](foobar.md` ) ，so don't worry about different between title in summary.md with H1 header  in foobar.md.


HU90m:
This is a cool feature. Would it be better placed in a preprocessor? Just to avoid unnecessary complexity in mdBook itself.

GuillaumeGomez:
Sorry for the delay. Currently going through the backlog. Closing this PR as a feature including this one was added in #3092. Thanks in any case!

PR #1986, opened 2023-01-15
Run preprocessors in `mdbook test`

mgeisler:
While adding support for [translations to Comprehensive Rust 🦀][1], I noticed that `mdbook test` doesn’t execute preprocessors the same way as `mdbook build`.

This PR makes the two commands use the same code to find and execute preprocessors.

[1]: https://github.com/google/comprehensive-rust/pull/130

### DIFF
--- a/src/book/mod.rs
+++ b/src/book/mod.rs
@@ -196,21 +196,26 @@ impl MDBook {
         Ok(())
     }
 
-    /// Run the entire build process for a particular [`Renderer`].
-    pub fn execute_build_process(&self, renderer: &dyn Renderer) -> Result<()> {
-        let mut preprocessed_book = self.book.clone();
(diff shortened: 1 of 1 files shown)


mgeisler:
We've run into our first uncaught translation error: https://github.com/google/comprehensive-rust/issues/415 :slightly_smiling_face: The problem was a code snippet which was translated incorrectly and this turn resulted in an error when trying to execute it. I believe merging this PR will allow us to prevent this in the future.

mgeisler:
@ehuss, would you be able to take a look?

ehuss (approved):
Thanks!  There is some risk that this may cause issues if a project is relying on preprocessors having a certain behavior.  However, I can't think of particular scenarios that this would break.

PR #1987, opened 2023-01-15
Make fonts part of the theme.

### LINKED ISSUE #1967: How to customize fonts in mdBook?
I'd like to override `theme/fonts/fonts.css` and add additional TTF files. Is that directly possible in mdBook (on `mdbook build`)?

ehuss:
This changes it so that fonts are part of the theme. This allows the user to override and customize the fonts.

The `output.html.copy-fonts` option is deprecated, and will be removed in a future release. If customizing the fonts, you should define a `theme/fonts/fonts.css` file with whatever settings you want, and include any fonts in the `theme/fonts` directory.

Note that FontAwesome is not part of the custom fonts. I will likely try to find some alternate solution (such as #1330) for loading (and customizing) that.

Closes #1967


hydroper:
Thanks!

PR #1991, opened 2023-01-18
Theme-dependent image variants

### LINKED ISSUE #1990: Theme-dependent image variants
### Problem

MdBook themes do currently not work well together with images.

Images with light background look weird in the dark and Rust themes, while images with transparent background can become invisible e.g. on the dark themes.

### Proposed Solution

Allow for different image versions for different themes. There are several ways to implement it, as well as to realize a syntax for it.

My ideas so far:

* Implement it in mdBook (using a custom syntax)
* Write a preprocessor for it (using a custom syntax)
* Use HTML classes and CSS and/or JS to hide specific classes (e.g., each theme could hide HTML tags that have a certain class, like `no-rust` or `no-light`)

For the first two options, a syntax would be required. could be something like an HTML comment on the line before the image.

Probably, the last option would be the most consistent one.

### Notes

_No response_

mlange-42:
Implements theme-dependent image variants using CSS and HTML classes.
Basically, HTML elements can be excluded on individual themes, or on themes grouped by light vs. dark.

See the added documentation for usage details. They also show the feature in action.

An alternative would be to let users do it via CSS themselves, but this solution might not be obvious.

Fixes #1990


guide/src/format/markdown.md
```diff
@@ -117,6 +117,28 @@ Which, of course displays the image like so:
...
 
+## Theme-dependent images
+
+Image variants for different themes can be used via CSS classes.
+
+Here is an example for different images for light vs. dark themes:
+
+```markdown
+<img src="images/rust-logo-blk.svg" class="no-dark-themes" />
+<img src="images/rust-logo-blk-dark.svg" class="no-light-themes" />
```
not-my-profile, line 128: I'd prefer `only-light-themes` and `only-dark-themes` for the class names ... since the negation is a bit unintuitive.

mlange-42: Thanks for the review, you are probably right here. Will rework this accordingly.

hayesall:
@mlange-42 I've been working with this for a little while now, and it seems like a good approach after not-my-profile's review:

```css
.light img[src$="#only-dark"],
.rust  img[src$="#only-dark"] {
    display:none;
}

.navy img[src$="#only-light"],
.ayu  img[src$="#only-light"],
.coal img[src$="#only-light"] {
    display:none;
}
```

plus it works directly in Markdown:

```markdown
![rust logo standard edition](images/rust-logo-blk.svg#only-light)
![rust logo dark mode](images/rust-logo-blk-dark.svg#only-dark)
```

guide/src/format/markdown.md
2bndy5: Technically, this fenced block is only `html` syntax.

Some MD linter prohibit use of raw HTML, so maybe change this to
~~~markdown
```markdown
![rust logo standard edition](images/rust-logo-blk.svg#only-light)
![rust logo dark mode](images/rust-logo-blk-dark.svg#only-dark)
```
~~~
as hayesall reported it working.

GuillaumeGomez:
Sorry for the delay, we now have a team. Nominating this PR for discussion.

cc @rust-lang/mdbook

Issue #1734, opened 2022-01-27
Fix sidebar layout on mobile screens


supfiger:
## Steps to repro:
* Watch the video https://monosnap.com/file/6NcwSNtDnBfYjTV34EbPWe1xc1psha

## Expected behavior:
* Elements don't overlap each other

## Environment:
* device: iPhone X
* OS: IOS 15+
* browser: Safari

ehuss:
Screenshot just to capture from the remote site.
<img width="343" alt="image" src="https://user-images.githubusercontent.com/43198/171324024-c5f25878-c6cc-459a-a972-37a273292bdc.png">

ehuss:
This should now be fixed via #1994 in 0.4.26.

Issue #1667, opened 2021-10-13
Content can overlap open menu in mobile Safari


vwkd:
Opening the menu in mobile Safari allows the main content to be scrolled around. It seems the overflow is not being hidden. I didn't test on other mobile browsers.

https://user-images.githubusercontent.com/33468089/140620139-56415c2e-e25d-4eea-94a8-ca2b287606cd.mp4

Safari on iOS 15.0.1 

Reference: https://github.com/rust-lang/book/issues/2905

ehuss:
This should now be fixed via #1994 in 0.4.26.

PR #1994, opened 2023-01-22
fix overflow-x on mobile

Skwodo:
The browsers on mobile devices ignore the `overflow-x: hidden` within the body tag if `<meta name="viewport">` tag is present. The fix is to add a container inside the body tag that contains everything the body would normally contain. This should solve issues #1667 and #1734.

### DIFF
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -12,7 +12,16 @@ a > .hljs {
     color: var(--links);
 }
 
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks!  I added some comments, please take a look if I misunderstood something.

I'm often nervous about making changes like this since it feels like everything is built on sand.

PR #1998, opened 2023-01-26
Remove dependency to time 0.1.44

dalance:
Now mdbook depend to time 0.1.44 which has CVE-2020-26235.
By adjusting feature, the dependency can be removed.


ehuss (approved):
Thanks!  I also believe this will go away in chrono 0.5.

PR #2001, opened 2023-01-28
Fix incorrect version command

iFreilicht:
The `version` command does not exist, but clap defines a default flag `--version` or `-V`.


ehuss (approved):
Oops, thank you!!

PR #2009, opened 2023-02-08
bump 'toml' dependency

danieleades:
bumps the 'toml' dependency to the latest version


danieleades:
any particular reason this crate isn't using dependabot?

The usual complaint is that for binaries (as opposed to libraries) it can be very noisy. But it's possiblye to throttle the PRs in the config. Should at least be used for the github workflow actions, right?

ehuss (approved):
Thanks!

ehuss:
I believe dependabot is configured, though maybe for only for special alerts?

I generally don't find dependabot helpful. The more interesting changes are the ones that need manual intervention (like this one) which it doesn't help with. Even with throttling it seems to generate a lot of unimportant noise. I also tend to lean towards updating dependencies when there is a need rather than constantly being on the latest.

PR #2013, opened 2023-02-09
Add heading extension support

### LINKED ISSUE #1141: Custom section (header) ID
It would be useful if the section IDs can be specified manually (such as `# About foo {#foo}`, supported by many processors including [zola](https://github.com/getzola/zola/issues/433)).

For example, it would be useful when:

* Some of section names are very long, but wants IDs to be short enough.
* Attempt to prevent future link breakage, which can be caused by sections renaming or mdbook algorithm change.
* Providing multiple sites with different languages, but wants URIs (especially paths and fragments) to be coherent English.

Users can already control paths by setting filename, then it is natural if users can also control fragment part of URIs (`/user/specified/path.html#THIS-PART`).

Is this feature simply not implemented or considered yet, or are there some reasons to object to this feature?

Maybe related: #880

ImUrX:
Closes #1141 
Code got a little complicated on the build header links code, had to add some extra logic :p


Eirenliel (approved):

ImUrX:
@badboy I use ``mdbook-toc`` and seeing the code I think it might require a change in the mdbook API or a change in your lib

badboy:
> @badboy I use `mdbook-toc` and seeing the code I think it might require a change in the mdbook API or a change in your lib

I take PRs.

ImUrX:
We will see if @ehuss asks me to modify how an API works first :P

ehuss:
Thanks! 

This seems to have a few issues:
* When I use a header like `## Test1 { #my-id }`, the content of the header gets converted to `my-id`, and the string `Test1` is lost.
* This normalizes the ID which might not what the author would want. For example, if I define a header with a specific ID, then that's probably because I would want to link to it using that specific ID. Though that could cause problems in the print page if there are ID conflicts if it weren't normalized, so there isn't an easy answer here, but I would probably lean towards not modifying their ID.
* The class information seems to be erased.
* This probably needs a variety of tests. Some unit tests, and some full tests using the `test_book`.
* The search index seems to be using the wrong ID.

ImUrX:
By the contents of the header, you mean ID, right?

I think this is very useful mainly for normalizing the ID, it let's you have an ID that never changes even when you change the heading's name. And I think it's a powerful tool that should be handled by the end user, if there are any ID conflicts.
This is useful for maintaining old links and for translations which would also be able to maintain through cross-language the same ID.

I don't know much about how mdBook works, can you define class information in the code for all headings? I don't understand what you mean by the class information being erased.

I agree on testing being needed, but I think most of the testing is already handled by the crate that we are using, so the testing needed should be for mdBook specific behavior.

ehuss:
> By the contents of the header, you mean ID, right?

A header:

```markdown
## Test1 { #my-id }
```

generates the html:

```html
<h2 id="my-id"><a class="header" href="#my-id">my-id</a></h2>
```

I would expect it to be:

```html
<h2 id="my-id"><a class="header" href="#my-id">Test1</a></h2>
```

notice the contents of the text of the header.

> I think this is very useful mainly for normalizing the ID

Sorry, I'm not following what you are saying here. I would expect the point of explicitly specifying the id with the `{ #id }` syntax is to explicitly set the id of the header. However, the code currently puts it through the normalization process which can potentially rewrite it unpredictably. For example, a user writing `{ #my.id}` will end up with an id of `myid`. 

> can you define class information in the code for all headings? I don't understand what you mean by the class information being erased.

I would expect a header like:

```markdown
## Test1 { .someclass }
```

to generate the HTML:

```html
<h2 id="test1" class="someclass"><a class="header" href="#test1">Test1</a></h2>
```

the current code deletes the class:

```html
<h2 id="test1"><a class="header" href="#test1">Test1</a></h2>
```

ImUrX:
I actually didn't notice all this, sorry :sob:

ImUrX:
I think it's more simple to just have an empty class attribute rather than optionally adding it (I accidentally named it ``classes`` :laughing:).

ImUrX:
oops, @ehuss sorry, do you mind doing a review so I can just re-request it when I make a fix?

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -816,15 +833,19 @@ fn build_header_links(html: &str) -> String {
 fn insert_link_into_header(
     level: usize,
     content: &str,
+    id: Option<String>,
+    classes: Option<String>,
     id_counter: &mut HashMap<String, usize>,
 ) -> String {
-    let id = utils::unique_id_from_content(content, id_counter);
+    let id = id.unwrap_or_else(|| utils::unique_id_from_content(content, id_counter));
+    let classes = classes.unwrap_or("".to_string());
```
ehuss, line 841: Can this skip adding `class=""` when there are no classes?

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -999,27 +1020,31 @@ mod tests {
...
-                r##"<h4 id="hï"><a class="header" href="#hï"><em>Hï</em></a></h4>"##,
+                r##"<h4 id="hï" class=""><a class="header" href="#hï"><em>Hï</em></a></h4>"##,
             ),
             (
                 "<h1>Foo</h1><h3>Foo</h3>",
-                r##"<h1 id="foo"><a class="header" href="#foo">Foo</a></h1><h3 id="foo-1"><a class="header" href="#foo-1">Foo</a></h3>"##,
+                r##"<h1 id="foo" class=""><a class="header" href="#foo">Foo</a></h1><h3 id="foo-1" class=""><a class="header" href="#foo-1">Foo</a></h3>"##,
+            ),
+            (
+                r##"<h1 id="foobar" class="class1 class2">Foo</h1>"##,
```
ehuss, line 1046: Can you also add another entry for a custom ID?

guide/src/format/markdown.md
```diff
@@ -220,3 +220,14 @@ To enable it, see the [`output.html.curly-quotes`] config option.
 [tables]: https://github.github.com/gfm/#tables-extension-
 [task list extension]: https://github.github.com/gfm/#task-list-items-extension-
 [`output.html.curly-quotes`]: configuration/renderers.md#html-renderer-options
+
+### Heading attributes { #headingattributes .class1 .class2 }
```
ehuss, line 224: There probably isn't a need to add a custom ID or class here. If you were intending this for testing purposes, that should probably go in the `test_book`.

guide/src/format/markdown.md
```diff
@@ -220,3 +220,14 @@ To enable it, see the [`output.html.curly-quotes`] config option.
...
+
+Headings can have a custom ID and classes. This let's you maintain the same ID even if you change the heading's text, it also let's you add multiple classes in the heading.
```
ehuss, line 226: For a reader who is unfamiliar with HTML, this might be helpful.

```suggestion
Headings can have a custom HTML ID and classes. This let's you maintain the same ID even if you change the heading's text, it also let's you add multiple classes in the heading.
```

ehuss (commented):
The search index still seems to be using the wrong id.

Can you add some custom ID and class headers to the `test_book` and add some tests to the integration tests?

src/renderer/html_handlebars/hbs_renderer.rs
ImUrX: Yes, it's possible but it would complicate the code more. Do you think it's really needed to make it have that kind of behavior?

ehuss: Yea, I think it would be better to keep the generated output cleaner. It shouldn't be too complicated, I think something like the following should work:

```rust
let classes = classes.map(|s| format!(" class=\"{s}\"")).unwrap_or_default();
```

ehuss:
I went through and applied the suggestions, updated some tests, and fixed the search support. I *think* everything should be working now, hopefully this won't cause too many issues. Thanks for working on this!

ImUrX:
oh, thank you for finishing the PR :heart:, I been busy sorry qwq

PR #2014, opened 2023-02-09
Fix clippy warnings

nickelc:


src/book/book.rs
```diff
@@ -77,10 +77,10 @@ fn create_missing(src_dir: &Path, summary: &Summary) -> Result<()> {
 /// [`iter()`]: #method.iter
 /// [`for_each_mut()`]: #method.for_each_mut
 #[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize)]
+#[non_exhaustive]
 pub struct Book {
     /// The sections in this book.
     pub sections: Vec<BookItem>,
-    __non_exhaustive: (),
```
ehuss, line 83: This is a breaking change, can you revert it please?

ehuss (commented):
Can you maybe put the different lints into different commits so that it is easier to review?

src/book/book.rs
nickelc: Isn't the added `#[non_exhaustive]` attribute the same as the private field?

Dylan-DPC: No, it is exposed to preprocessors. There was a similar attempt earlier which was reverted: https://github.com/rust-lang/mdBook/pull/1848

nickelc: I removed the change.

ehuss:
Closing as I have applied some clippy fixes in #2083.

PR #2020, opened 2023-02-13
Remove spammy `debug!` log from `path_to_root`

mgeisler:
The log statement is empty and simply fills up the logs when you run `mdbook` with `RUST_LOG=debug`.


ehuss (approved):
Thanks!

PR #2033, opened 2023-02-25
Only apply "key" styling to nested <kbd>s

ISSOtm:
It turns out that this is the recommended practice: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd#representing_keystrokes_within_an_input

### DIFF
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -148,7 +148,7 @@ blockquote {
     border-bottom: .1em solid var(--quote-border);
 }
 
-kbd {
+kbd > kbd {
     background-color: var(--table-border-color);
     border-radius: 4px;
     border: solid 1px var(--theme-popup-border);


ehuss:
I don't think this is a change we should make, since it breaks basic usage of something like `<kbd>X</kbd>`.  I'm not sure this is something that can work with or without nesting, but if that's possible then it seems reasonable. Otherwise I probably would prefer to not change it.

ISSOtm:
The problem is that `<kbd>` appears to be misusing the feature, accessibility-wise. With the current nesting, something like `<kbd><kbd>Ctrl</kbd>+<kbd>Tab</kbd></kbd>` creates a double layer of borders, which is not great; hence this change.

ehuss:
I wouldn't characterize it as misusing it. Per that linked document:

> Note: You don't need to do all this wrapping; you can choose to simplify it by leaving out the external `<kbd>` element. In other words, simplifying this to just `<kbd>Ctrl</kbd>+<kbd>N</kbd>` would be perfectly valid.

The non-nested style seems to be the dominant behavior that I have seen (such as GitHub <kbd>Ctrl</kbd>+<kbd>N</kbd> which would render nested as <kbd><kbd>Ctrl</kbd>+<kbd>N</kbd></kbd>).

ISSOtm:
Alright, then I suppose I will override that locally.

PR #2039, opened 2023-03-02
Change overflow-x hidden to clip

### LINKED ISSUE #2028: Sticky header bar isn't sticky any more
### Problem

I updated from mdBook 0.4.21 to 0.4.27 and it seems to have broken the sticky header bar.

### Steps

1. Go to https://tavianator.com/2022/ray_box_boundary.html
2. Scroll down a page or two
3. Scroll back up a few inches

When I do this, the header bar remains off screen.  I can only see it when I scroll all the way back up to the top of the page.

### Possible Solution(s)

_No response_

### Notes

I tried both Firefox and Chrome on Arch LInux.

### Version

```text
mdbook v0.4.27
```

Skwodo:
I'm sorry for my mistake, this should fix #2028. 

As it turns out, any ancestor of a sticky element that has `overflow: hidden`  becomes the scrolling container for the sticky element. Changing the `overflow` value from `hidden` to `clip` should fix it, since both are basically the same in this scenario and `clip` doesn't have this behavior. 
It's important to note that `overflow: clip` is a little bit new, but the [support](https://caniuse.com/mdn-css_properties_overflow-x_clip) is really good.

Let me know if I overlooked something else.

### DIFF
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -22,7 +22,7 @@ a > .hljs {
         the screen on small screens. Without it, dragging on mobile Safari
         will want to reposition the viewport in a weird way.
     */
-    overflow-x: hidden;
+    overflow-x: clip;
 }
 
 /* Menu Bar */


ehuss (approved):
Thanks! I don't have much capacity to test older browser versions. I do have some very old Safari versions which seemed to be ok with it.

Issue #1789, opened 2022-04-19
Blind user experience issues with code blocks


briankung:
(Originally reported on https://github.com/rust-lang/book/issues/3125)

I saw a comment on Reddit indicating that code blocks were difficult for blind readers to read:

> I’m blind and find the Rust books hard to read since the arrow keys cause the pages to flip when I’m attempting to read code examples character by character.  In the beginning this was very annoying because I was also learning syntax, but having passed that phase I’m mostly fine.

https://reddit.com/r/rust/comments/u63g90/_/i56o9jt/?context=1

I just wanted to surface it in case the topic hasn’t come up before.

- [x] I have checked the latest `main` branch to see if this has already been fixed
- [x] I have searched existing issues and pull requests for duplicates

URL to the section(s) of the book with this problem:

All / any code blocks

Description of the problem:

See above 

Suggested fix:

joshrotenberg:
I think I can reproduce this: if you don't click into a code block, right/left arrow keys go to the next/previous page, so presumably whatever method a blind user uses to advance through the page doesn't necessarily explicitly put the cursor into a code block.

tavianator:
Related to https://github.com/rust-lang/mdBook/issues/518

I [manually remove the buttons](https://github.com/tavianator/tavianator.com/blob/919773b518ebbcecb57fb4db354d2b960016b7ac/src/tavianator.js#L16-L24) on my site which also disables the left/right arrow keys, but a setting for this would be nice.

brettchalupa:
@ehuss I'd like to take this issue on, but I think some guidance before I dig in would be helpful. Do you have an opinion on how to proceed with this and #518? Add config option to disable or disable by default with option to enable or something else?

Issue #518, opened 2018-01-02
[Feature request] make keyboard nav optional


pygy:
Hi!

I'm currently reading the Rust book, which is awesome in many respects, but I regularly trigger the next/previous page events by pressing the arrows (it's kind of reflex, I'm not even sure of why I use those keys, but I suppose I rely on them usually being noops). It would be nice if there was a way to disable them permanently (short of forking mdRust and serving a custom build of the book on localhost).

Since they are AFAICT undocumented and non-standard, those shortcuts also have discoverability issues for people who would benefit from them.

While it would clobber the UI, a checkbox would also raise the awareness of the feature.

Michael-F-Bryan:
Technically this wouldn't be very difficult to do. You'd just add a checkbox and whenever the page loads or the checkbox is toggled,  the appropriate event handlers are updated. *However*, I'd be a little hesitant to add an extra thing to the UI for *just* deactivating the arrows. Just because it'll end up adding clutter to the UI and it's an extra thing which needs to be maintaned, my concern is the feature wouldn't be used enough to warrant the effort.

Another thought is to have a more general "config" overlay, similar to what [the playground] does. That way you've got a mechanism for UI tweaking, it'd be out of your way by default, and more maintainable in the long run.

i.e. Something like this:

![Playground Screenshot](https://i.imgur.com/XrUuwcT.png)

It may be useful to ask people who work on *The Book* or *Rust By Example* and see if they've seen similar UI issues raised before. That would help us get a feel for how desired the ability to tweak the UI is. 

cc: @carols10cents, @steveklabnik, @budziq, @projektir (feel free to cc in anyone else who may have an opinion)

[the playground]: https://play.rust-lang.org/

pygy:
I can see where you're coming from...

The thing I wonder is why you have it as it is in the first place, i.e. non-standard, hidden.

Who's the target audience for the feature?

I'd expect people with screen readers are used to `tab` + `enter` to activate the buttons.

The web sites/apps I use the most (GitHub, Reddit, HN, Gmail) all ignore the left/right arrows... is it common in reader apps?

Michael-F-Bryan:
I remember the first time I was reading through *The Book* and was surprised that hitting the right arrow did exactly what I wanted, go to the next chapter. It's a very intuitive thing to do, especially when all browsers let you use the up and down arrows as an alternative to scrolling around a page.

> The thing I wonder is why you have it as it is in the first place, i.e. non-standard, hidden.

[...]

PR #2043, opened 2023-03-05
Use modifier keys for chapter navigation

lionel-rowe:
Fixes https://github.com/rust-lang/mdBook/issues/518 and https://github.com/rust-lang/mdBook/issues/1789.

Change chapter navigation key combos from bare left/right arrow to ctrl/cmd + left/right arrow. The bare arrow key combos are bad for accessibility and very easy to press by mistake for users who expect arrow keys to scroll the page.

Edit: Actually not 100% certain it will fix https://github.com/rust-lang/mdBook/issues/1789 as I don't know if ctrl/cmd + arrow keys would also be used for navigating within code blocks when using screen reader software. Could be made less likely to clash by using more complex modifiers, e.g. ctrl+shift, at the expense of making the shortcut less easy to use.

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -561,9 +561,25 @@ function playground_text(playground, hidden = true) {
 })();
 
 (function chapterNavigation() {
+    const modifiers = Object.fromEntries(['ctrl', 'alt', 'shift', 'meta']
+        .map((k) => [k, `${k}Key`]));
+    const matches = (event) => (matcher) => Object.values(modifiers)
+        .every((v) => matcher.includes(v) === event[v]);
+
     document.addEventListener('keydown', function (e) {
-        if (e.altKey || e.ctrlKey || e.metaKey || e.shiftKey) { return; }
-        if (window.search && window.search.hasFocus()) { return; }
+        const { ctrl, meta } = modifiers;
+
+        // Rejected modifier key combos when combined with left/right arrow keys:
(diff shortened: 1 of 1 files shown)


lionel-rowe:
> I have custom editable elements on my page and ctrl-left/right moves the cursor one word at a time.
>
> Could there just be a simple toml option to disable navigation globally? Otherwise it seems this pull request is taking an Apple-esq approach to slowly muse on the perfect solution.

That still doesn't fix the issue from the user's end, though. Most likely, the majority of projects using mdbook to publish will keep the default configuration, which leaves the accessibility issues unsolved.

I think it's uncontroversial that navigation keyboard shortcuts shouldn't trigger while editable elements are selected. Something like this should fix that:

```js
function isTextInputMode() {
    return document.activeElement.isContentEditable || ['INPUT', 'TEXTAREA'].includes(document.activeElement.nodeName)
}

// ...then, inside key event listeners...
if (isTextInputMode()) return 
```

I'm also experimenting with having a user-configurable setting for the keyboard shortcuts, which seems like a sensible way to fix the issue of conflicts with assistive technology etc.

lionel-rowe:
Closing in favor of https://github.com/rust-lang/mdBook/pull/2061. @expikr could you confirm if that PR would work for your use case? In particular, does the new `isTextInputMode` logic correctly detect when your custom editable elements are focused and block the shortcuts in those cases?

PR #2044, opened 2023-03-08
Allow environment variables to be evaluated by the preprocessor

### LINKED ISSUE #2040: Providing your own command for a preprocessor doesn't evaluate $ENV_VARS
### Problem

Currently, there is no way to define a custom preprocessor in case the command-line arguments contain an environment variable which needs to be evaluated as a part of the script's path.

### Steps

Adding a preprocessor to a `book.toml`:

```toml
[preprocessor. example]
command = "python $HOME/example.py" # that doesn't work
```

mdbook build fails with an error:

```sh
Python: can't open file '$HOME/example.py' # please note that $HOME env var was not evaluated
```
Using a relative path or an absolute path works fine:

```toml
command = "python example.py" # that works
command = "python /absolute/path/example.py" # that works fine
command = "python ../../relative/path/example.py" # that works as well
```

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
mdbook v0.4.27
```

kumavale:
Fixes #2040 

Evaluate some environment variables ($ENV_VARS) in the script's path.
Strings surrounded by `%` and `%` are treated as environment variables for Windows, and strings starting with `$` are treated as environment variables for Linux.
Does nothing if the environment variable is not found.


slmgc:
Is there anyone who can review this small change? It's a real deal breaker for us 😔

PR #2057, opened 2023-04-03
fix(cli): init --force skips confirmation prompts

### LINKED ISSUE #1556: mdbook init force flag documentation is wrong

seanpoulter:
### Summary

This PR fixes #1556 by changing/fixing the behaviour of `mdbook init --force`.

The intent of `--force` is to skip the prompts. This is not how it currently behaves:

```
$ mdbook init --force

Do you want a .gitignore to be created? (y/n)

What title would you like to give the book? 

2023-04-02 23:17:48 [INFO] (mdbook::book::init): Creating a new book with stub content

All done, no errors...
```

### Proposed Changes

* Skip the .gitignore prompt when `--force` is set
* Skip the title prompt when `--force` is set
* Add docs to the guide for `--force`

### What should the reviewer focus on?

* I'm new to Rust and the repo. 🦀 
* I've found it hard to add a failing test that detects the CLI prompts. I did not find an example testing interactions with the CLI in the repo. The test I've added is actually a false positive that does not fail on `master` today. I'm hoping you may have some insight on how to improve it, or we could remove it since it does not add value/confidence. From what I've read, we might consider a tool like the [`rexpect` crate](https://crates.io/crates/rexpect) to help interact with the CLI.

### How have you reviewed these changes?

* I've run `cargo test`, `cargo fmt`, `cargo clippy`
* Manual validation from the debug build works as expected:
  ```
  $ target/debug/mdbook init --force
  2023-04-02 23:24:59 [INFO] (mdbook::book::init): Creating a new book with stub content

  All done, no errors...
  ```

### DIFF
--- a/guide/src/cli/init.md
+++ b/guide/src/cli/init.md
@@ -68,3 +68,7 @@ Create a `.gitignore` file configured to ignore the `book` directory created whe
 If not supplied, an interactive prompt will ask whether it should be created.
 
 [building]: build.md
+
+#### --force
+
+Skip the prompts to create a `.gitignore` and for the title for the book.
--- a/src/cmd/init.rs
+++ b/src/cmd/init.rs
@@ -56,7 +56,7 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
             "git" => builder.create_gitignore(true),
             _ => builder.create_gitignore(false),
         };
-    } else {
+    } else if !args.get_flag("force") {
         println!("\nDo you want a .gitignore to be created? (y/n)");
         if confirm() {
             builder.create_gitignore(true);
@@ -65,6 +65,8 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
 
     config.book.title = if args.contains_id("title") {
         args.get_one::<String>("title").map(String::from)
+    } else if args.get_flag("force") {
+        None
     } else {
         request_book_title()
     };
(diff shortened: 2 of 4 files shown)


ehuss (approved):
Looks good, thanks!

As for testing, yea writing a test that would fail on master could be a little tricky.  I think one approach would be to use something like comparing the entire stdout output and making sure it does not print "What title would you like to give the book?". However, since mdbook's output uses log lines with timestamps, that can be a little awkward, and we currently don't have a good setup to deal with that.

PR #2058, opened 2023-04-09
Use blockquote styles for collapsible-blockquote

expikr:
image

image

image

image


ehuss:
Can you say why this would be added instead of putting the `<details>` block inside a blockquote? For example:

> <details><summary>CLICK ME</summary>
> <p>
> Content goes here.
> </p>
> </details>

In general, when opening a pull request, it would be helpful to include some text explaining what a PR does and what the motivation is behind it.

PR #2059, opened 2023-04-09
Add additional-theme option

### LINKED ISSUE #2048: [Feature Request] Entrypoint to add additional themes ALONGSIDE the default (rather than overwriting/replacing them)
### Problem

According to the documentation, you either have to stick with the shipped themes or swap them out entirely/modifying them. 

I would like to keep the default themes, but only add extra ones alongside them.

I would like to be able to easily make simple color alterations without having to go through the entire swamp of CSS.

I would like to be able to just find themes other people have shared and paste them into the direction and have it automatically show up in the dropdown selector of the compiled book

### Proposed Solution

An entrypoint like Jekyll does with custom-variables would be neat, this way users don't need to deal with the entire CSS when all they want is a slightly altered color scheme.

Have the existing theme files be just a folder of stylesheets. Adding or removing them from the directory automatically changes the resulting dropdown listing to match what's in the folder.

Or perhaps something similar to SUMMARY.md but instead it's THEMES.md where you list out theme entries and their respective references.

Or some other ways of implementing that is ergonomic to the use case of "ok, it ships with these five themes, I want to remove two of the light ones, and add three of my own that is just a slight color tweak of Navy, without having to write an entire CSS sheet for every element"

### Notes

_No response_

expikr:
Closes https://github.com/rust-lang/mdBook/issues/2048

This allows users to specify _additional_ themes without replacing the presets.

image

### `book.toml`
```
[output.html]
additional-css = ["orange.css"]
additional-theme = ["Orange"]
```

### `orange.css`

<details><summary>Click to expand</summary>

```
.orange {
    --bg: #f6f6ef;
    --fg: Black;

    --sidebar-bg: #ff6600;
    --sidebar-fg: Black;
    --sidebar-non-existant: color-mix(in srgb, var(--sidebar-bg) 75%, Black);
    --sidebar-active: White;
    --sidebar-spacer: color-mix(in srgb, var(--sidebar-bg) 95%, Black);

    --scrollbar: #8F8F8F;

    --icons: #747474;
    --icons-hover: #000000;

    --links: #828282;

    --inline-code-color: Black

    --theme-popup-bg: #fafafa;
    --theme-popup-border: #cccccc;
    --theme-hover: #e6e6e6;

    --quote-bg: #e9e9ed;
    --quote-border: #8f8f9d;

    --table-border-color: hsl(0, 0%, 95%);
    --table-header-bg: hsl(0, 0%, 80%);
    --table-alternate-bg: hsl(0, 0%, 97%);

    --searchbar-border-color: #aaa;
    --searchbar-bg: #fafafa;
    --searchbar-fg: #000;
    --searchbar-shadow-color: #aaa;
    --searchresults-header-fg: #666;
    --searchresults-border-color: #888;
    --searchresults-li-bg: #e4f2fe;
    --search-mark-bg: #a2cff5;
}

```
</details>

## Limitations
- The theme class needs to specify ALL the variables, otherwise they default to no-style.
  Can we figure out an easy way to derive the variable values from the presets?
- The syntax highlighting defaults to that of an assumed light mode ([see here](https://github.com/rust-lang/mdBook/blob/master/src/theme/book.js#L341)). 
  What would be the most appropriate way for the theme class to declare itself as a dark-based or light-based theme and have the correct syntax highlight selected?

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -541,6 +541,8 @@ pub struct HtmlConfig {
     pub google_analytics: Option<String>,
     /// Additional CSS stylesheets to include in the rendered page's `<head>`.
     pub additional_css: Vec<PathBuf>,
+    /// Additional theme.
+    pub additional_theme: Vec<PathBuf>,
     /// Additional JS scripts to include at the bottom of the rendered page's
     /// `<body>`.
     pub additional_js: Vec<PathBuf>,
@@ -604,6 +606,7 @@ impl Default for HtmlConfig {
             copy_fonts: true,
             google_analytics: None,
             additional_css: Vec::new(),
+            additional_theme: Vec::new(),
             additional_js: Vec::new(),
             fold: Fold::default(),
             playground: Playground::default(),
@@ -822,6 +825,7 @@ mod tests {
         smart-punctuation = true
         google-analytics = "123456"
         additional-css = ["./foo/bar/baz.css"]
+        additional-theme = ["foobar"]
         git-repository-url = "https://foo.com/"
         git-repository-icon = "fa-code-fork"
 
@@ -869,6 +873,7 @@ mod tests {
             smart_punctuation: true,
             google_analytics: Some(String::from("123456")),
             additional_css: vec![PathBuf::from("./foo/bar/baz.css")],
+            additional_theme: vec![PathBuf::from("foobar")],
             theme: Some(PathBuf::from("./themedir")),
             default_theme: Some(String::from("rust")),
             playground: playground_should_be,
@@ -1049,6 +1054,7 @@ mod tests {
         smart-punctuation = true
         google-analytics = "123456"
         additional-css = ["custom.css", "custom2.css"]
+        additional-theme = ["barfoo"]
         additional-js = ["custom.js"]
         "#;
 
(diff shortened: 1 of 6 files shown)


xzi:
Is there anything that needs to be done before this can be merged upstream? Myself and a few people in a theming Discord were curious as to the status of this since this PR would greatly help our project out. 

I am more than willing to help out where I can with anything that needs to be finished up.

xzi:
@ehuss forgot to ping you in my original comment. Please see https://github.com/rust-lang/mdBook/pull/2059#issuecomment-2211182951

tmandry:
Thanks, I think this would help with my use case. Since I thought about implementing this independently I'll leave some comments for the benefit of the maintainers.

> * The theme class needs to specify ALL the variables, otherwise they default to no-style.
>   Can we figure out an easy way to derive the variable values from the presets?

I think this is fine for now. A theme inheritance mechanism would just add complexity and should be done later.

> * The syntax highlighting defaults to that of an assumed light mode ([see here](https://github.com/rust-lang/mdBook/blob/master/src/theme/book.js#L341)).
>   What would be the most appropriate way for the theme class to declare itself as a dark-based or light-based theme and have the correct syntax highlight selected?

The highlight CSS is easy enough to override with selectors like `.my_theme .hljs-type`. In a future change the builtin themes can be refactored to use this instead of depending on the javascript logic as they do now.

The only thing this doesn't work for is the editor theme. Maybe there is a CSS-only way to style that or maybe we'll need to extend the configuration to specify whether the theme is light or dark. Either way since the editor is an optional feature, I think it's fine to merge this PR as-is.

tmandry:
It would be ideal to decouple the theme name in the UI from the class name used for the theme.

tmandry:
I opened #2594 with a smaller diff that requires no JS logic.

PR #2061, opened 2023-04-13
Accessibility - Use modifiers for chapter nav and add shortkey settings

### LINKED ISSUE #518: [Feature request] make keyboard nav optional

lionel-rowe:
Fixes https://github.com/rust-lang/mdBook/issues/518 and https://github.com/rust-lang/mdBook/issues/1789. Supersedes https://github.com/rust-lang/mdBook/pull/2043.

This PR changes the default chapter navigation key combos from bare <kbd>←</kbd> and <kbd>→</kbd> to <kbd>Ctrl</kbd>+<kbd>←</kbd> and <kbd>Ctrl</kbd>+<kbd>→</kbd> (Windows) or <kbd>⌘</kbd>+<kbd>←</kbd> and <kbd>⌘</kbd>+<kbd>→</kbd> (Mac). The bare arrow key combos are bad for accessibility and very easy to press by mistake for users who expect arrow keys to scroll the page, which is the usual and expected behavior. Hitboxes of the chapter navigation buttons are also reduced to prevent accidental clicks.

It also adds options to make all shortkeys user-configurable. These options are included in the new Settings menu, which also allows for setting the theme and takes the place of the Theme menu in the menu bar. The new options make it easy for users to switch back to the old behavior if desired or choose their own key combinations to avoid any functionality clashes with screen readers, browser extensions, or other 3rd-party software.

Shortcuts for all keyboard-accessible elements are also surfaced to users on hover by appending them to the `title` of the relevant elements, and to assistive software by setting `aria-keyshortcuts`.

The options also allow configuring the shortkeys for _Toggle Table of Contents_, _Search_, and _Open settings_.

Finally, all shortkeys are temporarily disabled whenever the currently focused element is an `input`, `textarea`, or is `contentEditable`. This should fix issues of unexpected behavior in books that use custom editable elements.

image

image

### DIFF
--- a/guide/src/guide/reading.md
+++ b/guide/src/guide/reading.md
@@ -21,7 +21,7 @@ In that situation, the menu icon (three horizontal bars) at the top-left of the 
 
 The **arrow buttons** at the bottom of the page can be used to navigate to the previous or the next chapter.
 
-The **left and right arrow keys** on the keyboard can be used to navigate to the previous or the next chapter.
+Pressing <kbd>Ctrl</kbd>+<kbd>←</kbd> and <kbd>Ctrl</kbd>+<kbd>→</kbd> (<kbd>⌘</kbd>+<kbd>←</kbd> and <kbd>⌘</kbd>+<kbd>→</kbd> on Mac) on the keyboard can be used to navigate to the previous or the next chapter.
 
 ## Top menu bar
 
@@ -31,7 +31,7 @@ The icons displayed will depend on the settings of how the book was generated.
 | Icon | Description |
 |------|-------------|
 | <i class="fa fa-bars"></i> | Opens and closes the chapter listing sidebar. |
-| <i class="fa fa-paint-brush"></i> | Opens a picker to choose a different color theme. |
+| <i class="fa fa-cog"></i> | Opens a settings menu for setting a different color theme or shortcut keys. |
 | <i class="fa fa-search"></i> | Opens a search bar for searching within the book. |
 | <i class="fa fa-print"></i> | Instructs the web browser to print the entire book. |
 | <i class="fa fa-github"></i> | Opens a link to the website that hosts the source code of the book. |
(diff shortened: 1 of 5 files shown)


lionel-rowe:
> I think it would be better if the theme button is left as it currently is, and instead add an extra gear icon in the sidebar's bottom-left ...
> This placement signals that it's for general options about the whole reader, rather than something specific to the content like themes are.

I agree that simplicity is desirable, and I'm open to making changes, but I don't really follow the logic here — themes are also global to the reader rather than just the content, and placing settings under the ToC gives the false impression that the settings related to the ToC. It also makes the settings difficult to discover for books with ToCs taller than the screen height, unless the settings icon is a fixed element that obscures the bottom of the ToC. I also don't think people would typically expect to find settings in the bottom left corner, based on established UX patterns.

I'll wait for feedback from maintainers before making any further substantive changes.

maebli:
I have had some complaints about the none intuitive back forward buttons as well. When there is an input fiield and a user wants to make a correction, he navigates away from the page. To me this is ok if there is no input, but there should be a way to disable the back/forward

GuillaumeGomez:
Sorry for the time it took for us to get back to your PR. It's doing a lot of things at once (adding a settings menu for example) and each of these changes would need a discussion. Please open issue for earch of them (one for each shortcut modifier and one for the setting menu) so we can take things one step at a time. Closing this PR in the
 meantime.

PR #2070, opened 2023-04-16
Added missing tests for MathJax to the test book

expikr:
image

### DIFF
--- a/test_book/src/SUMMARY.md
+++ b/test_book/src/SUMMARY.md
@@ -22,6 +22,7 @@
   - [Tables](individual/table.md)
   - [Tasks](individual/task.md)
   - [Strikethrough](individual/strikethrough.md)
+  - [MathJax](individual/mathjax.md)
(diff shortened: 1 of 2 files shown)


test_book/src/individual/mixed.md
```diff
@@ -29,9 +29,16 @@ fn main(){
 
 <kbd>Ctrl</kbd> + <kbd>S</kbd> saves a file.
 
-A random image sprinkled in between
+---
+
+Although markdown does not have a native figure caption syntax, one can use block quote as a general-purpose "environment" to emulate it, and max-level heading inside the quote as caption:
+
```
ehuss, line 35: Can you say why this was modified? I think this test was just help with showing that images work at all and how styling works.  This seems unrelated to adding a test page for MathJax? Adding this seems to not be in the spirit of this test, which is to have simple direct examples of various markdown elements.

ehuss (approved):
Thanks!

Issue #2060, opened 2023-04-10
Broken relative links when using first chapter as an index page.


HU90m:
### Problem

If there is no index page in the source directory, mdBook uses the first chapter as the index page. When doing so, mdBook doesn't render the page any differently. This causes relative links on the page to break, if the first chapter is within a sub-directory.

### Steps

I've created a repo for easy recreation: <https://github.com/HU90m/mdbook-broken-landing-page-links>.

### Possible Solution(s)

I think the most elegant solution is to use a redirect to the first chapter, if no index page is given. I've put together a fix, before reading the [issue assignment section of the contributor's guide](https://github.com/rust-lang/mdBook/blob/master/CONTRIBUTING.md#issue-assignment) 🙃.

You can find my suggested fix at https://github.com/HU90m/mdBook/tree/landing-page-links-fix . I'll happily open a PR if this solution sounds good to you. (I thought I'd avoid bombarding you with both an issue and PR.

### Version

```text
`v0.4.28`
```

Leroy-X:
Hi, may I ask when it will be fixed? It is now version 0.4.35. The problem still exists. I look forward to fixing it soon.
Thanks.

Leroy-X:
@HU90m Thank you very much for your reply, I'm not a developer and don't use Rust.

J3rome:
I am also facing this issue (#2380). I haven't looked deeply at the code (I'm a rust noob) but I've observed that the links in the `print.html` page get properly rewritten. Perhaps the logic used when generating the `print.html` could be reused instead of doing a redirect ?

PR #2077, opened 2023-04-23
Fixed bug with the index page's relative links.

HU90m:
If there is no index page in the source directory, mdbook uses the first chapter as the index page. When doing so, mdbook doesn't render the page any differently. This causes relative links on the page to break, if the first chapter is within a sub-directory and there is not `index.md` (or `README.md`) in the root.

To fix this, a redirect to the first chapter is used instead of a copy of the page.

See bug report for more information: https://github.com/rust-lang/mdBook/issues/2060


KFearsoff (approved):
LGTM, thank you for your contribution!

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -120,22 +120,7 @@ impl HtmlHandlebars {
...
-                rendered_index,
-                &ctx.html_config.playground,
-                &ctx.html_config.code,
-                ctx.edition,
-            );
-            debug!("Creating index.html from {}", ctx_path);
-            utils::fs::write_file(&ctx.destination, "index.html", rendered_index.as_bytes())?;
-        }
-
-        Ok(())
```
nbdd0121, line 138: Could the redirection code be moved here, so we can avoid returning the URL and changing the is_index logic?

nbdd0121:
Not sure why the tag is removed...

@rustbot label +S-waiting-on-review

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -535,21 +520,36 @@ impl Renderer for HtmlHandlebars {
...
+                seen_index |= rendered_file == Path::new("index.html");
+                if first_non_draft.is_none() {
+                    first_non_draft = Some(rendered_file);
+                }
+            }
+        }
+        if !seen_index {
+            if let Some(first_file) = first_non_draft {
+                let redirect_to = first_file
+                    .to_str()
```
HU90m, line 546: Changing PR to draft because this implementation is buggy and shouldn't be merged. Simply converting path to string can produce bad URLs.

src/renderer/html_handlebars/hbs_renderer.rs
HU90m: Not really (discussed offline)

HU90m:
I haven't found time to make this merge-able for a little while, so closing.

PR #2080, opened 2023-04-28
Fix handling of copy-fonts=true when fonts.css is overridden

t2y:
I'm new to mdbook. This is my experience when I failed to configure it.

```bash
$ mdbook init --theme
$ mdbook serve
...
2023-04-28 12:47:53 [WARN] (mdbook::renderer::html_handlebars::hbs_renderer): output.html.copy_fonts is deprecated.
Set  and ensure the fonts you want are in the `theme/fonts/` directory.
...
```

After that, I added `copy_fonts=false` to the book.toml, but this warning is not resolved since it seems that an attribute name convension in book.toml is different.

So, I suggest showing this warning.

```
2023-04-28 12:39:46 [WARN] (mdbook::renderer::html_handlebars::hbs_renderer): output.html.copy_fonts is deprecated.
Set copy-fonts=false in book.toml and ensure the fonts you want are in the `theme/fonts/` directory.
```

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -294,7 +294,7 @@ impl HtmlHandlebars {
                 if html_config.copy_fonts {
                     warn!(
                         "output.html.copy_fonts is deprecated.\n\
-                        Set copy_fonts=false and ensure the fonts you want are in \
+                        Set copy-fonts=false in book.toml and ensure the fonts you want are in \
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks!  I pushed some changes since this wasn't working quite like I was intending. It definitely shouldn't warn when the theme directory is initialized. Since fonts.css is a new theme file, it was supposed to signal that the stock fonts aren't being used, but I didn't get things quite right when I added it.

PR #2089, opened 2023-05-23
Display line numbers for code examples

0x8000-0000:
This makes it easier to refer to a specific line of code while teaching or collaborating on a code snippet.


ehuss:
Thanks for the PR!  The documentation here is intended to document the defaults. If your intention is to change the default, that might require some discussion first.

ehuss:
I'm going to close since the examples are intended to show the defaults. Thanks!

Issue #443, opened 2017-09-14
Styling flash on RBE


steveklabnik:
This doesn't seem to be true 😄 Check out a page like https://rustbyexample.com/hello/comment.html; we get a flash of content.

azerupi:
By flash you mean the code blocks being styled?
I may be wrong, but to me, the most probable cause for this is highlight.js, because styling happens after the page loads. The only thing I can think of that would effectively get rid of this flash is to style the code beforehand with something like [syntect](https://github.com/trishume/syntect) which is something I would like to experiment with.

There may be other issues at play though, I don't know. :)

steveklabnik:
> By flash you mean the code blocks being styled?

That's the biggest offender, yeah.

> I may be wrong, but to me, the most probable cause for this is highlight.js, because styling happens after the page loads. 

Ahh bummer :/ Hmmm

projektir:
Some of this is also the Ace editor taking time to load?

There's a styling flash in the Rust Book, so that definitely seems to be highlight.js.

zqianem:
#1652 would fix the majority of this issue.

PR #2101, opened 2023-06-18
Avoid menu border flash during page navigation

zqianem:
Partially addresses #443

Previously when navigating between pages using the sidebar, the border of the menu bar would briefly appear before being hidden by JavaScript. This change fixes this flash by hiding the border initially and then letting JavaScript decide whether or not it should be shown.


ehuss (approved):
Thanks! I'm not sure if there was a reason it was done that way, but this seems good to me.

PR #2104, opened 2023-06-22
Fix flicker when setting sidebar scroll position

zqianem:
Previously, sidebar scroll was set in an external script which caused a flicker as the sidebar is initially rendered without any scroll before being scrolled to the desired location.

Switching to an inline script right after the HTML tags for the sidebar seems to avoid the flicker in most cases. In addition, logic is added to avoid scrolling jumps when navigating via links within the sidebar.

Partially addresses #443

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -551,13 +551,6 @@ function playground_text(playground, hidden = true) {
             firstContact = null;
         }
     }, { passive: true });
-
-    // Scroll sidebar to current active section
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks! I think I really like this! I always felt the scroll jumping was a little awkward.

I tried putting it under some stress tests, and still noticed a little bit of flicker in some cases, but I couldn't really figure out when because it was a little random. I also had Chrome running in 6x slow mode, because otherwise I couldn't ever see it.

PR #2108, opened 2023-06-29
Add `padding-bottom: 50vh` for <main>

Ddystopia:
Hello,
This PR introduces a simple tweak: adding extra padding to the bottom of the content area in mdBook.

Here's why it might be a good idea:

- More Comfortable Reading: This change lets users scroll the text up to the center of their screen. It's often easier to read from the middle of the screen and can help reduce eye-strain during long reads.

- Familiar Feel: This 'scroll-past-end' feature is common in popular code editors like VSCode and Atom. It might make mdBook feel a bit more like home for those folks.

This change doesn't break anything and can easily be overridden by custom stylesheets if needed.

Of course, we know changes like these can be a matter of personal preference. So, this PR is completely open for discussion. If you think it's not a great fit for the project, no worries at all if the PR gets closed.

Looking forward to hearing your thoughts!

### DIFF
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -90,7 +90,7 @@ h6:target::before {
 
 .content {
     overflow-y: auto;
-    padding: 0 5px 50px 5px;
+    padding: 0 5px 50vh 5px;
 }
 .content main {
     margin-left: auto;


GiorgioReale:
It seems like a good idea, but i think it's better to restore the previous printing style.

```css
.content {
    ...
    padding: 0 5px 50vh 5px;
}

@media print {
    .content {
        padding: 0 5px 50px 5px;
    }
}
```

Ddystopia:
> It seems like a good idea, but i think it's better to restore the previous printing style.
> 
> ```css
> .content {
>     ...
>     padding: 0 5px 50vh 5px;
> }
> 
> @media print {
>     .content {
>         padding: 0 5px 50px 5px;
>     }
> }
> ```

in file `src/theme/css/print.css` there is a style
```css
#content {
    max-width: none;
    margin: 0;
    padding: 0;
}
```
Id has more priority then class, so padding would already be zero.

GiorgioReale:
@Ddystopia 😅 Oops, sorry. I didn't see the ID `#content`, i only checked the class `.content`.

GuillaumeGomez:
Sorry for the time it took for us to get back to this PR. I'm not sure how good of an idea it is. Might be nicer if mdbook had JS settings like rustdoc so people could just enable this feature if they feel like it.

Anyway, I added the tag "needs discussion" for @rust-lang/mdbook to take a look.

camelid:
Yeah, I'm inclined to close. I don't normally see websites with this styling, and if I did see it, I'd think it was a bug.

GuillaumeGomez:
I think it's more common for text editors (I like having it there).

notriddle:
I double-checked a few websites that seem "good" to me, and they don't do this. Probably shouldn't be added as a option, *definitely* shouldn't be the default.

Ddystopia:
Well, closing or merging that's on you, but after I added it locally for myself via tamper monkey it became easier to read. That's all I can say. I really don't know what "should" "be right" or "good".

GuillaumeGomez:
None of the members of the team is positive about this, so let's close. Thanks for the proposition in any case!

PR #2116, opened 2023-07-08
Add oh-my-zsh quick exemple to shell completions

Stargateur:
I have trouble to find this information, doesn't cost must to add it here.


ehuss (approved):
Thanks!

PR #2118, opened 2023-07-09
Fix theme-color meta tag not syncing with the theme

### LINKED ISSUE #2117: status bar of phone is always white regardless of the theme
### Problem

The color of phone's status bar and web address bar should be the same as the background color of the webpage.

### Steps

1. Use a phone to browse a site that uses mdbook. For example: https://rust-lang.github.io/mdBook
2. Change theme to any but white. For example: `Ayu`
3. The phone's status bar and web address bar are always white.

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

_No response_

zica87:
Fixes #2117.

document.`body`.backgroundColor is always transparent (rgba(0, 0, 0, 0))
while
document.`documentElement`.backgroundColor gives the correct value.

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -346,7 +346,7 @@ function playground_text(playground, hidden = true) {
         }
 
         setTimeout(function () {
-            themeColorMetaTag.content = getComputedStyle(document.body).backgroundColor;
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

Issue #2142, opened 2023-07-25
Metadata on a per-page basis


cr0mll:
### Problem

For example, page keywords are only specified for the book as a whole - you cannot have separate keywords for each page which can make it difficult for people using search engines to find the correct page.

However, this can be augmented by also allowing one to specify for example author information for each page. This is useful for projects with multiple contributors which would like to give proper credit.

There should simply be a way to specify custom metadata for each page and maybe have some predefined metadata attributes such as keywords, author information, or time of creation/modification which mdbook handles on its own.

### Proposed Solution

The way https://chirpy.cotes.page/ do things seems like an appropriate way to implement. Metadata is inserted into each page as key-value pairs in a markdown comment at the top of each page.

The value of each metadata attribute should be accessible as a template in index.hbs with `{{ attribute_name }}`.

### Notes

This is _not_ a request to implement specific metadata tags such as keywords or author information, but simply a request to expose custom metadata for each page as templates in index.hbs.

noraj:
> Metadata is inserted into each page as key-value pairs in a markdown comment at the top of each page.

It's called front-matter (like in hexo https://hexo.io/docs/front-matter). Generally the easiest front-matter for markdown format is using YAML.

Potentially inspirational projects:

- https://lib.rs/crates/gray_matter
- https://github.com/r7kamura/fronma
- https://github.com/azdle/rust-frontmatter

Also [Zola](https://www.getzola.org/) Rust staticgen has some front-matters: https://github.com/search?q=repo%3Agetzola%2Fzola%20frontmatter&type=code

sspaeti:
I think this feature is very relevant if you want to share your book's subpage on socials and have an accurate description and feature image. I made a suggestion on how something like that could be implemented with front-matter in this comment: https://github.com/rust-lang/mdBook/issues/1416#issuecomment-1953878338.

I guess it wouldn't be too hard, we'd need a front-matter parser (here some I found):
- https://crates.io/crates/extract-frontmatter/
- https://lib.rs/crates/gray_matter ([GitHub](https://github.com/the-alchemists-of-arland/gray-matter-rs))
- https://github.com/r7kamura/fronma

And just include these in the `index.rbs` as illustrated in my [comment](https://github.com/rust-lang/mdBook/issues/1416#issuecomment-1953878338.). What do you think?

Issue #277, opened 2017-05-16
Allow chapters to have an optional TOML header


azerupi:
The conclusion of the discussion #176 is to support an optional TOML header in the markdown sources to provide chapter specific information. Chapter specific information that overrides the global configuration for this specific chapter.

A chapter source file could look like the following:

```md
------
author = "John Doe"
title = "A book of little words"
translator = "Jane Doe"
------

# Chapter title

The rest of the file is **normal** markdown
```

This will probably require some changes to the internal representation.

steveklabnik:
The only thing I don't like about this is that then, they don't become
standalone markdown files.

mkovarik:
Is it planned for key-value pairs in the TOML header to be available for handlebars templating?

Michael-F-Bryan:
@mkovarik, we haven't done much towards adding a TOML header just yet, but I'm assuming it would be implemented by adding a `metadata` field to the [`Chapter`] struct which gets filled in during the [`load_chapter()`] process. All renderers would then see this field, so it'd be trivial for the HTML renderer to pass the key-value pairs through to the templates.

Did you have a particular use case in mind? 

[`Chapter`]: https://github.com/rust-lang-nursery/mdBook/blob/b5382f49c6b4348222d09c13688299c9aff12409/src/book/book.rs#L96-L110
[`load_chapter()`]: https://github.com/rust-lang-nursery/mdBook/blob/b5382f49c6b4348222d09c13688299c9aff12409/src/book/book.rs#L155

yech1990:
@Michael-F-Bryan That is useful for migrating from markdown blog into mdbook. Meanwhile metadata  for 'create date', 'last modify date' and 'keywords' can be preserved.

brendanzab:
This could be really useful for letting us specify HTML metadata. I think this is what [docusaurus does, for example](https://v2.docusaurus.io/docs/markdown-features#markdown-headers).

PR #2149, opened 2023-07-31
Add data property to chapter struct

bgotink:
This data property makes it possible to share data between pre-processors, or between pre-processors and the renderer, including the "html" renderer which exposes the chapter data as "data" in the .hbs files.

I've explicitly not provided a built-in way to configure data. Any built-in approach for parsing data would have to be discussed. While I consider hugo/jekyll style [front-matter](https://gohugo.io/content-management/front-matter/) pretty standard for this kind of thing, others might disagree. Even if everyone agreed, the question would then become "what flavours of front-matter do we support?"

I've also explicitly not created any data properties that are used out of the box by mdBook itself.

I'm assuming this PR would need to go in 0.5.0, as it changes the JSON model and any pre-processors compiled with older versions of mdBook would strip the property.

For reference, my use case: I have (many) large JSON files containing data, and I want to use a pre-processor to insert the correct data in every chapter, to then use it in our theme's `.hbs` file to generate a metadata table, links, that kind of stuff.

Related:

- Request for chapter metadata: #2142 
- Request for TOML front-matter: #277
- Other PR adding a data property to chapters: #1465. That PR adds a pre-processor to extrat TOML front-matter, and it defines a few data properties that are built into mdBook. This PR is more limited in scope.
- Crate [`mdbook_frontmatter`](https://docs.rs/mdbook-frontmatter/0.0.4/src/mdbook_frontmatter/lib.rs.html#1-139) which processes a fixed, limited, set of data properties in YAML front-matter

### DIFF
--- a/examples/nop-preprocessor.rs
+++ b/examples/nop-preprocessor.rs
@@ -138,7 +138,8 @@ mod nop_lib {
                                 "sub_items": [],
                                 "path": "chapter_1.md",
                                 "source_path": "chapter_1.md",
-                                "parent_names": []
+                                "parent_names": [],
+                                "data": {}
                             }
                         }
                     ],
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -165,6 +165,9 @@ pub struct Chapter {
     pub source_path: Option<PathBuf>,
     /// An ordered list of the names of each chapter above this one in the hierarchy.
     pub parent_names: Vec<String>,
+    /// Data that can be shared between preprocessors, accessed in the renderer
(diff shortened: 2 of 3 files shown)


CakeVR:
Hello, the lack of front matter support in `mdBook` is unfortunate, and this pull request seems to add great power for preprocessors.
If I were able to parse the front matter of each chapter, I would be able to generate Open Graph Meta Tags  *per* chapter in the `index.hbs`.
This allows for great link embed previews.

I'm sorry to see your work has not been looked into yet.
I do hope it gets reviewed and evaluated, thanks for your effort.

PR #2156, opened 2023-08-05
Add additional-hljs-packages option

Evian-Zhang:
As a syntax highlighting tool, highlight.js supports 3rd party languages (see [instructions here](https://github.com/highlightjs/highlight.js/blob/main/docs/language-contribution.rst)) since they don't currently merge new grammars into core. This can be done by placing additional JavaScript files alongside hightlight.js's js file, and then call the syntax highlighting functions provided by `hljs`, such as `hightlightBlock` used in our `book.js`.

However, this can't be done currently in mdbook, since the `additional-js` field is designed for js files called **after** the main `book.js`, while placing the highlight.js 3rd packages there leads to `highlightBlock` called **before** those packages loaded. On the other hand, moving js files specified in `additional-js` field above the `book.js` may lead to unexpected behaviors, since it is supposed to be executed after the `book.js` is executed.

So adding another field, `additional-hljs-packages`, could solve this problem, which will place those js files between the highlight.js's js file and `book.js`.

### DIFF
--- a/guide/src/format/configuration/renderers.md
+++ b/guide/src/format/configuration/renderers.md
@@ -102,6 +102,7 @@ mathjax-support = false
 copy-fonts = true
 additional-css = ["custom.css", "custom2.css"]
 additional-js = ["custom.js"]
+additional-hljs-packages = ["custom.min.js"]
 no-section-label = false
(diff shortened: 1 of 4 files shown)


Evian-Zhang:
Oh yes, I forgot this way. Thank you!

PR #2162, opened 2023-08-08
Fix code blocks "indent" without highlight.js

ISSOtm:
The `.hljs` class added by highlight.js adds a `display: block` rule which makes `padding` apply correctly (left padding on all lines, not just the first one).
This change ensures that said rule is applied even if highlight.js doesn't run for whatever reason.

Comparison ([source](https://gbdev.io/pandocs/Audio_Registers.html#ff12--nr12-channel-1-volume--envelope)):

Without

With

### DIFF
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -260,6 +260,7 @@ pre > .buttons button {
     }
 }
 pre > code {
+    display: block;
     padding: 1rem;
 }
 


avivace (approved):

ehuss (approved):
Thanks!

ISSOtm:
Thank *you*!

PR #2168, opened 2023-08-16
Add prefetch to next link

### LINKED ISSUE #1975: Add prefetch to next link
https://github.com/rust-lang/mdBook/blob/41a6f0d43e1a2d9543877eacb4cd2a017f9fe8da/src/theme/index.hbs#L213

I was navigating through some books with short pages like https://google.github.io/comprehensive-rust/control-flow/break-continue.html, was trying to go through pages and pages through next button `]]` using vimium extension, I find it a bit slow, maybe prefetching would help and loading the user waiting time for the next page? Like `<a rel="next prefetch" ...>`.

I think maybe the previous don't need it since the chance of reading the previous page over the next pages is lesser, and books are usually not expected to read in the reversed order.

pickfire:
Fix #1975

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -217,7 +217,7 @@
                         {{/previous}}
 
                         {{#next}}
-                            <a rel="next" href="{{ path_to_root }}{{link}}" class="mobile-nav-chapters next" title="Next chapter" aria-label="Next chapter" aria-keyshortcuts="Right">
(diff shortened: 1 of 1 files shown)


Dylan-DPC:
Thanks. Sadly prefetch isn't supported on Safari so we need to take that into account as well

pickfire:
How would you suggest we go about this? I think if it isn't supported it is fine too, given that it works on the other browsers.

Dylan-DPC:
I am fine with this as, as long as it doesn't break on safari. Just wanted to mention it so it's noted :P

Dylan-DPC:
Ok i'm happy with this since it will do nothing on safari and that's fine till there is an interest in having an alternative. Thanks

PR #2170, opened 2023-08-21
Fixing clippy lints

gents83:

PR #2172, opened 2023-08-21
Adding LinksPreprocessor for SUMMARY.md

gents83:
This commit is intended to add the possibility to reuse the LinksPreprocessor inside SUMMARY.md

Example:
This would allow to use something like {{#include autogenerated.md}} to allow a more dynamic summary with an autogenerated md file that is left to the user


gents83:
Sorry - some modifications that appear here are just related to the other PR: https://github.com/rust-lang/mdBook/pull/2170

PR #2173, opened 2023-08-21
upgrade to notify 6.1.1 and watcher-mini 4.1

0xpr03:
@Dylan-DPC here you go


ehuss (approved):
Thanks! I don't really have much context on why this is updating, but I'll trust that you know what's best here.

Issue #859, opened 2018-12-22
Mobile - Remove the CSS sidebar


travankor:
I usually browse things on my mobile phone without javascript, since javascript eats my phone's battery life (thanks advertising/analytics companies?).

mdBook generally works well on mobile w/o JS, except for the large CSS sidebar that takes up half of my phone's screen. Instead of trying to auto-hide the sidebar using JS, could there be a better way to integrate the sidebar? FWIW, mkdocs does not have this issue.

I know this is a somewhat niche case, but it would be great to not have to enable JS just to read simple markdown.

zcdunn:
I would be in favor of a CSS solution, too, even though I don't disable JS. Less JS is always a good thing and the current JS sidebar is finicky on mobile.

Dylan-DPC:
@travankor is it possible to send a PR that makes this change? thanks

PR #2175, opened 2023-08-24
Sidebar but Pure CSS, fix rust-lang/mdBook#859

qaqland:
* change sidebar-toggle button by input:checked and label
* change nav-wrappers' CSS selectors
* remove loading animation when page opens with z-index

Without JavaScript, the webpage cannot cache the state of whether the sidebar is open or closed. To simplify removing the transition animation of the sidebar `translateX` when the page loads, I have set it to be closed by default in no JavaScript environment.

### DIFF
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -93,7 +93,7 @@ a > .hljs {
     display: flex;
     margin: 0 5px;
 }
-.no-js .left-buttons {
+.no-js .left-buttons button {
     display: none;
 }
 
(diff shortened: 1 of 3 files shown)


WhistlingZephyr (approved):
The code looks good to me, I also tested it locally (on Firefox and Chromium) with JS disabled to verify.

Dylan-DPC:
Thanks for the pr.

EzioTheDeadPoet:
I don't know which other change was possibly made recently to MD Book but the Wiki we host for Wabbajack is broken on mobile with no way of disabling the side bar and the side bar overlapping with the content of the website. This is the only change in the recent changelog that explicitly mentions changes to the side bar so I must assume that this is the change that broke the site for mobile devices.

EzioTheDeadPoet:
A little update on this:

After rolling back the mdBook version used by GitHub actions to both 0.4.33 and 0.4.34 the issue is gone.

WhistlingZephyr:
@EzioTheDeadPoet
It seems like your theme hasn't rolled out a release that supports this change yet. The fix has already been pushed to GitHub but the cargo release is still pending, <https://github.com/catppuccin/mdBook/pull/73>. So your best bet is to wait for the release and use a compatible version in the meantime.

P.S. I see that you have a raw .exe in your repo, I'd recommend managing it using cargo instead as it makes testing it on other OSes harder while also not maintaining complete open-sourceness.

EzioTheDeadPoet:
Pretty sure I tried that but had issues with getting that to run on my system when setting everything up.

PR #2187, opened 2023-09-08
Add `.warning` class for doc author use

notriddle:
This is designed to be compatible with rustdoc's version, in https://github.com/rust-lang/rust/pull/106561


ehuss (approved):
Thanks!

PR #2197, opened 2023-09-19
Update index.hbs

dluschan:
Trailing slash on void elements has no effect and interacts badly with unquoted attribute values.


ehuss (approved):
Thanks!

PR #2209, opened 2023-10-03
Add resize indicator to the sidebar

### LINKED ISSUE #2186: "I didn't know that the sidebar could be resized"
### Problem

https://github.com/rust-lang/rust/pull/115660#issuecomment-1712050928

Based on usability discussions about adding this same feature to rustdoc, at least two expert users didn't know you could resize the navigation panel in mdBook.

### Proposed Solution

Add a grab handle to the UI, next to the sidebar (like this, though the mockup shows rustdoc instead of mdbook).

### Notes

_No response_

cyevgeniy:
It fixes #2186.

Indicator has the same color as icons, which is theme-dependent. On devices with limited accuracy ( tablets and smartphones most of the time) indicator is hidden.

### DIFF
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -269,6 +269,11 @@ pre > .buttons button {
         /* On mobile, make it easier to tap buttons. */
         padding: 0.3rem 1rem;
     }
+
+    .sidebar-resize-indicator {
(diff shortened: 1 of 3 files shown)


ehuss (approved):
Thanks!

Issue #2215, opened 2023-10-09
google-analytics is not working


ettoreleandrotognoli:
### Problem

I added the google analytics tag but is was not collecting data

### Steps

add your google analytics tag to book.toml

```toml
[output.html]
google-analytics = "*****"
```

![Screenshot from 2023-10-09 16-40-07](https://github.com/rust-lang/mdBook/assets/4960560/fa5a23c5-7d5c-49a5-9109-202b0c5b9000)

The HTTP answer is 200 but the data is not been shown on google analytics dashboard

### Possible Solution(s)

Update the google analytics script to use collect?v=2 instead of collect?v=1

### Notes

Google is suggesting a new js code

### Version

```text
mdbook v0.4.35
```

ehuss:
Closing per https://github.com/rust-lang/mdBook/issues/2720 where this has been removed.

PR #2216, opened 2023-10-09
update google-analytics and add ipv6 localhost address

ettoreleandrotognoli:
resolves https://github.com/rust-lang/mdBook/issues/2215

### DIFF
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -268,20 +268,17 @@
         {{/if}}
 
         {{#if google_analytics}}
-        <!-- Google Analytics Tag -->
+        <!-- Google tag (gtag.js) -->
+        <script async="async" src="https://www.googletagmanager.com/gtag/js?id={{google_analytics}}"></script>
         <script>
-            var localAddrs = ["localhost", "127.0.0.1", ""];
-
+            var localAddrs = ["localhost", "127.0.0.1", "[::1]", ""];
             // make sure we don't activate google analytics if the developer is
             // inspecting the book locally...
             if (localAddrs.indexOf(document.location.hostname) === -1) {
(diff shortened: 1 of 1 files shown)


joshka (commented):
This looks reasonable based on the installation instructions at https://developers.google.com/tag-platform/tag-manager/datalayer#installation

It would be worth showing you've tested this with a couple of screenshots.

ettoreleandrotognoli:
The site that I'm using it this one:
https://ettoreleandrotognoli.gitlab.io/unimar-object-oriented-programming-i/

This print any one can reproduce inspecting the site:
Screenshot from 2023-10-18 18-19-46

This one is from the google analytics site, it needs authentication
Screenshot from 2023-10-18 18-21-50

joshka (approved):
LGTM

ehuss:
I'm reluctant to merge this since the google analytics option is deprecated and is intended to be removed (there is also a large warning telling you not to use it). I also don't particularly want to research to see what the possible consequences of changing this are. I have a vague understanding that Google has changed from one system to another, but I don't know if that migration would break existing configurations, or what the status is of that migration.

I'm going to suggest that instead of fixing the code here that you use a custom `theme/head.hbs` file instead.

KFearsoff:
> I'm going to suggest that instead of fixing the code here that you use a custom `theme/head.hbs` file instead.

I think it's a good idea, but it would be nice to include the code snippet into documentation (into the separate "user-contributed" section in mdbook's book, perhaps?). This is a good contribution and I think it shouldn't go to waste even if we pull the plug on supporting Google Analytics. After that we can consider closing the PR?

ehuss:
Closing per https://github.com/rust-lang/mdBook/issues/2720 where this has been removed.

PR #2225, opened 2023-10-22
💄 Dark and Light Image Variants

hayesall:
This picks up pull request #1991 but implements a different light/dark mode variant that should be easier in markdown.

Writers can append `#only-light` or `#only-dark` to the end of an image path:

```markdown
![rust logo standard edition](images/rust-logo-blk.svg#only-light)
![rust logo dark mode](images/rust-logo-blk-dark.svg#only-dark)
```

... which renders:

https://github.com/rust-lang/mdBook/assets/11916674/f01c03ef-854f-484e-8314-7ac78bc8c5ab


ehuss:
Thanks, this looks interesting!

I'm a little concerned about hard-coding the themes, as this won't work if the user adds their own themes. The `general.css` file should not have any theme values in it. Is it at all possible to make it so that the CSS rules can be based on the `color-scheme` of the theme? My quick searching didn't turn up anything useful, but perhaps the `color-scheme` could be exposed as a data attribute on the root tag? That might be a little more complex than this solution since the javascript code would need to update it when it updates the class list. I'm just wondering what kinds of options there are.

GuillaumeGomez:
Closing as @ehuss comment was not answered. Please feel free to reopen if you have more information to provide.

PR #2229, opened 2023-10-31
Fix `mdbook serve` unexpected panic

### LINKED ISSUE #2228: Unexpected panic 'path is expected to be under the root'
### Problem

When maintaining official `rustup` docs, I have noticed that `mdbook serve` will work at first but fail immediately after a file update, which can break my workflow quite often:

```sh
> mdbook serve ./doc/user-guide --open
2023-10-30 17:17:41 [INFO] (mdbook::book): Book building has started
2023-10-30 17:17:41 [INFO] (mdbook::book): Running the html backend
2023-10-30 17:17:41 [INFO] (mdbook::cmd::serve): Serving on: http://localhost:3000
2023-10-30 17:17:41 [INFO] (mdbook): Opening web browser
2023-10-30 17:17:41 [INFO] (warp::server): Server::run; addr=[::1]:3000
2023-10-30 17:17:41 [INFO] (warp::server): listening on http://[::1]:3000
2023-10-30 17:17:41 [INFO] (mdbook::cmd::watch): Listening for changes...
2023-10-30 17:21:24 [ERROR] (mdbook::cmd::serve): Unable to serve: panicked at 'path is expected to be under the root', /Users/brew/Library/Caches/Homebrew/cargo_cache/registry/src/index.crates.io-6f17d22bba15001f/ignore-0.4.20/src/gitignore.rs:227:9
```

### Steps

1. `git clone https://github.com/rust-lang/rustup.git && cd rustup`
1. `mdbook serve ./doc/user-guide --open`
1. Change anything under `./doc/user-guide` and observe the panic

### Possible Solution(s)

I'm not sure what this error message is about, but `.matched_path_or_any_parents(path, path.is_dir())` requires `path` to be absolute. Maybe at some point down the business logic we need to canonicalize the path so it can appear to "be under the root":

https://github.com/rust-lang/mdBook/blob/94e0a44e152d8d7c62620e83e0632160977b1dd5/src/cmd/watch.rs#L89-L99

### Notes

_No response_

### Version

```text
> mdbook --version
mdbook v0.4.35
```

leonzchang:
Currently the `paths` input of `filter_ignored_files` is absolute while `.matched_path_or_any_parents(path, path.is_dir())` has an assertion which requires path can not have root. This will break down the `mdbook serve`(#2228).
I have tested this issue locally with different path scope and this change ensures we convert the input path into a relative path so that won't break down `mdbook serve`.

https://github.com/rust-lang/mdBook/blob/94e0a44e152d8d7c62620e83e0632160977b1dd5/src/cmd/watch.rs#L89-L99

### DIFF
--- a/src/cmd/watch.rs
+++ b/src/cmd/watch.rs
@@ -4,6 +4,7 @@ use ignore::gitignore::Gitignore;
 use mdbook::errors::Result;
 use mdbook::utils;
 use mdbook::MDBook;
+use std::env;
 use std::path::{Path, PathBuf};
 use std::sync::mpsc::channel;
 use std::thread::sleep;
@@ -87,11 +88,16 @@ fn find_gitignore(book_root: &Path) -> Option<PathBuf> {
 }
 
 fn filter_ignored_files(ignore: Gitignore, paths: &[PathBuf]) -> Vec<PathBuf> {
+    let current_dir = env::current_dir().expect("Unable to determine the current directory");
+
     paths
         .iter()
         .filter(|path| {
+            let normalized_path = path
+                .strip_prefix(&current_dir)
+                .expect("Could not normalize path");
             !ignore
-                .matched_path_or_any_parents(path, path.is_dir())
+                .matched_path_or_any_parents(normalized_path, normalized_path.is_dir())
                 .is_ignore()
         })
         .map(|path| path.to_path_buf())
--- a/src/cmd/watch.rs
+++ b/src/cmd/watch.rs
@@ -87,8 +87,8 @@ fn find_gitignore(book_root: &Path) -> Option<PathBuf> {
         .find(|p| p.exists())
 }
 
-// Using `canonicalize` can fail on Windows sometimes, might be a potential risk.
-// see ([https://github.com/rust-lang/mdBook/pull/2229#discussion_r1408665981).
+// Note: The usage of `canonicalize` may encounter occasional failures on the Windows platform, presenting a potential risk.
+// For more details, refer to [Pull Request #2229](https://github.com/rust-lang/mdBook/pull/2229#discussion_r1408665981).
 fn filter_ignored_files(ignore: Gitignore, paths: &[PathBuf]) -> Vec<PathBuf> {
     let ignore_root = ignore
         .path()
--- a/src/cmd/watch.rs
+++ b/src/cmd/watch.rs
@@ -87,6 +87,8 @@ fn find_gitignore(book_root: &Path) -> Option<PathBuf> {
         .find(|p| p.exists())
 }
 
+// Using `canonicalize` can fail on Windows sometimes, might be a potential risk.
+// see ([https://github.com/rust-lang/mdBook/pull/2229#discussion_r1408665981).
 fn filter_ignored_files(ignore: Gitignore, paths: &[PathBuf]) -> Vec<PathBuf> {
     let ignore_root = ignore
         .path()
--- a/src/cmd/watch.rs
+++ b/src/cmd/watch.rs
@@ -5,7 +5,6 @@ use mdbook::errors::Result;
 use mdbook::utils;
 use mdbook::MDBook;
 use pathdiff::diff_paths;
-use std::env;
 use std::path::{Path, PathBuf};
 use std::sync::mpsc::channel;
 use std::thread::sleep;
@@ -89,13 +88,16 @@ fn find_gitignore(book_root: &Path) -> Option<PathBuf> {
 }
 
 fn filter_ignored_files(ignore: Gitignore, paths: &[PathBuf]) -> Vec<PathBuf> {
-    let current_dir = env::current_dir().expect("Unable to determine the current directory");
+    let ignore_root = ignore
+        .path()
+        .canonicalize()
+        .expect("ignore root canonicalize error");
 
     paths
         .iter()
         .filter(|path| {
             let relative_path =
-                diff_paths(&current_dir, &path).expect("One of the paths should be an absolute");
+                diff_paths(&path, &ignore_root).expect("One of the paths should be an absolute");
             !ignore
                 .matched_path_or_any_parents(&relative_path, relative_path.is_dir())
                 .is_ignore()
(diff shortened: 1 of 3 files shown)


leonzchang:
> Could we have some information in the PR description, such as
> 
> * Why this is a correct fix to the linked issue.
> * How to verify this fix, either manually or having some tests.
> 
> That would largely increase the chance for this PR getting reviewed.

Sure, I have updated some description.

src/cmd/watch.rs
```diff
@@ -87,11 +89,15 @@ fn find_gitignore(book_root: &Path) -> Option<PathBuf> {
 }
 
 fn filter_ignored_files(ignore: Gitignore, paths: &[PathBuf]) -> Vec<PathBuf> {
+    let current_dir = env::current_dir().expect("Unable to determine the current directory");
```
ehuss, line 92: I don't think using the current_dir is correct. The ignore builder is built relative to the book root, which may not be the same as the current_dir. Since paths in the ignore file are going to be relative to the book root, using paths relative to the current directory won't match.

Additionally, mdbook can watch directories outside of the book root (such as `extra_watch_dirs`).

src/cmd/watch.rs
```diff
@@ -87,11 +89,15 @@ fn find_gitignore(book_root: &Path) -> Option<PathBuf> {
...
+
     paths
         .iter()
         .filter(|path| {
+            let relative_path =
+                diff_paths(&current_dir, &path).expect("One of the paths should be an absolute");
```
ehuss, line 98: IIUC, the fundamental problem is that the original root path might have path components like `.` or `..`, and the paths we're iterating over sometimes don't have those. Since `ignore` is wanting the path prefixes to match exactly, it is failing.

The solution here is to always use relative paths which `ignore` seems to be happier with.

We can't use `strip_prefix` because it doesn't handle `..` very well.

Do I understand the problem correctly? Were there other considerations you made for why you chose this approach?

Can you maybe add a unittest for `filter_ignored_files` that includes a few wonky examples (for example, files outside the gitignore root, gitignore root with `.` or `..` components, where the given paths don't have those, or paths that do have `.` or `..`). For example, on Windows you could have a gitignore root of something like `C:\\example\\src\\../doc/user-guide"` and a changed path that looks like `C:\\example\\src\\../doc/user-guide\\src\\index.md`.  I would imagine these tests have a few different ignore patterns (like `*.tmp` or `src/foo.md`) and check how it behaves with paths outside the ignore root.

leonzchang: Sorry for didn't explain it clearly, I think the fundamental problem is [`ignore .matched_path_or_any_parents`](https://github.com/BurntSushi/ripgrep/blob/7099e174acbcbd940f57e4ab4913fee4040c826e/crates/ignore/src/gitignore.rs#L220-L242) input requires a relative path to where(`current dir`) you execute the `mdbook` command.


In this change, what I have tested with `rustup/doc` locally:
```bash
# same path level with book
mdbook serve . --open

# outer path level with book
mdbook serve ./user-guide --open

# inner path level with book
mdbook serve ../doc/user-guide --open

# absolute path
mdbook serve /Users/xxx/Desktop/rustup/doc/user-guide/ --open

# relative path
mdbook serve ../newfolder/../rustup/doc/user-guide  --open
```

ehuss:
Hi @leonzchang! I noticed you requested a review, but I don't see any new commits since the last review. The two issues above noted that this shouldn't be using the current directory, and there was a request to add some tests. Do you think you can work on those?

leonzchang:
> Sorry for not explaining it clearly. I think the fundamental problem is [`ignore.matched_path_or_any_parents`](https://github.com/BurntSushi/ripgrep/blob/7099e174acbcbd940f57e4ab4913fee4040c826e/crates/ignore/src/gitignore.rs#L220-L242) input requires a relative path to the current directory where you execute the `mdbook` command, but we are currently providing an absolute path to `ignore.matched_path_or_any_parents`.
> 
> In this change, here's what I have tested with `rustup/doc` locally:
> 
> ```shell
> # Same path level with the book
> mdbook serve . --open
> 
> # Outer path level with the book
> mdbook serve ./user-guide --open
> 
> # Inner path level with the book
> mdbook serve ../doc/user-guide --open
> 
> # Absolute path
> mdbook serve /Users/xxx/Desktop/rustup/doc/user-guide/ --open
> 
> # Relative path
> mdbook serve ../newfolder/../rustup/doc/user-guide  --open
> ```

@ehuss sure, I just want to confirm if it makes sense to you to use the current directory before I proceed.

ehuss:
Hm, maybe I'm not understanding correctly, but I believe that `matched_path_or_any_parents` needs a path relative to the ignore root (the path passed to `Gitignore::new`), or an absolute path that can be stripped from the ignore root. The paths it takes should not be relative to the current directory.

leonzchang:
@ehuss you are right. It should be ignore root, I have changed it and added some tests.

src/cmd/watch.rs
leonzchang: update in b0a001c

src/cmd/watch.rs
leonzchang: update in b0a001c

leonzchang:
friendly ping, @ehuss

src/cmd/watch.rs
```diff
@@ -179,3 +187,50 @@ where
...
+    fn test_filter_ignored_files() {
+        let current_dir = env::current_dir().unwrap();
+
+        let ignore = GitignoreBuilder::new(&current_dir)
+            .add_line(None, "*.html")
+            .unwrap()
+            .build()
+            .unwrap();
+        let should_remain = path_to_buf(current_dir.to_str().unwrap(), "record.text");
+        let should_filter = path_to_buf(current_dir.to_str().unwrap(), "index.html");
```
ehuss, line 213: I'm not entirely clear why the `path_to_buf` helper exists, as this should all be doable with the regular path APIs. This is taking a Path, converting it to a string, then converting it back to a PathBuf. Can you rework these to avoid that helper?

```suggestion
        let should_remain = current_dir.join("record.text");
        let should_filter = current_dir.join("index.html");
```

src/cmd/watch.rs
```diff
@@ -179,3 +187,50 @@ where
...
+    fn filter_ignored_files_should_handle_parent_dir() {
+        let current_dir = env::current_dir().unwrap();
+
+        let ignore = GitignoreBuilder::new(&current_dir)
+            .add_line(None, "*.html")
+            .unwrap()
+            .build()
+            .unwrap();
+
+        let parent_dir = format!("{}/../", current_dir.to_str().unwrap());
```
ehuss, line 229: Similarly, this can be:

```suggestion
        let parent_dir = current_dir.join("..");
```

src/cmd/watch.rs
leonzchang: Fix 40c034e. Thanks for feedback, this is much way cleaner.

src/cmd/watch.rs
leonzchang: Fix 40c034e.

src/cmd/watch.rs
```diff
@@ -87,11 +88,18 @@ fn find_gitignore(book_root: &Path) -> Option<PathBuf> {
...
+    let ignore_root = ignore
+        .path()
+        .canonicalize()
+        .expect("ignore root canonicalize error");
+
     paths
         .iter()
         .filter(|path| {
+            let relative_path =
+                diff_paths(&path, &ignore_root).expect("One of the paths should be an absolute");
```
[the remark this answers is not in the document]
leonzchang: I added comment for this.
https://github.com/rust-lang/mdBook/blob/8d4ae388fa237f2406e57b9ea37f449a95884b43/src/cmd/watch.rs#L90-L109

ehuss (approved):
Thanks!

PR #2232, opened 2023-11-08
Add Nim to default languages

arnetheduck:
Nim is a systems programming language (included in the highlight.js `system` group), and we're quite happily using `mdBook` in several of our documentation projects starting with our [style guide](https://status-im.github.io/nim-style-guide/).

While we can maintain our own highlight.js, including `Nim` in the default distribution would allow us to promote more mdBook usage in the Nim community at the cost of a ~2kb increase in the `highlight.js` size.


ehuss (approved):
Thanks!

PR #2235, opened 2023-11-10
Make sure page wrapper transform is removed in print mode

### LINKED ISSUE #1932: Print page on firefox includes blank column when sidebar is visible
On firefox, if the sidebar is visible, generating the print page will incorrectly include visible whitespace on the left side.

Actual:

image

Expected:

image

This doesn't seem to affect Chrome or Safari

Tested with Firefox 107.0.

ardcore:
resolves #1932 - this bug was actually present in all browsers (chrome, chromium, brave, firefox), not just firefox as stated in the issue.

### DIFF
--- a/src/theme/css/print.css
+++ b/src/theme/css/print.css
@@ -7,7 +7,7 @@
 }
 
 #page-wrapper.page-wrapper {
-    transform: none;
+    transform: none !important;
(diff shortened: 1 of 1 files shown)


Dylan-DPC:
Thanks. I've tested it with the help of someone else and it looks good :)

Issue #2035, opened 2023-02-27
Poll option for file watching


digitalcora:
### Problem

Some file systems one might reasonably want to use with mdBook don't support change events ([mounted Windows file systems under WSL2](https://github.com/microsoft/WSL/issues/4739) are rather infamous for this). If your book is on such a file system, the `watch` and `serve` commands cannot detect changes, making them not very useful.

### Proposed Solution

The `watch` and `serve` commands could have an option (e.g. `--poll`) to detect changes by periodically polling the file system.

### Implementation

I'm not at all familiar with Rust, but it appears the `notify` package used by mdBook already has a [`PollWatcher`](https://docs.rs/notify/latest/notify/poll/struct.PollWatcher.html) intended for just this situation. So, mdBook would not have to implement this logic from scratch.

### Prior Art

Polling is supported by some popular static site generators; I'm familiar with it from [Hugo](https://gohugo.io/commands/hugo_server/#options), but Jekyll apparently has it as well. Hugo requires the user to specify a poll interval, but suggests 700ms as a default. In my experience this strikes a decent balance between responsiveness and resource usage for small sites, but those with more files might warrant a higher value.

### Workarounds

The command-line tool [`fswatch`](https://github.com/emcrisostomo/fswatch) supports polling, with a default interval of 1s. I've gotten it working with mdBook like so (then opening the output files using `file://` in the browser):

```
fswatch -m poll_monitor -o -r src | xargs -I{} mdbook build
```

This is not quite as ergonomic as a built-in option would be — it deletes and recreates everything in the output directory on every change, which takes a second or two, and if you refresh before it's ready you can get a 404 or missing CSS/fonts.

KFearsoff:
@rustbot label +C-bug +Command-serve +Command-watch +S-Duplicate

See https://github.com/rust-lang/mdBook/issues/2102#issuecomment-1810982299

Issue #1441, opened 2021-01-12
Changes in the destinations of symbolic links not detected


FrankHB:
Changes made in the destinations of symbolic links are not detected by `mdbook serve` (tested in the target `x86_64-pc-windows-gnu`).

The [CHANGELOG](https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#mdbook-044) is not accurate for this case (although it can work in cases like #1323).

KFearsoff:
@rustbot label +C-bug +Command-serve +Command-watch +S-Duplicate

See https://github.com/rust-lang/mdBook/issues/2102#issuecomment-1810982299

PR #2240, opened 2023-11-18
Use PollWatcher to always get filesystem updates

### LINKED ISSUE #2102: mdbook serve doesn't work in Docker
### Problem

Hello, my idea was to change files on host machine, then it will changes files in container and rebuild docs. And of cource, there will be port exposing to open docs on host machine in browser. 
I created Docker file and run container. It didn't work. I checked up, that files, that were changed on my machine, were changed in container and vice versa. According logs md serve didn't respond to changes. 
I'm only begginer, so don't rule out the option, that I did smth wrong.

### Dockerfile

```
FROM ubuntu
RUN apt-get update && apt-get install -y curl tar
RUN curl -sSL https://github.com/rust-lang/mdBook/releases/download/v0.4.27/mdbook-v0.4.27-x86_64-unknown-linux-gnu.tar.gz | tar -xz --directory=/bin
COPY /docs /docs
```

### Commads:

* `docker build -f docker/Dockerfile.local.docs -t docs-local-image .`
* `docker run -dp 45000:45000 -w /docs -v "$(pwd)/docs:/docs" --name docs-local docs-local-image sh -c "mdbook serve -n 0.0.0.0 -p 45000"`

### Possible Solution(s)

Mb it's Docker bug?

### Notes

Host OS: Windows 10
Docker version: 24.0.2 

### Version

_No response_

KFearsoff:
See https://github.com/rust-lang/mdBook/issues/2102#issuecomment-1810982299 for explanation on this

Closes #2102, #2035, #383 and #1441

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -9,6 +9,8 @@ use mdbook::errors::*;
 use mdbook::utils;
 use mdbook::utils::fs::get_404_output_file;
 use mdbook::MDBook;
+use notify::PollWatcher;
+use notify::RecommendedWatcher;
 use std::net::{SocketAddr, ToSocketAddrs};
 use std::path::PathBuf;
 use tokio::sync::broadcast;
@@ -43,6 +45,9 @@ pub fn make_subcommand() -> Command {
                 .help("Port to use for HTTP connections"),
         )
         .arg_open()
+        .arg(arg!(--compat "Watch files in compatibility mode.\n\
+                            Use this if your environment doesn't support filesystem events (Windows, Docker, NFS),
+                            or if you encounter issues otherwise"))
 }
 
 // Serve command implementation
@@ -97,23 +102,48 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
     }
 
     #[cfg(feature = "watch")]
-    watch::trigger_on_change(&book, move |paths, book_dir| {
-        info!("Files changed: {:?}", paths);
-        info!("Building book...");
+    let polling = args.get_flag("compat");
 
-        // FIXME: This area is really ugly because we need to re-set livereload :(
-        let result = MDBook::load(book_dir).and_then(|mut b| {
-            update_config(&mut b);
-            b.build()
+    #[cfg(feature = "watch")]
+    if polling {
+        debug!("Using PollWatcher backend");
+        watch::trigger_on_change::<_, PollWatcher>(&book, move |paths, book_dir| {
+            info!("Files changed: {:?}", paths);
+            info!("Building book...");
+
+            // FIXME: This area is really ugly because we need to re-set livereload :(
+            let result = MDBook::load(book_dir).and_then(|mut b| {
+                update_config(&mut b);
+                b.build()
+            });
+
+            if let Err(e) = result {
+                error!("Unable to load the book");
+                utils::log_backtrace(&e);
+            } else {
+                let _ = tx.send(Message::text("reload"));
+            }
         });
+    } else {
+        debug!("Using RecommendWatcher backend");
+        watch::trigger_on_change::<_, RecommendedWatcher>(&book, move |paths, book_dir| {
+            info!("Files changed: {:?}", paths);
+            info!("Building book...");
 
-        if let Err(e) = result {
-            error!("Unable to load the book");
-            utils::log_backtrace(&e);
-        } else {
-            let _ = tx.send(Message::text("reload"));
-        }
-    });
+            // FIXME: This area is really ugly because we need to re-set livereload :(
+            let result = MDBook::load(book_dir).and_then(|mut b| {
+                update_config(&mut b);
+                b.build()
+            });
+
+            if let Err(e) = result {
+                error!("Unable to load the book");
+                utils::log_backtrace(&e);
+            } else {
+                let _ = tx.send(Message::text("reload"));
+            }
+        });
+    }
 
     let _ = thread_handle.join();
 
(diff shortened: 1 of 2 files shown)


ehuss:
I vaguely recall trying the PollWatcher and having issues with it, but I can't remember what they were. I think it was mostly performance problems.

One issue is that with this change, mdBook always prints an error when it starts (assuming you don't have a theme directory):

`2023-11-24 12:58:56 [WARN] (mdbook::cmd::watch): error while watching for changes: No such file or directory (os error 2) about ["/Users/eric/Temp/z30/theme"]`

IIUC, the way this works is that every second it scans all the files, and checks if any mtime changes are there. Is that correct?  I'm not sure I understand why this would need a debouncer if it scans every second, and the debouncer is removing duplicates once a second. How do the two of those interact?

For larger books, what is the expected performance concerns?  What if the source is over a network drive?

cc @0xpr03, do you perhaps have some guidance on our usage of notify and using a PollWatcher?

KFearsoff:
> One issue is that with this change, mdBook always prints an error when it starts (assuming you don't have a theme directory):

Makes sense. Shouldn't be hard to fix.

> IIUC, the way this works is that every second it scans all the files, and checks if any mtime changes are there. Is that correct?

Yes! That's completely right.

> I'm not sure I understand why this would need a debouncer if it scans every second, and the debouncer is removing duplicates once a second. How do the two of those interact?

Truth be told, I can't tell you this, because I simply don't know. Upstream provides [an example](https://github.com/notify-rs/notify/blob/c4d047078549c2fec0bbfada29653097776b8b2d/examples/debouncer_mini_custom.rs) of using debouncer with PollWatcher. I'm not sure if that makes sense, but my handwavy assumption is that debouncer may provide us with some safeguards against really weird edge cases like filesystem being extremely unresponsive for a second and then submitting two events simultaneously.

> For larger books, what is the expected performance concerns?

Hard to say without benchmarking it. Alacritty seems to use [1 second](https://github.com/alacritty/alacritty/blob/54889fc4ff117f59efb392955a18b414fee570c7/alacritty/src/config/monitor.rs#L16C20-L16C20), cargo-watch uses [5 seconds](https://github.com/watchexec/cargo-watch/blob/da7e7f5c631adffce74be97949e7aadfaff1c953/src/options.rs#L160C33-L160C33). Unfortunately, there doesn't seem to be much information about the performance hit.

> What if the source is over a network drive?

Actually, it seems like inotify doesn't support NFS anyway: https://github.com/notify-rs/notify/issues/64#issuecomment-216372446

NFS isn't the only network filesystem, of course, but I would cautiously assume it to be the most popular one that can be used with mdbook. 

Regarding large books and network drives in particular: I think it's a good idea to evaluate Kubernetes services for how they do live reloading. Overall, I've seen two approaches:

1. Don't do any kind of filesystem watching at all, and just reload on SIGHUP or HTTP POST request on `/-/reload` endpoint
2. Give up on live reloading completely and just re-create the container (often the case for static content, which is what we have)

Overall, I think supporting live reload on demand is a good idea. I'm not sure how high of a priority it has, considering that we haven't mapped out the user story for live reloading (as in, we don't know who actually needs this feature, what workarounds they use right now, and we don't even have a container image which is a prerequisite to talking with people who care about live reload). On a completely unrelated note: even if we don't support SIGHUP right now, there's a good reason to consider it, because we don't support SIGTERM either, which means containers end up getting SIGKILL'ed anyway because we don't have a graceful shutdown.

0xpr03:
There is no reason to use the debouncer with pollwatcher, unless you have timings that could make this work. Example would be to poll every 1 second and debounce every 2. So you can detect re-created files (depends on the editor of your choice), instead of just rendering nothing because in that second it was still gone.

It may also be useful in case you swap out debouncer backends depending on what your underlying filesystem is.

For bigger books it could be very resource intense. Though I do expect the OS to already have all of the relevant files cached from previous re-renders. On the flip side inotify can use up too many file handles in big workspaces, as it requires one file handle per watched node (folder,file). Which is not the case on windows, so those users won't see that problem.

As you said: It really depends on what your users actually expect from the live-reload. For instant and low-resource, low-idle live reloading, the pollwatcher might be a bad idea. But the moment someone comes around with networked files on linux (WSL -> Windows access), you've lost.

This ultimately just shows the need to get some kind of self-driven selection system into the full-debouncer, which somehow detects the underlying filesystem. Eventually, when I have time for that..

The easiest choice would be to allow both: Some flag so users can switch between native backends and the pollwatcher. Whether that still requires the debouncer I can't say. I definitely would use it for the native backends, too much chatter otherwise.

KFearsoff:
That's a great write-up!

I looked into what the world has to offer a little bit more, apparently `tsx` also has this issue, and it currently doesn't expose the command-line flag: https://github.com/privatenumber/tsx/issues/266

From my experience, there are two kinds of live-reloading: on-demand (for services e.g. in the cloud) and watching the filesystem directly (for dev environments). Generally, those are exposed by different commands: for on-demand, you have `build`, `release` and stuff, and for watching you use `dev`, `serve`. mdbook is a little bit weird in that it doesn't emphasize/provide a correct way to use the software in production, so people end up using `mdbook serve` for Docker containers (which doesn't really work without PollWatcher) and for local development.

There are a lot of things that can be done here, and I guess it ultimately comes down to user experience. I think I'll go with providing a flag for now.

KFearsoff:
I tried implementing a flag today, and failed miserably. I haven't figured out a way do provide `new_debouncer_opt` with a notifier that can be of either Watcher type, I'll have a look at it later too, but so far it seems impossible with current Rust features. I may be very wrong though!

KFearsoff:
Note:  #2030 is ready to merge, and its live-reload functionality depends on this PR

ehuss:
I'm going to close in favor of #2325. Thanks @KFearsoff!

PR #2259, opened 2023-12-06
Color test output and shorten chapter paths

stevecheckoway:
Currently, the output from `rustdoc --test` is not colored because `rustdoc`'s stdout is not a tty. The output of a failed `rustdoc` run is sent to `mdbook`'s stderr via the `error!()` macro. This commit checks if stderr is a tty using the standard `.is_terminal()` and if so, passes `--color always` to `rustdoc`.

The test output from `rustdoc` includes the full path that `rustdoc` was called with. This obfuscates the file with the error. E.g.,
```
---- /var/folders/9v/90bm7kb10fx3_bprxltb3t1r0000gn/T/mdbook-tnGJxp/lab0/index.md - Lab_0__Getting_Started (line 3) stdout ----
error: expected one of `!`, `.`, `::`, `;`, `?`, `{`, `}`, or an operator, found `code`
 --> /var/folders/9v/90bm7kb10fx3_bprxltb3t1r0000gn/T/mdbook-tnGJxp/lab0/index.md:4:6
  |
3 | this code has a bug
  |      ^^^^ expected one of 8 possible tokens
```

This commit runs `rustdoc` in the temp directory and replaces any relative library paths with absolute library paths. This leads to simpler error messages. The one above becomes
```
---- lab0/index.md - Lab_0__Getting_Started (line 3) stdout ----
error: expected one of `!`, `.`, `::`, `;`, `?`, `{`, `}`, or an operator, found `code`
 --> lab0/index.md:4:6
  |
3 | this code has a bug
  |      ^^^^ expected one of 8 possible tokens

error: aborting due to previous error
```
(with colors, of course).

### DIFF
--- a/src/book/mod.rs
+++ b/src/book/mod.rs
@@ -15,8 +15,9 @@ pub use self::init::BookBuilder;
 pub use self::summary::{parse_summary, Link, SectionNumber, Summary, SummaryItem};
 
 use log::{debug, error, info, log_enabled, trace, warn};
-use std::io::Write;
-use std::path::PathBuf;
+use std::ffi::OsString;
+use std::io::{IsTerminal, Write};
+use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::string::ToString;
 use tempfile::Builder as TempFileBuilder;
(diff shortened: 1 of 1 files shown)


stevecheckoway:
Here's a screenshot of the output mentioned in the commit message.
image

stevecheckoway:
I realized after creating this PR that I had accidentally included an unrelated change that causes `mdbook` to mimic how `cargo test` runs `rustdoc` by passing`-C embed-bitcode=no --error-format human`.

I split the commit into two different commits. The first commit makes the changes described above. The second commit just adds these command line options. I'm happy to drop the second commit, but the `-C embed-bitcode=no` part at least is probably useful to (very slightly) speed up tests, regardless of whether you want the colored output/shorter file paths.

ehuss (approved):
Thanks! I went ahead and dropped the second commit. `-C embed-bitcode=no` has no effect, since rustdoc is only generating executables, and those do not have bitcode. And `--error-format=human` is the default, so there shouldn't be a need to set that.

PR #2260, opened 2023-12-06
fix: remove double imports of css

KFearsoff:
Currently, the CSS stylesheets that are already imported from the parent HTML are also imported from the CSS stylesheets. As far as I can tell, it doesn't cause any major bugs, but it's just messy and causes minor issues (for example, Firefox doesn't fire extra requests to fetch CSS twice, but it loads it twice which makes navigating developer menu harder).

From my not-very-extensive testing, it should also allows us to remove this script: https://github.com/rust-lang/mdBook/blob/b7f46213c7df8f499eca0c82e7b41804cc15e369/src/theme/index.hbs#L80-L91

### DIFF
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -1,7 +1,5 @@
 /* CSS for UI elements (a.k.a. chrome) */
 
-@import 'variables.css';
-
 html {
     scrollbar-color: var(--scrollbar) var(--bg);
 }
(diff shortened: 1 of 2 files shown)


jhult (approved):
LGTM

@ehuss or @leonzchang, can we merge this?

jhult:
@ehuss, any thoughts on this?

ehuss:
PRs like these can take a while to review. In the future, it could help to provide more information. For example:
* What exact issue is this trying to solve? What "developer menu" is this referring to? What does a screenshot look like?
* What is an `@import`? How does it differ from `<link>`? Why would one choose one over the other? Can you cite MDN articles or others that would help explain the choices here?
* Why was `@import` added in the first place?

PR #2263, opened 2023-12-09
Remove warning check for theme directory as ./src/theme

jhult:
This was causing an unnecessary warning if `&src_dir` was the same as `ctx.root`.

Also, this code had a `TODO` to "remove some time after the 0.1.8 release".


Dylan-DPC:
Thanks

PR #2277, opened 2023-12-29
preprocess/links: fail for invalid links

### LINKED ISSUE #1094: Include directives to missing files do not return error
Our project used some `include` directives, that due to refactoring, inadvertently pointed to the wrong location, resulting in log error messages like the following:

```
[ERROR] (mdbook::preprocess::links): Error updating "{{#include ../tab_header.md}}", Could not read file for link {{#include ../tab_header.md}} (...)
```

These errors don't result in returning an error code from the process, so we missed them in CI.

Should these (and maybe all error log messages), result in an error code?

toxeus:
Before, `mdbook` would continue processing even when errors, such as invalid links, are encountered. Moreover, it would exit with a `0` return code. Such behavior is unexpected and leads to confusion when run in CI.

Now, when links that don't point to existing files are encounter and error is returned which yields to `mdbook` exiting with an error code. The change in behavior has revealed that some tests were run with invalid links.

Closes #1094

### DIFF
--- a/src/book/book.rs
+++ b/src/book/book.rs
@@ -102,11 +102,11 @@ impl Book {
     /// Unlike the `iter()` method, this requires a closure instead of returning
     /// an iterator. This is because using iterators can possibly allow you
     /// to have iterator invalidation errors.
-    pub fn for_each_mut<F>(&mut self, mut func: F)
+    pub fn for_each_mut<F>(&mut self, mut func: F) -> Result<()>
     where
-        F: FnMut(&mut BookItem),
+        F: FnMut(&mut BookItem) -> Result<()>,
     {
-        for_each_mut(&mut func, &mut self.sections);
+        for_each_mut(&mut func, &mut self.sections)
     }
 
     /// Append a `BookItem` to the `Book`.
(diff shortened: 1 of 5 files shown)


conorsch:
Took this branch for a spin, and worked well: my existing mdbook docroot with broken links threw an error, as I'd expected. After patching the include, no more errors. I support this change being merged.

franciszekjob:
Hi, will this be merged?

toxeus:
Thank you @rustbot. Because the maintainer has ignored this PR even though [he has acknowledged](https://github.com/rust-lang/mdBook/issues/1094#issuecomment-617460570) the issue I have little motivation to put more effort in vain.

GuillaumeGomez:
Hi, sorry it took us so long to come back to this PR. I don't see a test to ensure that when there is an invalid link, it stops building the book. Could you add one please?

toxeus:
Hi @GuillaumeGomez thanks for coming back after all those years. 

I think, given the history of this PR, we can conclude that there's a risk that after I sink more time and effort into this PR it will dangle for another 3 years. 

I fully understand that maintainers are volunteering in their free time and I cannot make demands towards them. But the same applies to contributors and now my motivation is low due to past experience.

With that said, do you accept contributions made with support of AI agents? I could see if this PR could be brought back into shape without too much intervention from my side.

GuillaumeGomez:
I understand. As for AI, the answer is "yes but with caveats". More information in the LLM policy: https://github.com/rust-lang/mdBook/blob/master/CONTRIBUTING.md#llm-policy

Issue #1936, opened 2022-11-24
Rename `curly-quotes` config option to `smart-punctuation`


virtualritz:
See [here](https://daringfireball.net/projects/smartypants/). What `pulldown-cmark` does is replace quotes **but also dashes and ellipses**.

Basically this turns something like:
```
"He took the boat Paris---Rome ... When he arrived he exclamated -- instantly -- 'Heureka!'."
```

Into:

“He took the boat Paris—Rome … When he arrived he exclamated – instantly – ‘Heureka, I found a secret waterway!’.”

virtualritz:
I updated this ticket. The feature is actually there (w/o support for foreign quotes but I'll open a ticket in `pulldown-cmark` for this).

PR #2280, opened 2024-01-02
Rename `config.curly-quotes` to `config.smart-punctuation`

virtualritz:
See #1936.

* Renamed `config.curly-quotes` to `config.smart-punctuation`.
* Ran `rustfmt`.
* Made line length consistently 80 characters in `renderer.md`.

### DIFF
--- a/guide/src/format/configuration/renderers.md
+++ b/guide/src/format/configuration/renderers.md
@@ -97,7 +97,7 @@ description = "The example book covers examples."
 theme = "my-theme"
 default-theme = "light"
 preferred-dark-theme = "navy"
-curly-quotes = true
+smart-punctuation = true
 mathjax-support = false
 copy-fonts = true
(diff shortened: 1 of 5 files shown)


ehuss:
Hi! Thanks for the PR!

Can you drop the commit that re-wraps the file? I have been moving things in the opposite direction to instead use [semantic line breaks](https://sembr.org/) which can make it easier to review changes.

Also, can you keep the original `curly_quotes` field, and emit a warning that it is deprecated and tell the user what the new field name is?

Finally, can you add a test to `rendered_output.rs` to verify the behavior when the value is and isn't set?

@rustbot author

virtualritz:
Will do, but I will have to find some time for this.

ehuss:
I've pushed PR #2327 to move this forward. Thanks!

PR #2285, opened 2024-01-06
feat: Upgrades to MathJax version 3 and supports setting different configuration files (defaults to 'tex-mml-chtml').

### LINKED ISSUE #2284: mdBook doesn't support `mathcal` symbols
### Problem

mdBook cannot render `mathcal` symbols correctly, e.g., `\mathcal{X}`.

### Steps

Type the following line in a `.md` file.

```markdown
For example, \\( \mathcal{X} \\).
```

Then, run `mdbook serve` and observe the output.

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

_No response_

HouJP:
## Introduction

This PR is used to upgrade to MathJax version 3 and supports setting different configuration files (defaults to `'tex-mml-chtml`).

It is inspired by the issue: [mdBook doesn't support `mathcal` symbols](https://github.com/rust-lang/mdBook/issues/2284).

Closes #2284.

## Background

This PR can fix a bug hidden in MathJax v2 introduced [here](https://meta.stackexchange.com/questions/384924/i-think-there-is-a-mathjax-bug-for-mathcal-typesetting-on-some-platforms/385789#385789).

### DIFF
--- a/guide/src/format/configuration/renderers.md
+++ b/guide/src/format/configuration/renderers.md
@@ -99,6 +99,7 @@ default-theme = "light"
 preferred-dark-theme = "navy"
 curly-quotes = true
 mathjax-support = false
+mathjax-config = "tex-mml-chtml"
 copy-fonts = true
 additional-css = ["custom.css", "custom2.css"]
 additional-js = ["custom.js"]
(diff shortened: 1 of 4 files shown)


ehuss:
Thanks for the PR! I believe #1918 is already open to address this. Can you give it a try and see how it works for you?

HouJP:
Thanks @ehuss!

I took a quick look and I think it will solve the problem. The advantage is that it's more flexible (e.g., local js), but its disadvantage is that it requires larger changes to the configuration.

I respect the owners' / reviewers' decision. Please feel free to let me know if you need me to close this PR.

edwardzcn-decade:
Can @ehuss  review this PR again? I meet the same problem in #2284 and I think it is common when using `mdbook` in macOS 13 Ventura and above (for me it is macOS 15) because Apple shipped the fonts which used by mathjax render v2 and left some mismatches since macOS 13.

Now your mathjax test markdown does not display italics properly on most Apple computers, and differs greatly from the schematic #2070. Also using `mathcal`symbols causes formulas to not show up at all (Mathematical Calligraphic Font mismatch).

This PR just upgrade to MathJax v3 with almost the same configuration

Issue #1453, opened 2021-01-25
Make JavaScript code examples runnable?


rpearce:
Hi! Thanks for this great project.

Is there a way to run editable and runnable JS examples like the Rust ones? It seems those go to a remote location, which could be doable with a nodejs backend somewhere, but I'm not sure how to do that in `mdBook`.

Is there a configuration option somewhere for what backend to send requests to? Is it something to do with Ace? Do I need a different editor entirely?

My use case: I'm writing a small book on [ramda.js](https://ramdajs.com) and want to have runnable examples of that but without having to pull in runkit or some other such thing.

I'm sure other people would want backends for Go, Python, Ruby, etc., too!

Thanks for your time. 💜

remexre:
I imagine you'd do some tinkering with book.js in the theme, specifically in the vicinity of:

https://github.com/rust-lang/mdBook/blob/4c951d530d7cf225fcca18d6e9146b28bf8dcf10/src/theme/book.js#L127

rpearce:
@remexre Thanks for the advice. I'll see if I can find a backend, how much I need to alter the code, and I will follow up here with results when I get them.

KritR:
I'm trying to get this to work with C. So far I've modified the book.js but I can't get the play buttons to show up yet. Will update if there's any progress.

KritR:
Ok juts got this working here.
https://github.com/KritR/my-intro-cs/blob/master/theme/book.js
One annoying thing that's happening right now is that #'s are being stripped from the beginning of my includes.
But besides that. Here's what's changed.

- Removed rust crates stuff
  - This seemed redundant but I don't know if this is necessary.
- [Replaced run-rust with run-c](https://github.com/KritR/my-intro-cs/blob/bc52c574d8d404044dac0f1b8f7d58aaaa9afaf7/theme/book.js#L20)
  - This is where the main c running logic is
- [Replaced all instances of language-rust with language-c](https://github.com/KritR/my-intro-cs/blob/bc52c574d8d404044dac0f1b8f7d58aaaa9afaf7/theme/book.js#L59)
  - In markdown the formatt is specified as `c,editable`
- [Added code to wrap language-c blocks with playground class](https://github.com/KritR/my-intro-cs/blob/bc52c574d8d404044dac0f1b8f7d58aaaa9afaf7/theme/book.js#L128)
  - This enables the play button support

That should be most of it.

If you wanted to tweak my version of book.js.
[...]

PR #2286, opened 2024-01-06
Add custom playground guide

dalance:
Custom playground for another language can be achieved through theme customization,
but the customization points are difficult and not documented.

This document gathers the existing workaround at #1453  and my trial.
In my environment (https://doc.veryl-lang.org/book/), the way described in this document achieved editable, syntax highlighted playground for my language.

I think some points seem to be hacky (ex. `editor.js` replacing).
It may be able to be simplified by mdBook's official support.

Related: #350 #1759


K0R7:
This guide is helpful! Thank you!

dalance:
@ehuss There seem to be a few people which needs this information. I think publishing it may be helpful until implementaion of more sophisticated way. How about it?

mio-19:
Yes, I found this information helpful

PR #2303, opened 2024-01-31
summary.md: clarify that part titles must be h1 headers

infogulch:
This formatting mistake was made in the kanidm book ([pr fix](https://github.com/kanidm/kanidm/pull/2463)), it may be common because markdown authors typically use the h1 header only for the title of the document.


ehuss (approved):
Thanks!

PR #2322, opened 2024-02-20
Fix incorrect theme documentation: Next/Previous should use `title`

wilwade:
The theme documentation for next and previous used name instead of title.

Per the code (and using it) `title` is the correct value.

https://github.com/rust-lang/mdBook/blob/5a4ac03c0de866c4cf30d41dd0c7ea02d8429a7a/src/renderer/html_handlebars/helpers/navigation.rs#L209


ehuss (approved):
Thanks!

PR #2332, opened 2024-02-28
Drinks - DRY Links Proof of Concept

Antiarchitect:
Just Proof of Concept of the feature I lacked in many Markdown -> HTML book builders.
Uses `<book root>/drinks.txt` for general links storage.
Replaces `{{#drink somedrink}}` placeholders with value from the dict.
It's just PoC written in an hour or so just to find out does the community have similar needs and should it be implemented as side preprocessor or the default one (current variant).

### DIFF
--- a/src/book/init.rs
+++ b/src/book/init.rs
@@ -83,6 +83,8 @@ impl BookBuilder {
 
         self.write_book_toml()?;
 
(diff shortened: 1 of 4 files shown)


schungx:
I use `mdbook-tera` for it...

Issue #1748, opened 2022-02-14
Feature Request: Multiple mdbook with navbar


laoshaw:
I have multiple mdbooks now that are all independent, is there a way for mdbook to support a navigation (on the top or somewhere) that I can have(or host) multiple books co-exist in one site?

something like the attached screenshot:

![image](https://user-images.githubusercontent.com/18351761/153950014-1a4c41b5-e7ae-4609-86d4-942a1919454e.png)

laoshaw:
vitepress is doing just that: https://github.com/vuejs/docs/blob/main/.vitepress/config.ts

kg4zow:
You can edit `theme/index.hbs` and add whatever you like within the `<body>` tag and before the `<nav>` tag.

Of course this means when you upgrade mdbook, you may have to upgrade your book's custom `theme/index.hbs` file as well. The original file is formatted *very* nicely, so it should be fairly simple to flag your changes within the file, so you can make the same changes when the new version's `index.hbs` file comes out. As an example, see [this file](https://github.com/kg4zow/mdbook-template/blob/8025537a922026589b0fd854a21ad84e9f7126da/theme/index-template.hbs#L108) lines 108-119.

Issue #1666, opened 2021-10-09
[Proposal] Nested books and book references


aiden-leong:
## Preface
This idea comes from `Makefile`. I use mdBook as a blog system, so the main book is called `blog`. At the same time I'm writing several independent books or documents on different topics. If I put them in a single book, the user experience won't be as good as we expect.

### SUMMARY.md
```bash
.
├── book1
│   └── SUMMARY.md
├── book2
│   └── SUMMARY.md
├── book3
│   └── SUMMARY.md
└── SUMMARY.md

```
We may have multiple nested `SUMMARY.md`. In the above case, we will generate 4 books. `localhost:3000`, `localhost:3000/book1`, `localhost:3000/book2` and `localhost:3000/book3`. If there's no cross reference, they are just 4 books. It doesn't matter if the `SUMMARY.md` in the top folder uses files in sub directories.

### Cross Reference
```markdown
<!-- # main book‘'s SUMMARY.md -->
- [Book 1](./book1)
```
* If there are cross references in `SUMMARY.md`, generate an entry for another book. 
* If the cross reference is in context, it's just a normal url, nothing special needed.

## Further
We may have nested `book.toml` as well, but that's another topic. I don't want to cover it in this proposal.

Issue #1347, opened 2020-09-30
mdBook for multi-repo github org


fmuyassarov:
Hi,
Thanks a lot first of all for the amazing work with mdBook.

I wanted to discuss/check out with mdBook contributors if there is a way to solve the problem we are facing currently.

Basically, we have multiple repositories under a single github organization. Each repository has its own documentation. What we want is that - a single mdBook based book that would fetch documentations from all our repositories and show it to the user as a single book.

I will try to use example repositories to be more clear. Let’s say we have `example` github organization, with `x, y, z` repositories under it.

```
example
 ├── x
 ├── y
 ├── z
```

Our wish is - whenever there is a pull request in any of the `x, y, z` repositories that introduces documentation changes, we want to have a preview with netlify on the corresponding repository. This brings the following problems:

1. If we want to have netlify preview on each PR - that means we need to have all the html, css, etc files (generated by mdbook) for each `x, y, z` repository
2. Whenever, we do the preview for a single repo, we don’t have anymore the global view of the book, it will be narrowed down only to a single repo part

One way that we came up was that - we have a separate repo w for storing documentation of all the other repositories. But we are not moving the docs from `x, y, x` to w but instead we clone `x, y,` and `z` as submodules into the `w` repo. The reason for not moving them to the `w` repo is because we would like to keep the documentation as close as possible to the repository that it corresponds to. This will work, we could start the mdbook from `w` repo + including content of `x, y, z` that we submoduled. And there will be automated CI that will be updating `w` repo for example every night with the changes introduced in `x, y, z`. But this also brings some problems when renewing a PR

1. Example: there is a PR on `x` repo and we want to show the changes on netlify to the PR author. So, netlify will try to fetch the content from the w repository(which doesn’t include current commit in `x` repo). But since w repo has no idea about the PR on `x` repo, we can’t really show the preview with the actual changes being introduced with the current commit.

![Untitled Diagram](https://user-images.githubusercontent.com/35802557/94665276-993be280-0314-11eb-9bdc-51e52bae1ee4.png)

In short it is problematic to have a single book that will be sourcing contents from multiple repositories.

I would really love to hear your inputs and open for a discussion. 
Thanks in advance.

frewsxcv:
[...]

Issue #1042, opened 2019-09-29
Feature request: serve multiple mdBooks in one site


rutrum:
I have a lot of mdbooks now.  In particular, I use them for notes in my classes.  I would love a way to have a single webpage serve multiple mdbooks.  In particular, I wish there was a way to structure an mdbook project to serve multiple books (each with a distinct SUMMARY.md), so I could view all my class notes in one repository.

GogoAkira:
Hello
Did you ever figure this one out, I just discovered mdBook today :) and so with it I found rust.
I don't know about the 'serve' command since it serves on the same port, but if you serve your books with nginx a path to index.html for all the books should do the trick, right?

GogoAkira:
Hello
I'm also thinking about this, I have nginx installed locally on computer. 
From the documentation/guide I served both books like the example below
`mdbook serve path/to/book -p 8000 -n 127.0.0.1 --websocket-hostname 192.168.1.100`
I used two different IP's but same port numbers, and also same IP different port numbers 8001 and 8002 but the port for incoming 3001 was already taken, it still worked though :)

Before this I did try serving the book via symbolic link to nginx root dir, and also at the same time I did 'mdbook serve' to serve a single book and it got served twice :) , idk, so I realised the built in engine is the way to go anyway, but since you have only one IP on a server, you could make different hostnames? And at this point i'm not sure if the 3001 websocket can be changed, it's getting late :).

If you did figure out anything in the mean time let me know too.

jgerrish:
I'd be interested in this feature.  My workflow involves switching between multiple books, and it's be great to have each subdirectory in a root directory have a different port number.

Possibly with a simple configuration file mapping subdirectories to ports or URLs. 

I could write a patch, but I don't want to start on a feature that is a dead end.  I don't have a good estimate of time and demand.

Thanks!  Great project.

GogoFC:
I now host multiple mdbooks on one server. I have nginx running and each book is on a different subdomain and therefore also different directory. Since https runs on port 443 everything is on that port.

rutrum:
I think my intention with this ticket was that the built side might have say, a top navigate bar, or a special drop down at the top, that would allow me to switch between books. Hosting multiple books behind nginx gets the job done, but it would be cool if the site itself also generated with links to navigate between books.

r12f:
This feature will be awesome! I am mostly being pushed to mkdocs because of lacking of this feature...

[...]

PR #2347, opened 2024-03-29
Bookshelf feature

Coi-l:
# Background
A while ago the need to bundle together multiple mdBooks arose for me. I did some digging but I could only find a number of feature requests in the issues list. So I decided to try and prototype something myself. Adding a 'shelf' command to the CLI.

# Strategy
The feature is based around a shelf.toml file which contains a number of books to include in your bookshelf. They can either be locally present on the filesystem or a git repository. If you use a git repository you can optionally provide a path inside git repository where the file is present and a branch.

These books are then build by the shelf command.

In addition to this one more book is built, the actual "shelf". This is also an mdBook. It will will contain an index and a separate page for each book containing a bit more information.

Every book built by the shelf command will have an additional icon in the top right corner of a book. Clicking this will bring you back to the bookshelf.
image
It is the fa-book icon. I would have liked to use something more like this: https://fontawesome.com/icons/books?f=classic&s=solid but all those seem to be "Pro".

## You can see a working example here:
https://coi-l.github.io/mdBook/bookshelf/index/book/index.html
or here:
https://coi-l.github.io/mdBook/bookshelf-multi/index/book/index.html

These to sites display the two options available. The first simple one with just one "shelf" in the bookshelf, and the seconds with multiple "shelves". This second option can be used to categorize the mdBooks.

# Trying it out
You can try this out for yourself by checking out the PR code and building the mdBook binary.
Then navigate to the 'test_shelf' directory where there is a shelf.toml file prepared. Run the mdBook binary with 'shelf' as the command. This will build the same bookshelf as provided in the link above.

# Summing it up
What I am looking for is comments regarding the PR.
First and foremost if this is something that is desired. There seems to be a number of feature requests for this, but maybe this strategy is not the way to go about it?

If you think this sounds like a good way forward I'd appreciate feedback on the code and on the actual design of the shelf book and UI in general. I don't mind polishing this to make it look good. Perhaps tables would be better or something.

Maybe the book icon in the books should produce a dropdown instead so you can easily select another book, or move back to the bookshelf.

The actual directory structure that the shelf command produces can also be adjusted.

# Issues I think are requesting something like this, but I might have misunderstood:
#1748
#1347 
#1666 
#1042


shenef:
Looking at your example, it seems there are some sub-pages where the "Return to shelf" URL is not generated correcly.
Instead of using
https://coi-l.github.io/mdBook/shelf/book/index.html some of them use
https://coi-l.github.io/mdBook/books/shelf/book/index.html which results in a 404.
In the rust documentation book it works correctly on every page i tried but in the mdbook docs and the test book there are some pages that have the wrong URL and some that have the correct one.

Coi-l:
> Looking at your example, it seems there are some sub-pages where the "Return to shelf" URL is not generated correcly. Instead of using https://coi-l.github.io/mdBook/shelf/book/index.html some of them use https://coi-l.github.io/mdBook/books/shelf/book/index.html which results in a 404. In the rust documentation book it works correctly on every page i tried but in the mdbook docs and the test book there are some pages that have the wrong URL and some that have the correct one.

Yes, you are correct. Using relative paths was not a good idea.
I've updated the code and the link to the example with a fix for this.

Thanks for the feedback!

ksd3:
Out of curiosity, was this ever merged? This is the exact thing I'm looking for.

Coi-l:
> Out of curiosity, was this ever merged? This is the exact thing I'm looking for.

No. I haven't really gotten any kind of response from the maintainers unfortunately.

Have you tried it? Did it work well for you?

ksd3:
> > Out of curiosity, was this ever merged? This is the exact thing I'm looking for.
> 
> No. I haven't really gotten any kind of response from the maintainers unfortunately.
> 
> Have you tried it? Did it work well for you?

I'm new to Rust, so I don't quite know how to even get it to work. I'm fine with downgrading my version of mdbook - could you provide instructions on how to install it?

Coi-l:
> > > Out of curiosity, was this ever merged? This is the exact thing I'm looking for.
> > 
> > 
> > No. I haven't really gotten any kind of response from the maintainers unfortunately.
> > Have you tried it? Did it work well for you?
> 
> I'm new to Rust, so I don't quite know how to even get it to work. I'm fine with downgrading my version of mdbook - could you provide instructions on how to install it?

I wouldn't recommend installing this instead of the official one. But if you want to try it out you can clone the repository and do a cargo build. That will give you the binary with the 'shelf' command included. And in the 'test_shelf' directory there is a toml file you can adapt.

ksd3:
> > > > Out of curiosity, was this ever merged? This is the exact thing I'm looking for.
> > > 
> > > 
> > > No. I haven't really gotten any kind of response from the maintainers unfortunately.
> > > Have you tried it? Did it work well for you?
> > 
> > 
> > I'm new to Rust, so I don't quite know how to even get it to work. I'm fine with downgrading my version of mdbook - could you provide instructions on how to install it?
> 
> I wouldn't recommend installing this instead of the official one. But if you want to try it out you can clone the repository and do a cargo build. That will give you the binary with the 'shelf' command included. And in the 'test_shelf' directory there is a toml file you can adapt.

Wasn't this implemented only half a year ago?...I don't think too many features have been added to mdbook after it - but yeah, worst case scenario, I just run a script that produces the bookshelf rather than having it in an executable

pdigiglio:
Any progress on this?

ksd3:
> Any progress on this?

@pdigiglio I switched to using Quarto instead. It has this feature built in, plus I needed a more permanent solution

PR #2353, opened 2024-04-06
Constrain `handlebars` version to satisfy MSRV

max-heller:
`handlebars` [raised its MSRV to 1.72](https://github.com/sunng87/handlebars-rust/commit/6426df0389e56acfc6b488e875e58614004a3dce) in its 5.1.2 release, which is incompatible with `mdbook`'s 1.71 MSRV. This restricts version selection to `5.0, <5.1.2` to avoid the increase


ehuss:
During the next update, I will probably raise the msrv instead of constraining the version. Thanks for the PR, though!

PR #2355, opened 2024-04-07
Reduce allocations in `fs::copy_files_except_ext`

johamster:
Above mentioned function copies files (recursively) from a source to a destination directory. For that, file/directory paths have to be created repeatedly. This allocates as directory and file names are concatenated into an owning path structure.

The number of allocations can be reduced by creating file/directory paths only once and borrowing them instead of cloning/recreating them.

In bigger projects, this reduces execution time noticeably. Please note that file system operations are dominant from performance POV.

### DIFF
--- a/src/utils/fs.rs
+++ b/src/utils/fs.rs
@@ -72,14 +72,12 @@ pub fn create_file(path: &Path) -> Result<File> {
 
 /// Removes all the content of a directory but not the directory itself
 pub fn remove_dir_content(dir: &Path) -> Result<()> {
-    for item in fs::read_dir(dir)? {
-        if let Ok(item) = item {
-            let item = item.path();
-            if item.is_dir() {
-                fs::remove_dir_all(item)?;
-            } else {
-                fs::remove_file(item)?;
-            }
+    for item in fs::read_dir(dir)?.flatten() {
+        let item = item.path();
+        if item.is_dir() {
+            fs::remove_dir_all(item)?;
+        } else {
+            fs::remove_file(item)?;
         }
     }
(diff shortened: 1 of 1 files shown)


src/utils/fs.rs
```diff
@@ -72,14 +72,12 @@ pub fn create_file(path: &Path) -> Result<File> {
...
 pub fn remove_dir_content(dir: &Path) -> Result<()> {
-    for item in fs::read_dir(dir)? {
-        if let Ok(item) = item {
-            let item = item.path();
-            if item.is_dir() {
-                fs::remove_dir_all(item)?;
-            } else {
-                fs::remove_file(item)?;
-            }
+    for item in fs::read_dir(dir)?.flatten() {
```
johamster, line 75: Btw, this is off-topic and just resolves a `clippy` warning.

src/utils/fs.rs
```diff
@@ -108,72 +106,41 @@ pub fn copy_files_except_ext(
...
+            copy_files_except_ext(&entry, &target_file_path, true, avoid_dir, ext_blacklist)?;
         } else if metadata.is_file() {
             // Check if it is in the blacklist
-            if let Some(ext) = entry.path().extension() {
+            if let Some(ext) = entry.extension() {
                 if ext_blacklist.contains(&ext.to_str().unwrap()) {
                     continue;
                 }
             }
-            debug!(
```
johamster, line 148: I removed this trace as it did not really add value, in my opinion. The target file path is traced also below.

ehuss (approved):
Thanks!

PR #2359, opened 2024-04-07
refactor: Use axum for "mdbook serve"

audunhalland:
This in my opinion brings mdbook's dependencies more in line with "the state of the art" of the Rust HTTP ecosystem.
It pulls in `hyper v1` and uses `tower-http` for filesystem serving functions.

I've manually tested serve with the `test_book`:
* serving pages from the book works
* websocket-based auto-reload works
* 404 page works

I can't guarantee though that there are no regressions. I did not find integrations tests for the serve command.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -139,6 +139,57 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
 
 [[package]]
+name = "axum"
+version = "0.8.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6d6fd624c75e18b3b4c6b9caf42b1afe24437daaee904069137d8bab077be8b8"
+dependencies = [
+ "axum-core",
+ "base64",
+ "bytes",
+ "futures-util",
+ "http",
+ "http-body",
+ "http-body-util",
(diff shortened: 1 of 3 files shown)


paolobarbolini:
FWIW this seems the right call, as `warp` is still on hyper v0.14 and looking at their repository it gets almost no maintenance.

audunhalland:
rebased, squashed commits; upgraded to axum 0.8.1.

szabgab (commented):
I think this change should be removed from the PR.
Especially as mdbook is now on 1.77 already.

szabgab:
Thanks for the update.

I am not familiar with axum so I can't comment on the code, but I tried the PR locally and it works on Linux.

My only thought was that by replacing a small framework with a larger one we might make the executable much larger.
So I checked the size of the `target/release/mdbook` executable on Ubuntu Linux after generating it with the `cargo build --locked --bin mdbook --release` command:

* on master 14,955,672
* with axum 13,146,536

So somehow this change also reduces the size of the executable by some 13%. Any idea how and why?

audunhalland:
Speculation:

* The `http` crate has been unified on version 1 (I see no other unification in the Cargo.lock diff)
* This change removes support for HTTP/2 as axum default features has been turned off

ehuss:
Thanks for your help with this! We have combined the changes here with https://github.com/rust-lang/mdBook/pull/2748, so closing this as merged.

PR #2381, opened 2024-05-16
Update pulldown-cmark to 0.11

ehuss:
Changelog: https://github.com/pulldown-cmark/pulldown-cmark/releases/tag/v0.11.0

Note that math is not yet supported. However, due to the new Event variant, I decided to include math script in the search index. I don't know if that is the right thing to do, but seems like something we can figure out later if we start using the math feature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1411,9 +1411,9 @@ dependencies = [
 
 [[package]]
 name = "pulldown-cmark"
-version = "0.10.3"
+version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "76979bea66e7875e7509c4ec5300112b316af87fa7a252ca91c448b32dfe3993"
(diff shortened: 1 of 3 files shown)


dylanowen:
From the view of a preprocessor this isn't a semver compatible update and probably needs a minor version bump:

https://github.com/dylanowen/mdbook-graphviz/actions/runs/9120113600/job/25076838288?pr=132

ehuss:
😦 Indeed. Thanks for letting me know. I did not realize the pulldown_cmark types were exposed, and we updated in 0.4.37 to 0.10 and nobody mentioned anything at the time.

I'll revert it for now. We'll probably need to do some kind of public re-export.

dylanowen:
Thanks for the quick fix! I think we ran into issues with 0.4.37 as well but I neglected to warn anyone here 😬:

https://github.com/dylanowen/mdbook-graphviz/issues/107#issuecomment-1939140745

Issue #2387, opened 2024-05-17
Figure out a strategy for updating public dependencies


ehuss:
We are currently not able to update public dependencies exposed in the API due to semver breakage. There probably should be some strategy for how to handle this. For example, public re-exporting the crates so that users can keep the version in sync. Another alternative is to hide these via wrappers or shrinking the number of `pub` items.

Currently this affects:

- anyhow
- toml
- handlebars
- pulldown-cmark

max-sixty:
Would upgrading them and publishing an mdbook 0.5.0 make sense?

We're currently locked on an older version of `pulldown-cmark` in PRQL since we're pinned to the mdbook version...

Thank you!

danieleades:
these are all leaky abstractions. It makes sense to release a new major version of md book that hides these abstractions behind interfaces so that future breaking changes can be avoided.

see https://github.com/rust-lang/mdBook/issues/2037

PR #2401, opened 2024-06-16
Upgrade pulldown cmark

Roms1383:
Hi and thanks for this beautiful piece of tooling, I simply love it since the first time I discovered it!

I updated `pulldown-cmark` but noticed the comment saying not to in `Cargo.toml`, so feel free to dismiss this PR if this is not supposed to happen from an external contribution.

If however you do accept this PR, I'd like to add a mathematical expression to the tests for the sake of completeness but I'd appreciate that you point me which markdown is the most appropriate, so putting this under draft for now.

Thanks!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1411,9 +1411,9 @@ dependencies = [
 
 [[package]]
 name = "pulldown-cmark"
-version = "0.10.3"
+version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "76979bea66e7875e7509c4ec5300112b316af87fa7a252ca91c448b32dfe3993"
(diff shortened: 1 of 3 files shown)


Roms1383:
Sorry for the ping @ehuss, but from a glance at latest commits it seem you're in charge of PRs.
Any update on this ? `pulldown-cmark` has released a new version `0.12.0` in the meantime.

> Also, saw mention in the guide of bumping version, am I allowed to do so ? Semver check recommends `0.5.0`, but I imagine you would gather the changes of many a contributor first ?

ehuss:
As the comment says, we can't update it at this time. See #2387. We will likely update this when we do a 0.5 release.

ehuss (approved):
Thanks, I have update it to 0.13.0.

Issue #2037, opened 2023-03-01
don't expose toml::Table in public API


danieleades:
### Problem

the `toml::Table` type is exposed in the public API (in the `Config::get_preprocessor` method), which means that changing the version of the toml dependency is a breaking change

### Proposed Solution

hide the toml::Table value by returning a newtype, or by returning a `impl Deserialize` anonymous type

### Notes

this change is itself a breaking change. Preprocessor developers who are accessing the Table values directly without deserialising into a custom struct will no longer be able to do so with the proposed change

PR #2407, opened 2024-06-26
disincentivize usage of functions that expose toml::Table in Config

JulianGmp:
hi there,

firstly, thank you for your continued work on mdbook, it's a great tool :)

This is related to #2037, which I encountered while [trying to deserialize the exposed `Map`/`Table` into a user defined struct](https://github.com/toml-rs/toml/discussions/744) in a preprocessor I'm writing.

Using `get_deserialized_opt` seems much more preferable to `get_preprocessor` (and `get_renderer` for that matter).
However, I merely happend to stumble across that function while looking through `config.rs`.  
I don't think I'm alone with this, I did a brief search over the preprocessors listed in [the third party plugins page](https://github.com/rust-lang/mdBook/wiki/Third-party-plugins) and the ones that do have options use `get_preprocessor`.

The one caveat for `get_deserialized_opt` is that you have to pass `"preprocessor.my_thing"` instead of `"my_thing"`, so I made `get_renderer_deserialized` and `get_preprocessor_deserialized` for convenience.
I added that to the example to incentivize people writing new preprocessors to use it.

I also marked `get_renderer` and `get_preprocessor` as deprecated. However this might be bold. There are probably valid use cases for these functions, unless mdbook were to remove the exposed toml type(s) like #2037 describes (which would significantly change `mdbook::config::Config`'s interface).

Note that this also results in 3 deprecation warnings in the tests.  
I haven't addresses these, because I think the decision whether to deprecate them at all should be done by the maintainers, I'm just a user of the library/tool.

### DIFF
--- a/examples/nop-preprocessor.rs
+++ b/examples/nop-preprocessor.rs
@@ -71,6 +71,7 @@ fn handle_supports(pre: &dyn Preprocessor, sub_args: &ArgMatches) -> ! {
 /// in your main `lib.rs` file.
 mod nop_lib {
     use super::*;
+    use serde::Deserialize;
 
     /// A no-op preprocessor.
     pub struct Nop;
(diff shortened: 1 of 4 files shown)


danieleades:
i think deprecating the methods that expose the toml type and pointing users towards `get_deserialized_opt` is the right way to go

JulianGmp:
I went ahead and rebased this branch from the current master, since the test structure was changed.
I still think this would be a good inclusion.
Whether or not the old functions should be deprecated I still want to leave to the maintainers.
Especially the [changes in `src/book/mod.rs`](https://github.com/rust-lang/mdBook/pull/2407/files#diff-d6ba2325b9fd12cfc43c45f9bf6664aed53830db7000a79030af8e3e752b48a6) aren't great imo, since they add an unnecessary copy. While I don't think this is a use case for people writing their own renderers or preprocessors, it still is a minor performance regression.
I could imagine adding some specific functions to get a specific config value like "preprocessor.some_value" by reference.

In any case, I want to kindly ask for a review, since this has gone a bit stale.

ehuss:
I'm going to close as `Table` is no longer publicly exposed (largely through https://github.com/rust-lang/mdBook/pull/2773). Thanks for the PR, though!

JulianGmp:
> I'm going to close as `Table` is no longer publicly exposed (largely through #2773). Thanks for the PR, though!

Ah, it achieves the same goal, thank you for working on this! :)
I'll update my preprocessors and renderer once the next release comes around.

PR #2410, opened 2024-07-02
gate few deps only used in bin by cli feature

klensy:
This adds `cli` feature (added to `default` list to prevent possible break) which gates deps used only by `mdbook` bin. This technically breaks users of bin with `default-features = false`, are they exist?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,15 +18,15 @@ rust-version = "1.74"
 
 [dependencies]
 anyhow = "1.0.71"
-chrono = { version = "0.4.24", default-features = false, features = ["clock"] }
+chrono = { version = "0.4.24", default-features = false, features = ["clock"], optional = true }
 clap = { version = "4.3.12", features = ["cargo", "wrap_help"] }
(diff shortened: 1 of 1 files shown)


ehuss:
Unfortunately we won't be able to make a change like this because it would break `cargo install --no-default-features`.

The main change in the 0.5 release is intended to fix this. I really want to get that rolling soon, it's just been hard to make time.

klensy:
Looking at 0.5 milestone (https://github.com/rust-lang/mdBook/milestone/11) and don't see similar change, is there list with planned breaking changes somewhere else?

klensy:
Something like #1937, which is not labeled :-(

ehuss:
I don't think there is a list besides the milestone. The primary thing is to split it into multiple packages, and do some API cleanup to fix backwards-compatibility issues to make it easier to update things.

ehuss:
Closing as this should be resolved by https://github.com/rust-lang/mdBook/pull/2766.

Issue #1238, opened 2020-05-29
Canonical Links Best Practice


markhildreth:
Is there a recommended practice for ensuring canonical URLs for a book? For example, the current rust book can be found at these URLs:

* `https://doc.rust-lang.org/book/`
* `https://doc.rust-lang.org/book`
  * which 302s to `/doc/stable/book/` and then `/stable/book/`.
* `https://doc.rust-lang.org/book/index.html`
* `https://doc.rust-lang.org/book/title-page.html` (although this one is __slightly__ different, in that the title page section in the ToC is styled as "active").

However, I don't notice a `<link rel="canonical" href="https://doc.rust-lang.org/book/">` on any of the resulting documents. Should mdbook offer any recommendations in this regard, and/or a configuration option to allow for a canonical link on pages?

ehuss:
I think it would probably use useful to have a canonical link.  Would probably need to add a `site-url` to the config to specify the canonical URL.  Would then need to feed that into the template (perhaps a computed value of site-url joined with the current page's canonical name).

(Note, also briefly discussed adding site-url at https://github.com/rust-lang/mdBook/pull/1221#issuecomment-635522550)

markhildreth:
@ehuss Thanks for the quick response. Based on your response and some more digging, I see a few ways forward:

1.) Expose the rendered path (e.g., `/index.html` for the root`/README.md`) to the template. This is technically all we need: then the user could just create the full URL in their template with the rendered path. Something like....

```
<link rel="canonical" href="https://example.com/{{ rendered_path }}" />
```

A bit hackish (especially with the "who is supposed to add the slash" confusion), but it works.

2.) In addition to the above, expose a `canonical-url` option (e.g., "https://example.com"). If this option is specified, then all pages will have the canonical link placed in their header by default, created by adding the rendered path onto the given canonical URL.

You mention a `site_url`, although I'm not sure that's what we want. For example, you may want to have a `latest.mydocs.com` and a `mydocs.com`. These might have different site-url values, but the same canonical-url value.

pinage404:
Hello,

I changed the domain name of a website generated by mdBook

Now, Google doesn't want to index the new URL because it is detected as duplicate content
 
> [Duplicate without user-selected canonical
These pages aren't indexed or served on Google](https://support.google.com/webmasters/answer/7440203#duplicate_page_without_canonical_tag)

Is there currently a solution ? (Even if it's a bit hacky)

PR #2415, opened 2024-07-21
add `rendered_path` in the template

pinage404:
pass the rendered path in the template, as suggested here https://github.com/rust-lang/mdBook/issues/1238#issuecomment-636270351

which allows to add the canonical URL https://gitlab.com/pinage404/git-gamble/-/commit/c7c2e601f9d42f1d2f88e81654787c4d0ba82545#4e2dfdde34a65a1bb4b51a5effb6635b41af0fbc_59_60

the intention of this PR is just to help the subject to advance, i don't care if it is not merged because prefered to another solution

the intention of this PR is just to help the topic move forward, i don't care if it is not merged because it is preferred to another solution

### DIFF
--- a/guide/src/format/theme/index-hbs.md
+++ b/guide/src/format/theme/index-hbs.md
@@ -23,6 +23,7 @@ Here is a list of the properties that are exposed:
 - ***book_title*** Title of the book, as specified in `book.toml`
 - ***chapter_title*** Title of the current chapter, as listed in `SUMMARY.md`
 
+- ***rendered_path*** Relative path to the rendered path from the source directory
 - ***path*** Relative path to the original markdown file from the source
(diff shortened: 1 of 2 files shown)


GuillaumeGomez (changes requested):
Please add a regression test, otherwise looks good to me.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/2415/ec996d350955a8974f8cae5a55c1374ae8da4fac..7baeccab43fbda4cd613a679a0fd4b54371d432b)*

pinage404:
This PR is better conceived than what i did https://github.com/rust-lang/mdBook/pull/2706

PR #2416, opened 2024-07-23
chore: update handlebars to v6

campeis:


ehuss (approved):
Thanks! When posting these kinds of PRs, it can be helpful to link to the release notes and provide any commentary about what might change or be broken by the release.

https://github.com/sunng87/handlebars-rust/blob/master/CHANGELOG.md#600---2024-07-20

I don't see anything particularly concerning there.

PR #2418, opened 2024-07-27
Added slash as additional trigger for search

### LINKED ISSUE #2379: Support '/' shortcut to search
### Problem

In rust-lang/rust#123355, we added `/` as an additional shortcut to open the searchbar. It's was already used by github, zulip, less, and google drive to focus/open the searchbar. Given that it's so common, I now find myself typing `/` by mussle memory, and expecting it to work in mdbook, before remembering that I need to use `s` here.

### Proposed Solution

`/` should behave like `s`, and open the searchbar.

### Notes

I'd be happy to contribute an implementation of this, assuming you're willing to accept the feature.

rik86189:
Closes #2379 

Given the language, I interpreted the request to mean that '/' should be an additional trigger. 
If this is not the desired behaviour let me know.


ehuss:
Closing as resolved by https://github.com/rust-lang/mdBook/pull/2698

PR #2421, opened 2024-08-01
Unify copy to clipboard icon with docs.rs, rustdoc and crates.io

GuillaumeGomez:
It looks like this:

image

All colors are the same. I computed the `filter` values with https://isotropic.co/tool/hex-color-to-css-filter/ (very nice website).

Another change you might have noticed: I increased the size of the buttons. There are two reasons for this change:
1. The new clipboard icon rendered badly with such small dimensions.
2. Since these buttons can be used on mobile devices, making them bigger will make things better for everyone.

### DIFF
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -267,7 +267,8 @@ pre > .buttons button.clip-button {
     padding: 2px 4px 0px 6px;
 }
 pre > .buttons button.clip-button::before {
-    /* clipboard <https://github.com/rust-lang/crates.io/commits/main/public/assets/copy.svg> */
+    /* clipboard image from octicons (https://github.com/primer/octicons/tree/v2.0.0) MIT license
+     */
     content: url('data:image/svg+xml,<svg width="21" height="20" viewBox="0 0 24 25" \
 xmlns="http://www.w3.org/2000/svg" aria-label="Copy to clipboard">\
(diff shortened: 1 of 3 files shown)


camelid:
The new button looks out of proportion with the others - maybe make the others a little smaller to match it better?

GuillaumeGomez:
Updated! You can see the new display in the first comment.

src/theme/css/chrome.css
```diff
@@ -262,6 +262,26 @@ pre > .buttons button {
     transition-property: color,border-color,background-color;
     color: var(--icons);
 }
+
+pre > .buttons button.clip-button {
+    padding: 2px 4px 0px 6px;
+}
+pre > .buttons button.clip-button::before {
+    /* clipboard <https://github.com/rust-lang/crates.io/commits/main/public/assets/copy.svg> */
```
ehuss, line 270: It looks like this was taken from [octicons](https://github.com/primer/octicons) (particularly, somewhere around the [2.0](https://github.com/primer/octicons/blob/v2.0.0/svg/clippy.svg) release). Can you update this to include the provenance and license from that version?

ehuss (commented):
Thanks!

src/theme/css/chrome.css
GuillaumeGomez: Done!

Issue #2403, opened 2024-06-21
If `!search_config.copy_js`, then perform useless search indexing, and search is disabled


SichangHe:
### Problem

https://github.com/rust-lang/mdBook/blob/3bdcc0a5a6f3c85dd751350774261dbc357b02bd/src/renderer/html_handlebars/search.rs#L42-L54

Here, if `!search_config.copy_js`, then `index` is not used, so the above expensive computation is wasted.

More importantly, since the search index is not written, the search feature does not work.
That is, search is only enabled if `search_config.enable && search_config.copy_js`, instead of the intended behavior of just depending on `search_config.enable`.

### Steps

N/A

### Possible Solution(s)

The name `copy_js` seems to suggest the use of a CDN if set to false, however, looking at https://github.com/rust-lang/mdBook/blob/3bdcc0a5a6f3c85dd751350774261dbc357b02bd/src/theme/index.hbs#L311, the links to a CDN seems to be missing. If so, then the solution would be:

1. Add the CDN links of the search JS to the `index.hbs` file and enable them when `search_config.enable && !search_config.copy_js`.
1. Correct `create_files` to write the index JSON file when `!search_config.copy_js`.

However, if `!search_config.copy_js` disabling the search entirely is the intended behavior (which would be weird), then maybe we should just get rid of it and fully rely on `search_config.enable`.

### Notes

Additionally, the search index is unnecessarily duplicated in both `searchindex.json` and `searchindex.js`, which are [sequentially tentatively queried by `searcher.js`](https://github.com/rust-lang/mdBook/blob/3bdcc0a5a6f3c85dd751350774261dbc357b02bd/src/theme/searcher/searcher.js#L471-L476).

### Version

```text
Please see the permalink above pointing to the latest master as of this writing.
```

PR #2425, opened 2024-08-09
de-duplicate the search index

nick1udwig:
Currently, the search index is duplicated in `searchindex.js` and `searchindex.json`. It seems the reason for the duplication is that the `.json` allows for async loading via `fetch()` and the `.js` provides a fallback in case CORS causes issues with the `fetch()`.

The problem with having the duplicated index is that for large books, the index becomes quite large (2.5MB for my company's book!). It would be nice if the index didn't have to be duplicated.

This PR removes `searchindex.json` from the site output. It replicates the async functionality by reading in the `searchindex.js` file and parsing out the JSON within. The fallback remains unchanged.

### DIFF
--- a/src/theme/searcher/searcher.js
+++ b/src/theme/searcher/searcher.js
@@ -468,9 +468,17 @@ window.search = window.search || {};
         showResults(true);
     }
 
-    fetch(path_to_root + 'searchindex.json')
-        .then(response => response.json())
-        .then(json => init(json))        
+    fetch(path_to_root + 'searchindex.js')
+        .then(response => response.text())
+        .then(text => {
+            const jsonMatch = text.match(/Object\.assign\(window\.search,\s*(\{[\s\S]*\})\s*\)/);
+            if (jsonMatch && jsonMatch[1]) {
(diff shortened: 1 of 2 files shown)


SichangHe:
Related: #2403.

src/theme/searcher/searcher.js
```diff
@@ -468,9 +468,17 @@ window.search = window.search || {};
         showResults(true);
     }
 
-    fetch(path_to_root + 'searchindex.json')
-        .then(response => response.json())
-        .then(json => init(json))        
+    fetch(path_to_root + 'searchindex.js')
```
GuillaumeGomez, line 471: Why keeping this part of the code instead of just using the content from the `.catch()` call? There is no point in parsing JS.

GuillaumeGomez:
I think we should indeed get rid of this duplication, however I don't think the code is correct currently (parsing JS, not a good idea).

GuillaumeGomez:
Closing in favour of https://github.com/rust-lang/mdBook/pull/2552.

PR #2134, opened 2023-07-18
Enhancing themes with `color-scheme: light | dark;` implementation in CSS

GiorgioReale:
This PR resolves issue #2132 by implementing `color-scheme: light | dark;` in CSS to improve the graphical interface of supported browsers based on the selected theme.

PR #2441, opened 2024-09-21
Fix automatic dark mode

### LINKED ISSUE #2440: Implement auto dark mode in CSS
### Problem

Currently, the page will load with a default light style, then JS will take over and if it detects that it is in "auto theme" mode and that the browser is in dark mode, it replaces the light colors with dark colors. This causes a few problems:

- If the internet connection is slow, the page will display the light colors first, destroying the dark mode user's eyes.
- If the user has JS disabled or if it breaks in any way, they will never be able to have dark mode support.

### Proposed Solution

You can use CSS media queries to perform the dark mode check inside the CSS.

### Notes

Related: #1504

atjn:
Automatic dark mode (in CSS) was implemented in #1069, then broken in  #1641 because it moved the `no-js` class to a different element. This PR restores the functionality. It also implements #2134 properly.

I also noticed that the original implementation depends on the `light` class to be set, which it is not if a different default theme has been chosen. I have also fixed this.

Closes #2440

### DIFF
--- a/src/theme/css/variables.css
+++ b/src/theme/css/variables.css
@@ -258,7 +258,7 @@
 }
 
 @media (prefers-color-scheme: dark) {
-    .light.no-js {
+    html:has(.no-js) {
         --bg: hsl(200, 7%, 8%);
         --fg: #98a3ad;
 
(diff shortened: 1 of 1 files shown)


ehuss:
I'm going to close as I believe this will be resolved by #2454. Thanks, though!

Issue #2437, opened 2024-09-11
Footnote line height adds too much space.


ehuss:
### Problem

When there is a footnote on a multi-line paragraph, the line with the footnote ends up adding some extra vertical space to the line:

<img width="653" alt="image" src="https://github.com/user-attachments/assets/67cc3f2e-ef02-49c4-9b1b-9acad16f040d">

I don't think it looks very good, and would like to see some kind of solution to avoid that.

### Steps

Have a multi-line paragraph with a footnote after the first line.

### Version

```text
mdbook v0.4.40
```

loliceo:
Hi! I was wondering if it was okay if I made a pull request for this issue?

I did read the contributions guideline and it said that I was free to claim the issue but this is my first time contributing so I was a bit nervous.

#### Possible Issue
I think this issue was because the all the text in the paragraph **already has a line-height of 1.5** and the **footnote-reference gets another line-height of 1.5** leading to this issue.


#### Solution
The only change I made was adding this piece of css

    .footnote-reference{
	    line-height: 0;
    }

![Before](https://github.com/user-attachments/assets/601e8e30-2e6c-4595-8116-dd1c509ff6e9)

![After](https://github.com/user-attachments/assets/16f9b5c7-557a-47a6-a722-fdc065bce662)

ehuss:
If it works on all the major browsers (chrome, safari, firefox, mobile safari, mobile chrome), and doesn't overlap with descenders of the previous line, then sounds great to me to get this fixed!

PR #2443, opened 2024-10-04
Fix line-height for footnote-reference

loliceo:
### ISSUE
#2437 Footnote reference adding some extra vertical space to the line

### FIX
Changed general.css file to include custom styling for footnote reference, removing inheritance of line height from parent class.
Tested for various different browsers. (Chrome, Firefox, mobile chrome, mobile firefox)

## Before
before

## After
after

### DIFF
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -202,6 +202,10 @@ kbd {
     display: inline;
 }
 
+.footnote-reference{
+    line-height: 0;
+}
+
 .tooltiptext {
     position: absolute;
     visibility: hidden;


src/theme/css/general.css
```diff
@@ -202,6 +202,10 @@ kbd {
     display: inline;
 }
 
+.footnote-reference{
+    line-height: 0;
```
notriddle, line 206: As far as I understand the standard, this does what you want because [a line box's final height is the maximum of the required line-height of all the inline fragment boxes within it](https://www.w3.org/TR/css-inline-3/#line-boxes). This means, if you set the line-height of something to 0, that element will have no effect on the height of the line.

It's fine, as long as the footnote-reference is never alone on a line. That would be bad typography anyway.

GitHub and rustdoc's CSS both do the exact same thing. You might want to apply this style to *all* superscripts, not just footnote references.

https://github.com/necolas/normalize.css/blob/master/normalize.css#L128

notriddle (approved):

notriddle:
What just happened?

This didn't get merged, did it?

ehuss:
I'm not sure. I reposted this at #2465.

Issue #2447, opened 2024-10-16
INI syntax highlighting not highlighting sections


SolidTux:
### Problem

`ini` is listed as a supported type in the [documentation](https://rust-lang.github.io/mdBook/format/theme/syntax-highlighting.html), but the syntax highlighting is very sparse and seems to only highlight numbers and quoted string. Taking the example from [Wikipedia](https://en.wikipedia.org/wiki/INI_file):

~~~ini
; last modified 1 April 2001 by John Doe
[owner]
name = John Doe
organization = Acme Widgets Inc.

[database]
; use IP address in case network name resolution is not working
server = 192.0.2.62     
port = 143
file = "payroll.dat"
~~~

is rendered as

![Image](https://github.com/user-attachments/assets/f2c54e43-2d8d-4739-9c02-4f05bb8a0209)

### Steps

1. Create file with some ini code
2. Compile book

### Version

mdbook v0.4.40

SolidTux:
That was fast, thank you! 🥳

PR #2448, opened 2024-10-17
Enhance syntax highlighting

jackieh:
Add syntax highlighting for `hljs-attr` and `hljs-section` CSS classes, consistent with the Ayu theme.

Fixes https://github.com/rust-lang/mdBook/issues/2447

Before:
before

After:
after

Affects syntax highlighting for the following languages: apache, ini, javascript, makefile, markdown, nginx, nix, properties, yaml.


ehuss (approved):
Thanks!

PR #2454, opened 2024-10-23
Improve theme support when JS is disabled

GuillaumeGomez:
Little PR to follow the system's light/dark theme when JS is disabled.

### DIFF
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -40,9 +40,9 @@ a > .hljs {
     border-block-end-style: solid;
 }
 #menu-bar.sticky,
-.js #menu-bar-hover-placeholder:hover + #menu-bar,
-.js #menu-bar:hover,
-.js.sidebar-visible #menu-bar {
+#menu-bar-hover-placeholder:hover + #menu-bar,
+#menu-bar:hover,
+html.sidebar-visible #menu-bar {
     position: -webkit-sticky;
     position: sticky;
     top: 0 !important;
@@ -91,7 +91,7 @@ a > .hljs {
     display: flex;
     margin: 0 5px;
 }
-.no-js .left-buttons button {
+html:not(.js) .left-buttons button {
     display: none;
 }
 
@@ -107,7 +107,7 @@ a > .hljs {
     overflow: hidden;
     text-overflow: ellipsis;
 }
-.js .menu-title {
+.menu-title {
     cursor: pointer;
 }
 
@@ -427,8 +427,7 @@ ul#searchresults span.teaser em {
     -ms-user-select: none;
     user-select: none;
 }
-.no-js .sidebar,
-.js:not(.sidebar-resizing) .sidebar {
+html:not(.sidebar-resizing) .sidebar {
     transition: transform 0.3s; /* Animation: slide away */
 }
 .sidebar code {
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -1,5 +1,5 @@
 <!DOCTYPE HTML>
-<html lang="{{ language }}" class="{{ default_theme }}" dir="{{ text_direction }}">
+<html lang="{{ language }}" class="{{ default_theme }} sidebar-visible" dir="{{ text_direction }}">
     <head>
         <!-- Book generated using mdBook -->
         <meta charset="UTF-8">
@@ -53,7 +53,7 @@
         <script async src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS-MML_HTMLorMML"></script>
         {{/if}}
     </head>
-    <body class="sidebar-visible no-js">
+    <body>
     <div id="body-container">
         <!-- Provide site root to javascript -->
         <script>
@@ -82,19 +82,16 @@
             var theme;
             try { theme = localStorage.getItem('mdbook-theme'); } catch(e) { }
             if (theme === null || theme === undefined) { theme = default_theme; }
-            var html = document.querySelector('html');
+            const html = document.documentElement;
             html.classList.remove('{{ default_theme }}')
             html.classList.add(theme);
-            var body = document.querySelector('body');
-            body.classList.remove('no-js')
-            body.classList.add('js');
+            html.classList.add("js");
         </script>
 
         <input type="checkbox" id="sidebar-toggle-anchor" class="hidden">
 
         <!-- Hide / unhide sidebar before it is displayed -->
         <script>
-            var body = document.querySelector('body');
             var sidebar = null;
             var sidebar_toggle = document.getElementById("sidebar-toggle-anchor");
             if (document.body.clientWidth >= 1080) {
@@ -104,8 +101,8 @@
                 sidebar = 'hidden';
             }
             sidebar_toggle.checked = sidebar === 'visible';
-            body.classList.remove('sidebar-visible');
-            body.classList.add("sidebar-" + sidebar);
+            html.classList.remove('sidebar-visible');
+            html.classList.add("sidebar-" + sidebar);
         </script>
 
         <nav id="sidebar" class="sidebar" aria-label="Table of contents">
--- a/src/theme/css/variables.css
+++ b/src/theme/css/variables.css
@@ -112,7 +112,7 @@
     --copy-button-filter-hover: invert(36%) sepia(70%) saturate(503%) hue-rotate(167deg) brightness(98%) contrast(89%);
 }
 
-.light {
+.light, html:not(.js) {
     --bg: hsl(0, 0%, 100%);
     --fg: hsl(0, 0%, 0%);
 
@@ -258,7 +258,7 @@
 }
 
 @media (prefers-color-scheme: dark) {
-    .light.no-js {
+    html:not(.js) {
         --bg: hsl(200, 7%, 8%);
         --fg: #98a3ad;
 
@@ -298,6 +298,8 @@
         --searchresults-border-color: #98a3ad;
         --searchresults-li-bg: #2b2b2f;
         --search-mark-bg: #355c7d;
+
+        --color-scheme: dark;
 
         /* Same as `--icons` */
         --copy-button-filter: invert(26%) sepia(8%) saturate(575%) hue-rotate(169deg) brightness(87%) contrast(82%);


GuillaumeGomez:
Finally fixed CI. ^^'

ehuss:
@notriddle Is this something you'd be willing to help review? No worries if not, I don't want to pressure anyone.

src/theme/css/noscript.css
```diff
@@ -0,0 +1,7 @@
+.left-buttons button {
+    display: none;
+}
+
+.sidebar {
+    transition: transform 0.3s; /* Animation: slide away */
+}
```
notriddle, line 7: The normal CSS file has this in it:

```css
body:not(.sidebar-resizing) .sidebar {
    transition: transform 0.3s; /* Animation: slide away */
}
```

Is there some reason why this doesn't work?

GuillaumeGomez:
I think it'd be nice for mdBook to get more reviewers, so if needed, don't hesitate to ask me too for other PRs.

src/theme/index.hbs
```diff
@@ -85,9 +88,7 @@
             var html = document.querySelector('html');
             html.classList.remove('{{ default_theme }}')
             html.classList.add(theme);
-            var body = document.querySelector('body');
-            body.classList.remove('no-js')
-            body.classList.add('js');
+            document.body.classList.add("js");
```
notriddle, line 91: If you're using a `noscript.css` file, do you need this at all?

notriddle (changes requested):
It looks like it's just code cleanup without any user-visible change, but your commit message sounds like it's saying that themes actually *work* better in noscript.

I don't think this PR allows you to actually switch themes when JS is disabled (I don't know how you'd even do that). It does allow you to see the menu bar, but the theme button isn't there.

The code itself seems okay, other than one question I asked here.

src/theme/css/noscript.css
GuillaumeGomez: I think this one can be removed indeed.

src/theme/index.hbs
GuillaumeGomez: For theme handling in case JS is actually enabled: you don't want the system theme to be applied in this case by default.

GuillaumeGomez:
> It looks like it's just code cleanup without any user-visible change, but your commit message sounds like it's saying that themes actually _work_ better in noscript.

It's now taking into account the system theme, which was not the case until now. So it is an improvement. ;)

> I don't think this PR allows you to actually switch themes when JS is disabled (I don't know how you'd even do that). It does allow you to see the menu bar, but the theme button isn't there.

It doesn't indeed.

notriddle:
> It's now taking into account the system theme, which was not the case until now. So it is an improvement. ;)

Okay. Please say that in the PR description. Instead of just "Little PR to improve theme support when JS is disabled.", I would prefer something like "Little PR to follow the system's light/dark theme when JS is disabled."

src/theme/index.hbs
notriddle: Ah, I understand now. This is specifically used by `variables.css`.

GuillaumeGomez:
Good idea, done!

ehuss:
This seems like quite a large change compared to #2441. Why is this approach preferred over the one in #2441?

GuillaumeGomez:
This PR removes some JS (which is why there are more changes, removing all `.no-js` items in CSS for example), introduces `noscript.css` which allows to have style specifically when JS is disabled.

Overall this PR is more complete I'd say.

src/theme/css/chrome.css
```diff
@@ -107,7 +104,7 @@ a > .hljs {
     overflow: hidden;
     text-overflow: ellipsis;
 }
-.js .menu-title {
+.menu-title {
     cursor: pointer;
```
notriddle, line 108: If I understand it correctly, this change will result in the :point_up_2: cursor being used on the menu title, even when javascript is disabled. I don't think clicking on the title actually does anything when javascript is disabled, so you'd need to do something to prevent that.

src/theme/css/noscript.css
```diff
@@ -0,0 +1,3 @@
+.left-buttons button {
+    display: none;
+}
```
notriddle, line 3: So, `noscript.css` is literally one CSS rule.

Is this change actually worth it? More specifically, do you expect to have a lot more JS-specific CSS than just this?

src/theme/css/chrome.css
GuillaumeGomez: Good point. We could emulate the same behaviour (scrolling to the top) without JS by replacing this element with `<a href="#">...</a>`. Should I do it in this PR?

src/theme/css/noscript.css
GuillaumeGomez: Might not be worth it currently. I didn't look at the rest yet but I think a lot more things could be done without JS. Anyway, gonna remove this file and move the rule into the main CSS file with `body:not(.js)` and should be fine.

GuillaumeGomez:
Removed the `noscript.css` file. Diff is much smaller now.

src/theme/css/chrome.css
notriddle: I think that should be a separate PR.

src/theme/css/general.css
```diff
@@ -14,6 +14,10 @@ html {
     -webkit-text-size-adjust: none;
 }
 
+* {
+    color: var(--fg);
+}
```
notriddle, line 19: Okay, so, why is this needed?

src/theme/css/chrome.css
GuillaumeGomez: :+1:

src/theme/css/general.css
GuillaumeGomez: Without this change, it looks like this:

image

notriddle: Yes, but why is it happening? This change seems like a workaround more than an actual fix.

GuillaumeGomez: I can't figure out why. In the inspector, it says that this is the color in use, but for reasons I can't understand it's not being applied.

notriddle: Here's a patch that fixes the problem in both Firefox and Ungoogled Chromium when I try it:

```diff
diff --git a/src/theme/css/chrome.css b/src/theme/css/chrome.css
index 9dbfe2c18..369d59ae0 100644
--- a/src/theme/css/chrome.css
+++ b/src/theme/css/chrome.css
@@ -91,7 +91,7 @@ body.sidebar-visible #menu-bar {
     display: flex;
     margin: 0 5px;
 }
-body:not(.js) .left-buttons button {
+html:not(.js) .left-buttons button {
     display: none;
 }
 
@@ -427,7 +427,7 @@ ul#searchresults span.teaser em {
     -ms-user-select: none;
     user-select: none;
 }
-body:not(.sidebar-resizing) .sidebar {
+html:not(.sidebar-resizing) .sidebar {
     transition: transform 0.3s; /* Animation: slide away */
 }
 .sidebar code {
diff --git a/src/theme/css/general.css b/src/theme/css/general.css
index 0e7443cc7..7670b087d 100644
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -14,10 +14,6 @@ html {
     -webkit-text-size-adjust: none;
 }
 
-* {
-    color: var(--fg);
-}
-
 body {
     margin: 0;
     font-size: 1.6rem;
diff --git a/src/theme/css/variables.css b/src/theme/css/variables.css
index a30b97eba..12d1db7a3 100644
--- a/src/theme/css/variables.css
+++ b/src/theme/css/variables.css
@@ -112,7 +112,7 @@
     --copy-button-filter-hover: invert(36%) sepia(70%) saturate(503%) hue-rotate(167deg) brightness(98%) contrast(89%);
 }
 
-.light, body:not(.js) {
+.light, html:not(.js) {
     --bg: hsl(0, 0%, 100%);
     --fg: hsl(0, 0%, 0%);
 
@@ -258,7 +258,7 @@
 }
 
 @media (prefers-color-scheme: dark) {
-    body:not(.js) {
+    html:not(.js) {
         --bg: hsl(200, 7%, 8%);
         --fg: #98a3ad;
 
diff --git a/src/theme/index.hbs b/src/theme/index.hbs
index ea19458ec..99d304fc1 100644
--- a/src/theme/index.hbs
+++ b/src/theme/index.hbs
@@ -85,7 +85,7 @@
             var html = document.querySelector('html');
             html.classList.remove('{{ default_theme }}')
             html.classList.add(theme);
-            document.body.classList.add("js");
+            html.classList.add("js");
         </script>
 
         <input type="checkbox" id="sidebar-toggle-anchor" class="hidden">
```

GuillaumeGomez: Wow. This is really impressive. Thanks a lot!

GuillaumeGomez:
Applied your suggestions. :)

src/theme/css/chrome.css
```diff
@@ -465,7 +464,7 @@ ul#searchresults span.teaser em {
     left: calc(var(--sidebar-resize-indicator-width) * -1);
     right: unset;
 }
-.js .sidebar .sidebar-resize-handle {
+.sidebar .sidebar-resize-handle {
```
notriddle, line 467: This change results in the sidebar handle being visible and having a cursor, but it doesn't actually work.

src/theme/css/chrome.css
```diff
@@ -427,8 +427,7 @@ ul#searchresults span.teaser em {
     -ms-user-select: none;
     user-select: none;
 }
-.no-js .sidebar,
-.js:not(.sidebar-resizing) .sidebar {
+html:not(.sidebar-resizing) .sidebar {
     transition: transform 0.3s; /* Animation: slide away */
```
notriddle, line 431: The `sidebar-resizing` and `sidebar-visible` classes are currently set on the body element. I would probably move them to the html element, just like all the other utility classes, so that you don't have to memorize which is which.

Right now, you get a very strange rubber-band effect when resizing the sidebar.

notriddle (changes requested):

src/theme/css/chrome.css
GuillaumeGomez: Good point, updated it.

src/theme/css/chrome.css
GuillaumeGomez: Ah indeed. Good catch!

GuillaumeGomez:
Updated!

notriddle (approved):

GuillaumeGomez:
@ehuss PR is now ready. :)

ehuss (approved):
Thanks!

PR #2455, opened 2024-10-25
feat(preprocessor): text tooltip on hover

tramondi:


ehuss:
Thanks for the PR! I would prefer to not add this at this time. I would recommend publishing this as a plugin.

PR #2463, opened 2024-11-01
fix: themes broken when localStorage has invalid theme id stored

### LINKED ISSUE #2461: Theme popup not opening when localStorage has an invalid theme id stored
### Problem

If `mdbook-theme` in localStorage contains a theme id that's not in the book, an uncaught exception causes theme popup to not function properly

### Steps

1. Create a new project with `--theme`, build and serve the book
2. Switch the theme so `mdbook-theme` gets set in localStorage (may not be needed?)
3. Edit `theme/index.hbs`, find `<ul id="theme-list"` and remove/change the `id` of the button so that the saved theme is no longer in the list
4. Rebuild the book
5. Observe that clicking on the theme button no longer opens the menu

### Possible Solution(s)

The root cause is `updateThemeSelected()` throws if the selected theme is not in the DOM

I think the cleanest fix is to update `get_theme()` to return `default_theme` if the saved theme is invalid
```js
    // cache the valid theme ids
    var themeIds = [];
    themePopup.querySelectorAll('button.theme').forEach(function (el) {
        themeIds.push(el.id);
    });

   // ...

    function get_theme() {
        var theme;
        try { theme = localStorage.getItem('mdbook-theme'); } catch (e) { }
        if (theme === null || theme === undefined || !themeIds.includes(theme)) { // <- add this check
            return default_theme;
        } else {
            return theme;
        }
    }
```
Verified that the solution works

### Notes

_No response_

### Version

mdbook v0.4.40

Pistonight:
Closes #2461

Check if stored theme is valid in `get_theme()` and defaults to default theme if not valid.

Validation steps:
- `cargo run -- build guide` and serve the book
- In browser devtools, run `localStorage.setItem("mdbook-theme", "foo")` to set an invalid theme `foo`
- Refresh the page, themes should work properly:
    -  the default theme should load
    - the theme popup functions properly
 
Manually tested in Google Chrome, MS Edge and Firefox (I don't have access to safari :(

Other:
- `localStorage` won't be updated (i.e. keeps the invalid value) until user manually changes the theme

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -289,6 +289,10 @@ function playground_text(playground, hidden = true) {
     var themeToggleButton = document.getElementById('theme-toggle');
     var themePopup = document.getElementById('theme-list');
     var themeColorMetaTag = document.querySelector('meta[name="theme-color"]');
+    var themeIds = [];
+    themePopup.querySelectorAll('button.theme').forEach(function (el) {
+        themeIds.push(el.id);
+    });
     var stylesheets = {
         ayuHighlight: document.querySelector("[href$='ayu-highlight.css']"),
         tomorrowNight: document.querySelector("[href$='tomorrow-night.css']"),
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

PR #2468, opened 2024-11-06
Add cairo support to highlight.js

wawel37:
Hello!

We wanted to support a [cairo](https://www.cairo-lang.org/) inside the `highlight.js`. We use mdBook inside our build / package manager tool called [Scarb](https://www.cairo-lang.org/), which has an option to generate markdown files with documentation and later be used by mdBook to generate a proper documentation html.


ehuss:
Can you say how you generated this? I do not see cairo as a supported language in highlight.js 10.1. If it isn't one of the built-in languages, I don't think I would want to include it. I think you should be able to include it as a js extension.

wawel37:
> Can you say how you generated this? I do not see cairo as a supported language in highlight.js 10.1. If it isn't one of the built-in languages, I don't think I would want to include it. I think you should be able to include it as a js extension.

Tbh, I just manually modified the rust rule in the minified highlight.js, as cairo is based upon rust. 

I am fully aware that we can include our own JS extension with the mdBook result, and that's what we are actually doing right now. And it works. But the problem is that we don't actually run a mdBook for a user, so later on, the user has to replace the `highlight.js` himself. We wanted mdBook to support the cairo out of the box. 

I also completely understand your concern. I will go and make a PR inside the `highlight.js` with cairo support, and later on come back here :)

ehuss:
> I will go and make a PR inside the `highlight.js` with cairo support, and later on come back here :)

Unfortunately that won't help for now, since we are stuck on an old version of highlight.js and can't update.

wawel37:
> > I will go and make a PR inside the `highlight.js` with cairo support, and later on come back here :)
> 
> Unfortunately that won't help for now, since we are stuck on an old version of highlight.js and can't update.

So just to be clear, there is no way to directly add a custom language (one that isn't natively supported by 10.1 highlight.js) to mdBook's highlight.js file, right?

ehuss:
Correct.

Issue #1927, opened 2022-11-14
Add footnote backlinks


ashutoshmjain:
Hi 

When I create a footnote as shown in the guide..

```
This is an example of a footnote[^note].

[^note]: This text is the contents of the footnote, which will be rendered
    towards the bottom.
```

I get the link from the number in maintext  to the footnote  which works normal, but back-links (footnote to the main text)  don't work ..So effectively in a superlong text , I can get to the footnotes easily but I need to scroll back up to find where was the origin .. wondering if I am doing something wrong?

ashutoshmjain:
To further elaborate on this issue -  When I see the preview in vscode, I do see the backlinks and they work fine as intended. But when I publish my markdown file to github pages via msbook , these backlinks simply vanish. 

<img width="1440" alt="Screen Shot 2022-11-18 at 4 16 52 PM" src="https://user-images.githubusercontent.com/17075281/202823837-0e697013-0bbe-498d-8bba-2ce135eff963.png">

ashutoshmjain:
If you want to test the behaviour live , please look at https://gita.shutri.com

PR #2475, opened 2024-11-07
footnote backreference links

gambhiro:
This adds an option to add backreference links at the end of footnote definitions, so that the user can return and continue reading the main body of the text.

The code is based on the [footnote-rewrite.rs](https://github.com/pulldown-cmark/pulldown-cmark/blob/master/pulldown-cmark/examples/footnote-rewrite.rs) example.

It can be enabled with a TOML option:

    [output.html]
    footnote-backrefs = true

This feature has been requested before in https://github.com/rust-lang/mdBook/issues/1927

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1139,7 +1139,7 @@ dependencies = [
 
 [[package]]
 name = "mdbook"
-version = "0.4.42"
+version = "0.4.43"
 dependencies = [
  "ammonia",
  "anyhow",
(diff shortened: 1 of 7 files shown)


ehuss:
Thanks for the PR! I went ahead and pushed https://github.com/rust-lang/mdBook/pull/2626. I didn't feel like I wanted to make this configurable, and there were a number of changes and fixes I wanted to include. I appreciate your help moving this forward, it is something I've wanted for some time.

Issue #2460, opened 2024-10-26
Support Djot


dcampbell24:
### Problem

It would be nice if you could write [Djot](https://djot.net/) files. There currently is a [library](https://github.com/hellux/jotdown) supporting Djot.

### Proposed Solution

This would work by files ending in `.md` are Markdown and files ending in `.dj` are Djot. The Djot files would be parsed into html by Jotdown. Otherwise, Djot files would be handled the same as Markdown files.

Dylan-DPC:
Thanks for creating the issue. A mdbook plugin might be a better option for this than supporting it out of the box at the moment as this does sound niche.

dcampbell24:
What do you mean by a plugin? Do you mean make it an optional feature?

dcampbell24:
Alright, I made a plugin:  https://github.com/dcampbell24/mdbook-djot.

PR #2483, opened 2024-11-14
Add djot support, make a test_djot_book folder.

dcampbell24:
Works on #2460. Tell me if you would accept an optional feature instead of this.


ehuss:
Thanks, but I don't think I'll be accepting additional input formats at this time. If possible, you can maybe try implementing it as a plugin if it is possible to convert the input to HTML (which should just pass-through mdbook's markdown processing).

dcampbell24:
Okay, thanks. I made a plugin: https://github.com/dcampbell24/mdbook-djot.

PR #2484, opened 2024-11-15
Use embedded Font Awesome SVG icons

uncenter:
Incorporates the changes from #1330, brings up to date with repository and adds various fixes & improvements.


notriddle (approved):

src/renderer/html_handlebars/helpers/fontawesome.rs
```diff
@@ -0,0 +1,53 @@
...
+        .or_else(|| name.strip_prefix("fas-"))
+        .unwrap_or(name);
+
+    if let Some(id) = h.param(2).and_then(|v| v.value().as_str()) {
+        out.write(&format!("<span class=\"fa-svg\" id=\"{}\">", id))?;
+    } else {
+        out.write("<span class=\"fa-svg\">")?;
+    }
+    out.write(
+        fa::svg(type_, name)
```
notriddle, line 47: I wonder if, because older versions of fontawesome didn't have the "brands" category, we should try to use a fallback to cover old books. For example:

```suggestion
        fa::svg(type_, name)
            .or_else(|| fa::svg(fa::Type::Brands, name))
```

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -855,6 +833,56 @@ fn insert_link_into_header(
...
+                        continue;
+                    }
+                }
+                other_classes += " ";
+                other_classes += class;
+            }
+
+            if icon.is_empty() {
+                text.to_owned()
+            } else if let Ok(svg) = fa::svg(type_, &icon) {
```
notriddle, line 871: I wonder if, because older versions of fontawesome didn't have the "brands" category, we should try to use a fallback to cover old books. For example:

```suggestion
            } else if let Ok(svg) = fa::svg(type_, &icon).or_else(|| fa::svg(fa::Type::Brands, &icon)) {
```

notriddle (commented):
Not sure if it's a good idea to try to work around this, but existing books are going to be written to the old fontawesome version.

uncenter:
Would it be possible to get a review on this PR? I don't want it to lose momentum and have to deal with conflicts again down the road.

src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -855,6 +833,58 @@ fn insert_link_into_header(
...
+                other_classes += class;
+            }
+
+            if icon.is_empty() {
+                text.to_owned()
+            } else if let Ok(svg) =
+                fa::svg(type_, &icon).or_else(|_| fa::svg(fa::Type::Brands, &icon))
+            {
+                format!(
+                    r#"<span{before}class="fa-svg{other_classes}"{after}>{svg}</span>"#,
```
GuillaumeGomez, line 875: Wouldn't it be better instead of generating the SVG in the DOM directly to rely on CSS instead? It brings multiple advantages:

1. Smaller DOM
2. If the same SVG is used more than once, it's not duplicated

GuillaumeGomez:
Overall I think it's a very good idea, but unless I'm wrong, you're inlining the SVGs directly into the HTML, and I don't think this is the right approach. If you use the CSS and font provided by font awesome, then no need for this.

src/renderer/html_handlebars/hbs_renderer.rs
notriddle: I don't want to build a giant stylesheet with every icon in it unless we have to. That file would be immense, and most of it wouldn't get used.

Besides, fontawesome SVGs almost all have a single `<path>` in them. The DOM is tiny.

uncenter: > FontAwesome is a font. Meaning that as long as we have the font, we have easily use their icons without needing to declare them ourselves.

But then you must include the entirety of the font (which is quite large iirc), unless you want to deal with subsetting? And then that is a whole other ordeal.

uncenter: CSS is a valid way to go about the inclusion of icons, and could be done in the future, but it should not be a blocker for this pull request to land.

uncenter:
> Overall I think it's a very good idea, but unless I'm wrong, you're inlining the SVGs directly into the HTML, and I don't think this is the right approach. If you use the CSS and font provided by font awesome, then no need for this.

CSS is a valid way to go about the inclusion of icons, and could be done in the future, but it should not be a blocker for this pull request to land. Much of the point of this pull request is to avoid the 77kb of fonts that must be loaded from FontAwesome, and to avoid the other downsides as mentioned in https://github.com/rust-lang/mdBook/pull/1330#issue-702340617.

GuillaumeGomez:
Then please inline the SVG into the CSS to prevent duplication in HTML.

PR #2487, opened 2024-11-18
Migrate to rinja

GuillaumeGomez:
Main argument for switching to rinja: compilation fails if your template is invalid. It also generates very performant code (as can be seen [here](https://github.com/mitsuhiko/minijinja/pull/588)).

Some extra advantages: it makes the code simpler to follow. Storing everything into a JSON map made it pretty tricky to understand what was going on and needed to have a lot more checks in the code.

With this JSON map gone, I also simplified code to handle chapter kinds: it's now a `Chapter` enum, making it easier to understand.

Last point: rustdoc, docs.rs and clippy lints page already switched to rinja so I think it tells a lot about how reliable `rinja` is. :)

Considering how big the PR is, please don't hesitate to ask me if anything is unclear.


notriddle:
In the current version of mdBook, `mdbook init --theme` drops copies of the hbs files that book authors can then modify, allowing them to change the HTML without recompiling.

This PR seems to remove that functionality. I don't think that's a good idea without first taking stock of customization that popular books use, helping them migrate over to some better way(s) to accomplish the same things, and only then deprecating and removing support for custom hbs.

In any case, you really need to mention stuff like that in the PR description.

GuillaumeGomez:
I didn't realize it was a feature. I saw a test checking that but didn't check further, should have. Gonna investigate.

GuillaumeGomez:
* [bevy cheat book](https://bevy-cheatbook.github.io/overview.html) is using JS to modify the pages.
 * Rust official books don't seem to update this file.
 * [rinja book](https://rinja.readthedocs.io/en/stable/) doesn't modify it.
 * [askama book](https://djc.github.io/askama/) doesn't modify it.
 * WIP

notriddle:
rust-by-example has a custom language picker, and [zhp-cn](https://github.com/zjp-CN/mdbook-theme) has a custom TOC

PR #2500, opened 2024-12-02
Add aarch64-apple-darwin release target

rukai:
we are now 4 generations into apples aarch64 desktop/laptop hardware, so we should definitely be providing builds for aarch64 macos.

We already test aarch64 macos as part of CI checks gating every checkin: https://github.com/rust-lang/mdBook/blob/74b2c79d46df45ab9c196646f98339dd7e0c00c3/.github/workflows/main.yml#L35


ehuss (approved):
Thanks!

PR #2508, opened 2024-12-19
Support web components for globalKeyHandler

### LINKED ISSUE #2507: Avoid triggering search when `s` is pressed on active input.
### Problem

We're trying to add a custom search element to mdbook docs, and the default search logic keeps stealing the `s` key when we try to type it into the search box of our alternate search. 

### Steps

1. To go https://test-builds.readthedocs.io/en/mdbook/
2. Press `/` to bring up our custom search
3. Try to type a word with `s` in it. 

### Possible Solution(s)

The mdbook search logic should not fire when there is an input element that has focus.

It seems like https://github.com/rust-lang/mdBook/blob/master/src/theme/searcher/searcher.js#L319 should be checking for input types as well as textarea types? 

### Notes

_No response_

### Version

```text
mdbook v0.4.43
```

ericholscher:
Fixes https://github.com/rust-lang/mdBook/issues/2507

### DIFF
--- a/src/theme/searcher/searcher.js
+++ b/src/theme/searcher/searcher.js
@@ -316,7 +316,9 @@ window.search = window.search || {};
     
     // Eventhandler for keyevents on `document`
     function globalKeyHandler(e) {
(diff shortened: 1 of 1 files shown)


binhle113 (approved):

binhle113 (approved):

ehuss:
Thanks for the PR! I'm going to close as this is resolved by https://github.com/rust-lang/mdBook/pull/3087 where it goes with a slightly different solution.

PR #2509, opened 2024-12-20
Make base_url available for .hbs to set base href if configured.

prasad83:
When site_url is configured make it available for hbs to set base href (using base_url)


prasad83:
Without this change I had to make workaround as follows:

theme/head.hbs

```
{{#if live_reload_endpoint}}
<!-- local run -->
{{else}}
<base href="https://website.url/">
{{/if}}
```

PR #2517, opened 2025-01-02
Ignore fragment when figuring out sidebar items

### LINKED ISSUE #2515: Fragments in URLs break sidebar auto-scrolling
### Problem

If you load a URL with a fragment in it, the sidebar does not auto-scroll correctly. When loading the page, it should automatically scroll the sidebar into the current page's position (in the middle of the screen IIRC).

@notriddle Is this something you can look at? I believe this regressed from #2414 (though I haven't bisected).

### Steps

Example that works: https://doc.rust-lang.org/1.83.0/reference/expressions.html#expression-precedence
Example that doesn't work: https://doc.rust-lang.org/nightly/reference/expressions.html#expression-precedence
Example that works in all versions (no fragment): https://doc.rust-lang.org/nightly/reference/expressions.html

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
mdbook v0.4.43
and
mdbook v0.4.42

Tested on multiple browsers.
```

notriddle:
Fixes #2515

### DIFF
--- a/src/theme/toc.js.hbs
+++ b/src/theme/toc.js.hbs
@@ -10,7 +10,7 @@ class MDBookSidebarScrollbox extends HTMLElement {
     connectedCallback() {
         this.innerHTML = '{{#toc}}{{/toc}}';
         // Set the current, active page, and reveal it if it's hidden
-        let current_page = document.location.href.toString();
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

PR #2518, opened 2025-01-03
Add configuration option to disable the default hidelines feature for rust

zgtm:
Simple change that adds a configuration option to disable the default hidelines feature for rust:

Usage:

```
[output.html.code]
default-hidelines = false
```

I need this in my book since I have several multi-line strings that have lines starting with `#`. (Embedded code snippets with preprocessor directives.)

~~Let me know if this needs a test.~~

If you want I can also add a line to the manual, but since this is a niche feature, the doc string  might just be enough. :)

[Edit: Added a test]

### DIFF
--- a/crates/mdbook-core/src/config.rs
+++ b/crates/mdbook-core/src/config.rs
@@ -605,12 +605,23 @@ impl Default for Playground {
 }
 
 /// Configuration for tweaking how the HTML renderer handles code blocks.
-#[derive(Debug, Default, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(default, rename_all = "kebab-case", deny_unknown_fields)]
 #[non_exhaustive]
 pub struct Code {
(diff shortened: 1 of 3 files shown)


max-heller:
> I need this in my book since I have several multi-line strings that have lines starting with `#`. (Embedded code snippets with preprocessor directives.)

[You can use two consecutive hashes `##` to keep the line visible.](https://doc.rust-lang.org/stable/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example)

zgtm:
Hi @max-heller, thanks for the hint!

Unfortunately this will not work in my case, as I'm trying to include the code from a source file via `{{#include filename}}`, and the source file will only work with this particular line starting with a single '#'.

(One could argue whether the line hiding mechanism should be disable completely for included sources via `{{#include filename}}`. But I did not want to open that can of worms. :grimacing:)

zgtm:
Updated to
- resolve merge conflicts
- conform to updated formatting guidelines
- fix creation of now-`non_exhaustive` structs

zgtm:
Rebased onto the new HTML rendering pipeline (#2844).

PR #2523, opened 2025-01-15
ci: move ubuntu-20 jobs to ubuntu-22

marcoieni:
ubuntu-20 will be removed in April. See the announcement [here](https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#ubuntu-20-image-is-closing-down).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -17,11 +17,11 @@ jobs:
       matrix:
         include:
           - target: aarch64-unknown-linux-musl
-            os: ubuntu-20.04
+            os: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks!

diegomrsantos:
Hi, thanks for all the work on mdBook! I just wanted to mention that this change introduced a dependency on glibc 2.34, which caused compatibility issues for some users on older systems. Since this was released as a patch version (0.4.44), it might be worth considering following semantic versioning more strictly and reserving breaking changes for minor releases instead. This could help users avoid unexpected issues when upgrading.

PR #2524, opened 2025-01-15
nicer style rules for margin around footnote defs

WaffleLapkin:
previous implementation used `:not(.footnote-definition) + .footnote-definition` and `.footnote-definition + :not(.footnote-definition)`.
the latter selector caused many problems:
- it doesn't select footnote definitions which are last children (this can be easily triggered in a blockquote)
- it changes the margin of the next sibling, rather than the footnote definition itself, which can also *shrink* margin for elements with big margins (this happens to headings)
- because it applies to the next sibling it is also quite hard to override in user styles, since it may apply to any element
  
this PR replaces the latter selector with `:not(:has(+ .fd))`, which fixes all of the mentioned problems.

see https://github.com/rust-lang/reference/pull/1710/commits/bc02bf95f937b2ea5675cd6eebf54f6ac989945d for concrete motivation.

See https://caniuse.com/mdn-css_selectors_not and https://caniuse.com/css-has for comparison of browser support of `:not` and `:has` (`:has` is slightly worse).

cc @ehuss

### DIFF
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -200,9 +200,11 @@ sup {
     line-height: 0;
 }
 
-:not(.footnote-definition) + .footnote-definition,
-.footnote-definition + :not(.footnote-definition) {
+:not(.footnote-definition) + .footnote-definition {
     margin-block-start: 2em;
+}
+.footnote-definition:not(:has(+ .footnote-definition)) {
+    margin-block-end: 2em;
 }
 .footnote-definition {
     font-size: 0.9em;


ehuss:
I don't think this works correctly. The problem is that if you have multiple sections with footnotes, only the footnotes in the first section will get the styling applied. You should be able to see the differences in something like:

```
# Chapter 1

This is a footnote in the first section[^1].

[^1]: Testing

## Next section

This is a footnote in the second section[^2].

[^1]: Testing

## Last section

Test
```

WaffleLapkin:
Ugh, I has wrong understanding of what `:last-of-type` means, I though "sibling" meant "`+`", not "`~`"...

I changed the rule to `:not(:has(.footnote-def)))`, which does do what I wanted. It actually slightly changes the style (new on the right):
image

The two differences are:
- More margin before headings after footnote defs (before footnote rule would override headings' margins to be smaller)
- More margin if the last footnote def has no following siblings (i.e. it's the last child)

WaffleLapkin:
ping @ehuss

ehuss (approved):
Thanks! I'm not really sure I understand the implications here, but I don't immediately see any issues.

Issue #2528, opened 2025-01-20
Sidebar renders badly when JS is disabled


GuillaumeGomez:
### Problem

As you can see:

![Image](https://github.com/user-attachments/assets/0e69bad2-80b4-4a54-8d7c-1aeb8f23f14d)

It should take the whole height but doesn't.

### Steps

Go to any page and disable javascript.

### Possible Solution(s)

Just a small CSS fix.

### Version

```text
v0.4.43
```

PR #2529, opened 2025-01-20
Fix display of sidebar when JS is disabled

GuillaumeGomez:
Fixes https://github.com/rust-lang/mdBook/issues/2528.

For the GUI test, I need to update the framework to allow to select an element inside a frame (opened https://github.com/GuillaumeGomez/browser-UI-test/issues/628 for it).

cc @notriddle


notriddle (approved):

ehuss (approved):
Thanks!

PR #2530, opened 2025-01-20
fix: make line hiding in Rust code blocks consistent with `rustdoc`

### LINKED ISSUE #2526: Line hiding in Rust code blocks is inconsistent with `rustdoc`
### Problem

Lines in Rust code blocks beginning with a `#` and followed immediately by a character other than `!`, `[`, `#`, or ` ` are hidden by `mdbook` but shown by `rustdoc`.

For example, the following code block is rendered as empty by `mdbook`:
````
```rust
#fn main() {}
```
````

Meanwhile, the following code block in a doc comment:
````rust
/// ```
/// #fn main() {}
/// ```
pub struct Foo;
````
is rendered as
Image

### Steps

1. `mdbook init`
2. `echo '```rust\n#fn main() {}\n```' > src/chapter_1.md`
3. `mdbook serve`
4. Observe an empty code block. When the "Show hidden lines" button is clicked, the code block contains `fn main() {}`

### Possible Solution(s)

`mdbook` should implement `rustdoc`'s behavior since [the docs claim it does](https://rust-lang.github.io/mdBook/format/mdbook.html?highlight=hide#hiding-code-lines):
> For the Rust language, you can use the # character as a prefix which will hide lines [like you would with Rustdoc](https://doc.rust-lang.org/stable/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example).

### Notes

[`rustdoc`'s rules](https://doc.rust-lang.org/stable/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example) are:
- If a line begins with `# ` (note the space following the `#`), the line is hidden
- Two consecutive hashes `##` can be used to prevent the hiding of a line that would otherwise be hidden
  - To write a non-hidden line that starts with `# `, write `## ` which will render as `# `

Is `mdbook` willing to change this in a non-breaking release since it seems like a bug according to the docs, or would you rather wait until 0.5?

### Version

```text
mdbook v0.4.43
```

max-heller:
Requires a space following a `#` for a line to be hidden.

Closes #2526

### DIFF
--- a/guide/src/format/mdbook.md
+++ b/guide/src/format/mdbook.md
@@ -4,7 +4,8 @@
 
 There is a feature in mdBook that lets you hide code lines by prepending them with a specific prefix.
 
-For the Rust language, you can use the `#` character as a prefix which will hide lines [like you would with Rustdoc][rustdoc-hide].
+For the Rust language, you can prefix lines with `# ` (`#` followed by a space) to hide them [like you would with Rustdoc][rustdoc-hide].
+This prefix can be escaped with `##` to prevent the hiding of a line that should begin with the literal string `# ` (see [Rustdoc's docs][rustdoc-hide] for more details)
 
 [rustdoc-hide]: https://doc.rust-lang.org/stable/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example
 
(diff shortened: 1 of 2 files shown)


GuillaumeGomez (approved):

ehuss (approved):
Thanks!

PR #2532, opened 2025-01-23
Prevent the real sidebar position from becoming unsynced from the JS

### LINKED ISSUE #2520: Sidebar opening/closing sometimes doesn't open/close it
### Problem

Sometimes the button to open sidebar seems to do nothing.

### Steps

1. Go to https://doc.rust-lang.org/nightly/reference/type-coercions.html
2. open dev tools
3. click on the hamburger in the left top corner until one of your clicks doesn't make the sidebar open/close
4. observe the fact that `<body>` has incorrect class (`sidebar-visible` is the sidebar is hidden and vice-versa)
5. click on the hamburger again and observe that the classes are still incorrect

### Possible Solution(s)

_No response_

### Notes

The bug is random, so sometimes you need to click a lot before it happens. And sometimes it happens once in about 4 clicks. :shrug:

Also is reproducible in both firefox and chrome.

### Version

I can reproduce with `mdbook v0.4.43`

notriddle:
This way, whatever behavior the browser might use for checkboxes will apply to the CSS class, localStorage, and the visible state.

Fixes #2520

### DIFF
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -449,6 +449,7 @@ function playground_text(playground, hidden = true) {
     var sidebar = document.getElementById("sidebar");
     var sidebarLinks = document.querySelectorAll('#sidebar a');
     var sidebarToggleButton = document.getElementById("sidebar-toggle");
+    var sidebarToggleAnchor = document.getElementById("sidebar-toggle-anchor");
     var sidebarResizeHandle = document.getElementById("sidebar-resize-handle");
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

Were you able to reproduce the problem? I'm not sure I fully understand what is going on here.

notriddle:
I was able to reproduce it. Maybe it only happens in Firefox?

PR #2542, opened 2025-02-01
Add clear_renderers / clear_preprocessors

dalance:
Now `Renderer` and `Preprocessor` can be added by `with_renderer` and `with_preprocessor`.
But there is no way to disable the existing renderers and preprocessors.
So this PR add `clear_renderers` and `clear_preprocessors`.
This addition enables the customization of renderer and preprocessor without modifying the original `book.toml`.

### DIFF
--- a/src/book/mod.rs
+++ b/src/book/mod.rs
@@ -256,6 +256,16 @@ impl MDBook {
         self
     }
 
+    /// Clear all registered renderers.
+    pub fn clear_renderers(&mut self) {
+        self.renderers.clear();
+    }
+
+    /// Clear all registered preprocessors.
+    pub fn clear_preprocessors(&mut self) {
+        self.preprocessors.clear();
+    }
+
     /// Run `rustdoc` tests on the book, linking against the provided libraries.
     pub fn test(&mut self, library_paths: Vec<&str>) -> Result<()> {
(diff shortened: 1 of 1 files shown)


ehuss:
Can you say more about why you can't modify the `Config`?

dalance:
@ehuss 

I want to use this API for Rust documents translation project: https://rust-lang-translations.org .

In most cases, the maintainers of the original version are negative to modify their repository for translation. Keeping forks for translation takes a lot of time because these documents are updated frequently.

So I need the way to customize renderer/preprocesser without modifing the original TOML.

ehuss:
If you are using the API, you should be able to load the config with [`Config::from_disk`](https://docs.rs/mdbook/latest/mdbook/config/struct.Config.html#method.from_disk). Then modify the config, for example:

```rust
    if let Some(renderers) = config.get_mut("output").and_then(|o| o.as_table_mut()) {
        renderers.clear();
    }
```

and then use [`MdBook::load_with_config`](https://docs.rs/mdbook/latest/mdbook/book/struct.MDBook.html#method.load_with_config) with whatever customizations you want. Does that work for you?

dalance:
Thank you for your suggestion.
I missed `load_with_config`.

PR #2552, opened 2025-02-20
Remove JSON search file

GuillaumeGomez:
Why bother having the search index in both a JSON and JS file when it always work with the JS file after all. :)

I took over https://github.com/rust-lang/mdBook/pull/2425 to implement this.

cc @notriddle

### DIFF
--- a/src/renderer/html_handlebars/search.rs
+++ b/src/renderer/html_handlebars/search.rs
@@ -60,7 +60,7 @@ pub fn create_files(
     let index = write_to_json(index, search_config, doc_urls)?;
     debug!("Writing search index ✓");
     if index.len() > 10_000_000 {
-        warn!("searchindex.json is very large ({} bytes)", index.len());
+        warn!("search index is very large ({} bytes)", index.len());
     }
 
     if search_config.copy_js {
--- a/src/renderer/html_handlebars/search.rs
+++ b/src/renderer/html_handlebars/search.rs
@@ -64,7 +64,6 @@ pub fn create_files(
     }
 
     if search_config.copy_js {
-        static_files.add_builtin("searchindex.json", index.as_bytes());
         static_files.add_builtin(
             "searchindex.js",
             format!("Object.assign(window.search, {});", index).as_bytes(),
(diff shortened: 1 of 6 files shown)


GuillaumeGomez:
Rebased and fixed merge conflict.

ehuss:
I don't quite understand this PR. Doesn't this introduce the problem of the index loading not being async? Can you provide a little more commentary with the PRs to help explain what they are doing and why, along with some historical information to explain why it is the way it is originally?

notriddle:
When a script tag is added using the DOM API, [it loads asynchronously](https://html.spec.whatwg.org/#script-force-async). Parsing and running blocks the UI thread, but that's true with JSON.parse too.

ehuss:
I see, thanks!

Is it true that the `fail-on-request-error` lines in the goml files can be removed now? Can you do that here?

Also, can you also update this:

```diff
--- a/src/renderer/html_handlebars/search.rs
+++ b/src/renderer/html_handlebars/search.rs
@@ -60,7 +60,7 @@ pub fn create_files(
     let index = write_to_json(index, search_config, doc_urls)?;
     debug!("Writing search index ✓");
     if index.len() > 10_000_000 {
-        warn!("searchindex.json is very large ({} bytes)", index.len());
+        warn!("search index is very large ({} bytes)", index.len());
     }
 
     if search_config.copy_js {
```

GuillaumeGomez:
> Is it true that the `fail-on-request-error` lines in the goml files can be removed now? Can you do that here?

I thought I did, gonna remove the remaining ones.

> Also, can you also update this:

Sure!

GuillaumeGomez:
In the `move-between` test, the `fail-on-request-error` is still required, but not because of the `search-index` anymore. I updated the comment.

ehuss (approved):
Looks good, thanks!

sgued:
> Parsing and running blocks the UI thread, but that's true with JSON.parse too.

That's true, but you should also be aware that parsing JSON is much faster than parsing object literals. Given the size of the index, this is most likely measurable performance regression.

Note that this could also be improved by having the `searchindex.js` just use `JSON.parse('{...}')`.

https://v8.dev/blog/cost-of-javascript-2019#json

GuillaumeGomez:
It's in my TODO list. We had this issue a long time ago in rustdoc.

PR #2553, opened 2025-02-20
Only load searchindex when needed

GuillaumeGomez:
This PR makes it so that the `searchindex.js` file is only loaded when the user arrives on a page with `?search=` in the URL or if the user opens the search. It means that the book size will be only the text content until the user actually needs to run a search.

cc @notriddle

### DIFF
--- a/src/front-end/css/chrome.css
+++ b/src/front-end/css/chrome.css
@@ -344,9 +344,34 @@ mark.fade-out {
     max-width: var(--content-max-width);
 }
 
+#searchbar-outer.searching #searchbar {
+    padding-right: 30px;
+}
+#searchbar-outer .spinner-wrapper {
+    display: none;
+}
+#searchbar-outer.searching .spinner-wrapper {
+    display: block;
+}
+
+.search-wrapper {
+    position: relative;
+}
+
+.spinner-wrapper {
+    --spinner-margin: 2px;
+    position: absolute;
+    margin-block-start: calc(var(--searchbar-margin-block-start) + var(--spinner-margin));
+    right: var(--spinner-margin);
+    top: 0;
+    bottom: var(--spinner-margin);
+    padding: 6px;
+    background-color: var(--bg);
+}
+
 #searchbar {
     width: 100%;
-    margin-block-start: 5px;
+    margin-block-start: var(--searchbar-margin-block-start);
     margin-block-end: 0;
     margin-inline-start: auto;
     margin-inline-end: auto;
--- a/src/front-end/css/variables.css
+++ b/src/front-end/css/variables.css
@@ -11,6 +11,7 @@
     --menu-bar-height: 50px;
     --mono-font: "Source Code Pro", Consolas, "Ubuntu Mono", Menlo, "DejaVu Sans Mono", monospace, monospace;
     --code-font-size: 0.875em; /* please adjust the ace font size accordingly in editor.js */
+    --searchbar-margin-block-start: 5px;
 }
 
 /* Themes */
--- a/src/front-end/searcher/searcher.js
+++ b/src/front-end/searcher/searcher.js
@@ -22,6 +22,7 @@ window.search = window.search || {};
     }
 
     const search_wrap = document.getElementById('search-wrapper'),
+        searchbar_outer = document.getElementById('searchbar-outer'),
         searchbar = document.getElementById('searchbar'),
         searchresults = document.getElementById('searchresults'),
         searchresults_outer = document.getElementById('searchresults-outer'),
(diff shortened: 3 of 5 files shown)


notriddle:
If you're going to mess with the search engine, can you also write some gui tests that exercise it?

GuillaumeGomez:
Very good point.

GuillaumeGomez:
Added the GUI test. :)

src/theme/searcher/searcher.js
```diff
@@ -259,6 +259,17 @@ window.search = window.search || {};
         doc_urls = config.doc_urls;
         searchindex = elasticlunr.Index.load(config.index);
 
+        searchbar.removeAttribute("disabled");
+        searchbar.focus();
```
notriddle, line 263: I don't like this behavior.

When I click the search button on a (simulated) slow connection, the search box appears, but I can't type anything into it until after the search index is loaded.

We've gone to a lot of hassle to avoid that in Rustdoc, because it's annoying to use when it works that way <https://github.com/rust-lang/rust/pull/90333#issuecomment-953166825>. Instead:

- If the user types something into the search box but the index isn't loaded yet, show a loading indicator of some sort. Hide it after the search index loads.
- After the search index loads, check if the search box is visible and has text in it. If they all check out, run the search.

GuillaumeGomez: Makes sense. I will add a spinner.

notriddle (approved):

GuillaumeGomez:
Rebased and fixed merge conflict.

notriddle:
It looks like you reverted the loading throbber, and now it disables the search input while it's loading again.

https://github.com/rust-lang/mdBook/commit/16cbfd6cfd766cdb3b981696f6b20c1b96b4e901 added it, but the pull request doesn't have it any more.

GuillaumeGomez:
Arg indeed. Rebase went wrong I guess. Great catch, thanks!

notriddle (approved):

notriddle:
Everything seems good here!

GuillaumeGomez:
Rebased.

GuillaumeGomez:
Re-rebased and also ran `eslint`.

notriddle (approved):

GuillaumeGomez:
Rebased. If you could take a look @ehuss. If you need help to understand what this PR is doing, don't hesitate to ask!

ehuss:
Can you say more about the reasoning behind this change? For me, it comes off as a worse experience, so we are somehow not seeing the same thing. For example, after opening a book and looking at it, I may want to search for something. Today this loads instantly (since it is eagerly loaded), but with this change I may need to wait 10+ seconds for anything to happen. Is this trying to save bandwidth? Or is it some performance issue?

I'm not completely against this kind of change, but the description here doesn't explain the motivation or acknowledge the potential downsides.

GuillaumeGomez:
> Today this loads instantly (since it is eagerly loaded), but with this change I may need to wait 10+ seconds for anything to happen. Is this trying to save bandwidth? Or is it some performance issue?

I'm very surprised. What book do you have this 10+ seconds wait?

The whole point of this PR is to actually speed up the page load and reduce the (page) size by default until you actually need the search. This is becomes more and more useful as the book size (and its search index) grows. This is how we allow rustdoc to always load extremely quickly, whatever the size of the crate (and of its search index).

GuillaumeGomez:
It's mostly for people with limited internet access (not in bandwidth but in data) that will get their life improved with this PR: it only loads extra content on demand.

> Can you help me understand how this speeds up page load? My understanding is that the index is loaded async in the background. My expectation is that since it is loading in the background, it shouldn't have any measurable effect on initial render time. What little page-load profiling I've done doesn't show a difference with this PR.

I wasn't clear, my bad. The load time should normally be close to not impacted (except I forgot to convert the JS to `JSON.parse()`, meaning that for big search indexes, all pages will have an impact when the JS will be parsed, because JS parsing is performed in the main thread, should be fixed in https://github.com/rust-lang/mdBook/pull/2633). Here, the impact is to reduce memory usage for the tab (until you actually need the search) and saving data.

ehuss:
I'm still a little concerned about the reduced responsiveness of the search box, though not enough to block this change. For example, given the RFC repo's 22MB search index:

| Speed | Time |
|-------|------|
| 5mbps | 37s  |
| 40mbps | 4.6s |
| 100mbps | 1.8s |
| 500mbps | 0.37s |
| 1000mbps | 0.18s |

I don't know what would be a typical network connection speed we should be aiming for. I think a 1 or 2s delay for slower connections although unfortunate, should be fine. And the delay should only happen on the first search while it stays cached.

Can you rebase and resolve the conflict, and then we can merge?

notriddle:
There's a reason I was so negative on disabling the search field: the index loads while you type, so there's no actual *waiting*.

(at least, if you're not on the 5Mb/s link)

GuillaumeGomez:
We should also work on reducing the size of the search index.

GuillaumeGomez:
Fixed merge conflict. :)

PR #2554, opened 2025-02-21
Fix eslint warnings and add eslint check in CI

GuillaumeGomez:
First commit comes from https://github.com/rust-lang/mdBook/pull/2552.

This PR fixed all eslint warnings and added it as a check in the CI. That is a first step in improving the JS quality.

cc @notriddle

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,95 @@
+{
+    "env": {
+        "browser": true,
+        "node": true,
+        "es6": true
+    },
+    "extends": "eslint:recommended",
+    "globals": {
+        "module": "readonly",
+        "require": "readonly"
+    },
+    "parserOptions": {
+        "ecmaVersion": 2018,
+        "requireConfigFile": false,
+        "sourceType": "module"
+    },
+    "ignorePatterns": ["**min.js", "**/highlight.js", "**/playground_editor/*"],
+    "rules": {
+        "indent": [
+            "error",
+            4
+        ],
+        "linebreak-style": [
+            "error",
+            "unix"
+        ],
+        "quotes": [
+            "error",
+            "single"
+        ],
+        "semi": [
+            "error",
+            "always"
+        ],
+        "brace-style": [
+            "error",
+            "1tbs",
+            { "allowSingleLine": false }
+        ],
+        "curly": "error",
+        "no-trailing-spaces": "error",
+        "no-multi-spaces": "error",
+        "keyword-spacing": [
+            "error",
+            { "before": true, "after": true }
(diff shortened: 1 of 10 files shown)


notriddle (approved):

src/renderer/html_handlebars/search.rs
```diff
@@ -64,7 +64,6 @@ pub fn create_files(
     }
 
     if search_config.copy_js {
-        static_files.add_builtin("searchindex.json", index.as_bytes());
```
ehuss, line 67: Why is this line changing?

.eslintrc.json
ehuss, line 1: Can you say how all the rules for this were created? Why are the styles different from the defaults?

tests/gui/runner.rs
```diff
@@ -26,12 +26,16 @@ fn get_available_browser_ui_test_version() -> Option<String> {
...
 
 fn expected_browser_ui_test_version() -> String {
-    let content = read_to_string(".github/workflows/main.yml")
-        .expect("failed to read `.github/workflows/main.yml`");
-    for line in content.lines() {
-        let line = line.trim();
-        if let Some(version) = line.strip_prefix("BROWSER_UI_TEST_VERSION:") {
-            return version.trim().replace('\'', "");
+    let content = read_to_string("package.json").expect("failed to read `package.json`");
+    let json = json::parse(&content).expect("invalid JSON");
```
ehuss, line 30: Can this use serde_json since it is already in the dependency tree?

ehuss (commented):
Can you add documentation to https://github.com/rust-lang/mdBook/blob/master/CONTRIBUTING.md with links to what this is, how to install it, and some information on how to run and use it?

src/renderer/html_handlebars/search.rs
ehuss: Oh, I see it's from another PR.

.eslintrc.json
GuillaumeGomez: It's the rules we use in rustdoc. "defaults" change from time to time, so with this set of rule, at least it's fixed.

src/renderer/html_handlebars/search.rs
GuillaumeGomez: Yep! From #2552 to be more exact.

tests/gui/runner.rs
GuillaumeGomez: Good idea.

GuillaumeGomez:
Rebased.

ehuss (approved):
Thanks!

PR #2558, opened 2025-03-03
Test downstream crates

szabgab:
Helper tool to easily check if changed made to mdBook don't break selected downstream crates.

See #1574

### DIFF
--- a/ci/test-downstream.sh
+++ b/ci/test-downstream.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+# Test selected downstream crates
+
+set -e
+
+root=$(pwd)
+echo $root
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks! However, this isn't quite what I had in mind in #1574. I would prefer to not have this reliance on external projects, and would also prefer to avoid too many shell scripts (and this isn't wired into CI, either). 

To make progress on #1574, we would probably need some kind of integration test as described in that issue.

szabgab:
Well, it was just an idea either to be part of CI if you liked, or just to have around to run once in a while to make sure downstream projects are not broken by changes to mdbook. It could also be written in any other language.

PR #2560, opened 2025-03-03
Make the error testing stricter

szabgab:
by checking the text of the error message


szabgab:
I'll send them separately to make it easier to review them

PR #2563, opened 2025-03-03
Enhanced wording for editable code blocks docs

jofas:
Corresponding URLO topic:

https://users.rust-lang.org/t/how-to-enable-editable-code-playgrounds-in-mdbook/126476

To me, it was not immediately clear from the docs that `output.html.playground.editable = true` only enables the `editable` attribute on code blocks. I thought it meant that every code block automatically becomes editable and if we only want a single code block to be editable, we use the `editable` attribute instead. I believe with minor changes to the second sentence it becomes clearer to the hasty reader that this is not the case.

### DIFF
--- a/guide/src/format/theme/editor.md
+++ b/guide/src/format/theme/editor.md
@@ -9,8 +9,8 @@ be added to the ***book.toml***:
 editable = true
 ```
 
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

Issue #2564, opened 2025-03-04
Make clippy happy


szabgab:
### Problem

I just ran `cargo clippy` and got a lot of warnings.

```
$ rustc -vV
rustc 1.85.0 (4d91de4e4 2025-02-17)
binary: rustc
commit-hash: 4d91de4e48198da2e33413efdcd9cd2cc0c46688
commit-date: 2025-02-17
host: x86_64-unknown-linux-gnu
release: 1.85.0
LLVM version: 19.1.7
```

IMHO it would be a good idea to fix these and then add `cargo clippy -- -D warnings` to the CI.

```
warning: this seems like a manual implementation of the non-exhaustive pattern
  --> src/book/book.rs:78:1
   |
78 | / pub struct Book {
79 | |     /// The sections in this book.
80 | |     pub sections: Vec<BookItem>,
81 | |     __non_exhaustive: (),
82 | | }
   | |_^
   |
help: remove this field
  --> src/book/book.rs:81:5
   |
81 |     __non_exhaustive: (),
   |     ^^^^^^^^^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_non_exhaustive
   = note: `#[warn(clippy::manual_non_exhaustive)]` on by default
help: use the `#[non_exhaustive]` attribute instead
   |
78 + #[non_exhaustive]
79 | pub struct Book {
   |

warning: doc list item without indentation
  --> src/preprocess/links.rs:22:4
   |
22 | ///.  lines, or only between the specified anchors.
   |    ^
   |
   = help: if this is supposed to be its own paragraph, add a blank line
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#doc_lazy_continuation
   = note: `#[warn(clippy::doc_lazy_continuation)]` on by default
help: indent this line
   |
22 | ///   .  lines, or only between the specified anchors.
   |    +++

warning: doc list item without indentation
  --> src/preprocess/links.rs:24:4
   |
24 | ///.  specified or the lines between specified anchors, and include the rest of the file behind `#`.
   |    ^
   |
   = help: if this is supposed to be its own paragraph, add a blank line
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#doc_lazy_continuation
help: indent this line
   |
24 | ///   .  specified or the lines between specified anchors, and include the rest of the file behind `#`.
   |    +++

[...]

PR #2565, opened 2025-03-04
Fix clippy issues See #2564

szabgab:


ehuss:
Thanks for the PR! As mentioned in https://github.com/rust-lang/mdBook/issues/2564#issuecomment-2697842757, I would prefer to not check clippy in CI.

Additionally, this includes some incorrect changes (such as adding non_exhaustive which is a breaking change). Smaller PRs are welcome, though.

szabgab:
Sure I'll send smaller PRs. I don't understand, however, why is `non_exhaustive` a breaking change? I mean the tests pass. I'd be happy to try to write a test that shows this breakage, I just don't know what would that be.

PR #2567, opened 2025-03-04
[refactor] use !is_empty() instead of != ""

szabgab:


ehuss:
Thanks for the PR! However, I'm going to close as this looks a bit like a sideways change to me. They both read equally to me, and in those situations I'm inclined to just leave things as they are.

szabgab:
OK, but then maybe we should tell clippy to allow this https://rust-lang.github.io/rust-clippy/master/index.html#comparison_to_empty

PR #2570, opened 2025-03-05
toc.js: marking active page in toc when redirected to url without `.html` suffix

Pistonight:
## Issue
When serving the book through services that redirects `/foo/bar.html` -> `/foo/bar`, the page in TOC isn't marked as active because the toc uses an exact match with the `href` on the `a` tag.

This is a common behavior for services such as GH pages, and tools like `npx serve` (which actually redirects to paths without `.html` and serves the html anyway. in this case the active page never gets marked)

## Fix
Refactored the check for if a link is current page, and loosened the condition so the `.html` suffix is optional:

- `/foo` and `/foo.html` will match `/foo.html` link
- `/foo/`, `/foo/index`, `/foo/index.html` will match `/foo/index.html` link

If the active page is still not found, it will try to map `/foo` to `/foo/index.html`, which is the redirect behavior of some services/tools

## Testing
```
cargo run -- build test_book
npx serve test_book/book
```
Verified each link in the TOC now gets highlighted when switching to the page

### DIFF
--- a/src/front-end/templates/toc.js.hbs
+++ b/src/front-end/templates/toc.js.hbs
@@ -7,14 +7,54 @@ class MDBookSidebarScrollbox extends HTMLElement {
     constructor() {
         super();
     }
+    // Set the current, active page, and reveal it if it's hidden
     connectedCallback() {
+        // Helper to compare if 2 link paths are equal:
+        // - `/foo/`, `/foo/index`, `/foo/index.html` are considered equal
+        // - `/foo`, `/foo.html` are considered equal
+        function pathEquals(a, b) {
+            // safety null checks
+            if (!a && !b) { return true; }
+            if (!a || !b) { return false; }
+            // normalize paths
+            if (a.endsWith("/")) {
+                a += "index.html";
+            } else if (!a.endsWith(".html")) {
+                a += ".html";
+            }
+            if (b.endsWith("/")) {
+                b += "index.html";
+            } else if (!b.endsWith(".html")) {
+                b += ".html";
+            }
+            return a === b;
+        }
+
+        // Helper to set a link element in TOC to be active and reveal its parent sections
+        function setActivePage(link) {
+            link.classList.add("active");
+            var parent = link.parentElement;
+            if (parent && parent.classList.contains("chapter-item")) {
+                parent.classList.add("expanded");
+            }
+            while (parent) {
+                if (parent.tagName === "LI" && parent.previousElementSibling) {
+                    if (parent.previousElementSibling.classList.contains("chapter-item")) {
+                        parent.previousElementSibling.classList.add("expanded");
+                    }
+                }
+                parent = parent.parentElement;
+            }
+        }
+
         this.innerHTML = '{{#toc}}{{/toc}}';
-        // Set the current, active page, and reveal it if it's hidden
         let current_page = document.location.href.toString().split("#")[0].split("?")[0];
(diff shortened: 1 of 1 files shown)


szabgab:
I did not notice that Github pages **redirects** `page.html` to `page`, but I do see that if can serve the content of `page.html` if we access `page`.

I personally prefer to direct people to the nice URLs (without the `.html`) so currently I post-process the generated html files and the toc.js removing all the `.html` extensions.  In the future we might have a way to tell mdbook to use this linking method so I won't need to do the post processing, but it will mean that toc.js will also contain `href` entries without `.html`.

I tried this PR locally, both with the files generated by mdbook and the files after my post-processing. It seems that the highlighting works in all the cases.

One thing I find odd in the PR is that `current_page` is used in `isCurrentPage` without passing it as an argument. It works, but I am not sure I like it.
Related to this I think it would be cleaner if the code

```javascript
        this.innerHTML = '{{#toc}}{{/toc}}';
        // Set the current, active page, and reveal it if it's hidden
        let current_page = document.location.href.toString().split("#")[0];
        if (current_page.endsWith("/")) {
            current_page += "index.html";
        }
        var links = Array.prototype.slice.call(this.querySelectorAll("a"));
```

came after all the function definitions and we would not have

```
code
function definitions
code
```

but

```
function definitions
code
```

Pistonight:
> I did not notice that Github pages **redirects** `page.html` to `page`, but I do see that if can serve the content of `page.html` if we access `page`. Do I have to set some configuration option for that redirection to work?

My bad for not making that clear, gh pages behaves like what you said (serves html without the extension and without redirect)

> I tried this PR locally, both with the files generated by mdbook and the files after my post-processing. It seems that the highlighting works in all the cases.

Yep that's intended, this code should require no change to continue working if the `a` tags are changed in the future to not include the html extension

> One thing I find odd in the PR is that `current_page` is used in `isCurrentPage` without passing it as an argument. It works, but I am not sure I like it.

I will update that

Pistonight:
Addressed the comments, and I also found a bug where in the test book, the prefix chapter gets incorrectly highlighted for the root path because it's the first chapter, while the root path opens Introduction instead. In this case the first chapter is not aliased to because an `index.html` path exists. I fixed it by doing the aliasing logic in an extra pass in the end

Pistonight:
@ehuss are you able to review this?

Pistonight:
Resolved the conflict, please review :)

qaqland:
I have doubts about why the logic here isn't generated during SSG. Handling path strings directly with JS now doesn't feel very elegant.

Pistonight:
@qaqland this PR has been here for a long time and I don't really have time to continue working on this. If this is still an issue in the latest version (I haven't had time to upgrad my projects due to other stuff happening in life), I think someone can make an issue to track, then there could be a better way to fix with redesigning this part. However I did not investigate too deeply, and I was only trying to patch at the surface level

jhult:
Related PR: #3038

ehuss:
Thanks for the PR! I've decided to go with a simpler solution in https://github.com/rust-lang/mdBook/pull/3028. Although this PR tries to handle cases like `/foo/index.html` redirecting to `/foo`, those kinds of redirects break the navigation (particularly for relative links). For now, I'd prefer to not try to handle that. There may be future extensions like #3047 which would provide first-class support for clean urls, which would hopefully deal with redirects by avoiding them.

Issue #2493, opened 2024-11-23
Can warnings be promoted to errors?


freeform-andre:
### Question

I made a change to my `book.toml` and typo'ed the `command` section of a preprocessor I was adding resulting in the command not being found. I assumed this would lead to an error but it didn't. Is it possible to promote these (and any) warnings to an error?

Thanks!

### Version

```text
0.4.42
```

freeform-andre:
@ehuss friendly ping on this question 😅

szabgab:
Could you provide an example that will generate a warning now and that you'd like to make it an error?

freeform-andre:
Providing `[preprocessor.*.command]` that points to a binary that doesn't exist
```
[preprocessor.mermaid]
command = "not-a-real-binary"
````

```
2025-03-06 05:29:06 [WARN] (mdbook::preprocess::cmd): The command wasn't found, is the "mermaid" preprocessor installed?
2025-03-06 05:29:06 [WARN] (mdbook::preprocess::cmd): 	Command: not-a-real-binary
```
I would consider this to have broken my docs but it's just a warining. I'd rather all warnings be treated as errors.

szabgab:
Oh, that's very different from what I thought.

As far as I understand this is considered only a warning in order to let `mdbook` work even if the given preprocessor is "not supported". For example if it is not installed.

The [documentaion](https://rust-lang.github.io/mdBook/format/configuration/preprocessors.html) does not actually indicate if configured preprocessors are required or if they are optional, but right now they are optional.

Changing this might be considered as a breaking change, but I don't see why would we want to allow `mdbook` to run if one of the preprocessors is not available. In other words I am in favor of implementing the change suggested in this issue.

freeform-andre:
I think this would be a very nice change to make. The warnings are generally ignored and I only find out something is broken when an end user tells me about it. If a plugin is added to the book, it should be required to run successfully or be removed.

freeform-andre:
> If it's just about changing the behavior to exit the application if a renderer is missing, I already have a patch that I can submit.

I know I would appreciate this patch!

Bktero:
Should I post the diff here or create a PR?

szabgab:
@Bktero create a PR and mention this issue

Bktero:
The PR is here https://github.com/rust-lang/mdBook/pull/2597

cusma:
[...]

Issue #1595, opened 2021-07-04
Warn on unknown configuration settings


ehuss:
When there are unknown settings in `book.toml`, mdbook just ignores them:

```toml
[book]
title = "foo"
abc = 123
```

This has occasionally caused problems and confusion for people when they make typos or mistakes.  I think it would be good for unknown options to issue a warning. [`serde_ignored`](https://crates.io/crates/serde_ignored) provides a facility for catching these unknown fields.

Some care will need to be taken to consider things like plugins which can add their own sections to the config.

joshrotenberg:
I'm looking into this a bit. 

`serde_ignored` works great when the top level struct (and any children) just derive `Deserialize`, but in our case `Config` has a custom deserialize implementation to, if I'm reading it correctly, handle the legacy `book.toml` format. It then goes on to pull the `book`, `build` and `rust` sections out and explicitly deserialize them with `try_into` and then finally builds the `Config` and tacks on the `rest` for anything left over. 

Here are some options:

1. If handling the legacy format is the only reason for the custom deserialization, a hard deprecation on that format would otherwise simplify the `Config` deserialization quite a bit. I haven't been around long enough to know what the ramifications of this would be. Looking briefly at the blame for `config.rs`, it looks like some of the legacy handling code was adding 4 years ago.

1. Pull the legacy format detection out of the `deserialize` somehow, try it first in `from_str` and `if_not_legacy` or whatever, do normal `derive(Deserialize)`.

1. Implement the `serde_ignored` method for the three known configs within `Confg`'s `deserialize` call individually. 

1. Make unknown fields an error in the three built in configs by adding `[serde(deny_unknown_fields)]` to them and skip `serde_ignored` altogether.

> Some care will need to be taken to consider things like plugins which can add their own sections to the config.

This makes me think the last option (also the easiest) makes sense since we have to have a `rest` field in this case anyway, and we have no way of validating other plugin's config aside from highly suggesting the implementor follow a similar pattern.

I'm happy to code up any (or all, for comparison) of the above scenarios if that would help.

ehuss:
I think removing the legacy support would be fine, though that would need to be deferred to 0.5.  

I don't think we can change them to hard errors as that would be a breaking change, so I think that would need to be deferred to 0.5, too.  I'm not sure if it makes sense to keep them as warnings or not long term, though.

[...]

PR #2572, opened 2025-03-06
deny_unknown_fields in the configuration file

szabgab:
To avoid typos and make the config reading stricter.

See #2493

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -899,7 +899,7 @@ mod tests {
     fn fail_on_invalid_config_field() {
         let src = r#"
         [book]
-        password = Secret
+        password = "Secret"
         "#;
 
         Config::from_str(src).unwrap();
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -894,6 +894,17 @@ mod tests {
...
+    #[test]
+    #[should_panic(expected = "Invalid configuration file")]
+    fn fail_on_invalid_config_field() {
+        let src = r#"
+        [book]
+        password = Secret
+        "#;
+
+        Config::from_str(src).unwrap();
+    }
```
ehuss, line 906: I don't think this test is actually checking the change in this PR. It seems to be checking that the TOML syntax is invalid.

ehuss (commented):
Thanks for the PR! However, I don't think we can go directly to making this an error. See https://github.com/rust-lang/mdBook/issues/1595 for a discussion about how this can be approached and some of the complexities.

ehuss:
Closing as this is now pulled into https://github.com/rust-lang/mdBook/pull/2787. Thanks!

PR #2576, opened 2025-03-08
Add "Auto" theme selection which switches between default light and dark theme automatically

### LINKED ISSUE #1504: [Feature request] Add a theme option like "auto" to reset the theme to auto dark mode
If change the theme manually, it can't be reset to auto dark mode that request the dark version of the site via the `prefers-color-scheme` CSS media query. 

The desktop browser can reset the theme by manually delete the local storage which is really troublesome. Also, many mobile browsers can't just delete a specific website data. It should add a theme option like "auto" to reset the theme to auto dark mode.

tmandry:
The first commit is pretty uncontroversial. It takes the existing behavior for auto-detecting light/dark mode and improves it so that the page automatically changes when the preferred mode changes without a refresh. This is useful on systems which have a scheduled light/dark mode change, which is a built-in option on many OSes.

The problem is that this functionality can only be accessed if the user has never selected a theme before. To allow restoring the default behavior, we now add an "Auto" theme menu option in the second commit.

Since some people customize their index.hbs we make both changes resilient to that, falling back to the old behavior if the index changes are not applied. Thankfully this is pretty simple to do. (Longer term it'd be nice to remove the need to do this, e.g. for custom themes... that's #2594 or #2059.)

Fixes #1504

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,7 +10,7 @@
         "require": "readonly"
     },
     "parserOptions": {
-        "ecmaVersion": 2018,
+        "ecmaVersion": 2021,
         "requireConfigFile": false,
         "sourceType": "module"
     },
--- a/src/theme/book.js
+++ b/src/theme/book.js
@@ -1,6 +1,6 @@
 'use strict';
 
-/* global default_theme, hljs, ClipboardJS */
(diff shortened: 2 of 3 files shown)


ehuss (approved):
Thanks!

PR #2579, opened 2025-03-09
[test] Check content of error message

szabgab:
testing error message of ask_the_preprocessor_to_blow_up

Apparently the `output.status` is different on windows and the rest of the OSes. That's why we need the `cfg!(windows)`
We could replace `output.status` by `output.status.code().unwrap_or(-1)` or by `output.status.code().unwrap()` thereby making sure the error message is the same on all (currently tested) OSes.


ehuss (approved):
Thanks!

PR #2582, opened 2025-03-09
[test] Check content of error message

szabgab:
in failure_on_missing_theme_directory

### DIFF
--- a/tests/rendered_output.rs
+++ b/tests/rendered_output.rs
@@ -732,6 +732,8 @@ fn failure_on_missing_theme_directory() {
     let md = MDBook::load(temp.path()).unwrap();
     let got = md.build();
     assert!(got.is_err());
+    let error_message = got.err().unwrap().to_string();
+    assert_eq!(error_message, "Rendering failed");
 }
 
 #[cfg(feature = "search")]


tests/rendered_output.rs
```diff
@@ -732,6 +732,8 @@ fn failure_on_missing_theme_directory() {
     let md = MDBook::load(temp.path()).unwrap();
     let got = md.build();
     assert!(got.is_err());
+    let error_message = got.err().unwrap().to_string();
+    assert_eq!(error_message, "Rendering failed");
```
Rabbival, line 736: I do find it a little offsetting that you check the string literal within the message, meaning that if someone happens to change a single char there in the future the test would fail. 
Also, and that one is more of a matter of personal taste, you might get a more informative test (and a slightly cleaner one) if you wrote it like:
```
match got{
    Err(e) => { assert_eq1(e.to_string(), "Rendering failed"); },
    Ok(content) => { panic!("Expected an error but got an Ok with content: {:?}", content); }
}
```

ehuss:
I'm going to close as superseded by https://github.com/rust-lang/mdBook/pull/2676, which is able to validate the entire output.

PR #2594, opened 2025-03-14
Support additional themes in config

### LINKED ISSUE #2048: [Feature Request] Entrypoint to add additional themes ALONGSIDE the default (rather than overwriting/replacing them)

tmandry:
This removes the need for replacing all of index.hbs just to have a custom theme.

Closes #2048.

### DIFF
--- a/guide/src/format/configuration/renderers.md
+++ b/guide/src/format/configuration/renderers.md
@@ -100,6 +100,7 @@ preferred-dark-theme = "navy"
 smart-punctuation = true
 mathjax-support = false
 copy-fonts = true
+additional-themes = [{ name = "Midnight Blue", class = "midnight" }]
 additional-css = ["custom.css", "custom2.css"]
 additional-js = ["custom.js"]
 no-section-label = false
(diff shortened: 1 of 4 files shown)


ehuss:
Some drive-by comments: 

One issue we've had is that the `id` generated for the theme selector can conflict with id's used in headers and other places. For example, you can't have a header like `## Rust` because it conflicts with the Rust color scheme. I don't know how we can better handle that, and I worry a little that supporting additional themes will make it harder to fix that (or make collisions more likely). That's probably not a blocker, but I figured I'd bring it up since I worry about it.

I'd also like to understand a little more how other generators work. #2048 discusses that a little, but I haven't had time to read and understand. I'm just trying to understand what the tradeoffs or alternate designs are here.

I also wonder what if the user wants to remove a stock theme, how do they do that?

I assume this works with all the options in `output.html` like `default-theme` and `preferred-dark-theme`?

I've been slightly uncomfortable that there is some muddled terminology here. "theme" in mdbook is a directory that allows you to completely customize how a page is rendered. Then there is also a "color theme" which is just a css class and a selector. The two are only loosely releated, and I worry about the potential confusion around them. I don't know why it was structured that way.

Otherwise seems like it would be fun to support this.

PR #2597, opened 2025-03-16
Exit if a preprocessor command is not found

Bktero:
A small change discussed in issue https://github.com/rust-lang/mdBook/issues/2493

When a command is not found, log an error and exit the application instead of simply logging a warning. Indeed, the generator of the book will probably be wrong anyway, so exiting seems legit.

### DIFF
--- a/examples/nop-preprocessor.rs
+++ b/examples/nop-preprocessor.rs
@@ -57,7 +57,7 @@ fn handle_supports(pre: &dyn Preprocessor, sub_args: &ArgMatches) -> ! {
     let renderer = sub_args
         .get_one::<String>("renderer")
         .expect("Required argument");
-    let supported = pre.supports_renderer(renderer);
+    let supported = pre.supports_renderer(renderer, false);
 
     // Signal whether the renderer is supported by exiting with 1 or 0.
     if supported {
@@ -99,7 +99,7 @@ mod nop_lib {
             Ok(book)
         }
 
-        fn supports_renderer(&self, renderer: &str) -> bool {
+        fn supports_renderer(&self, renderer: &str, _error_on_missing_preprocessor: bool) -> bool {
             renderer != "not-supported"
         }
     }
--- a/src/book/mod.rs
+++ b/src/book/mod.rs
@@ -605,7 +605,8 @@ fn preprocessor_should_run(
 ) -> bool {
     // default preprocessors should be run by default (if supported)
     if cfg.build.use_default_preprocessors && is_default_preprocessor(preprocessor) {
-        return preprocessor.supports_renderer(renderer.name());
+        return preprocessor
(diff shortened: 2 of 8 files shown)


szabgab:
I think we could add an optional field as @ehuss suggested that would turn these warnings into errors. By default it would be `false`, but when a new book is created with `init` then it would be added to the `book.toml` file and it would be `true`.
Suggested name of the field: `error_on_missing_preprocessor`

Bktero:
I have added a commit that tries to implement this flag. Is this something like this you had in mind?

There are still 2 points that I need to figure out:;

1. The flag isn't written to book.toml by the `init` command.
2. The flag is considered to be `true` is missing from the file.

Bktero:
I think I figured it out. I added another commit. Here is how I tested it.

First I created book:

```shell
$ ./target/debug/mdbook init my-new-book    

Do you want a .gitignore to be created? (y/n)
n
What title would you like to give the book? 
test
2025-03-22 14:56:11 [INFO] (mdbook::book::init): Creating a new book with stub content

All done, no errors...
```

The following `book.toml` was generated:

```toml
[book]
authors = ["Pierre Gradot"]
language = "en"
multilingual = false
src = "src"
title = "test"

[build]
build-dir = "book"
create-missing = true
error-on-missing-preprocessor = true
extra-watch-dirs = []
use-default-preprocessors = true
```

I add this at the end of te file (as in the linked issue):

```toml
[preprocessor.mermaid]
command = "not-a-real-binary"
```

I built the book and got an error:

```shell
$ ./target/debug/mdbook build my-new-book
2025-03-22 14:58:04 [INFO] (mdbook::book): Book building has started
2025-03-22 14:58:04 [ERROR] (mdbook::preprocess::cmd): The command "not-a-real-binary" wasn't found, is the "mermaid" preprocessor installed?
```

I changed the flag to `false` and I built again:

```shell
$ ./target/debug/mdbook build my-new-book
2025-03-22 14:58:19 [INFO] (mdbook::book): Book building has started
2025-03-22 14:58:19 [WARN] (mdbook::preprocess::cmd): The command "not-a-real-binary" wasn't found, is the "mermaid" preprocessor installed?
2025-03-22 14:58:19 [INFO] (mdbook::book): Running the html backend
```

Bktero:
I have synchronized my fork with the `rust-lang`'s repository and resolved the conflicts.

Any opinions on this PR?

Bktero:
Sorry for the noise @ehuss and @szabgab : did you get a chance to take a look at this PR?

janbridley:
Bumping for visibility - if there is anything else that needs to be done for this PR I'd be happy to jump in.

ehuss:
Closing in favor of https://github.com/rust-lang/mdBook/pull/2797.  Thanks!

Issue #2601, opened 2025-03-20
GUI tests stopped running in the CI


szabgab:
### Problem

After this commit fc7ef59dee00d0af6b675bc4c0aba74acc71450a was merged the GUI test in the CI started to skip all tests and started to report "No test found".

* Tests still running: https://github.com/rust-lang/mdBook/actions/runs/13763368023/job/38484205244
* No test found https://github.com/rust-lang/mdBook/actions/runs/13764711522/job/38488401535

### Possible Solution(s)

Besides fixing this, IMHO the GUI test should fail if "No test found".

ehuss:
Oh my, thanks for catching that!

cc @GuillaumeGomez

GuillaumeGomez:
Arf indeed. Sending a fix right away!

szabgab:
I just managed to fix it. Please review it.

GuillaumeGomez:
We wrote exactly the same fix, so I guess we're all good. Didn't see you commented.

ehuss:
Closing as fixed by https://github.com/rust-lang/mdBook/pull/2602

PR #2602, opened 2025-03-20
Select all the GUI tests if no filter was provided

szabgab:
The name of executable was taken as a filter and because of that nothing was selected by default.

Fixing #2601


ehuss (approved):
Thanks!

Issue #2607, opened 2025-03-21
Navigation - help with keyboard shortcuts


szabgab:
### Problem

In anticipation of allowing book authors do define / enable more keyboard shortcuts and even without that it would be a good idea to make
it easy for readers to find out the available keyboard shortcuts. AFAIK the button to activate this is usually `?`, though we might want to also add a help icon to the menu-bar.

See also #1408

### Proposed Solution

Let's start with a simple overlay with some fixed text and then make it based on the local configuration.

PR #2608, opened 2025-03-21
Add pop-up window showing the keyboard shortcuts

szabgab:
Make it display when the user presses `?`.

Implements #2607

### DIFF
--- a/src/front-end/css/chrome.css
+++ b/src/front-end/css/chrome.css
@@ -656,3 +656,46 @@ html:not(.sidebar-resizing) .sidebar {
     margin-inline-start: -14px;
     width: 14px;
 }
+
+/* The container for the help popup that covers the whole window. */
+#mdbook-help-container {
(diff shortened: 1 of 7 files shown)


ehuss:
Thanks! Would it be possible to update the styling for some of the dark themes? Right now it is unreadable there:

image

Also, can you resolve the conflicts?

shenef:
How would a regular user know about the `?` shortcut without seeing the list of shortcuts?

Looking at the code, it seems like the shortcut is only active when <kbd>shift</kbd> is being held. On custom keyboard layouts the user might not have to press <kbd>shift</kbd> to write `?` which in turn would make the shortcut inaccessible. (specifying "custom" here cause I, at a glance, wasn't able to find a language/region specific layout where that is the case)

ehuss (approved):
Thanks! I pushed some updates to adjust some of the styling and layout.

PR #2613, opened 2025-03-22
Avoid using the same file twice in SUMMARY.md

### LINKED ISSUE #2612: Reject duplicate entries in SUMMARY.md
### Problem

If the same `md` file is mentioned more than once in the `SUMMARY.md` they will still generate the same `html` file so the `prev` and `next` links will lead to the same page. From the point of view of the hierarchy this will seen to be "broken".
See #1629

```
- [A](./a.md)
- [B](./b.md)
- [A](./a.md)
- [C](./c.md)
```

Should the `next` link in `a.html` lead to B or C? Both will be considered broken.

### Proposed Solution

IMHO mdbook should report an error if the same file is mentioned twice.

### Notes

_No response_

szabgab:
Fixes #2612

### DIFF
--- a/src/book/summary.rs
+++ b/src/book/summary.rs
@@ -3,6 +3,7 @@ use log::{debug, trace, warn};
 use memchr::Memchr;
 use pulldown_cmark::{DefaultBrokenLinkCallback, Event, HeadingLevel, Tag, TagEnd};
 use serde::{Deserialize, Serialize};
+use std::collections::HashSet;
 use std::fmt::{self, Display, Formatter};
 use std::ops::{Deref, DerefMut};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


smilingpoplar:
I need to have duplicate files in `SUMMARY.md`. Can we make it an option?

szabgab:
@smilingpoplar Could you please comment on #2634 explaining your use-case for duplicate files?

smilingpoplar:
> @smilingpoplar Could you please comment on #2634 explaining your use-case for duplicate files?

https://github.com/rust-lang/mdBook/issues/2634#issuecomment-2777572930

Hoffenbar:
**I need an option to allow duplicate names, too**.  Additionaly, I'm not referencing the same file twice, only same named files **but in different locations**. Nevertheless, this error is generated:
```
2025-04-11 16:25:37 [ERROR] (mdbook::utils): Error: Summary parsing failed for file="d:\\data\\mdbook\\book1-src\\src\\SUMMARY.md"
2025-04-11 16:25:37 [ERROR] (mdbook::utils): 	Caused By: Duplicate file in SUMMARY.md: "dummy.md"
```
Switching back to 0.4.47 meanwhile. This changes breaks compatibility with existing source trees!

PR #2617, opened 2025-03-23
Fix `mode-rust.js` load error

GuillaumeGomez:
I was intrigued at the still failing `mode-rust.js` load. After some investigation, it's because ours has a hash at the end of its name, which `ace.js` doesn't know. So instead, I simply force it to load the correct name directly.

I'm not a big fan of this approach tough because this is a modification in minified JS which will need to be done again if we update `ace.js`. The best solution would be for us to be able to tell `ace.js` to load the file with a hash, which doesn't seem to be possible. So for now I think we're stuck with it.


src/theme/playground_editor/ace.js
```diff
@@ -25,7 +25,7 @@ ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
...
...+l+"px;"+"top:"+u+"px;"+"left:"+s+"px;"+(i||""))}o=(t.end.row-t.start.row-1)*r.lineHeight;if(o<=0)return;u=this.$getTop(t.start.row+1,r);var c=(t.start.column?1:0)|(t.end.column?0:8);this.elt(n+(c?" ace_br"+c:""),"height:"+o+"px;"+"right:0;"+"top:"+u+"px;"+"left:"+s+"px;"+(i||""))},this.drawSingleLineMarker=function(e,t,n,r,i,s){if(this.session.$bidiHandler.isBidiRow(t.start.row))return this.drawBidiSingleLineMarker(e,t,n,r,i,s);var o=r.lineHeight,u=(t.end.column+(i||0)-t.start.column)*r.characterWidth,a=this.$getTop(t.start.row,r),f=this.$padding+t.start.column*r.characterWidth;this.elt(n,"height:"+o+"px;"+"width:"+u+"px;"+"top:"+a+"px;"+"left:"+f+"px;"+(s||""))},this.drawBidiSingleLineMarker=function(e,t,n,r,i,s){var o=r.lineHeight,u=this.$getTop(t.start.row,r),a=this.$padding,f=this.session.$bidiHandler.getSelections(t.start.column,t.end.column);f.forEach(function(e){this.elt(n,"height:"+o+"px;"+"width:"+e.width+(i||0)+"px;"+"top:"+u+"px;"+"left:"+(a+e.left)+"px;"+(s||""))},this)},this.drawFullLineMarker=function(e,t,n,r,i){var s=this.$getTop(t.start.row,r),o=r.lineHeight;t.start.row!=t.end.row&&(o+=this.$getTop(t.end.row,r)-s),this.elt(n,"height:"+o+"px;"+"top:"+s+"px;"+"left:0
```
traviscross, line 28: There must be a better way than this.  I'd even take the compilation step doing a search/replace on the minified file as being an improvement over editing it like this.

traviscross: Also, `mode-rust.js` is tiny.  Could we not just inline it somewhere and avoid loading it at all?

notriddle:
If we need to fork ace.js, can we at least vendor an unminified copy?

traviscross:
In discussion, @ehuss mentioned perhaps the simplest immediate fix, which would be to just turn off the hashing for this one file.

ehuss:
Closing as I believe the correct solution is in https://github.com/rust-lang/mdBook/pull/2710. Thanks!

PR #2620, opened 2025-03-26
Added a way to hook into the serve process.

### LINKED ISSUE #2573: Add a way to ignore the source link file present in the SUMMARY.md while including the link.
### Problem

I want to add an auto generated part of a documentation that I am writing but the tool I am using generates HTML so i want to have a link to the page generated by the tool in the sidebar. Problem the book won't build because the source doesn't exist.

### Proposed Solution

Solution 1: Add a syntax to the SUMMARY parser to make it so it will ignore the file behind the link but generate the the link . 

Solution 2: If the link is not a `md` file it inserts the link into the sidebar without checking for a file in the source directory and leave it to the user to unsure the file at the link destination is present in the output directory.

### Notes

_No response_

AdrienDerobert:
Closes #2573
Added a way to hook into the serve process after the build.

```bash
mdbook serve -c <post build command>
# or
mdbook serve --post-build <post build command>
```

Tested on:
- [x] Linux
- [ ] Windows
- [ ] MacOs

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -10,12 +10,22 @@ use mdbook::utils::fs::get_404_output_file;
 use mdbook::MDBook;
 use std::net::{SocketAddr, ToSocketAddrs};
 use std::path::PathBuf;
+use std::process::Stdio;
 use tokio::sync::broadcast;
 use warp::ws::Message;
 use warp::Filter;
 
 /// The HTTP endpoint for the websocket used to trigger reloads when a file changes.
 const LIVE_RELOAD_ENDPOINT: &str = "__livereload";
+
+#[cfg(target_family = "unix")]
+const LAUNCH_SHELL_COMMAND: &str = "sh";
+#[cfg(target_family = "unix")]
+const LAUNCH_SHELL_FLAG: &str = "-c";
+#[cfg(target_family = "windows")]
+const LAUNCH_SHELL_COMMAND: &str = "cmd";
+#[cfg(target_family = "windows")]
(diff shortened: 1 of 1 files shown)


AdrienDerobert:
would close #2573

szabgab:
Thanks for the PR.

I also think we need a way to do this, but I thought it should be part of the configuration in the `book.toml` as it is more permanent. In which case would you prefer it to be a command-line flag?

In any case it would be better if there was a test that runs in the CI to check this feature.

AdrienDerobert:
I can add a configuration option for it in `book.toml` if you prefer. I was thinking on the easiest and quickest way to develop the feature and the cli flag for `serve` seemed the simplest to go. As for build you could easily run the build and your command/script right after. But if you prefer i can work a bit more on it to make it part of the configuration and add a test to the ci.

Just to make sure if it is in the `book.toml` it should also run in `build`. 
Am i right ?

AdrienDerobert:
Another problem with including this in the book toml config is that the command might be different for a windows user and a user of a posix system. Do I introduce a section per platform to define the command ?

AdrienDerobert:
Hello did you have any indications on how to progress this forward what would be the best course of action ? Do i add a new configuration option or only the command line argument flag would suffice.

In my opinion the command line argument is enough because the feature is only needed for the `serve` sub command in order to hook into the watch and rebuild mechanism. But for build nothing prevents you from running the commands after the build yourself. 

Or is it expected the build command to produce the final book with the post processing (eg: inclusion of extra HTML files) ?

PR #2621, opened 2025-03-28
elevate the log level of calling a preprocessor to info

szabgab:
To be the same as for the backends.

### DIFF
--- a/src/book/mod.rs
+++ b/src/book/mod.rs
@@ -212,7 +212,7 @@ impl MDBook {
         let mut preprocessed_book = self.book.clone();
         for preprocessor in &self.preprocessors {
             if preprocessor_should_run(&**preprocessor, renderer, &self.config) {
-                debug!("Running the {} preprocessor.", preprocessor.name());
+                info!("Running the {} preprocessor.", preprocessor.name());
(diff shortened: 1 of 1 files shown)


ehuss:
Hm, I'm a little reluctant to change this since I think mdbook is already too noisy, and the important things can be lost in the output. Can you say more about what kind of scenarios where this would be useful to know?

szabgab:
It is strange to me that you say it is too noisy, when I run `mdbook build` I only get two lines of output:

```
2025-04-01 07:39:15 [INFO] (mdbook::book): Book building has started
2025-04-01 07:39:15 [INFO] (mdbook::book): Running the html backend
```

Making the above change would add 2 more lines to this output.
Of course when there are more preprocessors and more output generators then the numbers multiply and when I run `mdbook serve` then I see all these lines on every change, so you might be right that it gets too noisy.

I needed this change when I was trying to figure out what's going on in a book that had

```toml
[preprocess.links]
```

(instead of `[preprocessor.links]`)

I also felt the level of logging should be the same for all the plugins.

Maybe we need to demote the "Running the html backend" do debug and instead of that add a "Book building has finished". And maybe we need some way to see when the plugins run without seeing all the "debug" messages.

PR #2622, opened 2025-03-28
warn on invalid fields in the root of book.toml

szabgab:
The first step for  #1595


ehuss (approved):
Thanks!

szabgab:
Thanks for accepting this!

PR #2625, opened 2025-03-29
Simplify resources location

GuillaumeGomez:
Now resources are grouped in folders based on their type. I'm planning to send a follow-up to remove all rust files from these folders as I think they should not be mixed.


GuillaumeGomez:
Fixed fmt. :)

ehuss:
I think this makes sense, but can you include some more information about why you want to make this change? Is it just an organizational preference to avoid mixing rust code with the theme files?

Can you resolve the conflicts?

GuillaumeGomez:
The new organization is much easier to navigate: you can find files grouped by kind instead of having to look for the ones that you actually want. So not just about removing the mix of Rust and other languages, just easier overall.

And yes, rebasing right away.

ehuss (approved):
Thanks!

PR #2627, opened 2025-03-31
test the command line without any parameters #1568

szabgab:


ehuss (approved):
Thanks!

Eventually I think we should move to something like `snapbox` where we can have automated snapshotting (so we don't have to update these stderr/stdout strings).

PR #2628, opened 2025-03-31
use empty folder; check more details in result #1568

szabgab:


szabgab:
@ehuss It would be nice if you could accept this so I can start writing other tests based on the empty temporary folder.

tests/dummy_book/mod.rs
```diff
@@ -38,12 +38,16 @@ impl DummyBook {
         self
     }
 
-    /// Write a book to a temporary directory using the provided settings.
-    pub fn build(&self) -> Result<TempDir> {
-        let temp = TempFileBuilder::new()
+    pub fn empty(&self) -> Result<TempDir> {
```
ehuss, line 41: This seems a little out of place, since it is no longer actually building the dummy book. That is, this method doesn't seem associated with the class DummyBook. Is there a reason that a test that needs an empty directory can't just call `TempFileBuilder` directly? That is done in quite a few other places. If we really need to shorten that, we can perhaps find a way to share it.

tests/cli/init.rs
```diff
@@ -6,21 +9,32 @@ use mdbook::config::Config;
...
+    assert_eq!(config.book.language, Some(String::from("en")));
+    assert_eq!(config.book.multilingual, false);
+    assert_eq!(config.book.src, PathBuf::from("src"));
 
     assert!(!temp.path().join(".gitignore").exists());
+    let summary = read_to_string(temp.path().join("src").join("SUMMARY.md")).unwrap();
+    assert_eq!(summary, "# Summary\n\n- [Chapter 1](./chapter_1.md)\n");
+    let chapter_1 = read_to_string(temp.path().join("src").join("chapter_1.md")).unwrap();
+    assert_eq!(chapter_1, "# Chapter 1\n");
+    assert!(temp.path().join("book").exists());
```
ehuss, line 37: Since this test is mostly about skipping prompts (and verifying the result of those skipped prompts), perhaps these should be moved to `tests/init.rs`?

tests/dummy_book/mod.rs
szabgab: Now creating the temporary directory directly.

tests/cli/init.rs
szabgab: I can move it there, but this is still a cli-based test so shouldn't it stay in the cli folder?

ehuss:
I'm going to close as superseded by https://github.com/rust-lang/mdBook/pull/2676, which now validates the full output of the init command and the files it generates.

PR #2630, opened 2025-04-02
Serve on a random port when user provides --port 0

### LINKED ISSUE #2599: Cannot use serve --open with --port 0
### Problem

When using the `mdbook serve` command with `--port 0` and `--open` option, the opened page is wrong, as the port is `0` instead of the random port attributed by the system.

### Steps

1. `mdbook init abc --force`
2. `mdbook serve abc --port 0 --open`

### Possible Solution(s)

The `serving_url` should contain the bound port instead of the asked port (0)

https://github.com/rust-lang/mdBook/blob/3a8faba6451d4513b059a5711829b7da10ddcfdc/src/cmd/serve.rs#L92

Maybe using something like [`bind_ephemeral`](https://docs.rs/warp/latest/warp/struct.Server.html#method.bind_ephemeral) could be used. It would maybe require to move the creation of the warp server in a separate function run before the thread spawn.

https://github.com/rust-lang/mdBook/blob/3a8faba6451d4513b059a5711829b7da10ddcfdc/src/cmd/serve.rs#L88-L90

### Notes

_No response_

### Version

```text
mdbook v0.4.47
```

szabgab:
Closes #2599

Manually tested on Linux.

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -54,7 +54,13 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
     let hostname = args.get_one::<String>("hostname").unwrap();
     let open_browser = args.get_flag("open");
 
-    let address = format!("{hostname}:{port}");
+    let mut address = format!("{hostname}:{port}");
+
+    if port == "0" {
+        let listener = std::net::TcpListener::bind(address).unwrap();
+        let port = listener.local_addr().unwrap().port();
+        address = format!("{hostname}:{port}");
+    }
 
     let update_config = |book: &mut MDBook| {
         book.config


src/cmd/serve.rs
```diff
@@ -54,7 +54,13 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
...
     let open_browser = args.get_flag("open");
 
-    let address = format!("{hostname}:{port}");
+    let mut address = format!("{hostname}:{port}");
+
+    if port == "0" {
+        let listener = std::net::TcpListener::bind(address).unwrap();
+        let port = listener.local_addr().unwrap().port();
+        address = format!("{hostname}:{port}");
+    }
```
GuillaumeGomez, line 63: Why not doing this when starting the server?

PR #2632, opened 2025-04-02
When running serve check if port is available

### LINKED ISSUE #2631: mdbook serve --port 80 --open panics, but opens the browser
### Problem

When we try to use a restricted port between 1-1023 we get a panic and the browser is still opened and reports a missing service.

```
mdbook serve test_book --open --port 83
```

I tried a few ports.

The same if we already have a book served on some good port (e.g. the default 3000) and we start a new book one the same port.
mdbook panics and opens the browser on that port, but that shows the other book.

### Steps

For the first one just run

```
mdbook serve test_book --open --port 83
```

For the second problem in one window run
```
mdbook serve test_book
```

And then in another window run

```
mdbook serve test_book --open
```

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
Both `mdbook v0.4.47` and what we have in `master`.
```

szabgab:
Give a nice explanation if it is used or if it is a protected port.

Closes #2631

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -69,6 +69,12 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
     update_config(&mut book);
     book.build()?;
 
+    let listener = std::net::TcpListener::bind(&address);
+    if listener.is_err() {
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Wouldn't it be better to do it where we start the server instead of doing an extra check?

PR #2633, opened 2025-04-02
Speed up search index loading

GuillaumeGomez:
Follow-up of https://github.com/rust-lang/mdBook/pull/2552.

`JSON.parse()` is much faster than parsing JS code for JS engines. I missed this case when I removed the original JSON file. This is also how we do in rustdoc. The bigger the search index, the more noticeable it is.

As for the second commit, it fixes this JS error:

Screenshot From 2025-04-02 11-14-15

@ehuss Just to be sure: does this PR solves your 10+ seconds issue for search load?

### DIFF
--- a/src/front-end/searcher/searcher.js
+++ b/src/front-end/searcher/searcher.js
@@ -3,7 +3,7 @@
 /* global Mark, elasticlunr, path_to_root */
 
 window.search = window.search || {};
-(function search(search) {
+(function search() {
     // Search functionality
     //
     // You can use !hasFocus() to prevent keyhandling in your key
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
Forgot to update the tests for loading the new JSON format, it's done fixed.

GuillaumeGomez:
@ehuss Did you have time to take a look? Do you need some extra explanations on the changes maybe?

ehuss (approved):
Thanks! I haven't done any performance tests, but I assume it should be ok.

> Just to be sure: does this PR solves your 10+ seconds issue for search load?

No, the concern was just around a low-bandwidth network connection trying to load a very large index. That's purely a network constraint.

Issue #2636, opened 2025-04-03
What to do with the book.multilingual field?


szabgab:
### Question

The code here 
https://github.com/rust-lang/mdBook/blob/4f698f813ce667d8d415434e23af87d731cdb29d/src/config.rs#L424
expects a field called `multilingual` in the `[book]` section of `book.toml`.

The `mdbook init` command even ads it to the `book.toml` file.

However, this field does not seem to be used anywhere in the code and it is not mentioned in the [documentation](https://rust-lang.github.io/mdBook/format/configuration/general.html#general-metadata) either.

Shall we remove it from the code and start warning about its presence?

ehuss:
Yea, I have no idea what that field is for, and it doesn't look like it was ever used. I'd be fine removing it.

PR #2637, opened 2025-04-03
Warn on invalid fields in the `[book]` section of `book.toml`

szabgab:
A step for #1595.

We might need to revise this if de decide to remove the `multilingual` field as described in #2636

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -313,6 +313,7 @@ impl<'de> serde::Deserialize<'de> for Config {
         }
 
         warn_on_invalid_fields(&raw);
+        warn_on_invalid_fields_in_book_section(&raw);
 
         use serde::de::Error;
         let mut table = match raw {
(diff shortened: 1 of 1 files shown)


ehuss:
Is it possible to implement this with serde_ignored?

szabgab:
I'll give it a try, though I thought this whole warning thing is a temporary solution to give the users the opportunity to fix their configuration before the strict checking is turned on.

szabgab:
I looked at `serde_ignored` and it is not clear to me how to combine it with the custom code we have to handle the legacy format. 
I'd prefer you integrate this change as it is.

Then, if you agree, I can send a PR to handle #2653 

Then I'll convert the whole thing to use `serde_ignored`.

ehuss:
Closing as I have decided to go ahead and just error on these fields instead of warn. We could potentially decide to switch it to a warning in the future if it ends up being an issue. Closing for https://github.com/rust-lang/mdBook/pull/2787. Thanks for the PR, though!

PR #2644, opened 2025-04-06
Mark more font files as binary

kg4zow:
Not having these files marked as binary causes problems for older versions of git (like 1.8.3.1 on CentOS/RHEL 7).


szabgab:
Looks like a reasonable change to me.

szabgab (approved):

ehuss (approved):
Thanks!

Issue #2639, opened 2025-04-04
Update highlight.js - support syntax highlighting for more languages


szabgab:
There are many issues and even PRs asking to add syntax highlighting support to other languages. Assuming we keep using [highlight.js](https://highlightjs.org/) the only way to accomplish that is to upgrade highlight.js and include more language. However each addition language will increase the size of the `highlight.js` file while most books only need support for a very limited number of languages and formats. (e.g. unlikely to have a book that has code snippets in rust, ruby, scala, and swift. Except the `test_book` of mdbook)

Currently highlight.js is 137,537. If we download one that includes support for all the languages it will be 1,078,246 minified! We don't want every book to load that file.

## Suggestion

My suggesting is that we generate a unique `highlight.js` file for every book that supports exactly the languages the book uses.

This will allow us to support every language `highlight.js` supports and keep the `highlight.js` file small. Maybe even smaller than what we have now.

## Implementation

* Download the bundle from [highlight.js](https://highlightjs.org/) and extract it in a folder called `hl`.
* It has a minimized JavaScript file that contains the core code and the code of each language.
* `cp hl/highlight.min.js src/front-end/js/highlight.js`
* We might need to pick one of the styles from the `styles` folder (though we might be able to keep the one we already have)
* During the book building phase we extract the list of languages used in the book.
* We extract the core and the languages from the source `highlight.min.js` file and build a new file including only the core and the selected languages and save that as `highlight.js`.

szabgab:
Apparently we cannot upgrade `highlight.js` because #1622 and the goal is to switch to `syntect` #1652

PR #2645, opened 2025-04-06
Upgrade highlight.js to 11.10

szabgab:
* highlightBlock was deprecated in favor of highlightElement
* Generate highlight.js based on the languages that are used in the book.

* See #2639


szabgab:
I'll fix it and send a new PR

PR #2646, opened 2025-04-06
Remove the book.multilingual field

### LINKED ISSUE #2636: What to do with the book.multilingual field?

szabgab:
As it is seems it has never been in real use.

Closes #2636

### DIFF
--- a/examples/nop-preprocessor.rs
+++ b/examples/nop-preprocessor.rs
@@ -117,7 +117,6 @@ mod nop_lib {
                         "book": {
                             "authors": ["AUTHOR"],
                             "language": "en",
-                            "multilingual": false,
                             "src": "src",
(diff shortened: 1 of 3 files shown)


ehuss (approved):
Thanks!

PR #2647, opened 2025-04-06
Upgrade highlight.js to 11.10

szabgab:
* highlightBlock was deprecated in favor of highlightElement
* Generate highlight.js based on the languages that are used in the book.

* See #2639


ehuss:
Does this deal with https://github.com/rust-lang/mdBook/issues/1622? That is what has been holding us back from updating.

In general I would prefer to not put any more energy into highlight.js and try to get https://github.com/rust-lang/mdBook/pull/1652 moving forward.

szabgab:
I just checked and this version of `highlight.js` does not hide the "boring" lines.

It's a pity that from all the highlighting-related issues I did not see that upgrading `highlight.js` is not possible because of changes in that engine and that you would prefer to move to `syntect` anyway. I just put in this **a lot** of time.

I am going to close this PR and the related issue as there is no point in wasting more time on this.

PR #2666, opened 2025-04-15
Sidebar improvements

### LINKED ISSUE #2665: Sidebar status is not correctly preserved between page changes, and other small sidebar improvements
### Problem

Hello, there is a small bug, and some possible improvements of the sidebar:

1. there is a bug where the visibility of the sidebar is not correctly kept between page changes/navigation, when the browser window was smaller than 1080 pixels wide (even if you explicitly clicked to show the sidebar, if you refresh or change page, it will disappear again)
2. the sidebar width could be saved as well as its visibility status
3. the sidebar “configuration” could be saved in `sessionStorage` instead or `localStorage`. Rationale: if you open multiple browser windows that have different sizes, you may want different configurations for the sidebar for each window.
4. when the page is first opened, the sidebar is either shown or hidden (by default) depending on the browser window’s width, but if you don’t interact with the sidebar visibility button, this default will not be stored anywhere, so if the user resizes the window then refreshes the page, the sidebar might unexpectedly disappear.

These all are minor issues, but annoying, and they can be fixed/improved relatively easily.

I actually worked on a fix for all that (I started working on it *before* I read the contributing guidelines and its warning to not work on a pull request before submitting an issue, sorry 😔 so I decided to finish), so I’ll still link a PR just in case.

### Steps

1. open a window smaller than 1080 pixels horizontally
2. click the “Toggle table of contents” button to show the sidebar
3. hit F5 to refresh the page
4. the sidebar has disappeared again

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
v0.4.48
```

njor:
Sidebar status is now moved to sessionStorage instead of localStorage.

Fix: sidebar visibility was not correctly loaded from storage when the browser window was smaller than 1080 pixels wide.

Sidebar width is now saved too.

Remove too large try-catch block that would hide some errors.

Resolves #2665

### DIFF
--- a/src/front-end/js/book.js
+++ b/src/front-end/js/book.js
@@ -1,6 +1,6 @@
 'use strict';
 
-/* global default_theme, default_dark_theme, default_light_theme, hljs, ClipboardJS */
+/* global default_theme, default_dark_theme, default_light_theme, hljs, ClipboardJS, mdBook */
 
 // Fix back button cache problem
 window.onunload = function() { };
--- a/src/front-end/js/book.js
+++ b/src/front-end/js/book.js
@@ -514,19 +514,6 @@ aria-label="Show hidden lines"></button>';
     });
 })();
 
-function setConfig(key, value, storage = localStorage) {
-    const fullKey = 'mdbook-' + key;
-    try {
-        storage.setItem(fullKey, value);
-    } catch (e) {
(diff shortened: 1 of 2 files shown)


njor:
Resolves 2665

PR #2675, opened 2025-04-21
Break off query string when comparing url for sidebar

### LINKED ISSUE #2671: folded chapters don't expand with `highlight` in URL
### Problem

When opening a page in a book with folded chapters, and the `highlight` value in the URL, the chapter sidebar does not expand to unfold to that chapter.

### Steps

Go directly to <https://doc.rust-lang.org/nightly/rust-by-example/custom_types/enum.html?highlight=enums#enums>

or

1. Go to <https://doc.rust-lang.org/nightly/rust-by-example/>
2. Type `s` to start a search, and enter "Enum"
3. Click on the first element

Notice that the sidebar does not expand:

Image

It should expand (like it does if `?highlight=enums` is not in the URL):

Image

### Possible Solution(s)

_No response_

### Notes

cc @notriddle This seems to have regressed via https://github.com/rust-lang/mdBook/pull/2414

### Version

```text
mdbook v0.4.48
```

notriddle:
Fixes #2671

### DIFF
--- a/src/front-end/templates/toc.js.hbs
+++ b/src/front-end/templates/toc.js.hbs
@@ -10,7 +10,7 @@ class MDBookSidebarScrollbox extends HTMLElement {
     connectedCallback() {
         this.innerHTML = '{{#toc}}{{/toc}}';
         // Set the current, active page, and reveal it if it's hidden
-        let current_page = document.location.href.toString().split("#")[0];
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks! Love the ability to test these.

PR #2679, opened 2025-04-25
fix(css): sidebar can no longer be larger than 80% of viewport width

### LINKED ISSUE #2678: Sidebar is not possible to close on certain mobile viewports
### Problem

with a 270x495 viewport with a DPR of 4, opening the sidebar of the rust reference looks like this:

screenshot

as you can see, the "close sidebar" button is not on screen.

### Steps

1. go to https://doc.rust-lang.org/reference/ in firefox
2. use Ctrl+Shift+M to enter responsive design mode
3. Device dropdown > Edit List > Add Custom Device > 270x495 viewport, DPR=4
4. open sidebar

### Possible Solution(s)

1. move the close sidebar button to align with how rustdoc works
2. use viewport units so this can't happen

### Notes

yes this corresponds to a real device, the HTC U23 Pro.

### Version

```text
unknown
```

lolbinarycat:
fixes #2678

### DIFF
--- a/src/front-end/css/variables.css
+++ b/src/front-end/css/variables.css
@@ -2,7 +2,8 @@
 /* Globals */
 
 :root {
-    --sidebar-width: 300px;
+    --sidebar-target-width: 300px;
+    --sidebar-width: min(var(--sidebar-target-width), 80vw);
     --sidebar-resize-indicator-width: 8px;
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks!

PR #2681, opened 2025-04-29
Warn on and ignore duplicate footnote definitions

### LINKED ISSUE #2649: Duplicate footnote names cause incorrect rendering and numbering
### Problem

When there are duplicate footnote definitions, it can cause the numbering to be off and generally cause confusion.

### Steps

Example page:

```
# Chapter 1

Test[^example]

Something else[^other]

Test2[^example]

[^example]: Definition one

[^other]: Other

[^example]: Definition two
```

This renders as:

Image

Several problems:

* The second `example` definition shows up, but nothing links to it.
* The "Something else" footnote number is 2, but clicking it takes you to footnote 3.

### Possible Solution(s)

Should probably warn if there are duplicate footnote definitions, and ignore/delete subsequent duplicates.

### Notes

mdbook v0.4.48

### Version

```text

```

krishanjmistry:
Rework the footnote rendering logic to handle cases where the same footnote tag is defined multiple times in a document.

When multiple definitions of the same footnote tag are encountered:
- a warning is logged
- only the _first_ definition for the tag is used for rendering
- subsequent definitions for the tag are ignored

Fixes #2649

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -235,10 +235,10 @@ pub fn render_markdown_with_path(
     // `count` is the number of references to this footnote (used for multiple
     // linkbacks, and checking for unused footnotes).
     let mut footnote_numbers = HashMap::new();
-    // This is a list of (name, Vec<Event>)
+    // This is a map of name -> Vec<Event>
     // `name` is the name of the footnote.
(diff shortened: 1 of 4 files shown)


ehuss (approved):
Thanks!

PR #2689, opened 2025-05-05
Revert: Remove the book.multilingual field and don't serialize it

ehuss:
This reverts https://github.com/rust-lang/mdBook/issues/2646 since I overlooked that this would be a breaking API change. Instead it just adds a `serde(skip_serializing)` so that it doesn't show up in `mdbook init`. We'll plan to remove it in the future (cc https://github.com/rust-lang/mdBook/issues/2636).

### DIFF
--- a/examples/nop-preprocessor.rs
+++ b/examples/nop-preprocessor.rs
@@ -117,6 +117,7 @@ mod nop_lib {
                         "book": {
                             "authors": ["AUTHOR"],
                             "language": "en",
+                            "multilingual": false,
                             "src": "src",
                             "title": "TITLE"
                         },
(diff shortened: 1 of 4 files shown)


szabgab:
Why is this a breaking API change and how could one (well, me) know?

ehuss:
Removing a field from a public type is a breaking change because code that constructs the type, such as:

```rust
BookConfig {
    title: Some(String::from("mdBook Documentation")),
    authors: vec![],
    description: Some(String::from("Description")),
    src: PathBuf::from("./source"),
    multilingual: false,
    language: None,
    text_direction: None,
};
```

This would fail to compile if the field was removed.

This can be checked with something like [`cargo-semver-checks`](https://crates.io/crates/cargo-semver-checks).

PR #2691, opened 2025-05-09
sidebar: use the same resize grip as rustdoc and playground

notriddle:
Follow up #2209 and https://github.com/rust-lang/rust/pull/139562

Since rustdoc and the playground both use similar resizing indicators, it only makes sense that mdbook should, too.

Before:

image

After:

image

### DIFF
--- a/src/front-end/css/chrome.css
+++ b/src/front-end/css/chrome.css
@@ -474,9 +474,24 @@ html:not(.sidebar-resizing) .sidebar {
 
 .sidebar-resize-handle .sidebar-resize-indicator {
     width: 100%;
-    height: 12px;
-    background-color: var(--icons);
+    height: 16px;
+    color: var(--icons);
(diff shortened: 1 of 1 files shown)


GuillaumeGomez (approved):

ehuss:
Can you say where this shows up in rustdoc? I don't see any resizing indicators on any of the doc pages.

Not really a blocker, but just FYI it renders a little awkwardly in Safari. Compare:

| Firefox | Chrome | Safari |
|--------|---------|-------|
| image |  image |  image |

ehuss (approved):
Thanks!

GuillaumeGomez:
> Can you say where this shows up in rustdoc?

You can see a screenshot in https://github.com/rust-lang/rust/pull/139562. It's really recent so you'll need to wait for nightly docs to be rebuilt but you should be able to see it in https://doc.rust-lang.org/nightly/std/.

PR #2698, opened 2025-05-11
Improve ergonomics of search

traviscross:
This PR improves the ergonomics of search by doing the following, and is best reviewed commit by commit:

- Allows for search by `/` as well as `s`.
- Fixes an uncaught exception when pressing `down` when there are no search results.
- Navigates automatically to the first search result when pressing `enter`.
- Uses `<kbd>` to render shortcut keys in the guide.

The commit messages contain more details and motivation.

Closes https://github.com/rust-lang/mdBook/issues/2379


ehuss (approved):
Thanks!

PR #2702, opened 2025-05-14
Add rel=edit attribute to "Suggest an edit" link

capjamesg:
This PR adds the rel="edit" attribute to the "Suggest an edit" edit link.

rel=edit lets a page indicate that the linked resource can be used to edit the page. It is defined at https://microformats.org/wiki/rel-edit. This can then be parsed by tools like the Universal Edit Button and custom bookmarklets to open the edit page corresponding with a website.

### DIFF
--- a/src/front-end/templates/index.hbs
+++ b/src/front-end/templates/index.hbs
@@ -181,7 +181,7 @@
                         </a>
                         {{/if}}
                         {{#if git_repository_edit_url}}
-                        <a href="{{git_repository_edit_url}}" title="Suggest an edit" aria-label="Suggest an edit">
+                        <a href="{{git_repository_edit_url}}" title="Suggest an edit" aria-label="Suggest an edit" rel="edit">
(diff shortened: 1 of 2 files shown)


ehuss:
Thanks, looks like the tests are failing, though.

@notriddle or @GuillaumeGomez are you familiar with `rel="edit"`? Does this seem like something worth adding?

GuillaumeGomez:
Hum, not 100% sure it's ok (although I think it's fine). This button is to suggest an edit whereas the `rel=edit` docs say:

> Tells the browser the URL for editing the current page

So not exactly the same but it's nitpicking here. Let's see if @notriddle has another opinion but I think it's ok.

notriddle:
> So not exactly the same but it's nitpicking here.

The Microformats wiki page links to the [Suprabase Docs](https://supabase.com/docs/guides/database/overview) as a positive example, and they're using `rel=edit` to link to a GitHub editor page just like this PR proposes.

That seems like a positive endorsement of the practice. I'm fine with this change.

Issue #1764, opened 2022-03-09
site-url not respected in links


mimoo:
Since I set the site-url (or base url in other frameworks):

```toml
[output.html]
site-url = "/proof-systems/"
```

I would expect all markdown links to be rewritten to respect that in my book. They aren't :o

ehuss:
All links in your book should be relative, so it shouldn't matter where it is located.  The `site-url` is only needed for generating the 404 page which has to be able to work when loaded from any location.  

If you are using absolute links, like `/foo.md`, then I recommend changing them to be relative (such as `foo.md` or `../foo.md`).  I'm not sure mdbook should rewrite absolute links, since they may legitimately be trying to link to a location outside of the book.

mimoo:
in my case, I am moving the generated rust documentation into the `/rustdoc` folder in the book. So that on github pages, I can access both [the book](https://o1-labs.github.io/proof-systems/) and [the rustdoc](https://o1-labs.github.io/proof-systems/rustdoc/). 

> I'm not sure mdbook should rewrite absolute links, since they may legitimately be trying to link to a location outside of the book.

I'm not sure I understand, all frameworks that expose a base-url do this: they rewrite all absolute links using the base url. What is a use case where you wouldn't want that?

schungx:
> I'm not sure I understand, all frameworks that expose a base-url do this: they rewrite all absolute links using the base url. What is a use case where you wouldn't want that?

I'd second this.

Writing a large book with `mdbook` means that I'm constantly shifting chapters from one place to another, sometimes across folders or up/down levels.  It would be a huge mess to manually fix all relative url links whenever I move a file.

Always using absolute url links resolve this issue nicely, resulting in a huge savings of time.  I believe this feature is sorely needed.

Right now, I use a templating engine `mdbook-tera` and put `{{rooturl}}/` in front of all my url links, but it is not as good as built-in support.

ehuss:
> What is a use case where you wouldn't want that?

This probably isn't used in practice very often, but I can imagine a book that wants to link to other pages on the same website (that aren't part of a book).  We don't use it for rust-lang, so I would be fine changing it, but it could possibly be a breaking change for other users, so it would need to wait for the next incompatible release.

mimoo:
> This probably isn't used in practice very often, but I can imagine a book that wants to link to other pages on the same website (that aren't part of a book)

[...]

PR #2706, opened 2025-05-15
Add `canonical-site-url` setting

### LINKED ISSUE #1238: Canonical Links Best Practice

roberth:
Fixes #1238.

Hi :wave:

This implements the suggestion by markhildreth to implement such a setting, based on the conversation in https://github.com/rust-lang/mdBook/issues/1238.
`<link rel="canonical">` plays an important role in avoiding an SEO penalty when users deploy multiple versions of their site.

I've additionally infixed `-site-` to highlight the relationship with the `site-url`, and to distinguish it from the canonical URL as it occurs in a page.

This is a more user-friendly alternative to #2415 as it's an "end to end" solution with documentation. (It may have usages beyond this use case though)

Let me know if there's anything I should improve.
When this is done, I'd like to implement a setting for opting in to clean URLs without the `.html` suffix so that we can close the whole issue.

Changelog suggestion (not included due to repeated conflicts...)
```md
- Added [`canonical-site-url`](https://rust-lang.github.io/mdBook/format/configuration/renderers.html?highlight=canonical-site-url#html-renderer-options) setting, to set `<link rel="canonical">` in the HTML output of each page.
```

### DIFF
--- a/crates/mdbook-html/front-end/templates/index.hbs
+++ b/crates/mdbook-html/front-end/templates/index.hbs
@@ -10,6 +10,9 @@
         {{#if base_url}}
         <base href="{{ base_url }}">
         {{/if}}
+        {{#if canonical_url}}
+        <link rel="canonical" href="{{ canonical_url }}">
+        {{/if}}
 
 
         <!-- Custom HTML head -->
--- a/crates/mdbook-html/src/html_handlebars/hbs_renderer.rs
+++ b/crates/mdbook-html/src/html_handlebars/hbs_renderer.rs
@@ -64,6 +64,13 @@ impl HtmlHandlebars {
             .to_str()
             .with_context(|| "Could not convert path to str")?;
         let filepath = Path::new(&ctx_path).with_extension("html");
+        let filepath_str = filepath
+            .to_str()
+            .with_context(|| format!("Could not convert path to str: {}", filepath.display()))?;
+        let canonical_url = ctx.html_config.canonical_site_url.map(|canon_url| {
+            let canon_url = canon_url.as_str().trim_end_matches('/');
+            format!("{}/{}", canon_url, self.clean_path(filepath_str))
+        });
 
         let book_title = ctx
             .data
@@ -80,6 +87,8 @@ impl HtmlHandlebars {
         };
 
         ctx.data.insert("path".to_owned(), json!(path));
+        ctx.data
+            .insert("canonical_url".to_owned(), json!(canonical_url));
         ctx.data.insert("content".to_owned(), json!(content));
         ctx.data.insert("chapter_title".to_owned(), json!(ch.name));
         ctx.data.insert("title".to_owned(), json!(title));
@@ -297,6 +306,15 @@ impl HtmlHandlebars {
         fs::write(original, rendered)?;
 
         Ok(())
+    }
+
+    /// Strips `index.html` from the end of a path, if it exists.
+    fn clean_path<'a>(&self, path: &'a str) -> &'a str {
+        if path == "index.html" || path.ends_with("/index.html") {
+            &path[..path.len() - 10]
+        } else {
+            path
+        }
     }
 }
 
--- a/tests/testsuite/rendering.rs
+++ b/tests/testsuite/rendering.rs
@@ -304,3 +304,25 @@ HTML tags must be closed before exiting a markdown element.
             str![[r##"<h3 id="option"><a class="header" href="#option">Option<t></t></a></h3>"##]],
         );
 }
+
+// Checks that a canonical URL is generated correctly.
+#[test]
+fn canonical_url() {
+    BookTest::from_dir("rendering/canonical_url")
+        .check_file_contains(
+            "book/index.html",
+            "<link rel=\"canonical\" href=\"https://example.com/test/\">",
+        )
+        .check_file_contains(
+            "book/canonical_url.html",
+            "<link rel=\"canonical\" href=\"https://example.com/test/canonical_url.html\">",
+        )
+        .check_file_contains(
+            "book/nested/page.html",
+            "<link rel=\"canonical\" href=\"https://example.com/test/nested/page.html\">",
+        )
+        .check_file_contains(
+            "book/nested/index.html",
+            "<link rel=\"canonical\" href=\"https://example.com/test/nested/\">",
+        );
+}
--- a/crates/mdbook-core/src/config.rs
+++ b/crates/mdbook-core/src/config.rs
@@ -476,6 +476,8 @@ pub struct HtmlConfig {
     pub input_404: Option<String>,
     /// Absolute url to site, used to emit correct paths for the 404 page, which might be accessed in a deeply nested directory
     pub site_url: Option<String>,
+    /// Canonical site url, used to emit <link rel="canonical"> tags in the HTML.
+    pub canonical_site_url: Option<String>,
     /// The DNS subdomain or apex domain at which your book will be hosted. This
     /// string will be written to a file named CNAME in the root of your site,
     /// as required by GitHub Pages (see [*Managing a custom domain for your
@@ -529,6 +531,7 @@ impl Default for HtmlConfig {
             git_repository_icon: None,
             input_404: None,
             site_url: None,
+            canonical_site_url: None,
             cname: None,
             edit_url_template: None,
             live_reload_endpoint: None,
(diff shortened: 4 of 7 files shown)


pinage404 (approved):
I tested locally, it works great

I hope a solution will be merged

whitequark:
I'd really like to have this functionality; is there a way I can help advance this PR? I want it for https://github.com/rust-lang/mdBook/issues/1238.

roberth:
I've rebased and put my suggestion for a changelog entry in the PR description, because it kept causing conflicts.

Let us know if there's anything else we can do!

crates/mdbook-html/front-end/templates/index.hbs
```diff
@@ -10,6 +10,9 @@
         {{#if base_url}}
         <base href="{{ base_url }}">
         {{/if}}
+        {{#if canonical_url}}
+        <link rel="canonical" href="{{ canonical_url }}">
```
pinage404: What do you think about adding also [the Open Graph URL](https://ogp.me/#metadata) ?

```hbs
<meta property="og:url" content="{{ canonical_url }}" />
```

roberth: I guess the more interesting question is would you want to rewrite social media without also hinting the search engines?
If that's a use case, maybe we should only define `canonical_url` but leave the socials-vs-search decision up to the custom HTML template, to keep things simple for the mdBook implementation?
If we don't have such a use case, I'd like to keep this built-in behavior right here, because otherwise you have a weird setting without any effect unless you customize your theme. That seems like an step too much for me in terms of UX.
So right back at ya, I guess :)
What do you think of removing `<link rel="canonical" .../>`?

pinage404 (approved):
It still works great

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks!

cc @ehuss

ehuss:
Can you please explain the relationship and picture for how this relates with:

- the existing `site-url`
- this new `rel=canonical` support
- sitemap (https://github.com/rust-lang/mdBook/issues/1491) which requires an absolute URL
- link support of site-url (https://github.com/rust-lang/mdBook/issues/1764)

When might these be different? How do we avoid adding four different settings that specify the "url" of the site? If we do need to add multiple URLs, is there a way we can do that so it isn't too confusing?

Some other questions:

- What are the consequences of setting this?
- How would this deal with the first page being translated to index.html?

@GuillaumeGomez What is your interest in this feature? Do you want this for doc.rust-lang.org?

GuillaumeGomez:
No interest for docs.rs. Just a feature that seems useful for people hosting their books (like we do for `askama` for example).

roberth:
Sitemap purpose: Tell search engines "here are all the pages that exist at this location"

Canonical purpose: Tell search engines "if you found this page, the authoritative version is at this other location"

`site-url`: The url where the book will be hosted. This is required to ensure navigation links and script/css imports in the 404 file work correctly, even when accessing urls in subdirectories. Defaults to /. If site-url is set, make sure to use document relative links for your assets, meaning they should not start with /. &mdash; [mdbook docs](https://rust-lang.github.io/mdBook/format/configuration/renderers.html?highlight=site-url#html-renderer-options)

---

> how this relates

### `site-url` <-> `canonical-site-url`
The latter is optional. If unspecified, pages are generated as usual, and crawlers assume that
- if the content is unique, it is canonical
- if the contents are not unique, it is duplicated, and all sites that host it will be considered less valuable.

### Sitemap URL <-> `site-url`

Not entirely clear to me; not the thing I've focused on.

1. I assume each site would contain a correct sitemap *file* based on `site-url`.
2. Each HTML page can reference a sitemap. I don't know what's best practice for that.
  - Maybe it's good SEO practice to point to the canonical sitemap instead of the local sitemap, but I don't really know.
  - I imagine for other tools that use the sitemap (browser extensions? accessibility? LLM harnesses?), those would prefer a reference to the local sitemap.

### Link support of `site-url`

My assumption would be that any links that are not relative (to the page base directory), would be interpreted against `site-url`. I'm surprised that `site-url` does not have to specify a full URL with authority part.

### My model for this

For the intrinsic structure of the site, in order: use relative paths, or `site-url`, preferably relative to `/`, alternatively relative to the absolute site URL.
I'm surprised that `site-url` can be relative, but I can see that be useful for e.g. having same content on staging and prod.

Only use a different URL where it matters for such things as SEO.
Sitemap content URL in the file itself: always the absolute site URL
Sitemap link in each page: default to absolute site URL or canonical URL + `/sitemap.xml`. Unsure, so maybe users should choose => add `sitemap-link-url` setting that defaults to either one (don't know which).
Canonical URL: separate URL, do not render unless specified.


---

> If we do need to add multiple URLs, is there a way we can do that so it isn't too confusing?

It's intrinsic complexity of the web domain, but users with a simple one-version one-location deployment can perhaps be somewhat shielded from this by clearly marking the extra URL settings as "optional" or "for multi-version deployments" in the docs.

---

> What are the consequences of setting this?

Higher ratings in search engines, and visitors are routed to your preferred version of the site.

---

> How would this deal with the first page being translated to index.html?

Not sure what's special about this?

jackpot51:
Is anything else required before this is merged?

crates/mdbook-html/src/html_handlebars/hbs_renderer.rs
```diff
@@ -64,6 +64,13 @@ impl HtmlHandlebars {
             .to_str()
             .with_context(|| "Could not convert path to str")?;
         let filepath = Path::new(&ctx_path).with_extension("html");
+        let filepath_str = filepath
+            .to_str()
+            .with_context(|| format!("Could not convert path to str: {}", filepath.display()))?;
+        let canonical_url = ctx.html_config.canonical_site_url.map(|canon_url| {
+            let canon_url = canon_url.as_str().trim_end_matches('/');
+            format!("{}/{}", canon_url, self.clean_path(filepath_str))
```
GuillaumeGomez, line 72: Instead of calling `clean_path` on a `&str`, please run it on `filepath` (a `Path`, which has an API to get the parent) and then convert it to a string.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/2706/cc78fb4ff18b129099ce9daec49613f0a6b513e0..a5ef49ba5792578ba0433a611b2e4f44cee6b2ad)*

tests/testsuite/rendering.rs
```diff
@@ -304,3 +304,25 @@ HTML tags must be closed before exiting a markdown element.
             str![[r##"<h3 id="option"><a class="header" href="#option">Option<t></t></a></h3>"##]],
         );
 }
+
+// Checks that a canonical URL is generated correctly.
+#[test]
+fn canonical_url() {
+    BookTest::from_dir("rendering/canonical_url")
```
GuillaumeGomez, line 311: Is it possible to add a test with a path containing only `index.html`? It'd be useful for askama.rs for example.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/2706/cc78fb4ff18b129099ce9daec49613f0a6b513e0..a5ef49ba5792578ba0433a611b2e4f44cee6b2ad)*

GuillaumeGomez (changes requested):
Few things to update, but otherwise looks very promising!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/2706/cc78fb4ff18b129099ce9daec49613f0a6b513e0..a5ef49ba5792578ba0433a611b2e4f44cee6b2ad)*

GuillaumeGomez (approved):

GuillaumeGomez:
Looks all good to me, thanks!

roberth:
Worth the wait. Thank you for the review and merge Guillaume!

PR #2707, opened 2025-05-19
Document the `edition2024` code block attribute

noritada:
This PR adds documentation on the `edition2024` code block attribute, which is missing.


ehuss (approved):
Thanks!

PR #2709, opened 2025-05-21
Update `opener` to 0.8.1

zeenix:
This removes a few indirect deps, especially system deps like `dbus` and `pkg-config`.


ehuss (approved):
Thanks!

Issue #2700, opened 2025-05-11
`mode-rust.js` is not found


traviscross:
### Problem

When loading a page with Rust editable code blocks, such as the "Rust Codeblocks" page in the test book, the browser throws a 404 on trying to access `/mode-rust.js`.  It's hitting this because the script tag isn't using the hashed version of the file name.

There's some strange interaction with the `ace` editor going on here.  The script tag that triggers this 404 is inserted dynamically.

This was introduced, I'm guessing, by https://github.com/rust-lang/mdBook/pull/1368.  A bisection points at commit 74200f7395002728d4599ff7b4e1712648574334, but I'm guessing that's just because it fixed another issue that may have been masking this one.

Is this why the `move-between-pages.goml` test had to keep `fail-on-request-error` disabled in https://github.com/rust-lang/mdBook/pull/2552?

cc @notriddle @GuillaumeGomez @ehuss

### Version

```text

```

GuillaumeGomez:
Yes and it's why I opened https://github.com/rust-lang/mdBook/pull/2617.

PR #2710, opened 2025-05-23
Fix loading of mode-rust.js

ehuss:
The reason the ACE editor was failing to load the rust syntax highlighting is because the syntax highlighting was being created *after* the editor was created. If the editor is created first, then ACE tries to load `ace/mode/rust`. Since it isn't already defined, it tried to compute the URL and load it manually. However, since the URLs now have a hash in it (via https://github.com/rust-lang/mdBook/pull/1368), it was unable to load.

The solution here is to make sure `ace/mode/rust` is defined before creating the editors. Then ACE knows that it can just load the module directly instead of trying to fetch it from the server.

Fixes https://github.com/rust-lang/mdBook/issues/2700

### DIFF
--- a/src/front-end/templates/index.hbs
+++ b/src/front-end/templates/index.hbs
@@ -297,8 +297,8 @@
 
         {{#if playground_js}}
         <script src="{{ resource "ace.js" }}"></script>
-        <script src="{{ resource "editor.js" }}"></script>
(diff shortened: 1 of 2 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Much better fix, thanks!

PR #2715, opened 2025-05-27
add another link type

tshepang:


ehuss (changes requested):
Thanks! This also needs to update the rendered version just below it.

In general I'm a little reluctant to add things like this since it is not intended to be a full Markdown tutorial. There are links at the top of the page for that. However, I think a small example here should be fine.

PR #2725, opened 2025-06-05
Hide the sidebar when collapsed to prevent browser search to find text from it

### LINKED ISSUE #2721: Browser search finds sidebar text when it is hidden
### Problem

When the sidebar is collapsed (hidden), using the browser's native search will find the hidden words in the sidebar. This is a little confusing because you cannot see the words. This also doesn't seem to be a problem on other sides with a sidebar that collapses. I'm not sure why this is happening.

### Steps

1. Go to https://rust-lang.github.io/mdBook/
2. Collapse the sidebar
3. Use your browser's "Find" command and search for "installation" and notice that it seems to find this word even though it isn't visible on the page.

This seems to happen with Firefox, Chrome, and Safari.

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
mdbook v0.4.51
```

GuillaumeGomez:
Fixes #2721.

For the GUI test, I need to add a new command to ensure the text is not searchable when collapsed. I opened an issue for that in browser-ui-test: https://github.com/GuillaumeGomez/browser-UI-test/issues/659.

### DIFF
--- a/src/front-end/js/book.js
+++ b/src/front-end/js/book.js
@@ -524,7 +524,6 @@ aria-label="Show hidden lines"></button>';
     let firstContact = null;
 
     function showSidebar() {
-        body.classList.remove('sidebar-hidden');
         body.classList.add('sidebar-visible');
         Array.from(sidebarLinks).forEach(function(link) {
             link.setAttribute('tabIndex', 0);
@@ -540,7 +539,6 @@ aria-label="Show hidden lines"></button>';
 
     function hideSidebar() {
         body.classList.remove('sidebar-visible');
-        body.classList.add('sidebar-hidden');
         Array.from(sidebarLinks).forEach(function(link) {
             link.setAttribute('tabIndex', -1);
         });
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
Thanks to the GUI tests, I uncovered the case where the window size is too small and therefore the sidebar should not be displayed by default. =D

GuillaumeGomez:
Anyway, it's ready for review.

ehuss:
Thank you very much!

This seems to introduce a few regressions:

- On Safari, the animation to display the sidebar seems broken. The sidebar pops into existence without animating, but the rest of the page animates over on top of it.
- On Firefox, if the window is below 1080px wide, and the sidebar is visible, reloading the page shows a blank sidebar. Normally it should reset so that the page loads with the sidebar hidden.

Would you be able to look into those issues?

GuillaumeGomez:
On safari, sadly no. But on firefox, definitely. :)

GuillaumeGomez:
Fixed. I also extended the GUI test to ensure it doesn't happen again.

ehuss:
Thanks, I think I have fixed the issue that was happening on Safari.

src/front-end/js/book.js
```diff
@@ -540,6 +540,10 @@ aria-label="Show hidden lines"></button>';
         /* To allow the sidebar expansion animation, we first need to put back the display. */
         if (!sidebarCheckbox.checked) {
             sidebar.style.display = '';
+            // Workaround for Safari skipping the animation when changing
+            // `display` and a transform in the same event loop. This forces a
+            // reflow after updating the display.
+            sidebar.offsetHeight;
```
GuillaumeGomez, line 546: I'm very confused about this fix considering it does nothing. Dark magic.

PR #2728, opened 2025-06-12
Add minification for CSS and JS files

GuillaumeGomez:
Here are the file size differences:

| file | before | after | diff |
|-|-|-|-|
| chrome.css | 16535 | 12014 | -37.4% |
| general.css | 6531 | 4122 | -36.9% |
| print.css | 661 | 505 | -23.6% |
| variables.css | 8836 | 6735 | -23.8%|
| book.js | 28707 | 17964 | -37.4% |

### DIFF
--- a/tests/testsuite/book_test.rs
+++ b/tests/testsuite/book_test.rs
@@ -205,7 +205,8 @@ impl BookTest {
 
     /// Builds the book in the temp directory.
     pub fn build(&mut self) -> &mut Self {
-        let book = self.load_book();
+        let mut book = self.load_book();
+        book.disable_minification();
         book.build()
             .unwrap_or_else(|e| panic!("book failed to build: {e:?}"));
         self.built = true;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1251,6 +1251,7 @@ dependencies = [
  "ignore",
  "log",
  "memchr",
+ "minifier",
  "notify",
  "notify-debouncer-mini",
  "opener",
@@ -1303,6 +1304,15 @@ checksum = "f7c44f8e672c00fe5308fa235f821cb4198414e1c77935c1ab6948d3fd78550e"
 dependencies = [
  "mime",
  "unicase",
+]
+
+[[package]]
+name = "minifier"
+version = "0.3.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "14f1541610994bba178cb36757e102d06a52a2d9612aa6d34c64b3b377c5d943"
+dependencies = [
+ "clap",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ handlebars = "6.0"
 hex = "0.4.3"
 log = "0.4.17"
 memchr = "2.5.0"
+minifier = "0.3.6"
 opener = "0.8.1"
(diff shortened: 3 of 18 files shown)


GuillaumeGomez:
I added comparisons as well in the first comment.

ehuss:
Thanks! I had a few questions:

- This isn't like a typical minifier that renames and does other compression, correct?
- Would there be any benefit to having source maps, or some plan to getting those in the future?
- Should this include searcher.js? Or mode-rust.js?
- I've been wondering about combining files. Is that an option for the future? For example, there are a bunch of CSS files, would it be possible to combine them somehow? I've often wondered, as it seems like it would be pretty inefficient to have a bunch of network requests for several small files.
- The debugging experience seems to be hindered a little by this. I'm wondering if it would be possible to turn this off while using `mdbook serve`?

GuillaumeGomez:
> This isn't like a typical minifier that renames and does other compression, correct?

Indeed, it's just to remove unneeded characters (so mostly whitespace) and comments (while keeping the license). Anything more advanced would require a full JS parser, which I plan to add "some day".

> Would there be any benefit to having source maps, or some plan to getting those in the future?

Never used source map so I can't answer this. If you used source maps, do you think it would be worthwhile to have them?

> Should this include searcher.js? Or mode-rust.js?

`searcher.js` yes, but `mode-rust.js` seems to be already minified?

> I've been wondering about combining files. Is that an option for the future? For example, there are a bunch of CSS files, would it be possible to combine them somehow? I've often wondered, as it seems like it would be pretty inefficient to have a bunch of network requests for several small files.

I suppose we could. First we need to check if some files are always on the same pages. In this case, we can merge their source code directly. However if some files are not always on the same pages, then either we generate different CSS files which include some CSS files. If there is a possibility here, it would reduce the number of files, so I think it'd be a very good thing to do.

> The debugging experience seems to be hindered a little by this. I'm wondering if it would be possible to turn this off while using `mdbook serve`?

There isn't currently but we can definitely add an option to not run minimization (although, since we have the "pretty format" feature in all web browsers, it's mostly ok imo). Gonna add it in this PR then.

GuillaumeGomez:
I added `searcher.js` and `toc.js` to the list of the minified files. I also added the `--disable-minification` option.

tests/testsuite/book_test.rs
```diff
@@ -205,7 +205,8 @@ impl BookTest {
 
     /// Builds the book in the temp directory.
     pub fn build(&mut self) -> &mut Self {
-        let book = self.load_book();
+        let mut book = self.load_book();
+        book.disable_minification();
```
ehuss, line 209: Why is this necessary? I'm worried about disabling this in tests, since this isn't testing what is actually generated.

ehuss (commented):
Sorry if I wasn't clear about how to enable/disable it. I was originally thinking that there would just be some non-overridable control when running `mdbook serve`. That is, I don't imagine anyone else will ever need the ability to control this, right? I'm hesitant to expose this in such a largely visible way. It can be something as basic as detecting the presence of the `live_reload_endpoint` setting.

tests/testsuite/book_test.rs
GuillaumeGomez: Because we do some string search in the tests. So for test code readability, I think it's better to disable it here.

We already check most things are working as expected in GUI tests (both CSS and JS) when minified. So I'm not sure it makes a big difference here.

GuillaumeGomez:
I removed the possibility to disable minification for now.

PR #2729, opened 2025-06-12
Show where the book was generated

GuillaumeGomez:
It now looks like this:

```
$ cargo run -- build test_book/
   Compiling mdbook v0.4.51 (/somewhere/mdBook)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4.48s
     Running `target/debug/mdbook build test_book/`
2025-06-12 21:33:28 [INFO] (mdbook::book): Book building has started
2025-06-12 21:33:28 [INFO] (mdbook::book): Running the html backend
2025-06-12 21:33:28 [INFO] (mdbook::renderer::html_handlebars::hbs_renderer): Book has been generated into `/somewhere/mdBook/test_book/book`
```

It's very convenient to see where the book is actually generated (even more when working on mdbook :laughing:).

### DIFF
--- a/src/renderer/html_handlebars/hbs_renderer.rs
+++ b/src/renderer/html_handlebars/hbs_renderer.rs
@@ -16,7 +16,7 @@ use std::sync::LazyLock;
 
 use crate::utils::fs::get_404_output_file;
 use handlebars::Handlebars;
-use log::{debug, trace, warn};
+use log::{debug, info, trace, warn};
 use regex::{Captures, Regex};
 use serde_json::json;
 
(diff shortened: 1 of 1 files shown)


src/renderer/html_handlebars/hbs_renderer.rs
```diff
@@ -482,6 +482,8 @@ impl Renderer for HtmlHandlebars {
         // Copy all remaining files, avoid a recursive copy from/to the book build dir
         utils::fs::copy_files_except_ext(&src_dir, destination, true, Some(&build_dir), &["md"])?;
 
+        info!("Book has been generated into `{}`", build_dir.display());
```
ehuss, line 485: Was there a particular reason to use `build_dir` instead of `destination`? It might be confusing, since this is specifically for the HTML renderer, but this is showing the root output directory which might not be where the HTML output actually resides. Maybe something like this?

```suggestion
        info!("HTML book written to `{}`", destination.display());
```

GuillaumeGomez: No particular reason, simply missed that HTML output folder might not be the same as build folder.

GuillaumeGomez:
Updated!

PR #2730, opened 2025-06-13
Fix sidebar links to use absolute paths to prevent 404 errors

### LINKED ISSUE #1764: site-url not respected in links

madyankin:
This change makes all sidebar navigation links absolute paths (starting with '/') instead of relative paths, preventing duplicate path segments when navigating between nested pages.

Added test to verify all sidebar links are absolute paths.

Fixes #1764

### DIFF
--- a/src/renderer/html_handlebars/helpers/toc.rs
+++ b/src/renderer/html_handlebars/helpers/toc.rs
@@ -117,7 +117,8 @@ impl HelperDef for RenderToc {
                         // Hack for windows who tends to use `\` as separator instead of `/`
                         .replace('\\', "/");
 
-                    // Add link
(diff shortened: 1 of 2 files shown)


ehuss:
I don't think this approach will work, since the book may not be rooted at any specific URL path. I believe something more along the lines of https://github.com/rust-lang/mdBook/pull/1802.

ehuss:
I'm going to close as I don't think this approach will work.   Thanks, though!

PR #2735, opened 2025-06-25
Fix JS error when `hasFocus` method is overwritten on search index load

GuillaumeGomez:
Fixes the following JS error:

image

It's because of https://github.com/rust-lang/mdBook/pull/2633: instead of adding new properties, I completely overwrote the `window.search` object, removing all the methods originally defined on it. This fixes it.


tests/gui/search.goml
```diff
@@ -1,5 +1,6 @@
 // This tests basic search behavior.
 
+fail-on-js-error: true
```
GuillaumeGomez, line 3: I'll make this value the default in `browser-ui-test` because it would have helped me catch this error much sooner.

GuillaumeGomez: Done in https://github.com/rust-lang/mdBook/pull/2736.

GuillaumeGomez:
CI passed. \o/

ehuss (approved):
Thanks!

PR #2743, opened 2025-07-07
Update renderers.md to included preferences for the markdown renderer

cboecking:


ehuss:
Thanks, but I would prefer to not add this content.

cboecking:
As you wish. Please feel free change/adapt/close as you wish. Chuck

PR #2746, opened 2025-07-08
Work around compat break in old index.hbs templates with new searcher.js

notriddle:
Problem reported in

https://github.com/rust-lang/mdBook/pull/2742#discussion_r2190930557


ehuss (approved):
Thanks!

Issue #2256, opened 2023-12-02
Support fragments in redirects 


ehuss:
### Problem

The [redirect table](https://rust-lang.github.io/mdBook/format/configuration/renderers.html?highlight=redirect#outputhtmlredirect) does not support fragments. It would be helpful when section headings are changed to provide a way to redirect to the new location (whether or the same page, or a different page).

### Proposed Solution

The redirect table should support `#` in the source and destination.
This is tricky because AFAIK native HTML does not support that. In my books, I have been using [javascript redirects](https://github.com/rust-lang/cargo/blob/58fb23140972092a12f7011d17a7db1d99e3eacf/src/doc/src/reference/manifest.md?plain=1#L650-L690). 
This would need to handle both scenarios where the original page still exists, and where it has been removed.
This should also consider any security issues like open redirects.

### Notes

cc original issue #430

KFearsoff:
Seems related: #1760,  #1715

While renames are different from permalinks, I feel like permalinks to the heading attributes would cover 80% of the use cases. Let's do that first, maybe?

PR #2747, opened 2025-07-08
Add support for fragment redirects

ehuss:
This adds the ability to redirect URLs with `#` fragments. This is useful when section headers get renamed or moved to other pages.

This works both for deleted pages and existing pages.

The implementation requires the use of JavaScript in order to manipulate the location. (Ideally this would be handled on the server side.)

This also makes it so that deleted page redirects preserve the fragment ID. Previously if you had a deleted page redirect, and the user went to something like `page.html#foo`, it would redirect to `bar.html` without the fragment. I think preserving the fragment is probably a better behavior. If the new page doesn't have the fragment ID, then no harm is really done. This is technically an open redirect, but I don't think that there is too much danger with preserving a fragment ID?

Closes https://github.com/rust-lang/mdBook/issues/2256

### DIFF
--- a/guide/src/format/configuration/renderers.md
+++ b/guide/src/format/configuration/renderers.md
@@ -310,13 +310,21 @@ This is useful when you move, rename, or remove a page to ensure that links to t
 [output.html.redirect]
 "/appendices/bibliography.html" = "https://rustc-dev-guide.rust-lang.org/appendix/bibliography.html"
 "/other-installation-methods.html" = "../infra/other-installation-methods.html"
+
+# Fragment redirects also work.
+"/some-existing-page.html#old-fragment" = "some-existing-page.html#new-fragment"
+
+# Fragment redirects also work for deleted pages.
+"/old-page.html" = "new-page.html"
+"/old-page.html#old-fragment" = "new-page.html#new-fragment"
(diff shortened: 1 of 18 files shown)


ehuss:
@GuillaumeGomez or @notriddle I don't suppose either of you would be willing to give this a light review?

One minor concern I had was about preserving the fragments on deleted pages, since that is somewhat of an open redirect. So for example, `old-page.html#foo` will now redirect to `new-page.html#foo` and preserve the `#foo` part in the redirect. Would you agree that should be safe?

GuillaumeGomez:
Is there a reason in particular we want to support this feature in the first place? Since `mdbook` is a static documentation generator, I'm not sure `mdbook` should be the one handling this case at all and should let it up to the server doing the rendering if any.

ehuss:
We don't have a reasonable ability to define server redirects with the rust-lang documentation.

GuillaumeGomez:
No we don't, and I'm not sure if we should. We keep older versions available (although my proposals on providing a GUI way to switch between versions were rejected) so that should cover this problem imo.

notriddle:
For the record, I don't have a problem with this change. :+1:

PR #2748, opened 2025-07-09
Replace warp with axum

jelmer:
This PR is more of a proof of concept and a question - but I figured I'd put it up since I'd already done to work to see if it can be done: I'm curious if you'd be open to a migration to axum (or perhaps something else?).

warp is problematic for Debian, since it has some outdated dependencies that we don't want to keep around just for warp; upstream is dormant (last change was about a year ago). mdbook is the only thing that actually uses warp nowadays in Debian.

(cc @plugwash)

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -2,6 +2,9 @@ use super::command_prelude::*;
 #[cfg(feature = "watch")]
 use super::watch;
 use crate::{get_book_dir, open};
+use axum::extract::ws::{Message, WebSocket, WebSocketUpgrade};
+use axum::routing::get;
+use axum::Router;
 use clap::builder::NonEmptyStringValueParser;
 use futures_util::sink::SinkExt;
 use futures_util::StreamExt;
@@ -11,8 +14,7 @@ use mdbook::MDBook;
 use std::net::{SocketAddr, ToSocketAddrs};
 use std::path::PathBuf;
 use tokio::sync::broadcast;
-use warp::ws::Message;
-use warp::Filter;
+use tower_http::services::{ServeDir, ServeFile};
 
 /// The HTTP endpoint for the websocket used to trigger reloads when a file changes.
 const LIVE_RELOAD_ENDPOINT: &str = "__livereload";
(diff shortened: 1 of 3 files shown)


src/cmd/serve.rs
```diff
@@ -116,38 +119,70 @@ async fn serve(
...
+    // WebSocket handler for live reload
+    let websocket_handler = move |ws: WebSocketUpgrade| async move {
+        let reload_tx = reload_tx_clone.clone();
+        ws.on_upgrade(move |socket| websocket_connection(socket, reload_tx))
+    };
+
+    // Serve static files with custom 404 fallback
+    let file_404_path = build_dir.join(file_404);
+    let serve_dir = ServeDir::new(&build_dir).not_found_service(tower::service_fn(
+        move |_: axum::http::Request<axum::body::Body>| {
```
ehuss, line 133: Thanks!

There is a similar PR to switch to axum in https://github.com/rust-lang/mdBook/pull/2359. That takes a slightly different approach for handling the `not_found_service` which seems a bit simpler:

```rust
.not_found_service(ServeFile::new(build_dir.join(file_404)))
```

Can you say more about why the difference here and why this needs to do mime guessing?

src/cmd/serve.rs
```diff
@@ -116,38 +119,70 @@ async fn serve(
...
+
+    axum::serve(listener, app).await.unwrap();
+}
+
+async fn websocket_connection(ws: WebSocket, reload_tx: broadcast::Sender<Message>) {
+    let (mut user_ws_tx, _user_ws_rx) = ws.split();
+    let mut rx = reload_tx.subscribe();
+
+    trace!("websocket got connection");
+    while let Ok(m) = rx.recv().await {
```
ehuss, line 182: I'm a little unclear why this is a loop. The original code doesn't use a loop, and https://github.com/rust-lang/mdBook/pull/2359 also does not use a loop. Can you say more about that?

jelmer: Just defensive - if the websocket is still open then we might as well forward reload triggers. Reverted.

ehuss (approved):
Thanks!

Issue #2697, opened 2025-05-11
Sidebar overlaps with a menu-bar


synalice:
### Problem

 When you resize the sidebar it goes under the menu-bar (the header).

https://github.com/user-attachments/assets/a17b482f-8579-492f-9716-45779a965617

### Steps

1. Open <https://rust-lang.github.io/mdBook/> or <https://doc.rust-lang.org/book/>.
2. Resize the sidebar and look at where it meets the menu-bar.

### Possible Solution(s)

We could make the sidebar be always on top of everything else. This means it would temporarily obscure the menu-bar.

Alternatively, make it so that menu-bar is resized linearly with the sidebar, without the "bounce" animation. With that the menu-bar and sidebar will never overlap. 

### Version

The one used for <https://rust-lang.github.io/mdBook/> at the time of writing this (a8aee21cd092e7d9bff3c7213c98adf1b9d13cfc).

synalice:
I've deleted [these 7 lines](https://github.com/rust-lang/mdBook/blob/a8aee21cd092e7d9bff3c7213c98adf1b9d13cfc/src/front-end/css/general.css#L89-L95) and here is the result. This might be a personal preference, but I do like it better this way.

https://github.com/user-attachments/assets/e406954e-8cea-4cb7-9088-a506f1f99b76

nikableh:
Found the same issue at [https://cxx.rs/#example](https://cxx.rs/#example):

https://github.com/user-attachments/assets/6fce1b78-909b-43fe-adbe-a9f80c80908b

nikableh:
Found the same issue at https://nixos.org/guides/nix-pills/01-why-you-should-give-it-a-try:

![Image](https://github.com/user-attachments/assets/7701b69d-60b9-4bd4-9279-569fdcdc5973)

PR #2749, opened 2025-07-13
Disable sidebar-resizing transition animation

nikableh:
This animation causes visual bugs (#2697) and makes the whole resize process appear less smooth (compared to not having this animation at all).

### Before this MR (with animation):

https://github.com/user-attachments/assets/25f24654-f4e4-4e28-aa29-c24174edfdd0

### After this MR (without animation):

https://github.com/user-attachments/assets/18814161-5124-4faf-8d97-ea693f4dfa4e

### DIFF
--- a/src/front-end/css/general.css
+++ b/src/front-end/css/general.css
@@ -86,13 +86,6 @@ h6:target::before {
     box-sizing: border-box;
     background-color: var(--bg);
 }
-.no-js .page-wrapper,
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks! There looks to be some larger issues with how the sidebar and resizing were working due to some recent changes. I have opened https://github.com/rust-lang/mdBook/pull/2750 that should fix this and several other issues. I'm going to close as that should resolve this.

Issue #2642, opened 2025-04-05
Sidebar State Affects Auto-Hiding Behavior of Menu Bar


WantenMN:
### Description  
In the documentation generated by mdBook, the auto-hiding behavior of the top Menu Bar is inconsistent depending on the Sidebar's state during page refresh. This issue is noticeable on long pages, such as the [Backends page](https://rust-lang.github.io/mdBook/for_developers/backends.html).  

### Steps to Reproduce  
1. **Scenario 1: Refresh Page with Sidebar Closed**  
   - Open a sufficiently long page, e.g., https://rust-lang.github.io/mdBook/for_developers/backends.html  
   - Ensure the Sidebar is closed (click the button to collapse the Sidebar).  
   - Refresh the page.  
   - Scroll down the page; the top Menu Bar auto-hides.  
   - Open the Sidebar.  
   - Continue scrolling down; the Menu Bar still hides with scrolling.  

2. **Scenario 2: Refresh Page with Sidebar Open**  
   - Open the same page (https://rust-lang.github.io/mdBook/for_developers/backends.html).  
   - Ensure the Sidebar is open.  
   - Refresh the page.  
   - Scroll down the page; the Menu Bar does not hide with scrolling.  
   - Close the Sidebar.  
   - Continue scrolling down; the Menu Bar fails to hide.

PR #2750, opened 2025-07-14
Fix sidebar animation and other behavior

ehuss:
This fixes several issues with how the sidebar was behaving:

- Manually resizing the sidebar was incorrectly applying transition animations to the page-wrapper causing awkward movement.
- Clicking the sidebar toggle caused the menu bar to behave differently compared to loading a page with the sidebar visible or hidden.
- page-wrapper animation wasn't working when JS was disabled.
- RTL sidebar animation was broken.

Most of these issues stem from
https://github.com/rust-lang/mdBook/pull/2454 which moved `js` and `sidebar-visible` classes from `<body>` to `<html>`, but failed to update some of the JS and CSS code that was still assuming it was on the body.

https://github.com/rust-lang/mdBook/pull/1641 previously moved `js` from `<html>` to `<body>` with the reasoning
"This will be necessary for using CSS selectors on root attributes.". However, I don't see how that is absolutely necessary, since selectors like `[dir=rtl].js` should work to select the root element. I might be missing something here, but I think everything looks to be working.

Fixes https://github.com/rust-lang/mdBook/issues/2697
Fixes https://github.com/rust-lang/mdBook/issues/2642

### DIFF
--- a/src/front-end/css/general.css
+++ b/src/front-end/css/general.css
@@ -86,11 +86,12 @@ h6:target::before {
     box-sizing: border-box;
     background-color: var(--bg);
 }
-.no-js .page-wrapper,
+html:not(.js) .page-wrapper,
 .js:not(.sidebar-resizing) .page-wrapper {
     transition: margin-left 0.3s ease, transform 0.3s ease; /* Animation: slide away */
 }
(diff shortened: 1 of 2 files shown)


nikableh:
Thanks! It's so pretty now :heart:

PR #2787, opened 2025-08-12
Deny all unknown config fields

ehuss:
This changes it so that it is an error if there is ever an unknown configuration field. This is intended to help avoid things like typos, or using an outdated version of mdbook. Although it is possible that new fields could potentially safely be ignored, setting up a warning system is a bit more of a hassle. I don't think mdbook needs to have the same kind of multi-version support as something like cargo does. However, if this ends up being too much of a pain point, we can try to add a warning system instead.

There are a variety of changes here:

- The top-level config namespace is now closed so that it only accepts the keys defined in `Config`.
- All config tables now reject unknown fields.
- Added `Config::outputs` and `Config::preprocessors` for convenience to access the entire `output` and `preprocessor` tables.
- Moved the unit-tests that were setting environment variables to the testsuite where it launches a process instead.

Closes https://github.com/rust-lang/mdBook/issues/1595


ehuss:
@szabgab FYI, I decided to go ahead with this for 0.5 to just error on these fields.

hmmh-dominik-saeume:
This also seems to affect Environment variables.

I noticed that, as I'm building my own docker image with renovate version variables like:

```dockerfile
# renovate: datasource=github-releases depName=rust-lang/mdBook
ENV MDBOOK_VERSION=v0.5.0
# renovate: datasource=github-releases depName=badboy/mdbook-toc
ENV MDBOOK_TOC_VERSION=0.14.2
```
maybe add that the migration guide?

I mean, for me the fix is easy, I just prefix the variablers with a R_ so that mdbook doesn't read them and tells me their invalid, but still its kindof unexpected to fail over the environment

ehuss:
Oh my, thanks for the report! I'll relax the restriction at the top level to allow that.

PR #2801, opened 2025-08-18
Keep preprocessors/backends execution order deterministic

HollowMan6:
There's a regression caused by recent refactor work, as it used to execute preprocessors/backends in a deterministic way, but now this is not the case, which causes trouble when some backends implicitly depend on the result from another backend and happen to work (e.g. mdbook-pdf). The root cause is that a HashMap has no order, so this PR switches this into `BTreeMap` instead.

### DIFF
--- a/crates/mdbook-driver/src/mdbook/tests.rs
+++ b/crates/mdbook-driver/src/mdbook/tests.rs
@@ -47,12 +47,10 @@ fn config_defaults_to_link_and_index_preprocessor_if_not_set() {
     // make sure we haven't got anything in the `preprocessor` table
     assert!(cfg.preprocessors::<toml::Value>().unwrap().is_empty());
 
-    let got = determine_preprocessors(&cfg, Path::new(""));
+    let got = determine_preprocessors(&cfg, Path::new("")).unwrap();
 
-    assert!(got.is_ok());
-    assert_eq!(got.as_ref().unwrap().len(), 2);
-    assert_eq!(got.as_ref().unwrap()[0].name(), "index");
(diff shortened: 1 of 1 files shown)


ehuss:
Just for clarification, this doesn't appear to preserve the order in `book.toml` (I don't think it did in the past, either, though it was alphabetical). I'm fine with merging this, but I just wanted to verify we're both understanding this correctly.

I'm wondering, if a renderer requires a certain order if it would make sense to instead have an explicit order like [preprocessors do](https://rust-lang.github.io/mdBook/format/configuration/preprocessors.html#require-a-certain-order)?

HollowMan6:
> Just for clarification, this doesn't appear to preserve the order in `book.toml` (I don't think it did in the past, either, though it was alphabetical). I'm fine with merging this, but I just wanted to verify we're both understanding this correctly.

Ah...😅 To be honest, I didn't really check the source code in the past until today I found `mdbook-pdf` is broken, and I imagined it used to preserve the order in `book.toml`. But in any case, it's good to have deterministic execution order, so that `mdbook-pdf` won't be broken.

> I'm wondering, if a renderer requires a certain order if it would make sense to instead have an explicit order like [preprocessors do](https://rust-lang.github.io/mdBook/format/configuration/preprocessors.html#require-a-certain-order)?

It can be good to have an explicit order like [preprocessors do](https://rust-lang.github.io/mdBook/format/configuration/preprocessors.html#require-a-certain-order) in the long run, but it's not necessary for `mdbook-pdf` then in the short term, as it solely relies on html backend's generated files, and `p` is after `h` if it's executed in alphabetical order.

PR #2810, opened 2025-08-22
Enable smart-punctuation by default

ehuss:
This enables the smart-punctuation setting by default. The long term plan is to continue to enable more markdown extensions by default across semver breaking releases.


your-diary:
What is the design decision behind enabling this by default?

I'm not complaining about it but just curious because I can easily disable the default via `output.html.smart-punctuation = false`.

PR #2816, opened 2025-08-24
test case: verify the more specific exception message

szabgab:


ehuss (approved):
Thanks!

I'm not sure how this field is going to work long-term, since the number of variants could increase every year, which could make this a little unwieldy. But for now this seems fine.

PR #2820, opened 2025-08-26
Enable hash-files by default

ehuss:
This enables the hash-files setting by default. We have been running it for a while, and it seems most of the issues have been resolved. This should help with more reliably loading content like the toc contents.


ehuss:
cc @notriddle FYI

PR #2822, opened 2025-08-27
Add sidebar heading navigation

ehuss:
This adds dynamic navigation of headers of the current page in the sidebar. This is intended to help the user see what is on the current page, and to be able to more easily navigate it. The "current" header is tracked based on the scrolling behavior of the user, and is marked with a small circle. This includes automatic folding to help keep it from being too unwieldy on a page with a lot of nested headers.

This includes the `output.html.sidebar-header-nav` option to disable it.

I'm sure there are tweaks, fixes, and improvements that can be made. I'd like to get this out now, and iterate on it over time to make improvements.

### DIFF
--- a/tests/gui/search.goml
+++ b/tests/gui/search.goml
@@ -41,7 +41,7 @@ assert-css: ("#mdbook-search-wrapper", {"display": "none"})
 press-key: 's'
 wait-for-css-false: ("#mdbook-search-wrapper", {"display": "none"})
 // We ensure the search bar has the focus.
-assert: "#mdbook-searchbar:focus"
+wait-for: "#mdbook-searchbar:focus"
 // Pressing a key will therefore update the search input.
 press-key: 't'
 assert-text: ("#mdbook-searchbar", "t")


shenef:
As I understand it, this is not supposed to act like a ToC and only shows a few upcoming and previous headers relative to the current scroll point?
Sadly I won't be able to take a look at this for a bit but I'd be curious how this looks like, especially with long pages and long headers (e.g. <https://github.com/shenef/Boktai-Speedrunning/blob/main/src/zoktai/any.md#ghouls--zombie-trap--short-sword-rng-manip>). We are currently using `mdBook-pagetoc` and I would gladly replace it with this if it works well.

ehuss:
> shows a few upcoming and previous headers relative to the current scroll point?

It shows all the headers on the page, though deeply nested headers are collapsed.

Here's a screen capture of an example:

[sidebar-scroll.webm](https://github.com/user-attachments/assets/7fda3671-2329-4ea6-ba57-5e47118d08c4)

Issue #1736, opened 2022-01-30
Convert all the links in the generated print.html for linking inside the book into URL fragment form


HollowMan6:
I'm requesting for a feature that convert all the links in the generated `print.html` that link to another html page inside the "book" into the [URL fragment form](https://en.wikipedia.org/wiki/URI_fragment), so that once a user click that link, they will be guided to the corresponding spot instead of another html page (This should be obvious as `print.html` includes all the generated book content, the current behavior is some kind of weird).

For one example of the mdBook Documentation (Notice URL change):
- Current Behavior:
![image](https://user-images.githubusercontent.com/43995067/151692222-d79c8d6d-121b-4bac-8cc2-03396892516b.png)

![image](https://user-images.githubusercontent.com/43995067/151692231-1ded9234-a2dc-4ab8-8a83-32df3dfd7aa5.png)

- Expected Behavior:
![image](https://user-images.githubusercontent.com/43995067/151692195-0fe45824-075b-4453-9f9d-bc3804787a44.png)

![image](https://user-images.githubusercontent.com/43995067/151692210-b38c86fa-1934-429d-8fb4-a6668e7d2cd7.png)

For use case: https://github.com/rust-lang/mdBook/issues/88#issuecomment-962613257

HollowMan6:
Well, I just noticed that developers have already noticed this issue: https://github.com/rust-lang/mdBook/blob/v0.4.15/src/utils/mod.rs#L73-L74 , it may be quite difficult, but still I hope someone can resolve this.

PR #2844, opened 2025-09-17
Add a new HTML rendering pipeline

ehuss:
This rewrites the HTML rendering pipeline to use a tree data structure, and implements a custom HTML serializer. The intent is to make it easier to make changes and to manipulate the output. This should make some future changes much easier.

This is a large change, but I'll try to briefly summarize what's changing:

- All of the HTML rendering support has been moved out of mdbook-markdown into mdbook-html. For now, all of the API surface is private, though we may consider ways to safely expose it in the future.
- Instead of using pulldown-cmark's html serializer, this takes the pulldown-cmark events and translates them into a tree data structure (using the ego-tree crate to define the tree). See `tree.rs`.
- HTML in the markdown document is parsed using html5ever, and then lives inside the same tree data structure. See `tokenizer.rs`.
- Transformations are then applied to the tree data structure. For example, adding header links or hiding code lines.
- Serialization is a simple process of writing out the nodes to a string. See `serialize.rs`.
- The search indexer works on the tree structure instead of re-rendering every chapter twice. See `html_handlebars/search.rs`.
- The print page now takes a very different approach of taking the same tree structure built for rendering the chapters, and applies transformations to it. This avoid re-parsing everything again. See `print.rs`.
    - I changed the linking behavior so that links on the print page link to items on the print page instead of outside the print page.
- There are a variety of small changes to how it serializes as can be seen in the changes to the tests. Some highlights:
	- Code blocks no longer have a second layer of `<pre>` tags wrapping it.
    - Fixed a minor issue where a rust code block with a specific edition was having the wrong classes when there was a default edition.
- Drops the ammonia dependency, which significantly reduces the number of dependencies. It was only being used for a very minor task, and we can handle it much more easily now.
- Drops `pretty_assertions`, they are no longer used (mostly being migrated to the testsuite).

There's obviously a lot of risk trying to parse everything to such a low level, but I think the benefits are worth it. Also, the API isn't super ergonomic compared to say javascript (there are no selectors), but it works well enough so far.

I have not run this through rigorous benchmarking, but it does have a very noticeable performance improvement, especially in a debug build.

I expect in the future that we'll want to expose some kind of integration with extensions so they have access to this tree structure (or some kind of tree structure).

Closes https://github.com/rust-lang/mdBook/issues/1736


max-heller:
This is exciting to see! [In `mdbook-pandoc`, I do something very similar](https://github.com/max-heller/mdbook-pandoc#html), I'd certainly be interested in seeing how much of this functionality could be shared between `mdbook` and backends.

PR #2858, opened 2025-09-21
Add cargo-semver-checks

ehuss:
This adds cargo-semver-checks to CI to help catch any unintended breaking changes to the API.


ehuss (commented):
I'm not 100% I've set this up correctly, but it seems to do what I expect. We may need to iterate on it. It seems to treat all pre-release versions as incompatible, so I don't think it will really do anything until 0.5 is released.

PR #2923, opened 2025-11-05
Hide the sidebar resize indicator when JS isn't available

LikeLakers2:
Hi there!

mdBook is set up to enable browsing of documentation, even when scripting isn't available. And for the most part, it works well enough.

However, #819 added a resize handle, and #2209 adds a small indicator for this handle. This resize handle requires Javascript in order to work properly - and otherwise does nothing but add visual clutter when scripting isn't available.

This PR adds a small style inside a `<noscript>` tag, that hides the resize indicator when scripting isn't available.

I'm unfortunately not too familiar with HTML/CSS conventions - most of my knowledge comes from browsing MDN. Plus, this is my first contribution to mdBook, and to the rust-lang org in general. So if there's a more preferred way of doing this, please let me know.

(P.S. I've allowed edits by maintainers on this PR, so feel free to use that if need be.)

### DIFF
--- a/crates/mdbook-html/front-end/css/chrome.css
+++ b/crates/mdbook-html/front-end/css/chrome.css
@@ -534,6 +534,11 @@ html:not(.sidebar-resizing) .sidebar {
     cursor: col-resize;
     width: calc(var(--sidebar-resize-indicator-width) - var(--sidebar-resize-indicator-space));
 }
+
+html:not(.js) .sidebar-resize-handle {
+    display: none;
+}
+
 /* sidebar-hidden */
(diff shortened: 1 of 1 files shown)


LikeLakers2:
Sorry for stopping the auto-merge - I just tested your change and it doesn't seem to work. Looking into why.

LikeLakers2:
Okay, I figured out what's going on. Apparently I was encountering a weird quirk with how I was blocking Javascript during the test, and how that interacts with `file:///` URLs.

When viewing the book via `mdbook serve` or `python -m http.server`, and blocking Javascript via that same method, the CSS you added seems to work.

@ehuss Please feel free to re-enable auto-merge.

PR #2069, opened 2023-04-16
Add AsciiMath support

expikr:
This pull request adds an option to let users additionally enable the more concise [AsciiMath](http://asciimath.org/) syntax supported by MathJax.

Usage:

```toml
[output.html]
mathjax-support = true
asciimath-support = true
```

To type an equation, the user simply use the <kbd>\\\`...\\\`</kbd> delimiter. 

So for example:

### Output:

image

### Input:

```
Euler's Formula \` e^(ix) = cosx + isinx \` defines fractional signs.

Matrix:

\`
[ [a,b],
  [c,d] ] 
\`

Fourier Transform:

\`
f(x) = int_(-infty)^(infty) F(s) e^(2 pi i x s) ds 
\`

\`
F(s) = int_(-infty)^(infty) f(x) e^(-2 pi i x s) dx
\`

Dirac Delta:

\`
delta (x) = {
    ( \infty , "," , x=0 ) ,
    ( 0      , "," , "otherwise" )
:} 
\`
```

PR #2968, opened 2025-11-29
Support dollar math syntax

### LINKED ISSUE #400: Improve MathJax support by enabling $$ for math equations
Currently, Mathjax is a little akward to use because we only support the `\( ... \)` and `\[ ... \]`method to indicate inline and block equations respectively. On top of that, backslashes need to be escaped so that the markdown parser doesn't remove them. Thus, equations need to be written like:

```
\\[ \mu = \frac{1}{N} \sum_{i=0} x_i \\]
```

So when mathjax support is enabled in the configuration, it would be great to add a preprocessor that rewrites the `\[ ... \]` forms to `\\[ ... \\]` and add support for the `$$ ... $$` math delimiters.

### LINKED ISSUE #662: Mathjax subscript colliding with markdown italics
## Output
screen shot 2018-04-03 at 2 10 52 pm

## Works
    ```
    \\[
    a_{n} = b_{n}
    \\]
    ```
    \\[
    a_{n} = b_{n}
    \\]

    ## Doesn't work
    ```
    \\[
    \bar{a}_{n} = b_{n}
    \\]
    ```
    \\[
    \bar{a}_{n} = b_{n}
    \\]

    ## Workaround
    ```
    \\[
    \bar{a}\_{n} = b_{n}
    \\]
    ```
    \\[
    \bar{a}\_{n} = b_{n}
    \\]

## Notes

- Sorry if this is a duplicate. I see that the MathJax support is in flux right now (though I couldn't find this particular issue).
- The workaround will not work in LaTeX. `$a\_{n}$` forces a literal underscore.
- You can reproduce the error with just `{a}_{n} = b_{n}` (I used `\bar{a}_{n} = b_{n}` because it's something someone might actually want to write).

### LINKED ISSUE #1402: MathJax Support works unreliable
Hello mdBook-Community!

I'm writing a technical documentation with many formulas. So far, it has all worked fine.
But now I have a problem with MathJax Support for only one formula. I try to type the following formula:
```md
\\( t_n = max({t_{rap}}_n, {t_{tar}}_{n} + t_{mis} \frac{{t_{res}}_n}{t_{res}}) \\)
```
On the rendere html-page I see this:
```
\( t_n = max({t_{rap}}n, {t{tar}}{n} + t{mis} \frac{{t_{res}}n}{t{res}}) \)
```
But expected is this:
formel

The formula works fine in DokuWiki with MathJax plugin but not in mdBook. So what is the problem? Or what doing I'm wrong?

Thank you!

jeanas:
Letting MathJax scrape the output HTML for math delimiters makes it awkward to write complex formulas because characters with a special meaning in Markdown must be escaped. This commit implements a new `output.html.math` option that uses the support for parsing `$...$` and `$$...$$` in Markdown introduced in pulldown-cmark 0.11.0. The old option `output.html.mathjax-support` is left for backwards compatibility.

pulldown-cmark renders formulas into `<span>` tags with special classes, so we have to configure MathJax to look for these instead of textual delimiters. The code for this was helpfully provided by David Cervone on <https://groups.google.com/g/mathjax-users/c/6cMuCH2dgmQ>.

The latest version of MathJax 4 is used rather than MathJax 2.7.1 like the mathjax-support option, because the way to do this configuration has changed between MathJax 2 and MathJax 3, I'm not sure how to do it on MathJax 2, and it is an old version anyway. The Cloudflare CDN doesn't seem to work for MathJax 4, so it uses the jsdelivr.net one recommended by the MathJax documentation. In the future, it would be good to use a bundled version of MathJax by default and/or make the CDN URL configurable.

Fixes #1402, fixes #662, fixes #400

### DIFF
--- a/guide/src/format/configuration/renderers.md
+++ b/guide/src/format/configuration/renderers.md
@@ -100,7 +100,7 @@ preferred-dark-theme = "navy"
 smart-punctuation = true
 definition-lists = true
 admonitions = true
-mathjax-support = false
+math = false
 additional-css = ["custom.css", "custom2.css"]
 additional-js = ["custom.js"]
 no-section-label = false
@@ -129,7 +129,7 @@ The following configuration options are available:
   Defaults to `true`.
 - **definition-lists:** Enables [definition lists](../markdown.md#definition-lists). Defaults to `true`.
 - **admonitions:** Enables [admonitions](../markdown.md#admonitions). Defaults to `true`.
-- **mathjax-support:** Adds support for [MathJax](../mathjax.md). Defaults to
+- **math:** Enables [support for math formulas](../mathjax.md). Defaults to
   `false`.
 - **additional-css:** If you need to slightly change the appearance of your book
   without overwriting the whole style, you can specify a set of stylesheets that
--- a/crates/mdbook-core/src/config.rs
+++ b/crates/mdbook-core/src/config.rs
@@ -460,7 +460,9 @@ pub struct HtmlConfig {
     pub definition_lists: bool,
     /// Support for admonitions.
     pub admonitions: bool,
-    /// Should mathjax be enabled?
+    /// Parse and render `$...$` and `$$...$$` as math formulas.
+    pub math: bool,
+    /// Legacy math support.
     pub mathjax_support: bool,
     /// Additional CSS stylesheets to include in the rendered page's `<head>`.
     pub additional_css: Vec<PathBuf>,
@@ -529,6 +531,7 @@ impl Default for HtmlConfig {
             smart_punctuation: true,
             definition_lists: true,
             admonitions: true,
+            math: false,
             mathjax_support: false,
             additional_css: Vec::new(),
             additional_js: Vec::new(),
--- a/crates/mdbook-html/front-end/templates/index.hbs
+++ b/crates/mdbook-html/front-end/templates/index.hbs
@@ -45,8 +45,36 @@
         <link rel="stylesheet" href="{{ resource this }}">
         {{/each}}
 
-        {{#if mathjax_support}}
-        <!-- MathJax -->
+        {{#if math}}
+        <!-- Load MathJax -->
+        {{! MathJax normally looks for math tags in the page content. In our case,
+            the math tags are parsed in the Markdown input by pulldown-cmark and
+            converted to <span> tags with special CSS classes. This bit of code
+            tells MathJax to render these and only these. Adapted from
+            <https://docs.mathjax.org/en/latest/upgrading/v2.html#math-script-example> }}
+        <script>
+        MathJax = {
+          options: {
+            renderActions: {
+              find: [10, function (doc) {
+                for (const node of document.querySelectorAll('span.math-inline, span.math-display')) {
+                  const display = node.classList.contains('math-display');
+                  const math = new doc.options.MathItem(node.textContent, doc.inputJax[0], display);
+                  const text = document.createTextNode('');
+                  node.parentNode.replaceChild(text, node);
+                  math.start = {node: text, delim: '', n: 0};
+                  math.end = {node: text, delim: '', n: 0};
+                  doc.math.push(math);
+                }
+              }, '']
+            }
+          }
+        };
+        </script>
+        <script defer src="https://cdn.jsdelivr.net/npm/mathjax@4/tex-mml-chtml.js"></script>
+        {{else if mathjax_support}}
+        <!-- Load MathJax -->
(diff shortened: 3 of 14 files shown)


jeanas:
Gentle ping? @ehuss

jeanas:
Ping?

GuillaumeGomez:
Hi, we created a new mdbook team to take over and I came across your PR. So a few questions: can this completely replace mathjax? Can you add some before/after differences (ie, how you would do it with this PR vs with mathjax)?

That would help a lot for the review.

jeanas:
@GuillaumeGomez Depends on what you mean by “completely replace MathJax”. In all cases, MathJax is being used in the reader's browser to render the math. The difference is that with the old option (which is still kept for compatibility), the author writes `\\( ... \\)`, the Markdown renderer doesn't recognize this as math or process it in any special way, this ends up as `\( ... \)` in the HTML, and then MathJax scrapes the whole HTML page on load to look for `\( ... \)` delimiters in text elements and replace them with rendered math in the DOM. On the other hand, with the new option, the author writes `$ ... $`, the Markdown renderer recognizes this as math and produces special HTML tags like `<span class="math math-inline">...</span>`, and then MathJax looks for these special tags.

The main benefit is that because this has support on the Markdown parser level, characters with a special meaning in Markdown are automatically escaped in `$...$`, so you can write e.g. `$\frac{u_n}{v_n}$` and not `\\(\\frac{u\_n}{v\_n}\\)` (and since `\` starts every LaTeX command and subscripts are used very often in formulas, this gets annoying quickly).

(There might also be a marginal benefit, which I didn't benchmark, from the fact that MathJax doesn't have to look for math delimiters through all text nodes in the HTML page anymore, it can use the browser's DOM APIs to just get all span tags identified by special CSS classes, which should be much faster, although it probably doesn't matter much in the grand scheme of things because that's usually not the most expensive step of math rendering.)

GuillaumeGomez:
Sounds quite close to https://github.com/rust-lang/rfcs/pull/3958. If it's what you're trying to achieve (sorry if I misunderstood), then we might want to get mdbook and rustdoc as close as possible for this feature.

cc @notriddle

jeanas:
(I rebased on `master` and made a tweak: after all there is no reason to prevent from using both the new support and the legacy one at the same time (might be useful to convert an existing book piecewise).)

GuillaumeGomez:
That's why I thought it was a complete replacement since it was an `else if` condition.

jeanas:
> Sounds quite close to [rust-lang/rfcs#3958](https://github.com/rust-lang/rfcs/pull/3958). If it's what you're trying to achieve (sorry if I misunderstood), then we might want to get mdbook and rustdoc as close as possible for this feature.
> 
> cc @notriddle

Interesting. I didn't know there were mature enough LaTeX to MathML renderers implemented in Rust, and that does sound like a superior option to me over MathJax (for error messages, avoiding flashes of raw syntax and readability in environments without JS), so I could update this PR to use the same library as `rustdoc` would with this proposal. Would you be happy with that?

GuillaumeGomez:
Yes, but maybe wait until the rustdoc side is implemented so you don't do work for nothing.

EDIT: Saying that because once RFC is merged, we might encounter unexpected issues in the implementation and might change a few things with some back and forth, and I don't want you to have to keep up.

notriddle:
I think, if this PR was changed to use math-core, I’d be happy to merge it now. Most of the discussion around rustdoc is about how it should be configured, which isn’t directly applicable to mdBook anyway, so waiting doesn’t buy you much.

GuillaumeGomez:
Just found out about https://github.com/rust-lang/mdBook/pull/2069 too.

guide/src/format/configuration/renderers.md
```diff
@@ -100,7 +100,7 @@ preferred-dark-theme = "navy"
 smart-punctuation = true
 definition-lists = true
 admonitions = true
-mathjax-support = false
+math = false
```
sanmai, line 103: It would be great to leave old `mathjax-support` config working for backward compatibility

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/2968/03018be55d62bce720c0a86e392039ce8a60809a..f6bbe0dd48816812135dc5d5ab6bb7e90731ce00)*

GuillaumeGomez: If we make a major release, do we care about backward compatibility?

jeanas: That's already what the PR does.

sanmai: My apologies, a very long day; indeed it does just that. Looking forward to see it merged.

PR #2991, opened 2025-12-19
Support for HTML Renderer to honor `.gitignore`

benfalk:
The Watcher and Poller already respected `.gitignore` files when monitoring for changes; however, the HTML Renderer did not. This update ensures that the HTML Renderer also honors `.gitignore`. This prevents ignored files from being processed and included in the generated HTML output.

I noticed this when working on a project where the mdbook source directory was the root of the project.  mdBook was taking a long time to copy the entire contents of the `target` directory into the book which was gigabytes in size.  After this change, the HTML renderer correctly ignored the `target` directory as specified in the `.gitignore` file, significantly speeding up the build process.


jhult:
Related PRs: #2247, #1908, and #3040.

PR #2993, opened 2025-12-24
Fix current page detection for clean URLs

### LINKED ISSUE #2962: Section Headers not Displaying if '.html' Stripped from URL
### Problem

If the built docs are being hosted on a site that automatically strips the `.html` from urls (e.g. `example.com/page.html` -> `example.com/page`) then the section headers aren't displayed for the current page and the current page isn't even highlighted.

Example with `.html` in URL:
Image

Same page w/o `.html` in URL:
Image

### Steps

1. Build mdBook docs
2. Host docs on website/service that removes `.html` from URLs
3. Navigate to a page with section headers

A good way to reproduce locally is `mdbook serve` (which has `.html`) vs. `npx serve` (which strips `.html`)
(My `npx serve` version is `14.2.0`)

### Possible Solution(s)

I fixed this problem with the following code in `toc.js`, albeit a bit hacky:

```diff
// Set the current, active page, and reveal it if it's hidden
let current_page = document.location.href.toString().split('#')[0].split('?')[0];
+ if (!current_page.endsWith(".html")) {
+   current_page += ".html";
+ }
if (current_page.endsWith('/')) {
    current_page += 'index.html';
}
```

This will essentially pretend the URL still has the `.html`, even if it doesn't actually. I saw no problems when testing this code, it does get wiped out whenever the docs are rebuilt though (understandably.)

### Notes

Another possible solution would be to add a flag that allows customization of the generated links. For example:
```toml
[output.html]
hrefs-include-extension=false # (where default would be true)
```

### Version

```text
mdbook v0.5.1
```

qaqland:
Previously, the sidebar JavaScript logic for detecting the "active" page relied on URL and assumed it would end with .html. When using clean URLs without this suffix, this logic would fail to correctly highlight the current page in the sidebar.

This change moves the current page identification to server-side rendering by adding a data-current-link attribute to the HTML root element.

Fixes https://github.com/rust-lang/mdBook/issues/2962

### DIFF
--- a/crates/mdbook-html/front-end/templates/toc.js.hbs
+++ b/crates/mdbook-html/front-end/templates/toc.js.hbs
@@ -10,10 +10,7 @@ class MDBookSidebarScrollbox extends HTMLElement {
     connectedCallback() {
         this.innerHTML = '{{#toc}}{{/toc}}';
         // Set the current, active page, and reveal it if it's hidden
-        let current_page = document.location.href.toString().split('#')[0].split('?')[0];
-        if (current_page.endsWith('/')) {
-            current_page += 'index.html';
-        }
+        const current_link = document.documentElement.dataset.currentLink;
         const links = Array.prototype.slice.call(this.querySelectorAll('a'));
         const l = links.length;
         for (let i = 0; i < l; ++i) {
@@ -23,10 +20,9 @@ class MDBookSidebarScrollbox extends HTMLElement {
                 link.href = path_to_root + href;
             }
             // The 'index' page is supposed to alias the first chapter in the book.
-            if (link.href === current_page
+            if (href === current_link
                 || i === 0
-                && path_to_root === ''
-                && current_page.endsWith('/index.html')) {
+                && current_link === 'index.html') {
                 link.classList.add('active');
                 let parent = link.parentElement;
                 while (parent) {
--- a/crates/mdbook-html/src/html_handlebars/hbs_renderer.rs
+++ b/crates/mdbook-html/src/html_handlebars/hbs_renderer.rs
@@ -85,6 +85,9 @@ impl HtmlHandlebars {
         ctx.data.insert("title".to_owned(), json!(title));
         ctx.data
             .insert("path_to_root".to_owned(), json!(fs::path_to_root(path)));
+        let current_link = path.with_extension("html").to_url_path();
+        ctx.data
+            .insert("current_link".to_owned(), json!(current_link));
         if let Some(ref section) = ch.number {
             ctx.data
                 .insert("section".to_owned(), json!(section.to_string()));
@@ -127,6 +130,8 @@ impl HtmlHandlebars {
             ctx.data.insert("path".to_owned(), json!("index.md"));
             ctx.data.insert("path_to_root".to_owned(), json!(""));
             ctx.data.insert("is_index".to_owned(), json!(true));
+            ctx.data
+                .insert("current_link".to_owned(), json!("index.html"));
             let rendered_index = ctx.handlebars.render("index", &ctx.data)?;
             debug!("Creating index.html from {}", ctx_path);
             fs::write(ctx.destination.join("index.html"), rendered_index)?;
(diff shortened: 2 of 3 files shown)


qaqland:
Ref: https://github.com/rust-lang/mdBook/pull/2570 and https://github.com/rust-lang/mdBook/pull/2415

qaqland:
@ehuss , could you please review this when you have a moment?

jhult:
Related PR: #3038

jonathanpallant:
I tested this and it works great.

See https://f74d75b4.ferrous-systems-rust-exercises.pages.dev/latest/book/nrf52/usb-exercise.

crates/mdbook-html/front-end/templates/toc.js.hbs
```diff
@@ -10,10 +10,7 @@ class MDBookSidebarScrollbox extends HTMLElement {
     connectedCallback() {
         this.innerHTML = '{{#toc}}{{/toc}}';
         // Set the current, active page, and reveal it if it's hidden
-        let current_page = document.location.href.toString().split('#')[0].split('?')[0];
-        if (current_page.endsWith('/')) {
-            current_page += 'index.html';
-        }
+        const current_link = document.documentElement.dataset.currentLink;
```
ehuss, line 13: Unfortunately this would be a breaking change if the user has a custom `index.hbs`. I try to be a little cautious about making those kinds of changes. 


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/2993/cc78fb4ff18b129099ce9daec49613f0a6b513e0..4d0d4e55a7f91a630578d759ffe539cbaf00846a)*

crates/mdbook-html/src/html_handlebars/hbs_renderer.rs
```diff
@@ -180,6 +185,7 @@ impl HtmlHandlebars {
         // Set a dummy path to ensure other paths (e.g. in the TOC) are generated correctly
         data_404.insert("path".to_owned(), json!("404.md"));
         data_404.insert("content".to_owned(), json!(html_content_404));
+        data_404.insert("current_link".to_owned(), json!("404.html"));
```
ehuss, line 188: Nothing to change here. Just noting that this isn't technically correct because the output filename can be customized. However, since there isn't anything in the sidebar for it to match, I don't think it matters.


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/2993/cc78fb4ff18b129099ce9daec49613f0a6b513e0..4d0d4e55a7f91a630578d759ffe539cbaf00846a)*

crates/mdbook-html/src/html_handlebars/hbs_renderer.rs
```diff
@@ -85,6 +85,9 @@ impl HtmlHandlebars {
         ctx.data.insert("title".to_owned(), json!(title));
         ctx.data
             .insert("path_to_root".to_owned(), json!(fs::path_to_root(path)));
+        let current_link = path.with_extension("html").to_url_path();
```
ehuss, line 88: I'm noticing that `current_link` isn't being set in the print page.

I don't think it matters too much because there is no sidebar navigation for the print page.

Offhand I don't think it is required, but I do have a little unease about having a missing variable.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/2993/cc78fb4ff18b129099ce9daec49613f0a6b513e0..4d0d4e55a7f91a630578d759ffe539cbaf00846a)*

ehuss:
Thanks for the PR! I've decided to go with a simpler solution in https://github.com/rust-lang/mdBook/pull/3028. 

I appreciate that this PR has a somewhat cleaner or more principled way of communicating the correct page to the TOC. However, I'm a little concerned with the additional complexity this adds. Hopefully the other change provides the functionality that you need. And if not, let me know.

There may be future extensions like https://github.com/rust-lang/mdBook/pull/3047 which would provide first-class support for clean urls, which would hopefully deal with redirects by avoiding them.

qaqland:
Thank you for your review; happy to see that the problem can be resolved.

PR #3001, opened 2026-01-03
Improve error when writing mdbook serve from nested folder

### LINKED ISSUE #3000: Improve error when writing `mdbook serve` from nested folder
### Problem

My girlfriend and I are using mdbook for our [personal cookbook](https://cooking.lquenti.de/) and it works great! She is non-technical, but I CI-automated the deployment so that once she pushes it automatically updates the deployment. The only issue is as follows:

- She opens the folder in kate, starts the embedded terminal
- Tries to `mdbook` something, with the help of `mdbook --help`
- Sees `mdbook serve`, and gets the following error

```bash
klara@zblech:~/Dokumente/data/GitProjekte/rezeptsammlung/src/Klara/Herzhaftes$ mdbook serve
2026-01-03 10:42:02 [ERROR] (mdbook::utils): Error: Couldn't open SUMMARY.md in "/home/klara/Dokumente/data/GitProjekte/rezeptsammlung/src/Klara/Herzhaftes/src" directory
2026-01-03 10:42:02 [ERROR] (mdbook::utils):    Caused By: No such file or directory (os error 2)
```

This is a confusing error message, as if the reader is not experienced with reading error traces, they think
- why does it fail, I can see the SUMMARY.md in  my text editor tree (as they are just skimming the path)
- Of course it fails, I don't have this `src` folder! (As they might not have an understanding how it internally builds the path)

I think this would be easy to improve with a hint

### Proposed Solution

on `ENOENT` error handling (i.e. we are already on the slot path), recursively walk up the path to see whether a parent directory contains a `src/SUMMARY.md`. If so, print another `[INFO]` log on whether the user might meant to start it from the parent folder instead.

### Notes

_No response_

lquenti:
Fixes #3000

I hope this is the correct place to fix, doing it directly in the serve execute function seemed weird as there its not trivial to me that it being a `NotFound` error implies that its about the SUMMARY.md.

### DIFF
--- a/crates/mdbook-driver/src/load.rs
+++ b/crates/mdbook-driver/src/load.rs
@@ -4,14 +4,41 @@ use mdbook_core::config::BuildConfig;
 use mdbook_core::utils::{escape_html, fs};
 use mdbook_summary::{Link, Summary, SummaryItem, parse_summary};
 use std::path::Path;
-use tracing::debug;
+use tracing::{debug, info};
 
 /// Load a book into memory from its `src/` directory.
 pub(crate) fn load_book<P: AsRef<Path>>(src_dir: P, cfg: &BuildConfig) -> Result<Book> {
     let src_dir = src_dir.as_ref();
     let summary_md = src_dir.join("SUMMARY.md");
 
-    let summary_content = fs::read_to_string(&summary_md)?;
+    let summary_content = match fs::read_to_string(&summary_md) {
+        Ok(content) => content,
+        Err(e) => {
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks! Can you please add a test for this? I think it would probably go in `tests/testsuite/build.rs`. The test would probably look something like `basic_build`, but would just need to call `current_dir()` to some subdirectory, and include `expect_failure()`. There's documentation at https://github.com/rust-lang/mdBook/blob/master/tests/testsuite/README.md on how to update the expected output.

PR #3013, opened 2026-01-30
Fix internal links involving `README` files resulting in 404 errors

### LINKED ISSUE #984: Internal links to README.md are broken

CCJK123:
Fixes #984.

Did my best to address the concerns mentioned in #1921, by opting to:
- Use the `mdbook-markdown`/`pulldown-cmark` and `url` crates instead of regex to identify and correct the relevant links ([comment][1] advising against regex usage)
- Modify the `index` preprocessor instead of `fix_link` in `mdbook-html` (previously `adjust_links`) to maintain clear separation between the preprocessor and the rest of the codebase ([comment][2] mentioning this consideration)

Also duplicated some code from `HtmlRenderOptions` to get the markdown options for creating a parser via `mdbook-markdown`, not sure whether this should be abstracted out elsewhere. The testcases I wrote just use the default options.

Hopefully this sufficiently addresses the issue, but do lmk if there's any changes/improvements I shd make!

[1]: https://github.com/rust-lang/mdBook/pull/1921#issuecomment-1384217793
[2]: https://github.com/rust-lang/mdBook/pull/1921#issuecomment-1614491392

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -394,6 +394,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "displaydoc"
+version = "0.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "97369cbbc041bc366949bc74d34658d6cda5621039731c6310521892a3a20ae0"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.112",
+]
+
+[[package]]
 name = "dunce"
 version = "1.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


CCJK123:
Realised that this approach seems to be lossy due to the current way `pulldown-cmark` and `pulldown-cmark-to-cmark` work, and can result in some information being lost, and possibly resulted in the following test failures:
- `markdown::basic_markdown`: A `&quot;` gets converted to `“`, though there are other `&quot;`s in the same test that aren't converted; disabling `smart_punctuation` does not resolve the issue
- `markdown::definition_lists`: `<p>` tag is not preserved within `<dd>` fsr
- `preprocessor::extension_compatibility` and `renderer::backends_receive_render_context_via_stdin`: Newlines at the end of markdown content get removed (see https://github.com/Byron/pulldown-cmark-to-cmark/issues/94)
- `rendering::header_links`: Nesting order of `<strong>` and `<em>` tags was reversed, inconsequential in this case but not necessarily so in other cases

Perhaps regex is still a viable approach, though it'll need to account for the various types of links. Tolerating some information loss is also an option. Otherwise, the issue will continue to be unresolved due to being blocked by `pulldown-cmark`/`pulldown-cmark-to-cmark`.

Would appreciate input on how best to proceed, thanks!

Issue #3015, opened 2026-01-31
Allow combining themes (output.html.theme and theme/)


matheusds4:
### Problem

I'm trying mdBook 0.5.2 with `output.html.theme = "third_party/theme"` and a `theme/` directory and the latter is completely ignored. The latter contains only the favicon.

### Proposed Solution

Allow `theme/` to override `output.html.theme` specified theme.

PR #3024, opened 2026-02-05
allow overriding subtheme

ghost:
This implements the feature requested in https://github.com/rust-lang/mdBook/issues/3015

- Continues allowing user to specify a theme directory different from `theme/`
  - Combines certain files from `theme/` if existing
    - Favicons (overrides entirely)
    - head.hbs (combines by the UTF-8 encoding)
    - highlight.js (overrides entirely)

This is useful when you're auto integrating an external theme, like [Metro design](https://github.com/sweaxizone/metro-mdbook), where you don't need to modify the external theme's files.


ghost:
@ehuss Hmm, sorry to annoy, but is there any reason why this can't be merged?

Issue #3026, opened 2026-02-08
HTML semantics for admonitions


orangecms:
### Problem

With #2851 came support for admonitions. :tada:
They are currently being rendered using `<blockquote>` elements, albeit being remarks/notes here.
That element is meant for actual quotes; see https://html.spec.whatwg.org/multipage/grouping-content.html#the-blockquote-element

GitHub themselves are using `<div>` elements, which has no semantics at all though; see https://gist.github.com/cseeman/8f3bfaec084c5c4259626ddd9e516c61 (open browser dev tools).

There is unfortunately no suitable semantic HTML element, as of now.
However a semantic _role_ attribute value exists:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/note_role (even mentions `blockquote` to be used where suitable)
this role is e.g. used in the WCAG guidlines; see e.g. the note under https://www.w3.org/TR/WCAG21/#name-role-value

... coming from https://github.com/kubernetes-sigs/kubebuilder/pull/5457 😅

### Proposed Solution

Switch to `<div role="note">`. I will happily file a PR. :)

Delapouite:
Related discussion happening in the WHATWG : https://github.com/whatwg/html/issues/10100

miikkas:
Although I've seen `<div>` used for admonitions in some existing implementations, I think `<aside>`[^1] could be considered as well. That would decrease the risk of creating a "div soup", and seems like it could be acceptable for the purpose[^2].

(I gave a similar comment for pulldown-cmark in <https://github.com/pulldown-cmark/pulldown-cmark/issues/1068#issuecomment-5378414964>.)

[^1]: <https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/aside>
[^2]: <https://stackoverflow.com/questions/7579826/which-semantic-html-tag-for-displaying-side-notes-and-admonitions>

PR #3027, opened 2026-02-08
use <div role="note"> instead of <blockquote> for admonitions

orangecms:
Also, render the 'title' in a `<span>` instead of a `<p>`. This is semantically closer to what assistive tooling supports. For reference, see also:
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/note_role
- https://www.w3.org/TR/WCAG21/#name-role-value (even has a note)

see #3026

### DIFF
--- a/crates/mdbook-html/front-end/css/general.css
+++ b/crates/mdbook-html/front-end/css/general.css
@@ -156,8 +156,7 @@ blockquote {
     border-block-end: .1em solid var(--quote-border);
 }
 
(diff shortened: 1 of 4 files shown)


orangecms:
I also filed this against the underlying parser:
https://github.com/pulldown-cmark/pulldown-cmark/pull/1069

ehuss:
Seems reasonable. I'm trying to think about this being somewhat of a breaking change since people may have custom styling that would then break. I'll think about it more, but this may need to wait for 0.6.

PR #3028, opened 2026-02-09
fix: pretty url toc highlighting

### LINKED ISSUE #3034: TOC highlighting breaks on prettified URLs
### Problem

Some static website hosting services (like Cloudflare Pages) reroute URLs to "pretty URLs", removing things like the .html suffix. 

See https://cloesce.pages.dev/ for an example of the TOC highlighting failing (works only on the index page)

### Steps

Host an mdBook website on a service that defaults to pretty URLs

bens-schreiber:
This fix compares the current page to the active page, ignoring the `.html` suffix.

fixes #3034

### DIFF
--- a/crates/mdbook-html/front-end/templates/toc.js.hbs
+++ b/crates/mdbook-html/front-end/templates/toc.js.hbs
@@ -23,7 +23,7 @@ class MDBookSidebarScrollbox extends HTMLElement {
                 link.href = path_to_root + href;
             }
             // The 'index' page is supposed to alias the first chapter in the book.
-            // We check both with and without the '.html' suffix to be robust against pretty URLs
+            // Check both with and without the '.html' suffix to be robust against pretty URLs
             if (link.href.replace(/\.html$/, '') === current_page.replace(/\.html$/, '')
                 || i === 0
                 && path_to_root === ''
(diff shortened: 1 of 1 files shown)


bens-schreiber:
It would be fair to say this is a server side problem and shouldn't be catered to. [Cloudflare Pages doesn't allow that](https://community.cloudflare.com/t/prevent-truncating-and-removal-of-page-name-extensions/388845) however. 

If this can't be merged, I might explore some kind of custom extension to mdbook.

jhult:
Related PRs: #3038, #2993, #2570

dv1x3r:
Thank you! As a workaround, it is also possible to apply this patch using the Makefile command:

```sh
.PHONY: docs
docs:
	mdbook build docs/
	find docs/book -name 'toc*.js' -print0 | xargs -0 sed -i 's@if (link.href === current_page@if (link.href.replace(/\.html$$/, "") === current_page.replace(/\.html$$/, "")@g'
```

ehuss (approved):
Thanks!

(Normally we would require tests for changes, but this is somewhat tricky to test since it would require a server that does the corresponding redirects. I'm fine with skipping that for now.)


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3028/b8c90970a63966154b19b4c0a1e4d6ba52fe6375..07cda662ded5268009acb7d38a5c7ce102369efb)*

PR #3040, opened 2026-02-25
feat: ignore files listed in `.mdbookignore` during build

jhult:
Rebase/update of https://github.com/rust-lang/mdBook/pull/1908 on top of `master`.

### DIFF
--- a/crates/mdbook-core/src/utils/fs.rs
+++ b/crates/mdbook-core/src/utils/fs.rs
@@ -238,6 +238,7 @@ fn copy<P: AsRef<Path>, Q: AsRef<Path>>(from: P, to: Q) -> Result<()> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use ignore::gitignore::GitignoreBuilder;
     use std::io::Result;
     use std::path::Path;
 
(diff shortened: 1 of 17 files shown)


jhult:
@ehuss, I'd appreciate a review when you get a chance.

jhult:
Other related PRs: #2247 and #2991.

jhult:
@ehuss - any thoughts on this?

crates/mdbook-html/src/html_handlebars/hbs_renderer.rs
```diff
@@ -444,7 +445,31 @@ impl Renderer for HtmlHandlebars {
...
-        fs::copy_files_except_ext(&src_dir, destination, true, Some(&build_dir), &["md"])?;
+        let mut builder = GitignoreBuilder::new(&src_dir);
+
+        // Add default ignores
+        builder.add_line(None, "/.mdbookignore")?;
+        builder.add_line(None, "/.gitignore")?;
+        builder.add_line(None, "/.git/")?;
+        builder.add_line(None, "/book.toml")?;
+        builder.add_line(None, "/theme/")?;
+        builder.add_line(None, "*.md")?;
```
ehuss, line 456: I'm not sure why all of these extra ignores were added. For example, this would break a book that has images or other content inside a directory called theme. Can this be trimmed down to just *.md and `.mdbookignore`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3040/df53063feaafede690c63c21877d3c2df454e3b1..9620641a18091ef61a3e69d552bb83c03ce0cd75)*

tests/testsuite/build.rs
```diff
@@ -85,3 +85,18 @@ fn dest_dir_relative_path() {
...
+    let mut test = BookTest::from_dir("build/mdbookignore");
+    test.build();
+    // Single file listed by name should not be copied.
+    assert!(!test.dir.join("book/ignored_file").exists());
+    // Files matching *.txt glob pattern should not be copied.
+    assert!(!test.dir.join("book/ignored.txt").exists());
+    // Directories listed in .mdbookignore should not be copied.
+    assert!(!test.dir.join("book/ignored_dir").exists());
+    // Non-ignored files should be copied.
+    assert!(test.dir.join("book/included.json").exists());
```
ehuss, line 101: There's some files in the book that this test isn't checking. For example, `file.json`, `custom.css`, etc. Can we make sure everything gets checked?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3040/df53063feaafede690c63c21877d3c2df454e3b1..9620641a18091ef61a3e69d552bb83c03ce0cd75)*

crates/mdbook-html/src/html_handlebars/hbs_renderer.rs
```diff
@@ -444,7 +445,31 @@ impl Renderer for HtmlHandlebars {
...
+
+        let mdbook_ignore = src_dir.join(".mdbookignore");
+        if mdbook_ignore.exists() {
+            if let Some(err) = builder.add(mdbook_ignore) {
+                warn!("Unable to load '.mdbookignore' file: {}", err);
```
ehuss, line 461: Why would this be a warning and not an error?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3040/df53063feaafede690c63c21877d3c2df454e3b1..9620641a18091ef61a3e69d552bb83c03ce0cd75)*

crates/mdbook-core/src/utils/fs.rs
```diff
@@ -131,16 +150,25 @@ pub fn copy_files_except_ext(
...
+                }
+            }
+
             // check if output dir already exists
             if !target_file_path.exists() {
                 fs::create_dir(&target_file_path)?;
             }
 
-            copy_files_except_ext(&entry, &target_file_path, true, avoid_dir, ext_blacklist)?;
+            copy_files_except_ignored(&entry, &target_file_path, true, avoid_dir, ignore)?;
```
ehuss, line 166: The way this handles relative paths seems to break root-anchored patterns like `/subdir/foo.txt`. 

When descending, `from` is now the subdirectory (e.g., `src/subdir/`), but `ignore` was built with the original root (e.g., `src`). In the recursive invocation, `pathdiff::diff_paths(&entry, from)` computes the path relative to the subdirectory, not relative to the gitignore root. So a root-anchored pattern like `/subdir/file.txt` will never match `file.txt` once inside `subdir/`. 

I think maybe the fix is to always compute relative paths against `ignore.path()` (the gitignore's own root), not against `from`.

Also, can you add some tests that include this scenario?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3040/df53063feaafede690c63c21877d3c2df454e3b1..9620641a18091ef61a3e69d552bb83c03ce0cd75)*

crates/mdbook-core/src/utils/fs.rs
```diff
@@ -131,16 +150,25 @@ pub fn copy_files_except_ext(
                 }
             }
 
+            if let Some(ignore) = ignore {
+                let relative_path =
+                    pathdiff::diff_paths(&entry, from).unwrap_or_else(|| entry.to_path_buf());
```
ehuss, line 155: Can you say more why pathdiff is needed? Would it be possible to drop the `pathdiff` dependency by using `entry.strip_prefix(from)`? Entries should always be under `from`, so I'm not sure what scenario this is needed.


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3040/df53063feaafede690c63c21877d3c2df454e3b1..9620641a18091ef61a3e69d552bb83c03ce0cd75)*

PR #3046, opened 2026-03-03
fix(html): rewrite CSS url() references to use hashed filenames

### LINKED ISSUE #2958: Local custom font files are serialized since version 0.5 - how to reference them?
### Question

Local custom font files are now serialized, which results in files like `fonts/fira-code-v27-latin-300-0fc02cbc.woff2` (`fira-code-v27-latin-300.woff2` is the original filename) in the output/html folder. The serialized CSS file (like fonts/fonts-3c66ea91.css) is linked correctly in the HTML files, but the fonts can't be loaded because the paths to the font files inside the CSS are not correct, they still point to the filenames without the serial. Maybe I just missed something, any hints?

### Version

```text
mdbook v0.5.1
```

jhult:
Fixes #2958

When font files are hashed for cache-busting, CSS `url()` references were not updated to use the hashed filenames, causing custom fonts to fail loading.

This fix rewrites `url()` references in CSS files during `write_files`, handling all three quoting styles (`url("…")`, `url('…')`, `url(…)`), resolving paths relative to the CSS file's location before looking up in the hash map, and leaving absolute/data URLs untouched.

### DIFF
--- a/crates/mdbook-html/src/html_handlebars/static_files.rs
+++ b/crates/mdbook-html/src/html_handlebars/static_files.rs
@@ -161,19 +161,23 @@ impl StaticFiles {
 
             this.static_files.push(StaticFile::Additional {
                 input_location,
+                // Always use forward slashes for consistent hash_map keys across platforms.
                 filename: custom_file
                     .to_str()
                     .with_context(|| "resource file names must be valid utf8")?
-                    .to_owned(),
+                    .replace('\\', "/"),
             });
         }
 
         for input_location in theme.font_files.iter().cloned() {
-            let filename = Path::new("fonts")
-                .join(input_location.file_name().unwrap())
-                .to_str()
-                .with_context(|| "resource file names must be valid utf8")?
-                .to_owned();
+            let filename = format!(
+                "fonts/{}",
+                input_location
+                    .file_name()
+                    .unwrap()
+                    .to_str()
+                    .with_context(|| "resource file names must be valid utf8")?
+            );
             this.static_files.push(StaticFile::Additional {
                 input_location,
                 filename,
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Sorry for the delay! We now have an mdbook team so review should hopefully be a bit faster. It's a big PR so I'll try to come back to it soon.

crates/mdbook-html/src/html_handlebars/static_files.rs
```diff
@@ -195,12 +256,19 @@ impl StaticFiles {
         // The `{{ resource "name" }}` directive in static resources look like
         // handlebars syntax, even if they technically aren't.
         static_regex!(RESOURCE, bytes, r#"\{\{ resource "([^"]+)" \}\}"#);
+        // CSS url() with double-quoted, single-quoted, or unquoted paths.
+        // Capture groups: 1=double-quoted path, 2=single-quoted path, 3=unquoted path.
+        static_regex!(
+            CSS_URL,
+            bytes,
+            r#"url\(\s*(?:"([^"]*?)"|'([^']*?)'|([^'"\s()]+))\s*\)"#
```
GuillaumeGomez, line 264: Instead of using a `regex`, why not add in the CSS files something like:

```css
property: /* hashed-url-file: original-filename */;
```

Like that we only need to look for `/* hashed-url-file: ` and replace the given filename with its hashed version.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3046/df53063feaafede690c63c21877d3c2df454e3b1..e375fc42987a09368cfdb67d2661ab93056e8dfa)*

PR #3047, opened 2026-03-08
feat: URLs with no index.html suffix 

### LINKED ISSUE #1251: Remove index.html suffix
As most static file servers will automatically serve `index.html`, it would be nice to have an option where you can use `/` instead of `/index.html`

jhult:
Fixes #1251

```toml
[output.html]
no-html-extension=true
```

```md
**no-html-extension:** When set to `true`, generates clean URLs without `.html` extensions. All pages are written to directory-based paths (e.g., `chapter/index.html` instead of `chapter.html`).

Special pages are placed in `_print/`, `_toc/`, and `_404/` directories.

All internal links use trailing-slash format (`chapter/` instead of `chapter.html`).

Works with all major static hosting providers (GitHub Pages, Netlify, Cloudflare Pages) and the built-in `mdbook serve` command.

**Note:** Chapter paths starting with `_print`, `_toc`, or `_404` will cause a build error.

Defaults to `false`.
```

Vaguely related PRs: https://github.com/rust-lang/mdBook/pull/2570, https://github.com/rust-lang/mdBook/pull/2993, https://github.com/rust-lang/mdBook/pull/3028, https://github.com/rust-lang/mdBook/issues/3034, https://github.com/rust-lang/mdBook/pull/3038

### DIFF
--- a/crates/mdbook-core/src/config.rs
+++ b/crates/mdbook-core/src/config.rs
@@ -518,6 +518,9 @@ pub struct HtmlConfig {
     /// If enabled, the sidebar includes navigation for headers on the current
     /// page. Default is `true`.
     pub sidebar_header_nav: bool,
+    /// If enabled, HTML files will not have the `.html` extension in URLs.
+    /// Defaults to `false`.
+    pub no_html_extension: bool,
 }
 
 impl Default for HtmlConfig {
@@ -548,6 +551,7 @@ impl Default for HtmlConfig {
             redirect: HashMap::new(),
             hash_files: true,
             sidebar_header_nav: true,
+            no_html_extension: false,
         }
     }
 }
--- a/crates/mdbook-html/front-end/templates/index.hbs
+++ b/crates/mdbook-html/front-end/templates/index.hbs
@@ -129,7 +129,7 @@
             <!-- populated by js -->
             <mdbook-sidebar-scrollbox class="sidebar-scrollbox"></mdbook-sidebar-scrollbox>
             <noscript>
-                <iframe class="sidebar-iframe-outer" src="{{ path_to_root }}toc.html"></iframe>
+                {{#if no_html_extension}}<iframe class="sidebar-iframe-outer" src="{{ path_to_root }}_toc/"></iframe>{{else}}<iframe class="sidebar-iframe-outer" src="{{ path_to_root }}toc.html"></iframe>{{/if}}
             </noscript>
             <div id="mdbook-sidebar-resize-handle" class="sidebar-resize-handle">
                 <div class="sidebar-resize-indicator"></div>
@@ -168,7 +168,7 @@
 
                     <div class="right-buttons">
                         {{#if print_enable}}
-                        <a href="{{ path_to_root }}print.html" title="Print this book" aria-label="Print this book">
+                        {{#if no_html_extension}}<a href="{{ path_to_root }}_print/" title="Print this book" aria-label="Print this book">{{else}}<a href="{{ path_to_root }}print.html" title="Print this book" aria-label="Print this book">{{/if}}
                             {{fa "solid" "print" "print-button"}}
                         </a>
                         {{/if}}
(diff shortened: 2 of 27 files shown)


bens-schreiber:
Certainly fixes my issue #3028 . Would love to get this in.

jhult:
@ehuss or @Dylan-DPC, I think many people would appreciate this being merged. I'd love any feedback you have to share.

CattoYT:
Just had to hop into the discussion and say that this fixed my issue with cloudflare pages

Before: https://ad2d70e6.csbook-a6u.pages.dev/
After: https://215b27f3.csbook-a6u.pages.dev/

Thanks for your work <3

jonathanpallant:
If I've read this right, the `no-html-extension` property doesn't remove the `.html` extension from the output files. Instead, it names file `foobar/index.html` instead of `foobar.html`, which most web servers will serve as `/foobar/`.

So, I wonder if the property needs a better name, to avoid confusing people who will see files on disk that still have a `.html` extension?

jhult:
@jonathanpallant - this is true. I'm open to suggestions for a better name if you have one.

szabgab:
It seems they are called [Clean URLS](https://en.wikipedia.org/wiki/Clean_URL)  so the option could be `clean_urls`

However, `toc.html` was replaced by `_toc`.  Same with `print.html`. Why is the `_` necessary?

jhult:
All URLs are "clean" including `toc`, `print`, and `404`. The question is, what do we call them? If we use `toc`, then no chapter can be named that. Thus, the appended underscore. However, I'm open to suggestions on how to better name these.

szabgab:
Oh, I though the question was how to call the option that controls this behavior. For that I suggested `clean_urls`.

jhult:
I'm open to `clean_urls` / `clean-urls` as the overall enablement for this feature.

My other question is what should we do with `toc`, `print`, and `404`. Should we prefix them with an underscore? Whatever URL we use, no chapter can be named the same. Thus, my reason for prefixing them with an underscore. However, I'm open to suggestions on how to better name these 3 pages.

szabgab:
Regarding the `toc`, `print`, and 404` pages, I am trying to understand what is the difference in behavior in the clean url version?

So I checked what happens now: (mdbook v0.5.2)

If I create a file called `src/404.md` it will replace the default 404 page. this is [documented](https://rust-lang.github.io/mdBook/continuous-integration.html?highlight=404#deploying).

If I create a file called `src/print.md` its content is included in the searchindex file and the build process will fail with an error `print.md is reserved for internal use`

If I create a file called `src/toc.md` it will replace the toc bar for people using no JavaScript, but otherwise will have no impact. I don't think this is documented.

So I think we should keep the names without the `_` but in a separate PR we should also at lest document the feature / bug. (We might even want to forbid the creation of `toc.md` and `print.md`, unless there is a value in them that I could not understand.

ps. Sorry for making two unrelated comments in one message earlier.

jonathanpallant:
So I've thought about this a bit, and I don't love how it's going to fundamentally change the on-disk structures of my rendered book. I have a *lot* of old copies of the book, and it seems useful that the changes between versions are all currently relatively small.

For my specific issue, #2993 is a much less invasive fix. Perhaps we can just do both - then I get a free choice about whether to enable this option or not?

jhult:
I didn't write the other PRs, but am not opposed to having both. Options can be nice.

GuillaumeGomez:
I'm not a big fan of this feature as it would be different from how rustdoc works too (and we're trying to get them kinda close).

@rfcbot fcp close

notriddle:
I'm not actually against the idea behind this PR.

If we're going to make every chapter it's own directory, we should "just do that." Don't even make it a config option (and generate redirects so that all the links that exist *now* aren't broken). Make `some-chapter/index.md` and `some-chapter.md` generate the same HTML, directly analogous to how `some_module/mod.rs` and `some_module.rs` are synonyms. "Clean URLs" aside, it's valuable to let authors restructure their book's filesystem layout without forcing them to break existing links or maintain redirect tables by hand.

If we make that change, then the config option to omit the `index.html` part of all the paths becomes almost trivial.

PR #3049, opened 2026-03-09
feat(core): expose `init_logger` publicly

### LINKED ISSUE #2049: Exposing `init_logger` to the public API for uniform logging across preprocessors
### Problem

Currently, all preprocessors need to call their own logging library. As calls to preprocessors are done within the main mdBook process, it would make sense to have a uniform logger across all preprocessors, so that when they log messages, it does not look weird compared to mdBook's own logging.

### Proposed Solution

Move `main::init_logger` to `utils::init_logger` and make it public, so that people can opt in for the same logger as mdBook.

https://github.com/rust-lang/mdBook/blob/efb671aaf241b7f93597ac70178989a332fe85e0/src/main.rs#L97-L121

### Notes

_No response_

jhult:
Preprocessor authors can now call `mdbook_core::utils::init_logger()` to use mdBook's logger configuration (`MDBOOK_LOG` env var, `INFO` default, silenced noisy deps like handlebars/html5ever).

Closes #2049

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1068,7 +1068,6 @@ dependencies = [
  "tokio",
  "tower-http",
  "tracing",
- "tracing-subscriber",
  "walkdir",
 ]
 
@@ -1087,6 +1086,7 @@ dependencies = [
  "tempfile",
  "toml",
  "tracing",
+ "tracing-subscriber",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


jhult:
@GuillaumeGomez, can you take a look at this?

GuillaumeGomez:
That needs to be documented too.

PR #3059, opened 2026-04-03
Try to make heading-nav-collapsed test more reliable

ehuss:
This test seems to be a little flaky. This tries to make it more reliable by using `wait-for-` commands instead of `assert-`. My guess is that clicking on an element doesn't necessarily wait for all the javascript code to finish and the DOM to be done updating.


ehuss:
Closing, merged in https://github.com/rust-lang/mdBook/pull/3074.

PR #3063, opened 2026-04-10
CI: Pin GitHub Actions to commit SHAs

Turbo87:
This pins all third-party actions to immutable commit SHAs, with the resolved version tag in a trailing comment. This prevents upstream tags from silently changing under us.

- actions/checkout → v6.0.2
  (bumped from v5 to reduce `artipacked` severity: v6 no longer stores credentials in `.git/config`)
- actions/setup-node → v6.3.0
- rust-lang/crates-io-auth-action → v1.0.4

See https://docs.zizmor.sh/audits/#unpinned-uses

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -30,7 +30,7 @@ jobs:
             os: windows-latest
     name: Deploy ${{ matrix.target }}
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
     - name: Install Rust
       run: ci/install-rust.sh stable ${{ matrix.target }}
     - name: Build asset
@@ -43,7 +43,7 @@ jobs:
     name: GitHub Pages
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
     - name: Install Rust (rustup)
       run: rustup update stable --no-self-update && rustup default stable
     - name: Deploy the User Guide to GitHub Pages using the gh-pages branch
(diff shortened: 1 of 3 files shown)


ehuss:
Thanks. I don't think we can update the checkout action to v6 yet without some other fixes.

Turbo87:
> I don't think we can update the checkout action to v6 yet without some other fixes.

hmm, how so? did you notice any issues with it?

ehuss:
> hmm, how so? did you notice any issues with it?

v6 changed the way git permissions are stored. There is a scheduled job here that pushes to a branch that I *think* may break (similar things broke in other repos). But I haven't had the time to test it or figure out how to fix it.

ehuss:
Also, do you happen to know if the renovate configuration needs to be updated to use this style?

Turbo87:
> v6 changed the way git permissions are stored. There is a scheduled job here that pushes to a branch that I _think_ may break.

I was aware of the change, which was the main reason for the simultaneous upgrade, but I wasn't aware that stuff would actually break due to it. I was under the impression that the git credentials path change was transparent/opaque/whatever-the-right-term-is and a regular `git push` would just keep working without changes.

https://github.com/rust-lang/mdBook/blob/05fbc5d3d0310100bb2138ebb849058ff35e882a/.github/workflows/update-dependencies.yml#L3-L4 is the only scheduled job I've found, so I'm assuming you're talking about that one. the workflow does not write anything itself but calls `ci/update-dependencies.sh`, which then uses `git push` and `gh pr create`, which I would assume should just keep working.

https://github.com/rust-lang/mdBook/blob/05fbc5d3d0310100bb2138ebb849058ff35e882a/ci/update-dependencies.sh#L40-L44

> similar things broke in other repos

do you have any examples that I could look at?

> do you happen to know if the renovate configuration needs to be updated to use this style?

the style should [match what renovate does natively](https://docs.renovatebot.com/modules/manager/github-actions/#digest-pinning-and-updating), and is AFAIU also supported by dependabot

ehuss:
Unfortunately the logs are long gone that had the authentication errors.

I did some testing, and this seems to be working, so let's go ahead and try.

ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3063/05fbc5d3d0310100bb2138ebb849058ff35e882a..2ab29ef789748d7c8b75e1ba4ebbd5b5b1f04381)*

PR #3065, opened 2026-04-10
fix: prevent '?' shortcut from intercepting input in code editors

### LINKED ISSUE #3064: Shortcuts intercepting '?' in interactive exercises/playgrounds
### Problem

The `?` keyboard shortcut (Help menu) intercepts the keypress globally. It is a "breaking" issue for **Rust by Example** because typing the `?` operator is required for many of the code exercises (error handling).

Currently, if you are solving a "Todo" in the embedded editor and try to use the `?` operator, the help popup opens and the character is never inserted into the code.

**Expected Behavior**
Shortcuts should be suppressed if the focus is within an editable element, including inputs, textareas, or `contenteditable` containers used by code editors.

### Steps

1. Open any Rust by Example page with an exercise (e.g., [Hello World](https://doc.rust-lang.org/rust-by-example/hello.html)).
2. Click into the interactive code block to edit.
3. Try to type `?`.
4. **Result:** The Help modal opens; the editor remains unchanged and the character is not typed.

### Possible Solution(s)

The `keydown` listener in `book.js` needs a guard clause to check the event target before calling `e.preventDefault()`.

```javascript
// Suggested guard at the start of the keydown handler
const target = e.target;
const isEditable = target.tagName === 'INPUT' || 
                   target.tagName === 'TEXTAREA' || 
                   target.isContentEditable ||
                   (target.closest && target.closest('.ace_editor'));

if (isEditable) { return; }

### Notes

* **Environment:** Affects all mdBook sites with interactive elements. Reproducible on Chrome/Firefox/Safari.
* **Willingness to contribute:** I have the fix ready and can submit a PR if this issue is confirmed.

### Version

```text
0.5.2
```

kartik-py12:
This adds a guard to the global keydown listener in book.js so that the '?' shortcut does not intercept keypresses when the user is focused inside an input, textarea, or Ace editor block.

Fixes #3064 and https://github.com/rust-lang/rust-by-example/issues/1984

### DIFF
--- a/crates/mdbook-html/front-end/js/book.js
+++ b/crates/mdbook-html/front-end/js/book.js
@@ -706,10 +706,20 @@ aria-label="Show hidden lines"></button>';
 
         // Usually needs the Shift key to be pressed
         switch (e.key) {
-        case '?':
+        case '?': {
+            const active = document.activeElement;
+            if (active &&
+                (
+                    active.tagName === 'INPUT' ||
+                    active.tagName === 'TEXTAREA' ||
+                    active.isContentEditable
+                )) {
+                return;
+            }
             e.preventDefault();
             showHelp();
             break;
+        }
         }
 
         // Rest of the keys are only active when the Shift key is not pressed


ehuss:
Thanks for the PR! I'm going to close as this is resolved by https://github.com/rust-lang/mdBook/pull/3087 where it goes with a slightly different solution.

Issue #2830, opened 2025-09-12
Multines include


hg-anssi:
### Problem

Multilines includes which are nested into markdown structures (like quote blocks for instance) brake the markdown compilation.

### Steps

When you preprocess this markdown

```md
> Quoted inclusion
> ```rust
> {{#include hello_world.rs}}
> ```
```

with multilines file `hello_world.rs` like

```rust
pub fn main() {
    println!("Hello, World")
}
```

you get

```md
> Quoted inclusion
> ```rust
> pub fn main() {
    println!("Hello, World")
}
> ```
```

which breaks the markdown compilation

### Version

```text
mdbook v0.5.0-alpha.1
```

miikkas:
In addition to nesting multiline includes in quotes (and admonitions), something I would find useful that doesn't currently work is indented includes within a list context. E.g.

````md
My list:

* Item 1, Hello World:
  
  ```rust
  {{#include hello_world.rs}}
  ```
* Item 2, ...
````

Issue #2521, opened 2025-01-09
invalid output while include code using {{ #include <path> }}


BiswajitThakur:
### Problem

when i include code using  `{{ #include <path> }}`
it does't produce valid output.![Image](https://github.com/user-attachments/assets/783fa420-3192-4393-9f37-afa44480b306)

### Steps

1. `mkdir dsa`
2.  `cd dsa`
3. `mdbook init`
4. `book.toml`
```toml
[book]
authors = ["BiswajitThakur"]
language = "en"
multilingual = false
src = "src"
title = "DSA Rust"
```
5. `src/SUMMARY.md`
```md
# Summary

- [Chapter 1](./chapter_1.md)
```
6. `src/chapter_1.md`
````md
# Chapter 1

- This is demo
  ```rust
  {{ #include ./../code/linear_search.rs }}
  ```
- Hello World
````
8.  `code/linear_search.rs`
```rust
pub fn linear_search<T: PartialEq>(arr: &[T], value: &T) -> Option<usize> {
    let mut i = 0;
    while i < arr.len() {
        if &arr[i] == value {
            return Some(i);
        }
        i += 1;
    }
    None
}

#[test]
fn test_linear_search() {
    let arr = [1, 3, 5, 7, 9, 11, 15];
    assert_eq!(linear_search(&arr, &1), Some(0));
    assert_eq!(linear_search(&arr, &3), Some(1));
    assert_eq!(linear_search(&arr, &5), Some(2));
    assert_eq!(linear_search(&arr, &11), Some(5));
    assert_eq!(linear_search(&arr, &15), Some(6));
    assert_eq!(linear_search(&arr, &0), None);
    assert_eq!(linear_search(&arr, &-1), None);
    assert_eq!(linear_search(&arr, &-3), None);

    let arr1 = [1.5, 77.8, -5.3, 1.3, -9.0, 2.5, 5.3];
    assert_eq!(linear_search(&arr1, &1.5), Some(0));
    assert_eq!(linear_search(&arr1, &-5.3), Some(2));
    assert_eq!(linear_search(&arr1, &2.5), Some(5));
    assert_eq!(linear_search(&arr1, &-9.0), Some(4));
    assert_eq!(linear_search(&arr1, &0.0), None);
    assert_eq!(linear_search(&arr1, &1.4), None);
}
```

### Possible Solution(s)

- It generate valid output when i include like this
````
<no space>```rust
<no space>{{ #include ./../code/linear_search.rs }}
[...]

PR #2831, opened 2025-09-12
Support multiline includes

hg-anssi:
Fix #2830

PR #3080, opened 2026-04-27
Include indent fixes

Omnikron13:
Should fix issues #1626, #1564, #2830, #2521, #1127, #619, #1601.

Would supersede PRs #1718, #2831, #1565, and I think the majority of uses for #1856 & #2333.

~~There is an open question of what exactly to do with non-whitespace before a link, e.g. `foo{{#include ...}}`~~

### DIFF
--- a/crates/mdbook-driver/src/builtin_preprocessors/links.rs
+++ b/crates/mdbook-driver/src/builtin_preprocessors/links.rs
@@ -371,13 +371,15 @@ impl<'a> Link<'a> {
             LinkType::Playground(ref pat, ref attrs) => {
                 let target = base.join(pat);
 
-                let mut contents = fs::read_to_string(&target).with_context(|| {
-                    format!(
-                        "Could not read file for link {} ({})",
-                        self.link_text,
-                        target.display()
-                    )
-                })?;
+                let mut contents = fs::read_to_string(&target).
+                    with_context(|| {
+                        format!(
+                            "Could not read file for link {} ({})",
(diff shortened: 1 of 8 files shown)


Omnikron13:
The [relevant documentation](https://rust-lang.github.io/mdBook/format/mdbook.html#including-files) doesn't need any _changes_ to reflect these changes, though would need an _addition_ if the new behaviour is considered non-obvious.

As a side note I ran into unit tests which cover undocumented behaviour while working on this. Anchored includes in the following form aren't covered atm:

```rust
// ANCHOR: name
println!("included");
// ANCHOR_END: name
println!("not included");
// ANCHOR: name
println!("included");
// ANCHOR_END: name
```

(As an additional detail, there were only actually unit tests for this for the `rustdoc_include` variant, but I have expanded the testing to cover the `include` version too.)

Omnikron13:
Would fix issue #2830, and supersede PR #2831 in turn.

Omnikron13:
Fixes #2521, #1127, #619, #1601

Omnikron13:
Note: some of the issues this would fix do so relying on the fact that _anything_ before the include link will be repeated as a prefix (because they are people wanting to use include within block quotes).

Omnikron13:
Would supersede PR #1565

GuillaumeGomez:
It was fixed in https://github.com/rust-lang/mdBook/pull/3196. Thanks for your work in any case!

PR #3084, opened 2026-05-01
Fix: Remove `?highlight=` from URL when highlights are dismissed via clicking

### LINKED ISSUE #1841: Search highlights don’t disappear from URL when clicked
In search results, you get URLs such as https://doc.rust-lang.org/book/ch04-01-what-is-ownership.html?highlight=ownership#what-is-ownership that highlight the search results. It’s possible to press ESC to remove the highlights *and* update the URL accordingly. There’s also the option to click on any of the hightlights, which makes them invisible but does *not* update the URL.

I would appreciate if that changed, so that the URL get’s updated (i.e. the `?highlight=foo` part removed) by clicking on a highlight, too. I only very recently learned about the ESC-key way in the first place and always had to edit the URL manually before, in order to share a link.

tonywu6:
Hi! This fixes #1841.

This should prevent search result highlights from reappearing when doing in-page navigation (like going to a different heading).

I also amended the `search` UI test to test this highlighting functionality.

### DIFF
--- a/crates/mdbook-html/front-end/searcher/searcher.js
+++ b/crates/mdbook-html/front-end/searcher/searcher.js
@@ -341,6 +341,11 @@ window.search = window.search || {};
                         marker.unmark();
                     }, 300);
                 }
+                // also removes the `?URL_MARK_PARAM=` search param so that
+                // in-page navigation doesn't make highlights unexpectedly appear again
+                const url = new URL(window.location.href);
+                url.searchParams.delete(URL_MARK_PARAM);
+                history.replaceState(null, '', url);
             };
 
             for (let i = 0; i < markers.length; i++) {
--- a/crates/mdbook-html/front-end/searcher/searcher.js
+++ b/crates/mdbook-html/front-end/searcher/searcher.js
@@ -343,9 +343,7 @@ window.search = window.search || {};
                 }
                 // also removes the `?URL_MARK_PARAM=` search param so that
                 // in-page navigation doesn't make highlights unexpectedly appear again
-                const url = new URL(window.location.href);
-                url.searchParams.delete(URL_MARK_PARAM);
-                history.replaceState(null, '', url);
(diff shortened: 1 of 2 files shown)


crates/mdbook-html/front-end/searcher/searcher.js
```diff
@@ -341,6 +341,11 @@ window.search = window.search || {};
                         marker.unmark();
                     }, 300);
                 }
+                // also removes the `?URL_MARK_PARAM=` search param so that
+                // in-page navigation doesn't make highlights unexpectedly appear again
+                const url = new URL(window.location.href);
+                url.searchParams.delete(URL_MARK_PARAM);
+                history.replaceState(null, '', url);
```
ehuss: I'm wondering if the following would be a better approach here:


```suggestion
                setSearchUrlParameters('', 'replace');
```

Can you say why this had a different approach?


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3084/9ea86e740b891f12c2af300cc9aa395391015b80..3a16879ddaf0f8f6ff3c2850caeedaa8267b1a86)*

tonywu6: Oops I didn't notice there was a dedicated function for this.

Although looking at the function, in case the URL had both `search=` and `highlight=` then this would remove both? (I imagine that would just be a minor side effect, but I'm not sure)

ehuss: I think that's fine. It seems reasonable to me to have the same behavior as the Escape key.

ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3084/9ea86e740b891f12c2af300cc9aa395391015b80..b4e674877fbf45c86c0ff4065fd32103bde33061)*

PR #3086, opened 2026-05-03
Add `ChapterMutThin` for parallel chapter access

SichangHe:
`Book` only allows sequential mutable traversal today, and lacks a way for the much more efficient parallel mutation that would speed up many preprocessed by #cores×.

This simple change adds `ChapterMutThin` and `Book::chapters_mut_thin()` to return per-chapter mutable views for parallel access.

I've been using this change from my fork `mdbook_fork4ls` for months, within `mdbook-katex` and `mdbook-header-footer`, together w/ Rayon, and haven't had any issues. I think it's a generally useful quality-of-life addition.

### DIFF
--- a/crates/mdbook-core/src/book.rs
+++ b/crates/mdbook-core/src/book.rs
@@ -89,6 +89,14 @@ impl Book {
         );
     }
 
+    /// Recursively collect all chapters in the book as mutable thin references,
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Do you have an example of how it would be used? I have no idea what this is about and what's the expected use of this change.

Issue #2760, opened 2025-07-22
How to add page numbering to the title of the pages?


szabgab:
### Question

In the   side-bar pages are numbered. Is there a way to ask mdbook to display the same numbers in the title of each page so one can see them even if the side-bar is closed?
If not, would this be a good idea to add this as an option to the html generator?

This would also help with  #2759

### Version

```text
mdbook v0.4.51
```

PR #3091, opened 2026-05-06
Rename no-section-label to section-label-toc

### LINKED ISSUE #2761: Rename no-section-label to be a positive flag e.g. section-label-toc
### Problem

IMHO negative configuration options are confusing as they use double negative `no-section-label` defaults to `false`.  It would be better to rename the option to something positive. e.g.  `section-label-toc`.

I suggest including the `-toc` part to indicate that this is about the table of contents.

This would allow for another option called `section-label-title` that would make mdbook add the numbers to the titles for #2760 

### Proposed Solution

1. Add the new flag and add a deprecation warning for the `no-section-label` flag. Update the documentation to indicate the new flag and the deprecation.
2. Open an issue to remove the new flag when breaking changes are allowed.

### Notes

_No response_

ChrisJr404:
Closes #2761.

The `no-section-label` HTML config key was a negative flag whose default was `false`, which made the meaning awkward (double negative). This renames it to `section-label-toc` with a default of `true`, preserving the existing behavior of showing numeric section labels in the table of contents.

The `-toc` suffix anticipates a possible future `section-label-title` companion flag for #2760.

Per @ehuss in #2761, this is a clean rename for 0.5 with no deprecation warning. The existing `HtmlConfig::Default` impl is updated to set `section_label_toc: true`. With `deny_unknown_fields` already on `HtmlConfig`, books that still set `no-section-label` will fail to deserialize with a message that lists `section-label-toc` among the expected fields, so the migration path is discoverable.

### Changes

- `crates/mdbook-core/src/config.rs`: rename the `HtmlConfig` field and flip the default to `true`.
- `crates/mdbook-html/src/html_handlebars/helpers/toc.rs`: rename `RenderToc::no_section_label` to `section_label_toc` and invert the gate.
- `crates/mdbook-html/src/html_handlebars/hbs_renderer.rs`: pass through the renamed field.
- `guide/src/format/configuration/renderers.md`: update the example and the docs entry.

### Verification

Built a small book locally with `section-label-toc = false`, `section-label-toc = true`, and the key omitted; confirmed that `aria-hidden` section-number spans appear only when the flag is `true` (or absent, since default is `true`), and disappear when set to `false`. With the old key, the build now fails with `unknown field 'no-section-label', expected one of ...section-label-toc...`.

`cargo build`, `cargo clippy --workspace --all-targets`, `cargo fmt --check`, and `cargo test --workspace` all pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1078,7 +1078,7 @@ version = "0.0.0"
 
 [[package]]
 name = "mdbook-core"
-version = "0.5.2"
+version = "0.6.0"
 dependencies = [
  "anyhow",
  "regex",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ hex = "0.4.3"
 html5ever = "0.39.0"
 indexmap = "2.14.0"
 ignore = "0.4.25"
-mdbook-core = { path = "crates/mdbook-core", version = "0.5.2" }
+mdbook-core = { path = "crates/mdbook-core", version = "0.6.0" }
 mdbook-driver = { path = "crates/mdbook-driver", version = "0.5.2" }
 mdbook-html = { path = "crates/mdbook-html", version = "0.5.2" }
 mdbook-markdown = { path = "crates/mdbook-markdown", version = "0.5.2" }
--- a/crates/mdbook-core/Cargo.toml
+++ b/crates/mdbook-core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mdbook-core"
-version = "0.5.2"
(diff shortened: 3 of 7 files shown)


GuillaumeGomez (changes requested):
Please remove the version change. Next release will be a major one in any case.

Otherwise looks good to me, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3091/9190b5d0933188c6fae6e1962e70e941b014d814..956dd64fb14996e4f8058a1e7be9f93aea6068d0)*

PR #3092, opened 2026-05-07
Fall back to "section" id for punctuation-only headings

### LINKED ISSUE #3002: Headings with just punctuation do not generate a linkable ID
### Problem

### Steps

### Possible Solution(s)

### Notes

_No response_

### Version

```text
mdbook v0.5.2
```

ChrisJr404:
Fixes #3002.

A heading whose text is entirely stripped by `id_from_content` (e.g. `## ::`, `## !!!`) currently renders as `<h2 id=""><a class="header" href="#"></a>`. The empty id can't be linked to and breaks the in-page anchor.

When the generated slug is empty, fall back to `"section"`. The existing `unique_id` helper then handles disambiguation, so two punctuation-only headings on the same page become `section` and `section-1`, matching what pandoc and kramdown do for the same case (thanks @stevecheckoway for pointing this out and providing the original sketch in the issue).

Changes:

- `crates/mdbook-html/src/utils.rs`: when `id_from_content` would return an empty string, return `"section"` instead. Added a unit test that exercises the per-page counter end-to-end through `unique_id` (two punctuation-only headings + a normal heading + a third punctuation-only heading produce `section`, `section-1`, `real-heading`, `section-2`).
- `tests/testsuite/rendering/header_links/expected/header_links.html`: the existing integration fixture already covers two empty-id headings (`###` and `####`) and previously documented the broken output (`id=""`, `id="-1"`); updated to the new `section` / `section-1` output.

Tested locally:

- `cargo test --workspace` — all green.
- `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- `cargo fmt --check` — clean.


tests/testsuite/rendering/header_links/expected/header_links.html
```diff
@@ -1,7 +1,7 @@
 <h1 id="header-links"><a class="header" href="#header-links">Header Links</a></h1>
 <h2 id="foobar"><a class="header" href="#foobar">Foo^bar</a></h2>
-<h3 id=""><a class="header" href="#"></a></h3>
-<h4 id="-1"><a class="header" href="#-1"></a></h4>
+<h3 id="section"><a class="header" href="#section"></a></h3>
```
GuillaumeGomez, line 3: Please add:

```md
### !.():

##### !!! {#explicit-punctuation}
```

in `tests/testsuite/rendering/header_links/src/header_links.md` to have more testcases (and don't forget to update `tests/testsuite/rendering/header_links/expected/header_links.html` afterward.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3092/9190b5d0933188c6fae6e1962e70e941b014d814..f895937f7134160b8383280ff6a6bfe2362e7c40)*

GuillaumeGomez (changes requested):

crates/mdbook-html/src/utils.rs
```diff
@@ -128,8 +137,32 @@ mod tests {
...
+            "section-1"
+        );
+        assert_eq!(
+            unique_id(&id_from_content("Real Heading"), &mut id_counter),
+            "real-heading"
+        );
+        assert_eq!(
+            unique_id(&id_from_content("***"), &mut id_counter),
+            "section-2"
+        );
```
GuillaumeGomez, line 166: Please add:

```suggestion
        assert_eq!(
            unique_id(&id_from_content("   "), &mut id_counter),
            "section-3"
        );
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3092/9190b5d0933188c6fae6e1962e70e941b014d814..f895937f7134160b8383280ff6a6bfe2362e7c40)*

GuillaumeGomez:
Fix looks good to me, thanks! Just a few missing tests and it's good to go.

ChrisJr404:
Thanks for the review! Added both requested cases:

- `### !.():` and `##### !!! {#explicit-punctuation}` in `header_links.md`, with the expected HTML updated — this shows the section fallback keeps counting (`section-2`) and that an explicit `{#id}` overrides the fallback on a punctuation-only heading.
- The extra `"   "` -> `section-3` assertion in the unit test.

`cargo test`, `cargo fmt --check`, and `cargo clippy` are all clean.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks for the PR. Like I said in #3185, please refrain from using LLM in your future contributions in mdbook. Merging this one but for such easy fixes, I'd rather have humans work on it so they can learn how the codebase works so they can eventually become part of the project.

PR #3093, opened 2026-05-07
Support socket activation for `mdbook serve`

Ericson2314:
Add a `--socket-activate` flag that adopts a pre-bound TCP listener from `LISTEN_FDS` instead of binding a new one. This allows process managers like foreman (with Socketfile support) or systemd to own the socket, so it survives server restarts.

Three modes:
- `--socket-activate`: require a passed-in socket, fail if absent
- `--port N` / `--hostname H`: always bind, ignore `LISTEN_FDS`
- Neither: try `LISTEN_FDS` first, fall back to binding the default

The listener is now bound in the main thread before spawning the server, so the actual address is always known for logging and `--open`.

Also parse `--port` as `u16` at arg-parse time instead of leaving it as a string.

Socket activation is normally associated with systemd. And indeed, it would be peculiar to wire up this development command with systemd, but it is also used in other contexts more appropriate to this command, like https://github.com/mitsuhiko/systemfd, a development tool. From a Capsicum/WASI perspective, it also is generally better when tools can consume resources provided by a more privileged caller, rather than having to open them themselves. For these reasons, I think everything should support socket-activation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -213,6 +213,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bumpalo"
+version = "3.20.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5d20789868f4b01b2f2caec9f5c4e0213b41e3e5702a50157d699ae31ced2fcb"
+
+[[package]]
 name = "bytes"
 version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
Sorry for the delay. No opinion on this feature, what do you think about it @rust-lang/mdbook ?

notriddle:
I’m in favor. It’s not very complicated, and people do use it.

GuillaumeGomez:
@rfcbot fcp merge mdbook

Ericson2314:
Thanks! Glad to hear there is acceptance of this sort of thing.

GuillaumeGomez:
Thanks everyone!

GuillaumeGomez (approved):

PR #3094, opened 2026-05-11
feat: reserve toc.md and document reserved source filenames

### LINKED ISSUE #3088: Update documentation to include `print.md` and `print.html`
### Problem

"
When the html pages are generated a file called `print.html` is also generated. When the user clicks on the printer icon mdbook will use this file to show the content that will be printed.

If the user creates a file called `src/print.md`  and references it from the `SUMMARY.md` the rendering of HTML pages will fail with the following error message: `print.md is reserved for internal use`.
"

Mentioning both print.md and print.html in the documentation will allow users to search for them and will help the users to understand the behavior.

See also the discussion in #3047

### LINKED ISSUE #3090: Document toc.md and toc.html and make toc.md a reserved file
### Problem

In normal circumstances mdbook will generate a file called `toc.html` that will contain the Table of Contents`. It is used as the TOC for people (and bots) who don't run JavaScript.

If the user creates a file called `src/toc.md` and references it in `SUMMARY.md` then mdbook will use it to generated the `toc.html` file replacing the normally generated one.

### Proposed Solution

IMHO this should be avoided. 
1. If there is a an `toc.md` file the build process should stop with an error `toc.md is reserved for internal use`.
2. Regardless of 1. being implemented or not, the behaviour should be described in the documentation so people can search for either `toc.html` , `toc.md` or `toc` and find it.

### Notes

See also #3047

ChrisJr404:
## Summary

Two related issues filed by szabgab on the same day, addressed together because they share the same underlying behavior:

- Closes #3090 (make `toc.md` a reserved file, document `toc.md` / `toc.html`)
- Closes #3088 (document `print.md` / `print.html`)

The HTML renderer already rejects `print.md` with `print.md is reserved for internal use`, because `print.html` is generated from the book's chapters and would otherwise be silently overwritten. The same situation applies to `toc.html`: it is the no-JavaScript table of contents fallback, generated from `SUMMARY.md`. Today, naming a chapter `toc.md` and referencing it from `SUMMARY.md` silently replaces the generated `toc.html`, which is confusing behavior (see issue #3090 for the original report and #3047 for the broader discussion).

## Changes

1. `crates/mdbook-html/src/html_handlebars/hbs_renderer.rs`: extend the existing reserved-filename check so `toc.md` is rejected the same way `print.md` is. The error message reuses the same template, so callers get `toc.md is reserved for internal use`.

2. `guide/src/guide/creating.md`: add a "Reserved file names" subsection under "Source files" that documents both `print.md` / `print.html` and `toc.md` / `toc.html`, with a brief explanation of what each generated file is for. Links to the existing `[output.html.print]` config section for cross-reference.

3. `guide/src/format/summary.md`: add a short "Reserved file names" subsection so users searching the SUMMARY.md reference page also find the constraint.

4. `tests/testsuite/build.rs` + fixture `tests/testsuite/build/no_reserved_filename_toc/`: new test `no_reserved_filename_toc` that mirrors the existing `no_reserved_filename` test for `print.md`. Creates `src/toc.md`, references it from `SUMMARY.md`, and asserts the build fails with `toc.md is reserved for internal use`.

## Test plan

- [x] `cargo test --test testsuite no_reserved_filename` passes (`no_reserved_filename` and the new `no_reserved_filename_toc` both green).
- [x] Full `cargo test` passes other than two preexisting environmental failures in `preprocessor::nop_preprocessor` / `preprocessor::failing_preprocessor`, which need a workspace `Cargo.toml` discoverable from the test's tmp dir (unrelated to this change; CI environment should not hit them).
- [x] Guide pages render without broken links: the new cross-reference to `#outputhtmlprint` matches the existing `### [output.html.print]` anchor in `guide/src/format/configuration/renderers.md`.


DanielEScherzer (approved):
community review: makes sense and looks good to me, a warning would be helpful instead of just silently replacing the file

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3094/9190b5d0933188c6fae6e1962e70e941b014d814..35057bf14c8fb4db1142085fdf6d6144a0f4dc0d)*

crates/mdbook-html/src/html_handlebars/hbs_renderer.rs
```diff
@@ -39,7 +39,8 @@ impl HtmlHandlebars {
 
         let path = ch.path.as_ref().unwrap();
         // "print.html" is used for the print page.
-        if path == Path::new("print.md") {
+        // "toc.html" is used for the no-JS table of contents fallback.
+        if path == Path::new("print.md") || path == Path::new("toc.md") {
```
GuillaumeGomez: Would be better to keep a list of reserved file names as it will very easily go out of date.

GuillaumeGomez (changes requested):

GuillaumeGomez:
@DanielEScherzer Doesn't it `bail!` before any rendering is done? (might have missed something obvious but maybe you know more ;) )

PR #3095, opened 2026-05-13
Clean up language

ericholscher:
I wasn't sure what this meant,
so seemed like it would be more clear to remove it.


ehuss:
Can you say more about what wasn't clear? It's giving examples of different ways you might encounter for publishing content. It's trying to emphasize that the formula for how to set up deployment will depend on your circumstances, and that those can vary significantly.

shenef:
The wording on the entire page seems quite awkward to be honest, as if it's trying to target a user that is very new to development activities but it's using terminology that is targeted at experienced users.
Imo it should be updated/rewritten and placed in the `for_developers` folder.

GuillaumeGomez:
Not too sure either. I'll close it for now but if you have more detailed information about what's wrong on how to improve it, feel free to open a new PR.

PR #3098, opened 2026-05-17
fix(html): skip HTML balance warnings when Handlebars is present

### LINKED ISSUE #2941: Warnings when using an include within an admonition
### Problem

The following causes warnings but renders correctly:
```html
> [!TIP]
> <details><summary>Click to open</summary>{{#include assets/tables/any_exp.html}}</details>
```
```
 WARN unclosed HTML tag `<details>` found in `zoktai/any.md` while exiting BlockQuote(Some(Tip))
HTML tags must be closed before exiting a markdown element.
 WARN unexpected HTML end tag `</details>` found in `zoktai/any.md`
Check that the HTML tags are properly balanced.
```

Image

All tags are properly closed but are not detected as such, replacing the include with plain text resolves the warnings.

### Version

```text
mdBook 0.5.0
```

wyf027:
## Summary

- Suppress unclosed/unexpected HTML tag warnings while parsing inline HTML that contains Handlebars syntax (`{{...}}`).
- Handlebars helpers such as `{{#include}}` are expanded after the HTML tree is built; validating tag balance on the pre-expansion source produced false positives (e.g. `<details>` wrapping an include inside an admonition).

Fixes #2941.

## Test plan

- [x] `cargo test -p mdbook html_with_handlebars`
- [x] Existing `unclosed_html_tags` test still passes (real unclosed tags still warn)

### DIFF
--- a/crates/mdbook-html/src/html/tree.rs
+++ b/crates/mdbook-html/src/html/tree.rs
@@ -199,6 +199,12 @@ pub(crate) struct MarkdownTreeBuilder<'opts, 'event, EventIter> {
     /// tag. After the document has been parsed, all the definitions are moved
     /// to the end of the document.
     footnote_defs: HashMap<CowStr<'event>, NodeId>,
+    /// Nesting depth of HTML fragments that contain Handlebars syntax.
+    ///
+    /// Handlebars helpers such as `{{#include}}` are expanded after the HTML
+    /// tree is built; validating tag balance on the pre-expansion source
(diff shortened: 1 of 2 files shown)


ehuss:
@leno23 I see that you've opened a large number of PRs using an AI tool. Please do not do that. Per the [contributing guide](https://github.com/rust-lang/mdBook/blob/master/CONTRIBUTING.md#issue-assignment), we are unable to review these.

PR #3104, opened 2026-05-17
fix: detect redirect loops during HTML build

### LINKED ISSUE #2901: Redirects should have loop detection
### Problem

If a redirect chain ends in a loop, the browser will reload the pages indefinitely.

### Steps

Set up a redirect loop, such as the following:

```toml
[output.html.redirect]
"/chapter_1.html#a" = "chapter_2.html#b"
"/chapter_2.html#b" = "chapter_1.html#a"
```

with a SUMMARY.md:

```markdown
# Summary

- [Chapter 1](./chapter_1.md)
- [Chapter 2](./chapter_2.md)
```

Navigate to `chapter_1.html#a`

### Possible Solution(s)

I'm not sure exactly what should be done. One thing to consider is that loops can happen in between books. Some ideas:

- During book building, detect a loop and report an error.
- When redirecting, use some mechanism to track the number of redirects. I don't know what this would look like, but url query parameters, cookies, or localstorage are some potential options. It looks like browsers allow roughly 20 redirects (when using http redirects).

### Notes

_No response_

### Version

```text
mdbook v0.5.0-alpha.1
```

wyf027:
## Summary

Fixes #2901.

Detects cycles in `[output.html.redirect]` at build time and fails with a clear error, instead of letting the browser reload redirect pages indefinitely.

## Changes

- Walk redirect entries to find cycles (including fragment redirects like `/page.html#frag`)
- Report a canonical cycle path in the error message
- Unit and integration tests for page-level and fragment-level loops

## Test plan

- [x] `cargo test -p mdbook-html redirect_loop`
- [x] `cargo test redirect_loop` (testsuite)
- [ ] CI

### DIFF
--- a/crates/mdbook-html/src/html_handlebars/hbs_renderer.rs
+++ b/crates/mdbook-html/src/html_handlebars/hbs_renderer.rs
@@ -11,7 +11,7 @@ use mdbook_core::config::{BookConfig, Config, HtmlConfig};
 use mdbook_core::utils::fs;
 use mdbook_renderer::{RenderContext, Renderer};
 use serde_json::json;
-use std::collections::{BTreeMap, HashMap};
+use std::collections::{BTreeMap, HashMap, HashSet};
 use std::path::{Path, PathBuf};
 use tracing::error;
 use tracing::{debug, info, trace, warn};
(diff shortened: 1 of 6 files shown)


wyf027:
Closing as duplicate of #3103 (same fix for #2901). Keeping #3103 for review.

PR #3110, opened 2026-05-17
fix(html): generate fallback id for punctuation-only headings

### LINKED ISSUE #3002: Headings with just punctuation do not generate a linkable ID

wyf027:
## Summary
- When a heading contains only punctuation (e.g. `## ::`), `id_from_content` returns an empty string
- This produced `id=""` and `href="#"`, breaking anchor links
- Fall back to a unique `section` slug (with numeric suffixes on collision)

## Test plan
- [x] `cargo test -p mdbook-html utils::tests`
- [ ] Build a book with a `## ::` heading and verify the generated HTML has a non-empty `id`

Fixes #3002

### DIFF
--- a/crates/mdbook-html/src/html/print.rs
+++ b/crates/mdbook-html/src/html/print.rs
@@ -6,7 +6,7 @@
 
 use super::Node;
 use crate::html::{ChapterTree, Element, serialize};
-use crate::utils::{ToUrlPath, id_from_content, normalize_path, unique_id};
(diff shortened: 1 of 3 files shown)


wyf027:
Closing as duplicate of #3092, which already addresses #3002 with the same approach. Thanks!

PR #3124, opened 2026-05-22
Add custom playground guide for v0.5

dalance:
This PR documents how to implement custom playgrounds.

This is a successor to PR #2286.
The method described in #2286 is no longer available in v0.5, so it has been rewritten for v0.5, and
it was possible to implement it with fewer patches than the previous version.

Related: #350 #1453 #1759

### DIFF
--- a/guide/src/SUMMARY.md
+++ b/guide/src/SUMMARY.md
@@ -33,6 +33,7 @@
     - [MathJax support](format/mathjax.md)
     - [mdBook-specific features](format/mdbook.md)
     - [Markdown](format/markdown.md)
+    - [Custom playground](format/custom-playground.md)
 - [Continuous integration](continuous-integration.md)
 - [For developers](for_developers/README.md)
     - [Preprocessors](for_developers/preprocessors.md)
--- a/guide/src/format/custom-playground.md
+++ b/guide/src/format/custom-playground.md
@@ -1,0 +1,230 @@
+# Custom Playground
+
+By default, mdBook supports playground for Rust only.
+If you want to add playground for another language, you can do it by
+overriding theme assets and adding a small initialization script.
+
+In this section, we will try to add a playground for the imaginary
+`example` language.
+
+## Add `playground` option to code blocks in the book
+
+`playground` marks a code block as runnable. `editable` makes the block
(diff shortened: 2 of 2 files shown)


GuillaumeGomez:
Sorry for the delay, we now have an mdbook team so things should move a bit faster here. This sounds like a very good idea to add an example! Do you mind adding a GUI test to ensure that the code you provided won't become outdated when we make updates?

GuillaumeGomez (changes requested):
Need a GUI test using the JS code example to ensure it doesn't become outdated.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3124/0ea415897758ea9b2904ed47ba9cb4901f9eb089..5dde35f2be7c24788d3cfcf06335d309baea4f12)*

dalance:
Thank you for your review.
I'm glad the mdBook team has been established.

When I wrote this documentation, I understood that review bandwidth was extremely limited,
so I wrote a procedure which can be implemented without making changes to mdBook itself.

If minor changes are acceptable, I can clean up some of the JavaScript hacks, what do you think?
For example, if we slightly adjusted the condition at https://github.com/rust-lang/mdBook/blob/74b14bb0ebd6a3425f53da40e6d4bff97a8c80dc/crates/mdbook-html/src/html/tree.rs#L1010,
the section titled "Restrict the `.playground` selector to `<pre>`" will no longer be necessary.

GuillaumeGomez:
If that can make things simpler, then it's worth a try!

dalance:
I wrote 2 issues to improve custome playground support.
If both of them are accepted, the way of this PR is significantly simplified because theme override is not necessary.
Especially it is important that `book.js` override is not necessary because it causes manual synchronization to mdBook update.

dalance:
@GuillaumeGomez
I created #3218.

Regarding #3216, since it would have a bit large impact, such as requiring the addition of a configuration value, 
I'd like to prepare it only after reaching a certain level of consensus.
What do you think?

If there is no prospect of it being accepted in the short term, I am also open to rewriting this guide on the assumption that #3216 can't be used.

PR #3127, opened 2026-05-26
Add zoomable images feature

### LINKED ISSUE #2075: Zoomable Images
### Problem

Currently, images in HTML rendered through mdbook are static and cannot be zoomed into which can make them difficult to see (such as [here](https://cr0mll.github.io/cyberclopaedia/Exploitation/Binary%20Exploitation/Stack%20Exploitation/Format%20String%20Vulnerabilities.html)).

### Proposed Solution

The images should appear as enlarged pop-ups when clicked on (similarly to how Discord or Obsidian do it). 

It should also be possible to configure this as an option in book.toml for the HTML renderer. For example
```
[output.html]
zoomable-images = true
```

### Notes

_No response_

GuillaumeGomez:
Fixes #2075.

The "zoomed-in" image looks like this:

image

The JS is only used to "zoom out" images with the "escape key", otherwise the feature works entirely without JS.

It works as follows: there is a (hidden) checkbox right before the image. The image itself is inside the label associated to the checkbox, meaning that when you click on the image, you switch the value of the checkbox. And thanks to `:checked` CSS selector, it changes the display of the zoomed out image.

To make it obvious that you can "zoom in"/"zoom out" on an image, I changed the cursor to look like a loop using the CSS `cursor: zoom-in|zoom-out` property.

I also realized that images had the `alt` attribute, but not the `title` one, so I added it while I was at it.


GuillaumeGomez:
Updated failing tests as well, ready for review. :)

GuillaumeGomez:
ping @ehuss

ehuss:
Won't this break images that are links?

GuillaumeGomez:
I think it would yes. Didn't think about that. If I only allow to zoom on non-link images, does it sound good to you?

shenef:
The description mentions the escape key for zooming back out, how would that work on a mobile device? Or does tapping it again toggle off the invisible checkbox?
I have a case where i currently use an image with a link to open a separate full-size version of the image. (740x638 -> 3712x3200)
So with this feature there would be no more need to do that.

GuillaumeGomez:
> The description mentions the escape key for zooming back out, how would that work on a mobile device? Or does tapping it again toggle off the invisible checkbox?

Tapping again will zoom-out.

GuillaumeGomez:
@ehuss Disabled the feature on images inside links and extended test to check it as well.

crates/mdbook-html/src/html/tree.rs
```diff
@@ -560,13 +570,53 @@ where
...
+
+                    let mut input = Element::new("input");
+                    input.insert_attr("id", img_input_id.to_tendril());
+                    input.insert_attr("class", "checkbox-img".to_tendril());
+                    input.insert_attr("type", "checkbox".to_tendril());
+                    self.append(Node::Element(input));
+
+                    let mut label = Element::new("label");
+                    label.insert_attr("class", "checkbox-label".to_tendril());
+                    label.insert_attr("for", img_input_id.to_tendril());
```
ehuss: The IDs end up getting broken on the print page because it rebuilds the IDs to be unique. I think the "for" here ends up being the wrong value in that case.

I'm not sure how to fix that, other than the print page also handling this attribute.

crates/mdbook-html/src/html/tree.rs
```diff
@@ -560,13 +570,53 @@ where
                 let mut img = Element::new("img");
                 let src = fix_link(dest_url).into_tendril();
                 img.insert_attr("src", src);
+                // This will eat TagEnd::Image
+                let alt = self.text_for_img_alt();
+                img.insert_attr("alt", alt.to_tendril());
                 if !title.is_empty() {
                     img.insert_attr("title", title.into_tendril());
+                } else {
+                    img.insert_attr("title", alt.into());
```
ehuss: Why was this fallback to add the alt as the title added?

crates/mdbook-html/src/html/tree.rs
```diff
@@ -560,13 +570,53 @@ where
...
+                    input.insert_attr("class", "checkbox-img".to_tendril());
+                    input.insert_attr("type", "checkbox".to_tendril());
+                    self.append(Node::Element(input));
+
+                    let mut label = Element::new("label");
+                    label.insert_attr("class", "checkbox-label".to_tendril());
+                    label.insert_attr("for", img_input_id.to_tendril());
+                    self.push(Node::Element(label));
+
+                    self.append(Node::Element(img.clone()));
```
ehuss, line 591: I'm a little concerned about this approach of using a duplicated `` tag. Screen readers may have difficulty handling this. I think at a minimum it should have `aria-hidden="true"`. I'm a bit uncomfortable with this approach in general, though I'm willing to give it a try.

crates/mdbook-html/src/html/tree.rs
```diff
@@ -560,13 +570,53 @@ where
...
+
+                    let mut wrapper = Element::new("span");
+                    wrapper.insert_attr("class", "img-wrapper".to_tendril());
+                    self.push_no_stack(Node::Element(wrapper));
+
+                    self.append(Node::Element(img));
+
+                    // We exit the `label` and the `div`.
```
ehuss: This comment seems off. There is no `div` AFAIK.

Perhaps it means:


```suggestion
                    // We exit the `label` and the `span` (which used push_no_stack).
```

crates/mdbook-html/front-end/css/general.css
```diff
@@ -406,3 +406,35 @@ dd > p {
     /* Add some space between the icon and the text. */
     margin-right: 8px;
 }
+
+.content .checkbox-img, .checkbox-label > .img-wrapper {
```
ehuss: Can you add comments explaining what these rules are for and what they do? It can be just a short comment on each selector to explain what it's for (image zoom) and what that selector is doing.

crates/mdbook-html/front-end/js/book.js
```diff
@@ -658,6 +658,12 @@ aria-label="Show hidden lines"></button>';
 })();
 
 (function chapterNavigation() {
+    function zoomOutImages() {
+        for (const elem of Array.from(document.querySelectorAll('input.checkbox-img'))) {
```
ehuss, line 666: I believe this can be simplified:


```suggestion
        for (const elem of document.querySelectorAll('input.checkbox-img')) {
```

tests/gui/image-zoom.goml
```diff
@@ -0,0 +1,60 @@
...
+
+go-to: |DOC_PATH| + "basic/index.html"
+call-function: ("check-image-not-zoomed-in", {})
+
+// We click on the image to "zoom in".
+click: ".checkbox-label > img"
+call-function: ("check-image-zoomed-in", {})
+
+// We click on the wrapper to "zoom out".
+click: ".checkbox-label > img"
```
ehuss, line 37: Is it intentional that this clicks on the image underneath the wrapper? That is, should this be something like `".image-wrapper > img"` or `".image-wrapper"` to simulate what the user would actually click on?

crates/mdbook-html/src/html/tree.rs
```diff
@@ -230,6 +233,13 @@ where
         builder.tree
     }
 
+    /// Returns the current `img_label_id` and increase its value.
+    fn get_current_img_label_id(&mut self) -> usize {
```
ehuss: Just a very minor nit, but this method name seemed a little confusing to me since it has a side effect. Perhaps something like `get_next_img_label_id` would be a little clearer?

ehuss (commented):
Thanks!

I'm not going to block on it, but this doesn't seem to support the ability to trigger with the keyboard. Generally we try to ensure to follow accessibility guidelines as much as we can.

Just curious, was it intentional that you can still scroll when the image is zoomed in? (I'm not sure if I have a preference, just checking.)

I'm slightly uneasy not having a way to turn this off. I'm fine with starting without it, but if people have issues, we may need to add something. In particular, small images like icons probably don't need or want to be zoomed in this way.


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3127/9873b35405ce8675d6c3ea4f5e853f2d78ed3ea1..83f42f97b906ee487e96784fe4a50e4822a3a727)*

ehuss:
Also, it's probably a good idea to include at least some note in the documentation that this exists. Perhaps on the [Images](https://rust-lang.github.io/mdBook/format/markdown.html#images) section?

GuillaumeGomez:
> I'm not going to block on it, but this doesn't seem to support the ability to trigger with the keyboard. Generally we try to ensure to follow accessibility guidelines as much as we can.

I was sure I tested that, I'll double-check and add a regression test to ensure it works.

> Just curious, was it intentional that you can still scroll when the image is zoomed in? (I'm not sure if I have a preference, just checking.)

I didn't consider it an issue but I can do it if you want as even with JS (it's required for that as I need to change a parent's style, so cannot be done with only CSS).

> I'm slightly uneasy not having a way to turn this off. I'm fine with starting without it, but if people have issues, we may need to add something. In particular, small images like icons probably don't need or want to be zoomed in this way.

I can add a CLI option for that if you want? As for small images, if I want to know their size, I would need another dependency for that. Not sure it's worth it but again, your call.

crates/mdbook-html/src/html/tree.rs
GuillaumeGomez: It's the print page, so I don't think it matters in this case, no?

crates/mdbook-html/src/html/tree.rs
GuillaumeGomez: I mentioned it in my first comment:

> I also realized that images had the alt attribute, but not the title one, so I added it while I was at it.

It's better for accessibility, and for other users, to have text when they hover an image without moving for a while.

crates/mdbook-html/src/html/tree.rs
GuillaumeGomez: I think screen readers all handle `display: none`, so I think we're fine (although if we could test it, would be even better). However, won't `aria-hidden=true` be an issue when the image is zoomed-in?

crates/mdbook-html/front-end/js/book.js
GuillaumeGomez: It can, but since we make modifications to the DOM, there is a non-zero chance that the selector list will recompute itself between each items (that was a huge issue in older rustdoc when we used it to generate `[-]` buttons).

crates/mdbook-html/src/html/tree.rs
GuillaumeGomez: :+1:

GuillaumeGomez:
I simplified the DOM generation: no need for an ID anymore. We can now make the zoom in feature work with the keyboard (still no JS, works with spacebar). Added a GUI test for it as well. And mentioned it in the mdbook guide.

I didn't add a new CLI option to disable it yet. Waiting for you to confirm if you want it or not.

GuillaumeGomez:
Updated tests with HTML comparisons. :)

tests/gui/image-zoom.goml
GuillaumeGomez: Well, they click on the image, but the image being part of the label, since there is no `preventDefault()` on the image, the event is propagated to the parent until the parent is the `label` which then handles the click event and stops its propagation. So it is how users will interact with it.

crates/mdbook-html/src/html/tree.rs
GuillaumeGomez: Fixed.

GuillaumeGomez:
Fixed merge conflict.

crates/mdbook-html/src/html/tree.rs
ehuss: Some people read book using the print page because it is a convenient way to browse on a single page. Marking this resolved since the latest update seems to use a different approach that fixes this.

crates/mdbook-html/src/html/tree.rs
ehuss: I think I would prefer to not do that. For example, MDN explicitly says:

> avoid duplicating the alt attribute's value in a title attribute declared on the same image. Doing so may cause some screen readers to announce the same text twice, creating a confusing experience.

I also haven't seen other markdown renderers do something like this.

crates/mdbook-html/src/html/tree.rs
ehuss: I tried with my screen reader. It does seem to ignore the "display: none".

It's a little awkward now since the screen reader navigates to the checkbox (and says you are on a checkbox, and tells you the image alt), and then the image (and the alt and title). But I suppose it should be fine. We can adjust it if we get further feedback.

ehuss:
> I can add a CLI option for that if you want?

No, I don't think it's necessary for now. Also, all of the options for rendering go in the config, not the CLI. So, for example, this would probably go in the `[output.html]` section.

GuillaumeGomez:
Removed the `title` attribute.

crates/mdbook-html/src/html/tree.rs
```diff
@@ -560,13 +560,46 @@ where
                 let mut img = Element::new("img");
                 let src = fix_link(dest_url).into_tendril();
                 img.insert_attr("src", src);
-                if !title.is_empty() {
-                    img.insert_attr("title", title.into_tendril());
-                }
```
ehuss, line 565: Looks like this deleted the title attribute entirely. It should probably just keep the original code here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3127/3ec0546d8b28870aaf6304ed7f0c0535b747c51a..c551c43cbc37d97e699e9da3b60c634f35cc62b7)*

GuillaumeGomez: Arf. I think I messed up when fixing the merge conflict when I removed the commit. Fixing that.

GuillaumeGomez: Ok it's back as it should. Sorry for the mess.

GuillaumeGomez:
Finally fixed my merge conflicts mess. ^^'

ehuss (approved):
Great, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3127/3ec0546d8b28870aaf6304ed7f0c0535b747c51a..beab2fa604f07697fb6160d42e5f46fb6261b4d0)*

Kroc:
Possibly of interest as a better way to do it, no JS required -- even to press ESC: https://oldbytes.space/@Kroc/116897812688116047

GuillaumeGomez:
Oh that's super neat! Want to give it a try in here?

Kroc:
That'll have to be for you or someone else; I'm not familiar with the codebase for it to be good for my health -- I'm just passing along a neat tip whilst I was searching for something unrelated :)

GuillaumeGomez:
Noted. All the code you need to update is the one in this PR but I understand. I'll try to get back to it at some point. Thanks for the tip in any case!

PR #3131, opened 2026-06-01
Fix GUI tests

GuillaumeGomez:


GuillaumeGomez:
Fixed the CI @ehuss.

ehuss:
Can you help me understand what is going on here? My understanding is that installing browser-ui-test also installs puppeteer. And puppeteer downloads a specific version of chromium (specified [here](https://github.com/puppeteer/puppeteer/blob/fd71fa85c20627a6f6d2b5e35f5432201772bc1d/packages/puppeteer-core/src/revisions.ts) I believe). Why does this need to download chromium? Wouldn't it make it less reproducible if the version of chromium isn't fixed?

GuillaumeGomez:
I have absolutely no clue why it suddenly stopped working. Maybe the `puppeteer` dependency version is too old. The layout rendering isn't supposed to change much between version so until I find a better solution, it will fix the CI for the time being.

ehuss:
Closing as this is resolved by https://github.com/rust-lang/mdBook/pull/3133. The issue is that there was a bug in node 24.16.0. A newer version of puppeteer has a workaround, so we updated that. (It wasn't clear to me if this will also be fixed in node in the future, but AFAIK it hasn't been published, yet.)

PR #3134, opened 2026-06-10
Fix previous-chapter nav hover overlapping sidebar resize handle

sharksforarms:
The nav arrow is position:fixed and starts at the sidebar's right edge, but the resize handle indicator dots protrude 8px beyond the sidebar. The nav's hover background was painting over that zone, hiding the dots.

Adding margin-inline-start equal to the resize indicator width pushes the nav just past the handle so the two no longer overlap.

Before/After:
Screenshot from 2026-06-10 10-49-31
image

Before/After:
Screenshot from 2026-06-10 11-00-25 Screenshot from 2026-06-10 11-02-03

### DIFF
--- a/crates/mdbook-html/front-end/css/chrome.css
+++ b/crates/mdbook-html/front-end/css/chrome.css
@@ -158,6 +158,7 @@ html:not(.js) .left-buttons button {
 }
 
 .nav-chapters.previous {
+    /* Ensure that the previous arrow does not overlap the resize indicator. */
     margin-inline-start: var(--sidebar-resize-indicator-width);
 }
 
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

I pushed a small update to add a comment explaining what it is doing.

There are some minor things, like this adds a margin even when the sidebar is hidden. However, the right "next" arrow also has a small margin. Also, when javascript is disabled, there is no resize indicator. However, I don't think it will be a problem. 


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3134/9873b35405ce8675d6c3ea4f5e853f2d78ed3ea1..96e0e6a8512b56c5def36f3d8bfb21b7f6fd113b)*

PR #3139, opened 2026-06-19
 Add absolute links via output.html.site-url

### LINKED ISSUE #1764: site-url not respected in links

JesusPerez:
Closes #1764.
Supersedes #1802 (a reimplementation of @joaofreires' approach on the current `crates/` workspace layout, with the review feedback addressed).

## Problem

When a book is hosted in a subdirectory (or any page is reached at a deeply nested path), mdBook's depth-relative links can resolve incorrectly. `output.html.site-url` already exists but only feeds the `<base href>` of the 404 page — ordinary navigation, sidebar, asset and in-content links are not affected. See #1764.

## What this does

When `output.html.site-url` is set (e.g. `https://example.com/docs/`), every generated link is rooted at that URL so the book resolves correctly regardless of the page's own depth:

- **In-content links** — root-relative `./path` links in chapter bodies (and images / raw HTML) are anchored to `site-url`. Scheme links (`https:`, `mailto:`) and fragments are left untouched.
- **Page chrome + JS sidebar** — `path_to_root` is set to `site-url`, so prev/next, the print link, search and the `toc.js`-injected sidebar all become absolute.
- **No-JS sidebar fallback** — `toc.html` (the iframe fallback) gets a `<base href>` of `site-url`; this is scoped to that render so it never leaks onto regular pages (which would break their page-relative content links).
- **Static assets** — the `{{resource}}` helper honors the page's `path_to_root`, so hashed CSS/JS are emitted absolutely too.
- **Print page** — `print.html` chrome/assets are rooted at `site-url`, and cross-chapter references between chapters present on the page are still folded into intra-page `#id` anchors, keeping the consolidated page self-contained.

During `mdbook serve` the book is hosted at the local server root, so `site-url` is overridden to `/` (host-agnostic, resolves locally). This was previously silent; it now logs once, and a new `--preserve-site-url` flag keeps the configured value to preview the production absolute links locally.

## Addressing the #1802 review

| ehuss's review point on #1802 | Status here |
| --- | --- |
| "This will need some tests to be added." | 7 integration tests + 4 unit tests (see below). |
| "links on the print page aren't using the adjusted links." | `print.rs::rewrite_links` is now `site-url`-aware; cross-refs fold to anchors, externals stay absolute. |
| "there are some merge conflicts." | N/A — clean reimplementation on current `master`. |
| "`HtmlConfig` is public API → a new field is a semver break; use a wrapper." | No longer applies: in the `crates/` layout `site_url` is **already** a field on `HtmlConfig`. No public API fields were added; threading is via internal `Option<&str>` parameters. `cargo semver-checks` passes. |
| "use `if let` over `match`, `Option<&str>` over `Option<&String>`." | Uses `Option<&str>` (`as_deref()`) and let-chains throughout. |
| "the new option needs to be in the TOML config docs." | Updated (see Docs). |

## Implementation

All link handling lives in `mdbook-html`; no public API was added.

- `crates/mdbook-html/src/html/tree.rs` — `fix_link`/`fix_html_link` take `site_url`; root-relative `./` links are anchored. Threaded at the three call sites via `self.options.config.site_url.as_deref()`.
- `crates/mdbook-html/src/html_handlebars/hbs_renderer.rs` — `path_to_root = site_url` per chapter (and the `index.html` variant); `base_url = site_url` scoped to the `toc.html` render; print page rooted at `site_url`.
- `crates/mdbook-html/src/html_handlebars/helpers/resources.rs` — `{{resource}}` honors `@root/path_to_root` with a fallback (byte-identical output when `site-url` is unset).
- `crates/mdbook-html/src/html/print.rs` — `render_print_page`/`rewrite_links` take `site_url`; the `site_url` prefix is stripped and the remainder resolved as root-relative so in-print cross-refs still fold to `#id`.
- `src/cmd/serve.rs` — `--preserve-site-url` flag; the `/` override is now logged and opt-out.

## Tests

- `tests/testsuite/rendering.rs` — `site_url_*` (7): content link rewriting, external links untouched, absolute `path_to_root`, absolute assets, `<base href>` only on `toc.html`, no `<base href>` leak onto chapter pages, opt-in behavior when unset, and the print page (chrome absolute + cross-refs folded to anchors).
- `crates/mdbook-html/src/html/tree.rs` — `fix_link_tests` (4): `.md`→`.html`, schemes/fragments left alone, `./`→`site-url`, bare-relative links unchanged.
- Fixture: `tests/testsuite/rendering/site_url/`.

`cargo test --workspace`, `cargo clippy --all-targets` and `cargo fmt --check` all pass.

## Docs

- `guide/src/format/configuration/renderers.md` — the `site-url` bullet now describes whole-site absolute links and the serve override.
- `guide/src/cli/serve.md` — `--preserve-site-url` section.

## Notes

- The root-relative convention is the explicit `./path` form, matching the original patch's semantics; bare relative links keep their page-relative meaning.
- A `./` link to the first chapter (which mdBook aliases `README` → `index`) stays an absolute link to its standalone page in the print output rather than folding to an anchor. This is a pre-existing `README`/`index` aliasing quirk, independent of this feature.
- Relationship to `canonical-site-url` (#2706): the two settings are complementary and do not overlap. `site-url` is the deployment location and, with this change, the root for every emitted link/asset so the book resolves under a sub-path. `canonical-site-url` only emits `<link rel="canonical">` for SEO and does not affect link resolution. They touch different code paths and can land independently.
- CHANGELOG.md intentionally left untouched (maintained at release time).

---

Supersedes #1802; builds on @joaofreires' original approach there.

### DIFF
--- a/crates/mdbook-html/src/html/print.rs
+++ b/crates/mdbook-html/src/html/print.rs
@@ -9,15 +9,18 @@ use crate::html::{ChapterTree, Element, serialize};
 use crate::utils::{ToUrlPath, id_from_content, normalize_path, unique_id};
 use mdbook_core::static_regex;
 use std::collections::{HashMap, HashSet};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 /// Takes all the chapter trees, modifies them to be suitable to render for
 /// the print page, and returns an string of all the chapters rendered to a
 /// single HTML page.
-pub(crate) fn render_print_page(mut chapter_trees: Vec<ChapterTree<'_>>) -> String {
+pub(crate) fn render_print_page(
+    mut chapter_trees: Vec<ChapterTree<'_>>,
+    site_url: Option<&str>,
+) -> String {
     let (id_remap, mut id_counter) = make_ids_unique(&mut chapter_trees);
     let path_to_root_id = make_root_id_map(&mut chapter_trees, &mut id_counter);
-    rewrite_links(&mut chapter_trees, &id_remap, &path_to_root_id);
+    rewrite_links(&mut chapter_trees, &id_remap, &path_to_root_id, site_url);
 
     let mut print_content = String::new();
     for ChapterTree { tree, .. } in chapter_trees {
@@ -125,6 +128,7 @@ fn rewrite_links(
     chapter_trees: &mut [ChapterTree<'_>],
     id_remap: &HashMap<PathBuf, HashMap<String, String>>,
     path_to_root_id: &HashMap<PathBuf, String>,
+    site_url: Option<&str>,
 ) {
     static_regex!(
         LINK,
@@ -149,13 +153,21 @@ fn rewrite_links(
                 continue;
             }
             for attr in ["href", "src", "xlink:href"] {
-                let Some(dest) = el.attr(attr) else {
+                let Some(dest) = el.attr(attr).map(str::to_string) else {
                     continue;
                 };
-                let Some(caps) = LINK.captures(&dest) else {
+                // Links emitted under `site-url` are absolute (`{site_url}path`)
+                // and root-relative (anchored at the book root). Strip the
+                // prefix so the path resolves against the print page like any
+                // other chapter link, instead of being skipped as a scheme.
+                let (search, root_relative) = match site_url {
+                    Some(site_url) if dest.starts_with(site_url) => (&dest[site_url.len()..], true),
+                    _ => (dest.as_str(), false),
+                };
+                let Some(caps) = LINK.captures(search) else {
                     continue;
                 };
-                if caps.name("scheme").is_some() {
+                if !root_relative && caps.name("scheme").is_some() {
                     continue;
                 }
                 // The lookup_key is the key to look up in the remap table.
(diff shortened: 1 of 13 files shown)


crates/mdbook-html/src/html/print.rs
```diff
@@ -149,13 +153,21 @@ fn rewrite_links(
...
+                let Some(dest) = el.attr(attr).map(str::to_string) else {
                     continue;
                 };
-                let Some(caps) = LINK.captures(&dest) else {
+                // Links emitted under `site-url` are absolute (`{site_url}path`)
+                // and root-relative (anchored at the book root). Strip the
+                // prefix so the path resolves against the print page like any
+                // other chapter link, instead of being skipped as a scheme.
+                let (search, root_relative) = match site_url {
+                    Some(site_url) if dest.starts_with(site_url) => (&dest[site_url.len()..], true),
```
GuillaumeGomez: What happens if the link is absolute? It won't work locally, right? I guess it should be replaced with `../` until we reach the book root. Works both locally and on websites.

GuillaumeGomez (changes requested):
I'm not sure this is the right path for making this work. Maybe I missed a few things, but absolute paths don't seem to be handled, only ones which start with the site-url.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3139/9873b35405ce8675d6c3ea4f5e853f2d78ed3ea1..0784554c88a63cd47b71fa67d755364932cfa22c)*

PR #3141, opened 2026-06-20
Invalid default-theme and preferred-dark-theme now cause an error

### LINKED ISSUE #3140: Invalid `theme` causes Rendering failed, but an invalid `default-theme` does not
### Problem

Invalid `theme` causes an error and building the book fails, however an invalid `default-theme` does not. I would expect that an invalid `default-theme` would also fail to build.
 
### Steps

1. In `book.toml` add
```
[output.html]
theme = "does-not-exist"
```
2. Try to run `mdbook serve`
3. As expected it gives an error
```
$ mdbook serve
 INFO Book building has started
 INFO Running the html backend
ERROR Rendering failed
        Caused by: theme dir /scratch/does-not-exist does not exist
```
4. 
5. Now, in `book.toml` change `theme` to `default-theme`
``` 
[output.html]
default-theme = "does-not-exist"
```
5. Run `mdbook serve`
6. It succeeds but the css is broken with links looking the same as normal text and a weird header
```
$ mdbook serve
 INFO Book building has started
 INFO Running the html backend
 INFO HTML book written to `/scratch/book`
 INFO Serving on: http://localhost:3000
 INFO Watching for changes...
 ```

Image

Image

### Possible Solution(s)

Make `default-theme` fail to build the book like a normal `theme` does

### Notes

_No response_

### Version

```text
mdbook v0.5.3
```

hollow-owl:
This causes building to fail if `default-theme` or `preferred-dark-theme` is not one of the built-in themes.

If `book.toml` has an invalid `default-theme`:

`mdbook build` outputs an error saying why:
```
 INFO Book building has started
 INFO Running the html backend
ERROR Failed to deserialize `output.html`
        Caused by: unknown variant `does-not-exist`, expected one of `ayu`, `coal`, `light`, `navy`, `rust`
in `default-theme`

 INFO HTML book written to `/scratch/book`
```

Fixes  #3140

### DIFF
--- a/crates/mdbook-core/src/config.rs
+++ b/crates/mdbook-core/src/config.rs
@@ -450,10 +450,10 @@ pub struct HtmlConfig {
     /// The theme directory, if specified.
     pub theme: Option<PathBuf>,
     /// The default theme to use, defaults to 'light'
-    pub default_theme: Option<String>,
+    pub default_theme: Option<BuiltinTheme>,
     /// The theme to use if the browser requests the dark version of the site.
     /// Defaults to 'navy'.
-    pub preferred_dark_theme: Option<String>,
+    pub preferred_dark_theme: Option<BuiltinTheme>,
     /// Supports smart quotes, apostrophes, ellipsis, en-dash, and em-dash.
     pub smart_punctuation: bool,
     /// Support for definition lists.
@@ -728,6 +728,40 @@ trait Updateable<'de>: Serialize + Deserialize<'de> {
 
 impl<'de, T> Updateable<'de> for T where T: Serialize + Deserialize<'de> {}
 
+/// A built-in HTML theme supported by mdBook.
+///
+/// These themes are defined in `crates/mdbook-html/front-end/css/variables.css`.
+///
+/// The enum variants correspond to the theme names accepted by the
+/// `[output.html]` configuration keys `default-theme` and
+/// `preferred-dark-theme`.
+#[derive(Debug, Copy, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase", deny_unknown_fields)]
+#[non_exhaustive]
+pub enum BuiltinTheme {
+    /// A dark theme with a balanced color palette.
+    Ayu,
+    /// A dark theme with a gray-on-gray aesthetic.
+    Coal,
+    /// A light theme optimized for readability.
+    Light,
+    /// A dark navy-themed variant.
+    Navy,
+    /// A rust-colored theme.
+    Rust,
+}
+
+impl BuiltinTheme {
+    /// An array of all built-in themes, in the order they should be displayed in the theme selector.
+    pub const ALL: [BuiltinTheme; 5] = [
+        BuiltinTheme::Light,
+        BuiltinTheme::Rust,
+        BuiltinTheme::Coal,
+        BuiltinTheme::Navy,
+        BuiltinTheme::Ayu,
+    ];
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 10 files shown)


Chocorean (approved):
The string config for default-themes is now replaced by an enum, and deser of the config now fails on an unknown value.  
The list of available themes is now injected into the view rather than hardcoded.  
The documentation is updated, and a few tests are added to config.

Seems good to go! 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3141/3b216a3c654f30baa9f33ed85b2544ee1258fd2a..a15cea74d069d6f00ea56af5c514d48e17c4e56d)*

crates/mdbook-core/src/config.rs
```diff
@@ -728,6 +728,40 @@ trait Updateable<'de>: Serialize + Deserialize<'de> {
...
+    Light,
+    /// A dark navy-themed variant.
+    Navy,
+    /// A rust-colored theme.
+    Rust,
+}
+
+impl BuiltinTheme {
+    /// An array of all built-in themes, in the order they should be displayed in the theme selector.
+    pub const ALL: [BuiltinTheme; 5] = [
```
GuillaumeGomez: Seems too easy to forgot to update this array if `BuiltinTheme` is updated... Do you mind adding a `Max` variant and make `BuiltinTheme` as `repr(usize)` with `Ayu = 0` so then it can become:

```rust
pub const ALL: [BuiltinTheme; BuiltinTheme::Max as usize] = [
```

Like that we can never forget. :)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3141/3b216a3c654f30baa9f33ed85b2544ee1258fd2a..a15cea74d069d6f00ea56af5c514d48e17c4e56d)*

GuillaumeGomez (changes requested):
Small code improvement nit but otherwise looks good to me!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3141/3b216a3c654f30baa9f33ed85b2544ee1258fd2a..a15cea74d069d6f00ea56af5c514d48e17c4e56d)*

crates/mdbook-core/src/config.rs
hollow-owl: Good point, updated it

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks! One last thing remain to be added: a ui test to check the book building fails. <del>However we don't have such a testsuite currently. Do you want to add it or are you ok with me pushing a commit in your PR adding this testsuite?</del> We actually do have a way to check ui, shame on me for forgetting.

tests/testsuite/theme.rs
```diff
@@ -175,4 +175,37 @@ book/fonts/fonts-[..].css
...
+ INFO Running the html backend
+ERROR Failed to deserialize `output.html`
+[TAB]Caused by: unknown variant `does-not-exist`, expected one of `ayu`, `coal`, `light`, `navy`, `rust`, `max`
+in `default-theme`
+
+ INFO HTML book written to `[ROOT]/book`
+
+"#]]);
+    });
+}
```
GuillaumeGomez: Missing a backline at the end of the file, but otherwise, big 👍 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3141/3b216a3c654f30baa9f33ed85b2544ee1258fd2a..f815708e28cce0469b4e195e30b845e7504f6c8b)*

GuillaumeGomez (approved):

GuillaumeGomez:
Looks all good to me, thanks!

PR #3142, opened 2026-06-22
Fix weird layout bug with images in links

GuillaumeGomez:
While working on https://github.com/rust-lang/mdBook/pull/3127, I uncovered this bug when pressing tab around:

Screenshot From 2026-06-22 18-36-19

What you see is the link being focused (an image cannot be focused unless we add `tabindex` to it, which we don't want) when you have an image inside a link (an `` inside a `<a>`). To fix, we simply need to make `` `display: block`. Once done, we have the expected result:

Screenshot From 2026-06-22 18-36-34

(still not the `` focused, it's the `<a>`, but with the correct layout)


crates/mdbook-html/front-end/css/general.css
```diff
@@ -111,6 +111,7 @@ html:not(.js) .page-wrapper,
 .content a { text-decoration: none; }
 .content a:hover { text-decoration: underline; }
 .content img, .content video { max-width: 100%; }
+.content img { display: block; }
```
ehuss: This doesn't quite seem like the correct fix to me because it will cause all images to be block displayed instead of inline. This might change the rendering when it is intended for the image to be inline. As a basic example, look at the badges in this repo's README file, and how they are rendered side-by-side. This would force them to be stacked.

How about instead we change the `a` to have inline-block display instead, like:

```suggestion
/* When an image is inside an anchor, ensure the focus ring covers the entire
   image instead of the anchor's line box. */
.content a:has(img) { display: inline-block; }
```


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3142/3b216a3c654f30baa9f33ed85b2544ee1258fd2a..d54869e047518d731e6d8e1d6c42cde3773172ef)*

GuillaumeGomez: Ah interesting, didn't think about that. I'll add a GUI test as well.

GuillaumeGomez:
Added the GUI test to ensure the links containing images have the expected size and also that they work correctly side by side.

ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3142/cb49cc5523e609a731f27dea1af4395a504815a5..0cbf0dd60e41f422e4b9257997e1867b1ffa92f0)*

Issue #2956, opened 2025-11-20
mdbook-html: missing license texts for favicon (CC-BY-4.0) and JavaScript libraries (MIT, BSD-3-Clause)


decathorpe:
We've been looking into packaging mdbook for Fedora Linux since it's a nice and approachable tool for writing pretty documentation (and it's also used for generating docs for the `nix` package manager, the package for which currently needs to skip this step because mdbook is not yet available as a package from our repositories).

During review of the package sources, we noticed that as-is, several license texts and SPDX license expressions seem to be missing from the repository - and from the contents of published `mdbook-html` crates:

- `crates/mdbook-html/front-end/images/favicon.{svg,png}`: `CC-BY-4.0`
- `crates/mdbook-html/front-end/js/clipboard.min.js`: `MIT`
- `crates/mdbook-html/front-end/js/highlight.js`: `BSD-3-Clause`
- `crates/mdbook-html/front-end/playground_editor/ace.js`: `BSD-3-Clause` - includes a copy of the license text in the .js source file, so this one should be fine
- `crates/mdbook-html/front-end/searcher/elasticlunr.min.js`: `MIT`
- `crates/mdbook-html/front-end/searcher/mark.min.js`: `MIT`
- `crates/mdbook-html/front-end/searcher/elasticlunr.min.js`: `MIT`

All of `CC-BY-4.0`, `MIT`, and `BSD-3-Clause` licenses require that (re)distributed sources contain a copy of the original license text - this is not the case for the files redistributed by the `mdbook-html` crate.

Please ensure that license texts for the `CC-BY-4.0`, `MIT`, and `BSD-3-Clause` licenses are present in this git repository *and* included when publishing the `mdbook-html` crate to crates.io to be compliant with the license terms of the listed projects.

Additionally, it would be good to adapt the `package.license` from `license.workspace = true` for this crate to ensure that consumers of the crate are aware that it not only contains content that is available under the `MPL-2.0` license, but also content that is available under the other three listed licenses (i.e. use `license = "MPL-2.0 AND BSD-3-Clause AND CC-BY-4.0 AND MIT"` in `crates/mdbook-html/Cargo.toml`).

PS: I also see references to the MIT-licensed "fontawesome" project in project sources and other places, but I don't see any actual files included from it. It appears that the fontawesome icon font used to be included in this repository in <v0.5.0 but has now been moved to an external dependency (`font-awesome-as-a-crate`), so it appears that these references are stale.

PR #3144, opened 2026-06-24
mdbook-html: Account for licenses of bundled JavaScript libraries and artwork

decathorpe:
This is an attempt to address issue #2956

The first commit adds the missing license texts:

- The MIT and BSD-3-Clause license texts are taken verbatim from the respective projects.
- The CC-BY-4.0 license text is the canonical copy from creativecommons.org.

favicon license (canonical copy of the CC-BY-4.0 text):
https://creativecommons.org/licenses/by/4.0/legalcode.txt

clipboard.js license text:
https://github.com/zenorocha/clipboard.js/blob/v2.0.5/LICENSE

highlight.js license text:
https://github.com/highlightjs/highlight.js/blob/10.1.1/LICENSE

elasticlunr.js license text:
https://github.com/weixsong/elasticlunr.js/blob/v0.9.5/LICENSE

mark.js license text:
https://github.com/julkue/mark.js/blob/8.11.1/LICENSE

The second commit updates the `package.license` of the mdbook-html crate to account for all included code and content.

### DIFF
--- a/crates/mdbook-html/Cargo.toml
+++ b/crates/mdbook-html/Cargo.toml
@@ -3,7 +3,7 @@ name = "mdbook-html"
 version = "0.5.3"
 description = "mdBook HTML renderer"
 edition.workspace = true
-license.workspace = true
+license = "MPL-2.0 AND Apache-2.0 AND BSD-3-Clause AND CC-BY-4.0 AND MIT AND OFL-1.1"
 repository.workspace = true
 rust-version.workspace = true
 
--- a/crates/mdbook-html/front-end/images/LICENSE-CC-BY-4.0
+++ b/crates/mdbook-html/front-end/images/LICENSE-CC-BY-4.0
@@ -1,0 +1,396 @@
+Attribution 4.0 International
+
+=======================================================================
+
+Creative Commons Corporation ("Creative Commons") is not a law firm and
+does not provide legal services or legal advice. Distribution of
+Creative Commons public licenses does not create a lawyer-client or
+other relationship. Creative Commons makes its licenses and related
+information available on an "as-is" basis. Creative Commons gives no
+warranties regarding its licenses, any material licensed under their
+terms and conditions, or any related information. Creative Commons
+disclaims all liability for damages resulting from their use to the
+fullest extent possible.
+
+Using Creative Commons Public Licenses
+
+Creative Commons public licenses provide a standard set of terms and
+conditions that creators and other rights holders may use to share
+original works of authorship and other material subject to copyright
+and certain other rights specified in the public license below. The
+following considerations are for informational purposes only, are not
+exhaustive, and do not form part of our licenses.
+
+     Considerations for licensors: Our public licenses are
+     intended for use by those authorized to give the public
+     permission to use material in ways otherwise restricted by
+     copyright and certain other rights. Our licenses are
+     irrevocable. Licensors should read and understand the terms
+     and conditions of the license they choose before applying it.
+     Licensors should also secure all rights necessary before
+     applying our licenses so that the public can reuse the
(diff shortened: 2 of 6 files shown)


crates/mdbook-html/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "mdbook-html"
 version = "0.5.3"
 description = "mdBook HTML renderer"
 edition.workspace = true
-license.workspace = true
+license = "MPL-2.0 AND Apache-2.0 AND BSD-3-Clause AND CC-BY-4.0 AND MIT AND OFL-1.1"
```
GuillaumeGomez, line 6: I wonder if the license field is for the whole crate or for the rust code. And even worse: if we don't enable the "editable codeblock" feature, some of the licenses are not in use.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3144/cb49cc5523e609a731f27dea1af4395a504815a5..921a74efee227398a1aebb39cb0fdde1f8400520)*

decathorpe: > I wonder if the license field is for the whole crate or for the rust code.

Right, I have not ever seen a conclusive decision about this - but since this is the license string that is prominently displayed in various places (like the crates.io page) ... 🤷🏼‍♂️

For what it's worth, the legal guidance we have received from Red Hat lawyers for Fedora packaging is that our license metadata should just enumerate the licenses of everything that's included in a package.

> And even worse: if we don't enable the "editable codeblock" feature, some of the licenses are not in use.

Barring conditional crate metadata ("if feature foo is enabled then license this else that") this is not expressible. Additionally, this seems to be related to the kind of book content that mdbook generates *at runtime* -  but that has no bearing on whether this content (JavaScript, etc.) gets compiled / statically embedded into the mdbook executable, which happens unconditionally.

GuillaumeGomez: Yeah no clue.

GuillaumeGomez:
@Manishearth Should we bring this topic up to the project?

crates/mdbook-html/Cargo.toml
decathorpe: Including license texts for all bundled components should be the uncontroversial part here though, I can drop changing the `package.license` metadata for the mdbook-html crate from this PR for now, if that helps?

GuillaumeGomez: I pinged someone who might know more about, so let's just wait for their input. Sorry for they delay here, it's the kind of things you don't want to mess up so rather than making you do more changes for nothing, better to wait for their input.

joshtriplett: > And even worse: if we don't enable the "editable codeblock" feature, some of the licenses are not in use.

The only way I know of to fix that is to split the bits with different licenses into a separate crate that you only pull in with a given feature. Then, tools that check licenses can check which crates you actually pull in as part of dependency resolution, and not check crates you *don't* pull in.

joshtriplett:
This looks reasonable to me.

crates/mdbook-html/Cargo.toml
decathorpe: Right, in general, this is the approach we use. But this comes back to the "editable codeblock" feature being a feature of mdbook that's (currently) not disable-able through cargo feature flags.

GuillaumeGomez (approved):

crates/mdbook-html/Cargo.toml
GuillaumeGomez: Then I guess it's fine. We can always "fine-tune" it later on. Thanks to both of you!

decathorpe:
Thank you!

PR #3146, opened 2026-06-26
Fix empty sidebar entry for a heading that is a link

### LINKED ISSUE #3126: Sidebar header navigation renders empty anchor when heading is a Markdown link
### Problem

A heading is a Markdown link causes the sidebar header navigation item to render as an empty anchor.

For example, a heading like:

```md
## [fzf](https://github.com/junegunn/fzf)
```

can produce sidebar HTML like:

```html
<li class="header-item expanded">
  <span class="chapter-link-wrapper">
    <a href="#fzf" class="header-in-summary current-header"></a>
  </span>
</li>
```

The `<a>` is present, but it has no text content, so the sidebar entry appears blank.

The sidebar generates a matching anchor element for the heading, but the rendered text content is empty.

See https://sigoden.github.io/mynotes/collections/cli.html for an example.

Image

### Steps

1. run `mdbook init`
2. modify src/chapter_1.md
```
# Chapter 1

## [fzf](https://github.com/junegunn/fzf)

An interactive filter program for any kind of list.

## [rg](https://github.com/BurntSushi/ripgrep)

Fast line-oriented search tool.
```
3. run `mdbook serve`

### Possible Solution(s)

The sidebar header navigation should display the heading text, even if the heading is a Markdown link.

At minimum, headings with links should not produce an empty sidebar entry.

### Notes

_No response_

### Version

```text
mdbook v0.5.3
```

chatman-media:
Closes #3126.

When a heading's whole content is a Markdown link (e.g. `## [fzf](url)`), the injected `<a class="header">` permalink wraps the link, producing nested anchors in the generated HTML. Browsers hoist the inner link out as a sibling of the (now empty) permalink anchor, so the sidebar's `filterHeader(header.children[0], …)` copied the empty anchor and rendered a blank entry.

The sidebar entry is now built from the whole heading with the permalink anchor unwrapped, so the heading text (and any embedded link) comes through. This also fixes the previously-silent loss of an embedded link in headings like `## Heading with a [link](…)`.

Verified the parsed-DOM behavior and the fix against jsdom for normal / markup / `<mark>` / link-only / text+link headings (link-only goes from `""` to `"fzf"`; others unchanged). Extended `heading-nav-markup.goml` with a link-only heading plus assertions for both link cases.

### DIFF
--- a/crates/mdbook-html/front-end/templates/toc.js.hbs
+++ b/crates/mdbook-html/front-end/templates/toc.js.hbs
@@ -336,10 +336,17 @@ window.customElements.define('mdbook-sidebar-scrollbox', MDBookSidebarScrollbox)
         });
     }
 
-    // Takes the nodes from the given head and copies them over to the
-    // destination, along with some filtering.
+    // Takes the rendered content of the given heading and copies it over to
+    // the destination, along with some filtering.
     function filterHeader(source, dest) {
         const clone = source.cloneNode(true);
+        // Unwrap the injected permalink anchor (`<a class="header">`), keeping
+        // its children. When the heading itself is a link, the browser hoists
+        // that link out as a sibling of the (now empty) permalink anchor, so
+        // unwrapping here lets the actual heading text come through.
+        clone.querySelectorAll('a.header').forEach(headerLink => {
+            headerLink.replaceWith(...headerLink.childNodes);
+        });
         clone.querySelectorAll('mark').forEach(mark => {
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
I don't think this is the right fix. Currently, the whole heading is the anchor. We could do like rustdoc and add an anchor alongside the heading, but it's a major UI change. Adding label for it to be discussed with the team.

Issue #2677, opened 2025-04-23
Improve test coverage


ehuss:
With the introduction of https://github.com/rust-lang/mdBook/pull/2676, I'd like to make some progress on improving the test coverage. Some ideas for additions and improvements:

- Try to get full coverage for all features in the integration testsuite. Some examples;
    - `{{#playground example.rs editable}}`
    - TOC fold tests
    - `{{#title}}`
    - All CLI flags (very few are currently covered)
    - `init` with git author configured
    - `init` check contents of `.gitignore`
    - Check that renderer or preprocessor uses PATH to locate `mdbook-FOO`
    - Any config option in book.toml that isn't in a test.
    - Tests for configuring book.toml via environment variables.
    - Hunt for error cases not covered (bonus to use coverage tools to drive this)
- Check the cli help output for every command (see `cli::help`)
    - Fix the term.svg files so that they have color. Offhand I don't remember how to force color to enable.
- The `toc` tests could use some review and cleanup.
    - Instead of using `select`, would it make sense to validate the contents of the raw HTML instead? (This is also hampered by the lack of the ability to prettify the HTML, which is otherwise all on one line.)
- Can pretty_assertions be replaced with snapbox assert? Or should we just remove it?

coyaSONG:
@rustbot claim

I am taking the remaining integration-coverage item for `{{#playground example.rs editable}}`. I will keep the change scoped to the include fixture and its rendered-output assertion.

PR #3160, opened 2026-07-14
test: cover editable playground includes

coyaSONG:
## Summary

- extend the existing playground include fixture with the `editable` form
- assert that the rendered code block preserves the `editable` class

This covers one of the remaining integration-test cases listed in #2677 without changing runtime behavior.

## Validation

- red proof: the targeted test failed after adding the fixture because the snapshot did not contain the editable heading and `language-rust editable` class
- `cargo test --test testsuite includes::playground_include -- --exact`
- `cargo test --test testsuite`
- `cargo test --workspace`
- `cargo test --workspace --no-default-features`
- `cargo clippy --workspace --all-targets --no-deps -- -D warnings`
- `cargo fmt --all --check`
- `git diff --check`

## Disclosure

This change was prepared with OpenAI Codex assistance and manually reviewed and validated with the commands above.

### DIFF
--- a/tests/testsuite/includes.rs
+++ b/tests/testsuite/includes.rs
@@ -83,6 +83,13 @@ fn playground_include() {
 </span><span class="boring">   // You can even hide lines! :D
 </span><span class="boring">  println!("I am hidden! Expand the code snippet to see me");
 </span>}</code></pre>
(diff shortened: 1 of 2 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks!

PR #3162, opened 2026-07-15
Support percent-encoded chapter paths

### LINKED ISSUE #1640: Please make SUMMARY.md support urlencode markdown file name
Example,There is a markdown file with a question mark in its file name. 
``` text
How can I get data from AsyncStorage and add it to initial state when using useReducer?.md
```

now i have to show it in SUMMARY.md
``` markdown
- [How can I get data from AsyncStorage and add it to initial state when using useReducer?](React/How can I get data from AsyncStorage and add it to initial state when using useReducer?)
```
and the result in toc is
``` html
<a href="React/How can I get data from AsyncStorage and add it to initial state when using useReducer?.html">How can I get data from AsyncStorage and add it to initial state when using useReducer?</a>
```
after click the link, the page will jump to the wrong url, the part "?.html" of the url will be the search part of url.

I hope I can write the markdown as
``` markdown
- [How can I get data from AsyncStorage and add it to initial state when using useReducer?](React/How%20can%20I%20get%20data%20from%20AsyncStorage%20and%20add%20it%20to%20initial%20state%20when%20using%20useReducer%3F.md)
```
and the result is
```
<a href="React/How%20can%20I%20get%20data%20from%20AsyncStorage%20and%20add%20it%20to%20initial%20state%20when%20using%20useReducer%3F.html">How can I get data from AsyncStorage and add it to initial state when using useReducer?</a>
```
so, please support 
1. someone can urldecode the file name in SUMMARY.md;
2. urldecode the file path before check file exists;
3. urlencode the name of markdown file before output as html;

ychampion:
## Summary

- decode percent-encoded chapter paths before reading source files
- encode generated chapter URLs in navigation, TOC, redirects, and search output

## Why

`SUMMARY.md` only decoded `%20`, so a path such as `question%3F.md` could not load `question?.md`. Emitting the decoded path unchanged would also turn `?` and `#` filename characters into URL delimiters.

Closes #1640.

## Validation

- `cargo test -p mdbook-summary -p mdbook-html` — passed
- `cargo test --test testsuite toc::percent_encoded_chapter_paths -- --exact` — passed
- `cargo test --test testsuite toc::percent_encoded_chapter_paths --no-default-features -- --exact` — passed
- `cargo clippy --workspace --all-targets --no-deps -- -D warnings` — passed
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --document-private-items --no-deps` — passed
- `cargo fmt --all --check` — passed


crates/mdbook-html/src/utils.rs
```diff
@@ -113,6 +133,14 @@ mod tests {
         assert_eq!(unique_id("Über", &mut id_counter), "Über-2");
     }
 
+    #[test]
+    fn percent_encodes_url_paths() {
+        assert_eq!(
+            Path::new("nested/spatiëring ?#%^.html").to_encoded_url_path(),
```
GuillaumeGomez, line 139: Is it really worth it to encode more than the characters not allowed in paths? Not sure it's worth encoding unicode characters for example.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3162/69287f26827e160e99d30c838d2dafa6ca9bb015..6267748ba7d13b6eabe44d0531c00304cc93cf8c)*

PR #2664, opened 2025-04-15
Add typos for CI, include typos.toml

miguelraz:
Addresses #2660

PR #3163, opened 2026-07-16
fix(links): support spaces in #include file paths

### LINKED ISSUE #2812: #include fails with a space in the file name
### Problem

When using {{#include file}}, if the file name has a space if fails to find the file and include it

Confirmed that path was correct with multiple files.

Tried
```markdown
{{#include ../source/fila a.md}}
{{#include ../source/fila%20a.md}}
{{#include <../source/fila a.md>}}
```

### Steps

1. add include for file with space
2. get error: Error Updating ... Could not read file for link
3. [WARN] No such file or directory

### Possible Solution(s)

_No response_

### Notes

using macOS 15.2

### Version

```text
mdbook v0.4.52
```

alexchen-sys:
## Summary

`{{#include}}`, `{{#rustdoc_include}}`, and `{{#playground}}` now support file paths containing spaces by wrapping the path in double quotes (e.g. `{{#include "file with spaces.md"}}`). Line ranges and anchors still work (`{{#include "file with spaces.md":1:10}}`). Unquoted paths keep full backward compatibility.

Backslash escaping only covers `\"` (literal quote) and `\\` (literal backslash). Any other `\X` preserves the backslash, so Windows paths like `C:\Users\file.md` work correctly.

Fixes #2812.

## Test plan

- [x] Unit tests in `links.rs` for quoted paths with spaces (with and without line ranges)
- [x] Escape semantics tests: `\t`, `\n`, `\\`, `\"`, trailing backslash, Unicode + backslash
- [x] Cross-platform tests: Windows absolute, mixed separators, Cyrillic, colons in name
- [x] Architecture tests: DRY consistency between `parse_include_path` and `parse_rustdoc_include_path`
- [x] Extended `includes/all_includes` testsuite case
- [x] `cargo test -p mdbook-driver --lib builtin_preprocessors::links::tests` — 88 passed
- [x] `cargo fmt --all` and `cargo clippy -p mdbook-driver --lib --no-deps -- -D warnings`
- [ ] CI

### DIFF
--- a/crates/mdbook-driver/src/builtin_preprocessors/links.rs
+++ b/crates/mdbook-driver/src/builtin_preprocessors/links.rs
@@ -285,13 +285,21 @@ impl<'a> Link<'a> {
             (_, Some(typ), Some(title)) if typ.as_str() == "title" => {
                 Some(LinkType::Title(title.as_str()))
             }
+            // Include directives take the full remaining capture as the path so
+            // file names may contain spaces. Playground still splits on whitespace
+            // because it accepts trailing properties after the path.
+            (_, Some(typ), Some(rest)) if typ.as_str() == "include" => {
+                Some(parse_include_path(rest.as_str().trim()))
+            }
+            (_, Some(typ), Some(rest)) if typ.as_str() == "rustdoc_include" => {
+                Some(parse_rustdoc_include_path(rest.as_str().trim()))
+            }
             (_, Some(typ), Some(rest)) => {
                 let mut path_props = rest.as_str().split_whitespace();
                 let file_arg = path_props.next();
                 let props: Vec<&str> = path_props.collect();
 
                 match (typ.as_str(), file_arg) {
-                    ("include", Some(pth)) => Some(parse_include_path(pth)),
                     ("playground", Some(pth)) => Some(LinkType::Playground(pth.into(), props)),
                     ("playpen", Some(pth)) => {
                         warn!(
@@ -301,7 +309,6 @@ impl<'a> Link<'a> {
                         );
                         Some(LinkType::Playground(pth.into(), props))
                     }
-                    ("rustdoc_include", Some(pth)) => Some(parse_rustdoc_include_path(pth)),
                     _ => None,
                 }
             }
@@ -640,6 +647,60 @@ mod tests {
                     RangeOrAnchor::Range(LineRange::from(..))
                 ),
                 link_text: "{{#include file.rs}}",
+            }]
+        );
+    }
+
+    #[test]
+    fn test_find_links_with_space_in_path() {
+        let s = "Some random text with {{#include fila a.md}}...";
+        let res = find_links(s).collect::<Vec<_>>();
+        assert_eq!(
+            res,
+            vec![Link {
+                start_index: 22,
+                end_index: 44,
+                link_type: LinkType::Include(
+                    PathBuf::from("fila a.md"),
+                    RangeOrAnchor::Range(LineRange::from(..)),
+                ),
+                link_text: "{{#include fila a.md}}",
+            }]
+        );
+    }
+
+    #[test]
+    fn test_find_links_with_space_in_path_and_range() {
+        let s = "Some random text with {{#include fila a.md:1:2}}...";
+        let res = find_links(s).collect::<Vec<_>>();
+        assert_eq!(
+            res,
+            vec![Link {
+                start_index: 22,
+                end_index: 48,
+                link_type: LinkType::Include(
+                    PathBuf::from("fila a.md"),
+                    RangeOrAnchor::Range(LineRange::from(0..2)),
+                ),
+                link_text: "{{#include fila a.md:1:2}}",
+            }]
+        );
+    }
+
+    #[test]
+    fn test_find_links_rustdoc_include_with_space_in_path() {
+        let s = "Some random text with {{#rustdoc_include fila a.rs}}...";
+        let res = find_links(s).collect::<Vec<_>>();
+        assert_eq!(
+            res,
+            vec![Link {
+                start_index: 22,
+                end_index: 52,
+                link_type: LinkType::RustdocInclude(
+                    PathBuf::from("fila a.rs"),
+                    RangeOrAnchor::Range(LineRange::from(..)),
+                ),
+                link_text: "{{#rustdoc_include fila a.rs}}",
             }]
         );
     }
--- a/crates/mdbook-driver/src/builtin_preprocessors/links.rs
+++ b/crates/mdbook-driver/src/builtin_preprocessors/links.rs
@@ -261,22 +261,89 @@ fn parse_range_or_anchor(parts: Option<&str>) -> RangeOrAnchor {
     }
 }
 
-fn parse_include_path(path: &str) -> LinkType<'static> {
-    let mut parts = path.splitn(2, ':');
+fn parse_quoted_path(input: &str) -> Option<(PathBuf, &str)> {
+    let input = input.trim_start();
+    let after_quote = input.strip_prefix('"')?;
+    let mut path = String::new();
+    let mut chars = after_quote.char_indices();
+    let mut closed = false;
+    let mut end_index = 0;
 
-    let path = parts.next().unwrap().into();
-    let range_or_anchor = parse_range_or_anchor(parts.next());
+    while let Some((idx, ch)) = chars.next() {
+        if ch == '\\' {
+            if let Some((_, next_ch)) = chars.next() {
+                path.push(next_ch);
+            }
+        } else if ch == '"' {
+            closed = true;
+            end_index = idx + 1;
+            break;
+        } else {
+            path.push(ch);
+        }
+    }
 
-    LinkType::Include(path, range_or_anchor)
+    if closed {
+        Some((PathBuf::from(path), &after_quote[end_index..]))
+    } else {
+        None
+    }
 }
 
-fn parse_rustdoc_include_path(path: &str) -> LinkType<'static> {
-    let mut parts = path.splitn(2, ':');
+fn parse_include_path(path: &str) -> Option<LinkType<'static>> {
+    let trimmed = path.trim();
+    if trimmed.starts_with('"') {
+        let (path, after) = parse_quoted_path(trimmed)?;
+        let after = after.trim_start();
+        let range_or_anchor = if let Some(range_part) = after.strip_prefix(':') {
+            parse_range_or_anchor(Some(range_part))
+        } else {
+            parse_range_or_anchor(None)
+        };
+        Some(LinkType::Include(path, range_or_anchor))
+    } else {
+        let mut path_props = trimmed.split_whitespace();
+        let file_arg = path_props.next()?;
+        let mut parts = file_arg.splitn(2, ':');
+        let path = parts.next().unwrap().into();
+        let range_or_anchor = parse_range_or_anchor(parts.next());
+        Some(LinkType::Include(path, range_or_anchor))
+    }
+}
 
-    let path = parts.next().unwrap().into();
-    let range_or_anchor = parse_range_or_anchor(parts.next());
+fn parse_rustdoc_include_path(path: &str) -> Option<LinkType<'static>> {
+    let trimmed = path.trim();
+    if trimmed.starts_with('"') {
+        let (path, after) = parse_quoted_path(trimmed)?;
+        let after = after.trim_start();
+        let range_or_anchor = if let Some(range_part) = after.strip_prefix(':') {
+            parse_range_or_anchor(Some(range_part))
+        } else {
+            parse_range_or_anchor(None)
+        };
+        Some(LinkType::RustdocInclude(path, range_or_anchor))
+    } else {
+        let mut path_props = trimmed.split_whitespace();
+        let file_arg = path_props.next()?;
+        let mut parts = file_arg.splitn(2, ':');
+        let path = parts.next().unwrap().into();
+        let range_or_anchor = parse_range_or_anchor(parts.next());
+        Some(LinkType::RustdocInclude(path, range_or_anchor))
+    }
+}
 
-    LinkType::RustdocInclude(path, range_or_anchor)
+fn parse_playground_args(rest: &str) -> Option<(PathBuf, Vec<&str>)> {
+    let trimmed = rest.trim_start();
+    if trimmed.starts_with('"') {
+        let (path, after) = parse_quoted_path(trimmed)?;
+        let props: Vec<&str> = after.split_whitespace().collect();
+        Some((path, props))
+    } else {
+        let mut path_props = trimmed.split_whitespace();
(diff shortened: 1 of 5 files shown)


tests/testsuite/includes/all_includes/src/includes.md
```diff
@@ -2,3 +2,7 @@
 
 {{#include sample.md}}
 
+## File with space in name
+
+{{#include fila a.md}}
```
GuillaumeGomez: I think the right approach should be to put the file name between quote `"`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3163/69287f26827e160e99d30c838d2dafa6ca9bb015..bcca7955ff7debeca01f2ede41c0199508445ad7)*

alexchen-sys:
Updated to require double quotes for file paths containing spaces across `include`, `rustdoc_include`, and `playground` (e.g. `{{#include "file with spaces.md":1:10}}` or `{{#playground "file with spaces.rs" editable}}`). Unquoted paths retain full backwards compatibility, and backslash escaping (`\"`) inside quoted paths is supported.

alexchen-sys:
Rebased and fixed the conflict with #2664.

GuillaumeGomez:
I didn't see a testcase where the quoted path is unclosed. It should emit an explicit error for such cases.

alexchen-sys:
Done in 14712f25 — unclosed quotes now emit an explicit `unclosed quote in path` error, covered by unit and integration tests.

crates/mdbook-driver/src/builtin_preprocessors/links.rs
```diff
@@ -261,22 +262,103 @@ fn parse_range_or_anchor(parts: Option<&str>) -> RangeOrAnchor {
...
+        Err("unclosed quote in path")
+    }
 }
 
-fn parse_rustdoc_include_path(path: &str) -> LinkType<'static> {
-    let mut parts = path.splitn(2, ':');
+fn parse_include_path(path: &str) -> Option<LinkType<'static>> {
+    let trimmed = path.trim();
+    if trimmed.starts_with('"') {
+        match parse_quoted_path(trimmed) {
```
GuillaumeGomez:
```suggestion
    if let Some(trimmed) = trimmed.strip_prefix('"') {
        match parse_quoted_path(trimmed) {
```

Meaning that in `parse_quoted_path` you can remove this code:

```rust
    let input = input.trim_start();
    let Some(after_quote) = input.strip_prefix('"') else {
        return Err("expected opening quote");
    };
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3163/78cd8333bf0f354d846ca0eb7545aea168a6d659..14712f257c23eef42724f8ba6883b2d23d598eb4)*

crates/mdbook-driver/src/builtin_preprocessors/links.rs
```diff
@@ -261,22 +262,103 @@ fn parse_range_or_anchor(parts: Option<&str>) -> RangeOrAnchor {
...
+    }
+}
 
-    let path = parts.next().unwrap().into();
-    let range_or_anchor = parse_range_or_anchor(parts.next());
+fn parse_rustdoc_include_path(path: &str) -> Option<LinkType<'static>> {
+    let trimmed = path.trim();
+    if trimmed.starts_with('"') {
+        match parse_quoted_path(trimmed) {
+            Ok((path, after)) => {
```
GuillaumeGomez: Please write a function to avoid duplicating the code between here and `parse_include_path` for the handling of the quoted path.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3163/78cd8333bf0f354d846ca0eb7545aea168a6d659..14712f257c23eef42724f8ba6883b2d23d598eb4)*

crates/mdbook-driver/src/builtin_preprocessors/links.rs
```diff
@@ -261,22 +262,103 @@ fn parse_range_or_anchor(parts: Option<&str>) -> RangeOrAnchor {
...
 }
 
-fn parse_rustdoc_include_path(path: &str) -> LinkType<'static> {
-    let mut parts = path.splitn(2, ':');
+fn parse_include_path(path: &str) -> Option<LinkType<'static>> {
+    let trimmed = path.trim();
+    if trimmed.starts_with('"') {
+        match parse_quoted_path(trimmed) {
+            Ok((path, after)) => {
+                let after = after.trim_start();
```
GuillaumeGomez, line 313: Btw, we should really not `trim` inside a quoted string. If it's quoted, we keep it's whole content as is (whitespace characters are allowed in paths).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3163/78cd8333bf0f354d846ca0eb7545aea168a6d659..14712f257c23eef42724f8ba6883b2d23d598eb4)*

crates/mdbook-driver/src/builtin_preprocessors/links.rs
```diff
@@ -261,22 +262,103 @@ fn parse_range_or_anchor(parts: Option<&str>) -> RangeOrAnchor {
...
+    };
+    let mut path = String::new();
+    let mut chars = after_quote.char_indices();
+    let mut closed = false;
+    let mut end_index = 0;
+
+    while let Some((idx, ch)) = chars.next() {
+        if ch == '\\' {
+            if let Some((_, next_ch)) = chars.next() {
+                path.push(next_ch);
```
GuillaumeGomez: Just thinking: do we need to keep the `\`? Like what happens for `\t` or `\n` for example?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3163/78cd8333bf0f354d846ca0eb7545aea168a6d659..14712f257c23eef42724f8ba6883b2d23d598eb4)*

alexchen-sys:
Thanks for the review! Updated:

- `parse_quoted_path` uses `strip_prefix` now, no more trimming inside quotes (whitespace in a quoted path is part of the path)
- Extracted `parse_quoted_or_plain` since the include and rustdoc_include parsers share the same logic, just different `LinkType` (playground left as-is, needs `Vec<&str>`)
- Only `\"` and `\\` are escapes. Other backslashes stay literal so Windows paths like `C:\Users\file.md` do not break. Tests added

GuillaumeGomez:
Thanks! Nominating it for team discussion.

cc @rust-lang/mdbook

crates/mdbook-driver/src/builtin_preprocessors/links.rs
```diff
@@ -261,22 +262,102 @@ fn parse_range_or_anchor(parts: Option<&str>) -> RangeOrAnchor {
...
+    let trimmed = raw.trim();
+    if trimmed.starts_with('"') {
+        match parse_quoted_path(trimmed) {
+            Ok((path, after)) => {
+                let after = after.trim_start();
+                let range_or_anchor = if let Some(range_part) = after.strip_prefix(':') {
+                    parse_range_or_anchor(Some(range_part))
+                } else {
+                    parse_range_or_anchor(None)
+                };
```
Chocorean: Can't we just do this :

```rust
let range_or_anchor = parse_range_or_anchor(after.strip_prefix(':'));
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3163/54e4f3bf6d25b4b1b5082d9979f126f0b4129831..3f2c5e491e017040f2d68de861665f84468c5e0b)*

tests/testsuite/includes.rs
```diff
@@ -13,6 +13,9 @@ fn include() {
 <h1 id="basic-includes"><a class="header" href="#basic-includes">Basic Includes</a></h1>
 <h2 id="sample"><a class="header" href="#sample">Sample</a></h2>
 <p>This is a sample include.</p>
+<h2 id="file-with-space-in-name"><a class="header" href="#file-with-space-in-name">File with space in name</a></h2>
+<h1 id="file-with-space"><a class="header" href="#file-with-space">File With Space</a></h1>
```
alexchen-sys, line 17: That's just source order — the h2 is written in includes.md above the include directive, and `fila a.md` carries its own `#` heading. I thought about tidying the hierarchy (demoting the included heading to `###`) but left it as is on purpose: the fixture is there to show the included content lands verbatim exactly where the directive sits, whatever heading levels the book author picked. Includes are textual, so heading structure stays the author's business.

crates/mdbook-driver/src/builtin_preprocessors/links.rs
alexchen-sys: Yes, that works — `strip_prefix` already returns exactly the `Option<&str>` the parser takes. Applied, thanks

Chocorean (commented):
Would be great to document the improvements in the guide at `guide/src/format/mdbook.md#including-files`

Good work!*

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3163/54e4f3bf6d25b4b1b5082d9979f126f0b4129831..76a5ab1a72d7e1cbd2f12d3a5cc40a372eb2a5eb)*

alexchen-sys:
Good point, added a short section on quoted paths to the guide, including escapes and the unclosed-quote error

GuillaumeGomez:
https://github.com/rust-lang/mdBook/pull/3163#discussion_r3847768432 and https://github.com/rust-lang/mdBook/pull/3163#discussion_r3847783275 are still pending btw. And please add a test to ensure that `"  spaces   "` keeps the whitespace characters.

crates/mdbook-driver/src/builtin_preprocessors/links.rs
alexchen-sys: Addressed in 3f2c5e4 — dropped the trim entirely; the call site already guarantees the opening quote, `parse_quoted_path` just consumes it and keeps everything after it verbatim

crates/mdbook-driver/src/builtin_preprocessors/links.rs
alexchen-sys: Agreed — the trim is gone, quoted content is kept as is. Pinned it down in 2297747: `"  spaces   "` parses with the whitespace untouched

alexchen-sys:
Both applied back in 3f2c5e4, I just never replied on those threads — my bad. Added the whitespace test in 2297747: `parse_quoted_path_preserves_whitespace` checks that `"  spaces   "` keeps all whitespace characters

crates/mdbook-driver/src/builtin_preprocessors/links.rs
GuillaumeGomez: No, this is not what I asked for. `parse_quoted_path` shouldn't have to trim any `"` character.

crates/mdbook-driver/src/builtin_preprocessors/links.rs
```diff
@@ -941,4 +1139,792 @@ mod tests {
...
+
+    // ========================================================================
+    // DRY / ARCHITECTURE: A/B unit-tests
+    // PR mdBook#3163 — reviewer GuillaumeGomez flagged copy-paste between
+    // parse_include_path and parse_rustdoc_include_path. These tests
+    // verify the architectural property that all three parsers behave
+    // consistently and quantify the duplication.
+    // ========================================================================
+
+    // --- 1. Consistency: parse_include_path vs parse_rustdoc_include_path ---
```
GuillaumeGomez: Which LLM did you use? Because these comments are not very useful.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3163/54e4f3bf6d25b4b1b5082d9979f126f0b4129831..22977474cbbdce3039f2032efc8665061ff883b4)*

crates/mdbook-driver/src/builtin_preprocessors/links.rs
```diff
@@ -941,4 +1139,792 @@ mod tests {
...
+            LinkType::Include(p, _) => Some(p),
+            LinkType::RustdocInclude(p, _) => Some(p),
+            LinkType::Playground(p, _) => Some(p),
+            LinkType::Invalid(_) => None,
+            LinkType::Escaped => None,
+            LinkType::Title(_) => None,
+        }
+    }
+
+    // ========================================================================
```
GuillaumeGomez: Same here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3163/54e4f3bf6d25b4b1b5082d9979f126f0b4129831..22977474cbbdce3039f2032efc8665061ff883b4)*

crates/mdbook-driver/src/builtin_preprocessors/links.rs
```diff
@@ -941,4 +1139,792 @@ mod tests {
...
+        // A (bug):  `\"` consumes next char, potential shift of end_index
+        // B (correct): `\"` -> `"`, scanning continues, rest correct
+        let result = parse_quoted_path(r#""a\"b\\c.md":2:4"#);
+        assert!(result.is_ok());
+        let (path, rest) = result.unwrap();
+        assert_eq!(path, PathBuf::from(r#"a"b\c.md"#));
+        assert_eq!(rest, ":2:4");
+    }
+
+    // ========================================================================
```
GuillaumeGomez: Same.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3163/54e4f3bf6d25b4b1b5082d9979f126f0b4129831..22977474cbbdce3039f2032efc8665061ff883b4)*

crates/mdbook-driver/src/builtin_preprocessors/links.rs
alexchen-sys: Kimi K3, if that matters. The code and test logic are mine: I use the model as a second pair of eyes, to run my tests and double-check my reasoning. The A (bug) / B (correct) labels are how I write regression tests; the banner blocks around them aren't — the model added those on its own this time, and I pushed without noticing, that's on me. Banners are gone, tests are plain assertions, and I've gone through the final diff line by line.

crates/mdbook-driver/src/builtin_preprocessors/links.rs
alexchen-sys: Right, both callers strip the opening quote with `strip_prefix` now, so `parse_quoted_path` gets the input right after the quote.

crates/mdbook-driver/src/builtin_preprocessors/links.rs
GuillaumeGomez: It's the first time I saw this kind of comments so I was wondering. But yeah, better without it. :+1:

GuillaumeGomez (approved):

GuillaumeGomez:
Looks all good to me, thanks a lot! And thanks @Chocorean for reviewing as well!

PR #3168, opened 2026-07-26
Update elasticlunr-rs to v3.1.0

mattico:
This release of elasticlunr-rs fixes a number of panics on different inputs, and fixes an issue which caused tokens to leak into the index of later fields: <https://github.com/mattico/elasticlunr-rs/issues/59>

The field leak fix caused a change to the search index snapshot and an assertion. I manually verified that the updated assertion is correct for the content in question.


ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3168/4f8c9460977e18974b37bb9a2292219bfa317632..4afa69afd8ab5aab64b3fbf7227bdc7103c329be)*

PR #3169, opened 2026-07-27
Fix print output inheriting screen theme colors

### LINKED ISSUE #1934: Print page is inheriting some theme colors
On Chrome and Safari, the print page is inheriting some theme colors. This can make the print page unreadable, or somewhat ugly. Firefox does not seem to be affected.

Safari Light:
image

Safari Coal:
image

giri256:
Fixes #1934.

## Summary

- reset print-only content variables to mdBook's existing light palette
- keep all selected screen themes unchanged
- cover page, text, quotes, tables, keyboard keys, footnotes, and callouts through the shared variables

## Verification

- Chrome print-media computed-style matrix for ayu, coal, light, navy, rust, and no-JavaScript navy
- navy screen-media negative control
- `cargo test -p mdbook-html --lib`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/crates/mdbook-html/front-end/css/print.css
+++ b/crates/mdbook-html/front-end/css/print.css
@@ -1,4 +1,26 @@
 
+/* Always use the light content palette when printing, regardless of the
+   selected screen theme. */
+html:root {
+    --bg: hsl(0, 0%, 100%);
+    --fg: hsl(0, 0%, 0%);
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Instead of redefining the variables, would be simpler to find a way to reuse the `light` variables from `variables.css`.

PR #3184, opened 2026-08-05
tests/testsuite/main.rs: remove unused `glob_one` import

DanielEScherzer:
Noticed in the CI logs


DanielEScherzer:
Hmm, it is used by other files, maybe a clippy false positive? `pub use` should be allowed even when not used within the current file

GuillaumeGomez:
Closing it then if you don't mind. ;)

PR #3185, opened 2026-08-08
Generate linkable IDs for punctuation-only headings

### LINKED ISSUE #3002: Headings with just punctuation do not generate a linkable ID

fzlzjerry:
## Summary

- use `section` when heading text has no characters that can form an automatic HTML ID
- keep generated IDs unique as `section`, `section-1`, and so on
- cover punctuation-only, empty, and explicitly identified headings

## Why

Punctuation-only headings currently render with `id=""` and `href="#"`, so the generated header link points to the page root instead of the heading. A second such heading becomes `-1`.

This follows the fallback used by Pandoc and kramdown while leaving explicit IDs unchanged.

Fixes #3002.

## Validation

- `cargo test --workspace`
- `cargo test --workspace --no-default-features`
- `cargo clippy --workspace --all-targets --no-deps -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --document-private-items --no-deps`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/crates/mdbook-html/src/utils.rs
+++ b/crates/mdbook-html/src/utils.rs
@@ -73,6 +73,8 @@ pub(crate) fn unique_id(id: &str, used: &mut HashSet<String>) -> String {
 }
 
 /// Generates an HTML id from the given text.
+///
+/// Falls back to `section` when no valid ID characters remain.
 pub(crate) fn id_from_content(content: &str) -> String {
     // This is intended to be close to how header ID generation is done in
     // other sites and tools, but is not 100% the same. Not all sites and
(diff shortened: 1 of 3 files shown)


DanielEScherzer (approved):
community review: looks good to me, I was worried that the repeated use of "section" would lead to duplicate IDs but that seems to already be handled

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3185/b90df240a318da0c59ec3efe6b75a58f63c6c459..da60cc1bd0dae750d52e7e49170b86e445b6340d)*

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me as well, thanks both of you!

GuillaumeGomez:
Sorry I'll close this PR in favour of https://github.com/rust-lang/mdBook/pull/3092 for two reasons:

1. #3092 was open weeks before this one.
2. This PR was made with LLM, and we're trying to onboard new members on the team. For that we need newcomers to level up in understanding the codebase so they can potentially become able to review PRs and join the newly created `mdbook` team.

PR #3187, opened 2026-08-10
Fix optional extensions with inaccessible PATH entries

### LINKED ISSUE #3170: `optional = true` renderer fails with Broken pipe mdBook 0.5.4
### Problem

On Linux, mdBook 0.5.4 fails when a configured renderer is not installed, even when the renderer is marked as optional.

The same setup worked with mdBook 0.5.1 and started failing after upgrading to 0.5.4.

According to the documentation, `optional = true` should demote a missing renderer error to a warning:

https://rust-lang.github.io/mdBook/format/configuration/renderers.html#optional-backends

### Actual behavior

```text
INFO Book building has started
INFO Running the html backend
INFO HTML book written to `/book/book/html`
INFO Running the missing backend
INFO Invoking the "missing" renderer
WARN Error writing the RenderContext to the backend, Broken pipe (os error 32)
ERROR Renderer exited with non-zero return code.
ERROR Rendering failed
      Caused by: The "missing" renderer failed
```

mdBook exits with code 101.

### Expected behavior

mdBook should emit a warning that `mdbook-missing` was not found, skip the optional renderer, and complete the HTML build successfully.

For example:

```text
WARN The command `mdbook-missing` for backend `missing` was not found, but is marked as optional.
```

### Steps

Create a book with this configuration:

```toml
[book]
title = "Optional renderer reproduction"
src = "src"

[output.html]

[output.missing]
optional = true
```

With a minimal `src/SUMMARY.md` and chapter, ensure `mdbook-missing` is not installed, then run:

```console
mdbook build
```

The same failure occurs during the initial build performed by:

```console
mdbook serve
```

### Possible Solution(s)

_No response_

### Notes

Failing environment:

```text
mdbook v0.5.4
rustc 1.97.1
Linux x86_64
Debian trixie container
Installed with: cargo install --locked mdbook --version 0.5.4
```

For comparison, the same mdBook version and configuration on Darwin arm64 correctly detects that the executable is missing, emits the optional-renderer warning, and exits successfully.

### Version

```text
mdbook v0.5.4
```

tianrking:
Fixes #3170.

## Problem

A missing optional renderer or preprocessor can still fail a build when the
platform reports command lookup differently from `ErrorKind::NotFound`, or when
the child exits before mdBook writes its JSON input. This regressed the expected
`optional = true` behavior on Linux.

## Change

- Preflight optional extension commands on both Unix and Windows before spawning
  them, so the optional-command decision is not dependent on child-process timing.
- Preserve the child command's working directory when resolving relative command
  paths and relative `PATH` entries.
- On Unix, treat an inaccessible `PATH` entry with no usable command as absent,
  while direct command paths and existing/non-executable files still go through
  `spawn` so their real errors are retained.
- On Windows, mirror `std::process::Command`'s documented bare-command search:
  the executable directory, system directory, Windows directory, then `PATH`;
  an omitted `.exe` extension is also considered.
- Apply the common preflight to renderer execution and both preprocessor paths
  (`run` and `supports`).

## Regression coverage

- Renderer and preprocessor: a missing optional command with an inaccessible
  `PATH` entry is skipped on Unix.
- Unix: an existing non-executable direct command is not misclassified as
  missing.
- Windows: an extensionless command with an existing `.exe` is not
  misclassified as missing.

## Validation

Local Windows validation with Rust 1.88.0:

- `cargo test --workspace --locked`
- `cargo test --workspace --no-default-features --locked`
- `cargo test -p mdbook-driver`
- `cargo test --test testsuite missing_optional -- --nocapture`
- `cargo clippy -p mdbook-driver --all-targets --no-deps -- -D warnings`
- `cargo clippy -p mdbook-driver --all-targets --no-default-features --no-deps -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`


GuillaumeGomez:
Would be better to handle unix and windows the same.

tianrking:
@GuillaumeGomez I updated this in 915ae729 to use the same optional-command preflight on Unix and Windows.

- Windows now follows `std::process::Command`'s documented bare-command search locations and `.exe` elision.
- Direct paths and existing/non-executable files still go through `spawn`, so their real errors are preserved.
- The Unix inaccessible-`PATH` renderer and preprocessor regressions remain covered; I also added Unix direct-command and Windows extensionless-command coverage.

The branch has been rebased onto current `main`. Both fresh CI runs for 915ae729 are green (38 successful checks total). Could you please take another look?

GuillaumeGomez (changes requested):
I don't think this is the right approach. Why not fixing the existing error in `spawn` and `crate::handle_command_error` instead?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3187/6a75867463828eb5b0b00f78feb0224c1bbc39aa..915ae72913e10c657837f162a151230201e65dab)*

PR #3189, opened 2026-08-14
fix: don't highlight the first chapter when another one is the index

### LINKED ISSUE #2883: README that is not the first chapter confuses the current page highlight
### Problem

If a chapter named `README.md` is not the first chapter, the sidebar gets confused as to which page is currently active.

### Steps

With a SUMAMRY.md outline such as:

```toml
# Summary

[Prefix 1](prefix-1.md)
[Prefix 2](prefix-2.md)

- [Introduction](README.md)
```

when loading index.html, both "Prefix 1" and "Introduction" are highlighted as the current page.

Image

Only "Introduction" should be highlighted.

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
mdbook v0.5.0-alpha.1
```

VXNCXNX:
Fixes #2883.

## The problem

The sidebar treats the first chapter as an alias for `index.html`:

```js
|| i === 0 && path_to_root === '' && current_page.endsWith('/index.html')
```

That holds when the first chapter *is* the index page. But a `README.md` chapter placed later in `SUMMARY.md` generates `index.html` itself, so loading the index highlights **both** it and the first chapter:

```
SUMMARY.md:  [Prefix 1](prefix-1.md)
             [Prefix 2](prefix-2.md)
             - [Introduction](README.md)

before:  ACTIVE: [ 'Prefix 1', '1. Introduction' ]
after:   ACTIVE: [ '1. Introduction' ]
```

which matches the screenshot in the issue.

## The fix

Skip the alias when some chapter already links to `index.html`:

```js
const has_index_chapter = links.some(link => link.getAttribute('href') === 'index.html');
```

In that case the real index chapter matches `current_page` directly on the line above, so the first-chapter fallback isn't needed and is actively wrong.

`getAttribute` rather than `link.href` is deliberate: `link.href` is reassigned inside the loop (`link.href = path_to_root + href`), and the browser-resolved absolute form would never equal a bare string on a `file://` book. Reading the raw attribute before the loop is the consistent choice.

## Why the exact string comparison holds up

This was the part I most expected to be brittle, so I checked rather than assumed. The href comes from `Chapter::path` verbatim (`helpers/toc.rs:119-125`), and the index preprocessor rewrites that path with `set_file_name("index.md")` (`builtin_preprocessors/index.rs:35-40`), so the filename is always literally `index`. Across 14 books I built to try to break it — `./README.md` in SUMMARY, `site-url` set, `ReadMe.markdown`, a `README.md#top` fragment, a prefix chapter, the index preprocessor disabled — every one produced exactly `index.html`. `site-url` only affects the 404 page's `base_url`; `--dest-dir` and `no-section-label` don't touch hrefs.

The subdirectory case is correctly *not* matched: `src/sub/README.md` gives href `sub/index.html`, and the root `index.html` really is generated from the first chapter there, so the alias should still apply.

I also looked for a false positive — a book with both `src/index.md` and `src/README.md`. That configuration is already broken and warned about (it emits duplicate `index.html` hrefs and highlighted three entries before this change); with the patch it highlights two rather than three. Not made worse.

## Tests

Added a case to `tests/gui/sidebar-active.goml` with a new fixture book whose index chapter is not first.

I ran the GUI tests for real rather than reasoning about them — no browser was present, so I installed puppeteer's chrome:

```
cargo test --test gui         23 succeeded, 0 failed
cargo test --workspace        all green
npm run lint                  clean
```

and confirmed the new case is load-bearing by stashing only the `toc.js.hbs` change:

```
[ERROR] line 23: expected 1 elements, found 2: assert-count: (".sidebar-scrollbox a.active", 1)
[ERROR] line 24: `Prefix 1` isn't equal to `1. Introduction`
<= doc-ui tests done: 22 succeeded, 1 failed
```

No CHANGELOG entry, since CONTRIBUTING says those are generated at release time with `cargo xtask changelog` — let me know if you'd like one anyway.

## One pre-existing thing I noticed

A SUMMARY link written as `sub/../README.md` produces href `sub/../index.html`, which this comparison misses. That book already has a broken `path_to_root` and highlights nothing before or after the change, so it looks like a separate pre-existing bug rather than something this should handle.


xhon-pelushi (approved):
Verified the two things I'd worry about with this fix:

**Operator precedence** — `&&` binds tighter than `||` in JS, so `!has_index_chapter` lands inside the existing `i === 0 && ... && ...` group as intended, not accidentally gating the `hrefsEqual` branch. Confirmed with a quick standalone truth-table check against the three cases in the issue (aliased first chapter suppressed, real index chapter still matches directly, and the no-conflict case is unaffected).

**The "pretty URLs" comment two lines above your change** — worth double-checking since `has_index_chapter` also does a raw string comparison against `'index.html'`. Traced `current_page` back to `document.location.href` in the same file: it's the live browser URL (which a proxy could rewrite to be extensionless), whereas `link.getAttribute('href')` comes from mdBook's own statically-rendered `<a>` tags, which are always literally `.html`-suffixed at build time. So that robustness concern only applies to `current_page`, not to the hrefs you're scanning — your exact-match check is safe.

Also appreciate that you flagged the `sub/../README.md` case as a separate pre-existing bug instead of folding a fix for it in here — keeps the diff focused on what the issue actually reported. Test setup (fixture book + goml assertions, verified load-bearing via stash) is exactly what I'd want to see for a sidebar-JS change like this.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3189/b90df240a318da0c59ec3efe6b75a58f63c6c459..589c738d43aa54f26e3ed0808ac7627277cb3592)*

tests/gui/sidebar-active.goml
```diff
@@ -15,3 +15,10 @@ assert-text: (".sidebar-scrollbox a.active", "Prefix 1")
 go-to: |DOC_PATH| + "all-summary/part-1/chapter-1.html?highlight=test"
 
 assert-text: (".sidebar-scrollbox a.active", "3. P1 C1")
+
+// When a chapter (like `README.md`) generates the index page, only that chapter
```
GuillaumeGomez: Please put this test into its own file.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3189/b90df240a318da0c59ec3efe6b75a58f63c6c459..589c738d43aa54f26e3ed0808ac7627277cb3592)*

tests/gui/sidebar-active.goml
```diff
@@ -15,3 +15,10 @@ assert-text: (".sidebar-scrollbox a.active", "Prefix 1")
...
+// should be highlighted, not the first chapter of the summary.
+go-to: |DOC_PATH| + "index-chapter/index.html"
+
+assert-count: (".sidebar-scrollbox a.active", 1)
+assert-text: (".sidebar-scrollbox a.active", "1. Introduction")
```
GuillaumeGomez: Can you also check when you go to the first chapter of the summary?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3189/b90df240a318da0c59ec3efe6b75a58f63c6c459..589c738d43aa54f26e3ed0808ac7627277cb3592)*

GuillaumeGomez (changes requested):
A few small nits but otherwise looks good to me!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3189/b90df240a318da0c59ec3efe6b75a58f63c6c459..589c738d43aa54f26e3ed0808ac7627277cb3592)*

VXNCXNX:
Both done. The index-chapter checks live in `tests/gui/index-chapter.goml` now, and that file also navigates to `prefix-1.html`, the first entry in the summary, and asserts it is the only active link there.

GuillaumeGomez:
Failing CI

VXNCXNX:
Not from this PR. `nightly linux` broke on new cargo manifest lints leaking into the `preprocessor` snapshot stderr, and master's own merge queue hit the same failure in the run for #3206 at 10:05. Nightly went green again at 10:36 on that PR, so I rebased on master to pick up a fresh run.

GuillaumeGomez:
Yes indeed. Sorry for the noise. Once https://github.com/rust-lang/mdBook/pull/3206 is merged, just rebase on it and CI should be happy.

GuillaumeGomez:
Please rebase.

VXNCXNX:
Correction: nightly is green on #3206 because that PR carries the `Cargo.toml` lint fixes from the closed #3213, not because the toolchain moved. So this stays red on nightly until #3206 lands, and a rerun after that will clear it.

VXNCXNX:
Rebased onto `main`. #3206 brought in the `[lints] workspace = true` fix, so the nightly `preprocessor` snapshot failures are gone locally too.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3191, opened 2026-08-15
Stop emitting infostring comments as HTML classes

### LINKED ISSUE #3188: Unknown code blocks attributes are still added as HTML class-names
Given unknown Rust code block attributes, mdBook still uses them as class names:
<pre><code>```rust,foo
fn bar() {}
```</code></pre>
```html
<code class="language-rust foo hljs hide-boring"> ... </code>
<!--                       ~~~                            -->
```

This causes issues with the rustc book; a lint uses `rust,ignore (requires next solver)`, which generates `class="language-rust ignore (requires next solver) hljs"`, and inadvertently adds the `.next` class who adds a `float: right;` CSS property:

Two code blocks are superposed

Lint code: https://github.com/rust-lang/rust/blob/ba28ff76f353a722f31c4f3dd2ac4e437d36411b/compiler/rustc_lint_defs/src/builtin.rs#L5699-L5709
Lint documentation: https://doc.rust-lang.org/nightly/rustc/lints/listing/warn-by-default.html#recursion-depth-exceeding-limit (can't permalink this one)

Note: rustdoc codeblocks allow parenthesized (comments): https://github.com/rust-lang/rust/blob/ba28ff76f353a722f31c4f3dd2ac4e437d36411b/src/librustdoc/html/markdown.rs#L914-L953

We should probably strip comments and/or only add classes for the predefined attributes. Rustdoc uses another syntax for custom classes: rust-lang/rust#79483.

Version: `mdbook v0.5.4`. Present at least since 2020 [^1], didn't check for earlier.

@rustbot label +A-Code-Blocks +C-bug

[^1]: <https://doc.rust-lang.org/1.48.0/rustc/lints/listing/allowed-by-default.html#unused-crate-dependencies> has a `ignore(needsexterncrate)` class. In Rust 1.56 (2021), the spaces were removed, resulting in the current `ignore (needs extern crate)` classes: <https://doc.rust-lang.org/1.56.0/rustc/lints/listing/allowed-by-default.html#unused-crate-dependencies>.

VXNCXNX:
Fixes #3188.

## What's broken

Every infostring token after the language becomes a raw HTML class:

```
```rust,ignore (requires next solver)
->  <code class="language-rust ignore (requires next solver)"
```

So `(requires`, `next` and `solver)` all become classes. The `next` one collides with the navigation rule at `crates/mdbook-html/front-end/css/chrome.css:184`:

```css
.next {
    float: right;
    right: var(--page-padding);
}
```

The code block floats right and overlaps, which is the screenshot in the issue. This breaks the real rustc book, whose lint emits exactly that infostring.

## The fix

Stop at the opening paren, since rustdoc treats the rest of the infostring as a comment. Two lines.

## Why not an allowlist

That was the other candidate, and I think it is your call rather than mine. The documented attribute list already exists in `guide/src/format/mdbook.md`, and `tree.rs` already consumes exactly those names, so the list would not have to be invented. But `tests/testsuite/markdown/basic_markdown/src/code-blocks.md:15` deliberately asserts that `text cls1,,cls2\tcls3` passes arbitrary classes through on a non-Rust block. An allowlist would either break that or have to be scoped to `language-rust` blocks only, which is a behaviour change to a documented escape hatch.

Comment stripping is the half that fixes the rustc book without touching that. Say the word if you would rather have the allowlist and I will send it instead.

## Verification

```
$ mdbook build

before:  <code class="language-rust ignore (requires next solver)"
after:   <code class="language-rust ignore"
```

The `cls1 cls2 cls3` fixture is unchanged, which I checked both in the expected file and by building a book with both fences:

```
<code class="language-rust ignore"
<code class="language-text cls1 cls2 cls3"
```

Full CI gates pass: `cargo fmt --check`, `cargo build --locked`, `cargo test --workspace --locked` (110 passed in testsuite, no failures anywhere), `cargo test --workspace --no-default-features`, and `cargo clippy --workspace --all-targets --no-deps -- -D warnings` with no warnings.

Reverting just the `tree.rs` change with the fixture kept fails the `basic_markdown` test on exactly the expected line.

## Edge cases I checked

`rust,ignore(no_space)` gives `language-rust ignore`. An unbalanced `(` truncates there, which matches rustdoc. An empty infostring still emits no class attribute.

One honest behaviour loss: a language token containing `(`, as in ` ```ru(st `, now truncates to `language-ru`. No real language name uses a paren and rustdoc would read it the same way, but it is the only thing this change takes away.

### DIFF
--- a/crates/mdbook-html/src/html/tree.rs
+++ b/crates/mdbook-html/src/html/tree.rs
@@ -422,6 +422,8 @@ where
                 let mut code = Element::new("code");
                 match kind {
                     CodeBlockKind::Fenced(info) => {
+                        // Rustdoc treats everything from an opening paren as a comment.
+                        let info = info.split('(').next().unwrap_or_default();
                         let mut infos =
                             info.split([' ', '\t', ',']).filter(|info| !info.is_empty());
                         if let Some(lang) = infos.next() {
--- a/crates/mdbook-html/src/html/tree.rs
+++ b/crates/mdbook-html/src/html/tree.rs
@@ -422,8 +422,7 @@ where
                 let mut code = Element::new("code");
                 match kind {
                     CodeBlockKind::Fenced(info) => {
-                        // Rustdoc treats everything from an opening paren as a comment.
-                        let info = info.split('(').next().unwrap_or_default();
+                        let info = strip_infostring_comments(&info);
                         let mut infos =
                             info.split([' ', '\t', ',']).filter(|info| !info.is_empty());
                         if let Some(lang) = infos.next() {
@@ -1107,6 +1106,33 @@ where
     }
 }
 
+/// Removes rustdoc-style `(comment)` runs from a code block infostring.
+///
+/// Rustdoc's grammar is `comment = OPEN_PAREN *<all characters except closing
+/// parentheses> CLOSE_PAREN`, and a lang-string may interleave comments with
+/// tokens, so `rust (why),ignore` keeps both `rust` and `ignore`. An unclosed
+/// paren runs to the end, matching rustdoc's own recovery.
+fn strip_infostring_comments(info: &str) -> Cow<'_, str> {
+    if !info.contains('(') {
+        return Cow::Borrowed(info);
+    }
+    let mut out = String::with_capacity(info.len());
+    let mut depth = 0usize;
+    for ch in info.chars() {
+        match ch {
+            '(' => depth += 1,
+            ')' => depth = depth.saturating_sub(1),
+            // Keep a separator so tokens either side of a comment stay distinct.
+            _ if depth == 0 => out.push(ch),
+            _ => {}
+        }
+        if ch == ')' && depth == 0 {
+            out.push(' ');
+        }
+    }
+    Cow::Owned(out)
+}
+
 /// Traverse the given node, emitting any plain text into the output.
 ///
 /// This is used to generate the `id` of a header.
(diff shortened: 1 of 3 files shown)


crates/mdbook-html/src/html/tree.rs
```diff
@@ -422,6 +422,8 @@ where
                 let mut code = Element::new("code");
                 match kind {
                     CodeBlockKind::Fenced(info) => {
+                        // Rustdoc treats everything from an opening paren as a comment.
+                        let info = info.split('(').next().unwrap_or_default();
```
GuillaumeGomez: It's not "from an opening paren" but everyting inside parens. So you can have:

`````rust
````rust (comment),ignore
`````

And then it would be both `rust` and `ignore`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3191/b90df240a318da0c59ec3efe6b75a58f63c6c459..198f8360a67f88726c88258d9bbdbc831c33b2b1)*

**Review by GuillaumeGomez — CHANGES_REQUESTED** [review]

File: crates/mdbook-html/src/html/tree.rs (hunk 1 above)
Comment (VXNCXNX) [reply crates/mdbook-html/src/html/tree.rs]: You are right, thanks. I read rustdoc's grammar after your comment and it is explicit:

```
lang-string = *(token-list / delimited-attribute-list / comment)
comment     = OPEN_PAREN *<all characters except closing parentheses> CLOSE_PAREN
```

Comments interleave with tokens, so truncating at the first paren threw away everything after it. Your example was silently losing `ignore`.

Now removes each parenthesized run in place, leaving a separator so tokens either side stay distinct:

```
rust (comment),ignore               -> [rust, ignore]
rust,ignore (requires next solver)  -> [rust, ignore]
rust (a) (b) ignore                 -> [rust, ignore]
rust (unclosed                      -> [rust]
```

An unclosed paren still runs to the end, which matches rustdoc's own recovery.

Added your exact case to the `basic_markdown` fixture, and a unit test on the helper. Reverting to the old truncation fails the fixture by dropping `ignore` from the emitted class, so the case is actually pinned rather than just described. `cargo test --workspace` is clean, fmt and clippy too.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks! Do you mind taking a look @notriddle as well? :)

notriddle (commented):
Looks fine.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3191/b90df240a318da0c59ec3efe6b75a58f63c6c459..c4e2769e57f4a12933db68d2101b856dc214e020)*

notriddle (approved):

GuillaumeGomez:
Then let's merge!

PR #3198, opened 2026-08-26
Use capital letters to indicate the default value in (y/n) prompts

### LINKED ISSUE #2395: Use capital letters to indicate the default value in (y/n) prompts
### Problem

currently there is no way to tell the default value in a yes/no question.
for example, in the prompt `Do you want a .gitignore to be created? (y/n)` it is not clear what pressing the return key will do

### Proposed Solution

generally the default option is specified with a capital letter
the previous example would turn into: `Do you want a .gitignore to be created? (y/N)`

### Notes

_No response_

Chocorean:
Fix #2395

Improve the two yes/no prompts by capitalizing the default choice, No

In both scenarios, nothing happens if the user types anything else than the few strings mentioned in the confirm() function in src/cmd/init.rs

### DIFF
--- a/src/cmd/init.rs
+++ b/src/cmd/init.rs
@@ -41,7 +41,7 @@ pub fn execute(args: &ArgMatches) -> Result<()> {
         // Skip this if `--force` is present
         if !args.get_flag("force") && theme_dir.exists() {
             println!("This could potentially overwrite files already present in that directory.");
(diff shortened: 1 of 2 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks for the improvement!

PR #3203, opened 2026-08-30
fix: pin next-chapter button to the viewport edge

### LINKED ISSUE #3202: "Next chapter" button doesn't touch the right edge of the screen
### Problem

The "Next chapter" button didn't get clicked.
The button looks right-aligned, so you expect it to be right-aligned. Annoyingly, you have to move the cursor a bit to the left, wait for the highlight to appear and only then click it, when you're sure that you're actually hovering over the button.

See https://github.com/rust-lang/mdBook/pull/993

Image

Even when the scrollbar is present, I don't think that the button has to be not right-aligned, because the scrollbar is prioritized for clicks.

### Steps

1. Open https://doc.rust-lang.org/stable/book/title-page.html
2. Move cursor to the very right edge of the screen
3. Click

### Possible Solution(s)

Just alight the button to the right, as in https://github.com/rust-lang/mdBook/pull/993.

### Notes

_No response_

### Version

```text

```

QuarkOS:
fixes #3202

The next chapter button is position fixed with a 15px right padding, so clicks on the actual viewport edge miss it.

We pin the next control to right 0 (left 0 in RTL).

I could not run the gui harness here (rustc 1.85, mdBook wants 1.88). CSS before was 15px, after is 0.


Chocorean:
Would not that be inconsistent with the previous chapter button? There is also a margin at the left of the previous chapter button:

image
image
image

On a personal note, I don't think sticking the buttons to the edges is super relevant because the scrollbar takes the priority anyway on the right edge, as that was mentioned in the original issue

QuarkOS:
The left arrow is inset on purpose so it does not sit on the sidebar drag strip also the right arrow had leftover page padding, thats why edge clicks missed. id leave the left arrow as it is right now but im also new to contributing so might not have the best judgement. #3202 already mentioned that the scrollbar is prioritized on edge clicks

GuillaumeGomez:
| current look | with this PR |
|-|-|
| image | Screenshot From 2026-09-02 22-08-40 |

> Would not that be inconsistent with the previous chapter button? There is also a margin at the left of the previous chapter button:

Not sure if it's important either. ^^'

I don't really have an opinion here. What do you think @rust-lang/mdbook ?

notriddle:
<details>Embarrassingly, my first response to this PR assumed it was introducing fixed-position as they weren’t always fixed-position. That was dumb of me. These buttons in mdbook are slowest fixed-position, and this PR only removes a bit of margin!</details>

GuillaumeGomez:
So what's your final opinion on this @notriddle ?

notriddle:
> On a personal note, I don't think sticking the buttons to the edges is super relevant because the scrollbar takes the priority anyway on the right edge, as that was mentioned in the original issue

This concern was never really addressed. We don't want people to accidentally click the next chapter button when they intended to use the scrollbar, so making it flush doesn't seem good.

GuillaumeGomez:
Fair enough, let's start the FCP then.

@rfcbot fcp merge mdbook

GuillaumeGomez:
@rfcbot fcp close

GuillaumeGomez:
I commented on the wrong PR...

GuillaumeGomez:
@rfcbot fcp cancel

GuillaumeGomez:
Not clicking on the next chapter button by mistake when trying to click on the scrollbar makes more sense to me and is better UX overall. So let's close it.

PR #3206, opened 2026-08-31
Prepare to rename primary branch to `main`

DanielEScherzer:

PR #3213, opened 2026-09-03
Fix new cargo lints

GuillaumeGomez:


GuillaumeGomez:
Arf, I need a review for that. I'll push my commit on https://github.com/rust-lang/mdBook/pull/3206, will be simpler...

PR #3217, opened 2026-09-08
Add citation option in configuration

Evian-Zhang:
Follow up the discussion in [this thread in Rust Internal Forums](https://internals.rust-lang.org/t/add-citations-for-official-rust-books/24507). More and more academic papers tend to cite official Rust books, and I think it would be better if the HTML version can directly give people the citation files.

In this PR, a new configuration key "citation" is added to the `[output.html]`. This key points to a file, and a new button will be added to the up right corner, which will link to the pointed file for user to download.

For now, the download button simply link to the original file given by book authors. In the future, this functionality can become: author provide either bib or cff, user can download bib, plaintext, APA, etc.. But since there is no such a mature crate for this functionality, I think we can temporarily keep it simple, and such functionality should not reside in this crate, but in an individual crate.


GuillaumeGomez:
Please add a screenshot of what the page looks like with the citation button. Otherwise seems like a good idea to me overall.

Evian-Zhang:
截屏2026-09-09 00 36 44
截屏2026-09-09 00 37 22

It looks like this, in the upper right corner.

PR #3218, opened 2026-09-09
Fix playground selector

### LINKED ISSUE #3215: Code block with `python,playground` cause wrong playground behaviour
### Problem

This issue is a part of #3124.

Now code block with `python,playground` is not supported, so it is expected no-op.
But actually it causes some strange behaviours.

### Steps

```toml
[output.html.playground]
editable = true
```

````markdown
```python,playground
print("hello")
```

```rust
fn main() { println!("hello"); }
```
````

* broken copy button is shown at the first python code block
* play/copy button is not show at the second rust code block

I think the expected behaviour is the python code block is not playground (because `python,playground` is not supported) and the rust code block works fine.
If this issue is fixed, user can insert `python,playground` support by the way of #3124.

### Possible Solution(s)

The necessary code change is probably like below:

https://github.com/rust-lang/mdBook/blob/a57975d499a660fd11da05a9010e93fe245525ba/crates/mdbook-html/front-end/js/book.js#L45

```diff
-    const playgrounds = Array.from(document.querySelectorAll('.playground'));
+    const playgrounds = Array.from(document.querySelectorAll('pre.playground'));
```

https://github.com/rust-lang/mdBook/blob/a57975d499a660fd11da05a9010e93fe245525ba/crates/mdbook-html/front-end/js/book.js#L273

```diff
-    Array.from(document.querySelectorAll('.playground')).forEach(function(pre_block) {
+    Array.from(document.querySelectorAll('pre.playground')).forEach(function(pre_block) {
```

### Notes

_No response_

### Version

```text
mdbook v0.5.3
```

dalance:
Fixes #3215

## Problem

Now playground selector in `book.js` is `.playground`, so arbitrary code block like `lang,playground` is targeted of playground code injection.

## Fix

More strict selector is `pre.playground` because mdBook generates `<pre class="playground">` for Rust playground.


tests/gui/books/playground/src/chapter_1.md
```diff
@@ -0,0 +1,13 @@
+# Chapter 1
+
+```python,playground
```
GuillaumeGomez, line 3: Should we emit a warning in `playground` used on other languages than rustdoc?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3218/6a75867463828eb5b0b00f78feb0224c1bbc39aa..536a5aa965166543fa572b03769a7d912aee51cd)*

GuillaumeGomez (changes requested):

tests/gui/books/playground/src/chapter_1.md
dalance: Now, `playground` is not included in the list of code block attributes in the document.
https://rust-lang.github.io/mdBook/format/mdbook.html?highlight=noplay#rust-code-block-attributes

Since users can assign arbitrary attributes (such as `rust,large_font`) and have them processed by plugins or custom JS/CSS,
I see no reason to add a special check specifically for `playground`.
It would likely be better to consider this in the future if custom playgrounds and the `playground` attribute become official features.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks all good to me, thanks!

PR #3219, opened 2026-09-09
fix: definition list in Chrome

pinage404:
Hello,

When there are several definitions (`<dd>`) for a term (`<dt>`), definitions (`<dd>`) are prefixed with a number

When there are several terms with both several definitions, the number was not incremented the same way in Firefox and Chrome

Tested with version `0.5.4` and the current `main` branch

The behaviour was good with Firefox

But the number wasn't resettled with definition term (`<dt>`) in Chrome

I saw this warning on MDN

> Warning: There is a difference between `counter-reset` and `counter-set` properties. After creating a counter using `counter-reset`, you can adjust its value by using the `counter-set` property. This is counterintuitive because, despite its name, the `counter-reset` property is used for creating and initializing counters, while the `counter-set` property is used for resetting the value of an existing counter.
>
> <https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/counter-reset#description>

I tried to change in the code, it aligned the behaviour of Chrome on the Firefox

Code to reproduce the bug

```md
# Example

Foo
  : should be 1
  : should be 2
  : should be 3

Bar
  : should be 1
  : should be 2
  : should be 3

Before this patch
  : Works with
    : Zen Browser 1.21.16b (Firefox 154.0.1) (64 bits)

  : Does not work with
    : Chromium Version 152.0.7977.64 (Build officiel)

After this patch
  : Works with
    : Zen Browser 1.21.16b (Firefox 154.0.1) (64 bits)
    : Chromium Version 152.0.7977.64 (Build officiel)
```

Render :

|Before|After|
|:--:|:--:|
|before|after|

I only tested on Linux with :
* Zen Browser 1.21.16b (Firefox 154.0.1) (64 bits)
* Chromium Version 152.0.7977.64 (Build officiel)

I didn't do any others checks


GuillaumeGomez:
Seems good to me, thanks! Please add a GUI test.

pinage404:
I don't know what should i check neither where / how i can check

Do you have clues ?

GuillaumeGomez:
If it's present in the text, you can check with `assert-text`. If not, maybe the element itself has the information?

GuillaumeGomez:
I couldn't figure out a way to check that with the GUI test as it's a CSS only value, computed by the web browser. So yeah... Please remove the last commit, I'll make one last review pass afterwards to confirm I didn't miss anything and then merge.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks all good to me, thanks!

PR #3222, opened 2026-09-10
fix: normalize path separators for cross-platform resource lookup

### LINKED ISSUE #3221: Path handling in hashed-asset resource replacement is OS dependent
### Problem

It might just be fonts but anywhere `Path::join` is used is susceptible to this bug
https://github.com/rust-lang/mdBook/blob/6a75867463828eb5b0b00f78feb0224c1bbc39aa/crates/mdbook-html/src/html_handlebars/static_files.rs#L113-L118

The resource syntax replacement look-up requires the path to match exactly

https://github.com/rust-lang/mdBook/blob/6a75867463828eb5b0b00f78feb0224c1bbc39aa/crates/mdbook-html/src/html_handlebars/static_files.rs#L206-L213

This means:
  - if I declare `{{ resource "fonts\my_font.woff2" }}` somewhere, the book can only be built properly on Windows. On other platforms, it cannot match the asset and skips adding the hash and ends up with invalid paths.
  - if I declare `{{ resource "fonts/my_font.woff2" }}` somewhere, the book can only be built properly on non-Windows. 

### Steps

```
mdbook init
mkdir -p themes/fonts
echo '{{ resource "fonts/my_font.woff2" }}' > themes/fonts/my_fonts.css
echo 'random stuff' > themes/fonts/my_font.woff2
mdbook build
```
Observe the output `book/fonts/my_fonts-<hash>.css` points to the hashed path on non-Windows and the non-hashed path on Windows. Change `/` to `\` in `my_fonts.css`, and observe the opposite happens

### Possible Solution(s)

I think the cleanest is to require `/` in the source asset files, and ensure that in static_files.rs, we always use `/` when registering files into the hash map. This would break someone who is currently using `\` in their source files on Windows, but it will be a temporary pain and it can be somewhat mitigated (if it matters at all) by printing a warning or making it a hard failure, if `\` is detected. The replacement is also purely mechanical and doesn't allow escape anyway, so there should never be a `\` in the path.

Alternatively the asset matching needs to be relaxed to allow both windows and unix style path separators, but I feel like that's when dealing with escapes and stuff would cause more complications

### Notes

_No response_

### Version

```text
mdbook 0.5.4
```

mmustafasenoglu:
Fixes #3221

The `{{ resource }}` directive in templates uses `/` as separator, but `Path::join` on Windows produces `\`. This means resource hash lookups fail when templates use forward slashes on Windows (or backslashes on non-Windows).

Changes:
- `static_files.rs`: Add `normalize_path()` to normalize filenames to `/` when registering static files into the hash map
- Same normalization applied to resource directive lookups in `replace_all`

This ensures that `{{ resource "fonts/my_font.woff2" }}` works correctly on all platforms, regardless of OS path separator.


crates/mdbook-html/src/html_handlebars/static_files.rs
```diff
@@ -102,12 +115,14 @@ impl StaticFiles {
         for custom_file in custom_files {
             let input_location = root.join(custom_file);
 
+            let filename = custom_file
+                .to_str()
+                .with_context(|| "resource file names must be valid utf8")?
+                .to_owned();
+            let normalized = normalize_path(&filename);
```
GuillaumeGomez:
```suggestion
            let filename = custom_file
                .to_str()
                .with_context(|| "resource file names must be valid utf8")?;
            let normalized = normalize_path(filename);
```

crates/mdbook-html/src/html_handlebars/static_files.rs
```diff
@@ -117,9 +132,10 @@ impl StaticFiles {
                 .to_str()
                 .with_context(|| "resource file names must be valid utf8")?
                 .to_owned();
```
GuillaumeGomez: Remove the `.to_owned()` here too.

crates/mdbook-html/src/html_handlebars/static_files.rs
```diff
@@ -11,6 +11,19 @@ use std::collections::HashMap;
...
+/// The `{{ resource "name" }}` directive in templates uses `/` as the
+/// separator, but `Path::join` on Windows produces `\`. This function
+/// ensures consistent keys regardless of OS.
+fn normalize_path(path: &str) -> String {
+    if cfg!(windows) {
+        path.replace('\\', "/")
+    } else {
+        path.to_owned()
+    }
+}
```
GuillaumeGomez:
```suggestion
fn normalize_path(path: &str) -> String {
    #[cfg(windows)]
    {
        path.replace('\\', "/")
    }
    #[cfg(not(windows))]
    {
        path.to_owned()
    }
}
```

I also wonder if we should return a `Cow` instead to prevent the unneeded allocations on non-windows...

GuillaumeGomez (changes requested):
Please also add a regression test (with `cfg(windows)`).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3222/6a75867463828eb5b0b00f78feb0224c1bbc39aa..6d5e93d06878eb75476563d979128f224a773f88)*

mmustafasenoglu:
Thanks for the detailed review! Done in the latest push: dropped both .to_owned() calls (the font path needed a let binding to live long enough), switched to #cfg(windows) blocks, normalize_path now returns Cow so non-Windows doesn't allocate, and added regression tests including a cfg(windows)-gated backslash case. Tests, fmt and clippy all pass locally.

@rustbot ready

crates/mdbook-html/src/html_handlebars/static_files.rs
```diff
@@ -317,4 +338,26 @@ mod tests {
...
+            "css/general.css"
+        );
+        assert_eq!(normalize_path("book.js").into_owned(), "book.js");
+    }
+
+    #[test]
+    #[cfg(windows)]
+    fn test_normalize_path_windows_separators() {
+        assert_eq!(
+            normalize_path("css\\general.css").into_owned(),
```
GuillaumeGomez, line 355: Please also test the equivalent paths with `/` (the `/` is also accepted for windows paths, just like `\`).

GuillaumeGomez (changes requested):
Just one nit and then it's good to go! Please also [squash](https://blog.guillaume-gomez.fr/Git-tips/1/2) your commits.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/mdBook/3222/6a75867463828eb5b0b00f78feb0224c1bbc39aa..e933ecad9f25cae11122ff3ec0c9efac3e298b6f)*

mmustafasenoglu:
Done — the Windows test now covers both separators for each path, and the branch is squashed into a single commit. Tests and fmt pass.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

